PR #28, opened 2020-08-04
feat(visualize): add capability to visualize ontology

ManuelRosenthaler:


BalduinLandolt:
@ManuelRosenthaler this branch has been stale for a long time and is quite far behind main. To what extent is this still relevant? Do we really need this visualization?

ManuelRosenthaler:
@BalduinLandolt Absolutely true ^^ we dont "need" it at all. It's just a "nice to have". But if it's too much work to integrate into the new versions, then i guess we can just throw it in the bin. But if it would be easy to integrate - why not, right? ^^

BalduinLandolt:
true... I'll have a look

PR #49, opened 2021-03-17
test(xml upload): try to provoke error with date (DSP-1409)

tobiasschweizer:
resolves DSP-1409

### DIFF
--- a/knora/dsplib/models/value.py
+++ b/knora/dsplib/models/value.py
@@ -298,24 +298,6 @@ class DateValue(Value):
         if self._y1 is None:
             raise BaseError("Invalid date format! " + str(value))
 
-        '''
-        if self._y2 is not None:
-            date1 = self._y1 * 10000
-            if self._m1 is not None:
-                date1 += self._m1 * 100
-            if self._d1 is not None:
-                date1 += self._d1
-            date2 = self._y2 * 10000
-            if self._m2 is not None:
-                date2 += self._m2 * 100
-            if self._d2 is not None:
-                date2 += self._d2
-            if date1 > date2:
-                self._y1, self._y2 = self._y2, self._y1
-                self._m1, self._m2 = self._m2, self._m1
-                self._d1, self._d2 = self._d2, self._d1
-        '''
-        
         super().__init__(iri=iri,
                          comment=comment,
                          permissions=permissions,
--- a/knora/dsplib/models/value.py
+++ b/knora/dsplib/models/value.py
@@ -297,6 +297,8 @@ class DateValue(Value):
         self._d2 = None if dp[8] is None else int(dp[8].strip('-: '))
         if self._y1 is None:
             raise BaseError("Invalid date format! " + str(value))
+
+        '''
         if self._y2 is not None:
             date1 = self._y1 * 10000
             if self._m1 is not None:
@@ -312,6 +314,8 @@ class DateValue(Value):
                 self._y1, self._y2 = self._y2, self._y1
                 self._m1, self._m2 = self._m2, self._m1
                 self._d1, self._d2 = self._d2, self._d1
+        '''
+        
         super().__init__(iri=iri,
                          comment=comment,
                          permissions=permissions,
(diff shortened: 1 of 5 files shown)


test/test_tools.py
```diff
@@ -49,7 +49,7 @@ def test_xmlupload(self):
                    imgdir="testdata/bitmaps",
                    sipi="http://0.0.0.0:1024",
                    verbose=True,
-                   validate=True)
+                   validate=False)
```
tobiasschweizer, line 52: @lrosenth @subotic To me it seems that the xml upload was not executed against DSP-API because `validate` was set to true. Hence, the upload was never tested automatically.

testdata/test-data.xml
```diff
@@ -58,7 +58,7 @@
             <boolean permissions="prop-default">true</boolean>
         </boolean-prop>
         <date-prop name=":hasDate">
-            <date permissions="prop-default" >JULIAN:CE:1401-05-17:CE:1402-01</date>
+            <date permissions="prop-default" >JULIAN:BCE:0700:BCE:0600</date>
```
tobiasschweizer, line 61: in the failed test it complains: 

> Message:{"knora-api:error":"org.knora.webapi.exceptions.BadRequestException: Start date 0600 BCE is after end date 0700 BCE","@context":{"knora-api":"http://api.knora.org/ontology/knora-api/v2#"}}!

https://github.com/dasch-swiss/dsp-tools/runs/2129275872#step:8:1448

tobiasschweizer:
Ok, the error must happen here:

https://github.com/dasch-swiss/dsp-tools/blob/d6d7e853fb28d3a0483ac350d1a52b5d3c79b4a2/knora/dsplib/models/value.py#L280-L314

value string:  JULIAN:BCE:0700:BCE:0600

DateValue before line 298: y1  700 y2  600

DateValue after line 314:  y1  600 y2  700

subotic:
here? https://github.com/dasch-swiss/dsp-tools/blob/d6d7e853fb28d3a0483ac350d1a52b5d3c79b4a2/knora/dsplib/models/value.py#L311

The reordering doesn’t take into account if BC or AD. It just compares the year values.

tobiasschweizer:
I think the swapping happens here:

https://github.com/dasch-swiss/dsp-tools/blob/d6d7e853fb28d3a0483ac350d1a52b5d3c79b4a2/knora/dsplib/models/value.py#L311-L314

The era is not taken into consideration.

subotic:
😉

tobiasschweizer:
Tests fail:

> FileNotFoundError: [Errno 2] No such file or directory: 'testdata/bitmaps/TEMP11.TIF'

https://github.com/dasch-swiss/dsp-tools/pull/49/checks?check_run_id=2138136550#step:8:1469

Since validate was set to true before for `xml_upload` in `test_xmlupload` I am not sure if this ever worked correctly.

knora/dsplib/models/value.py
```diff
@@ -297,6 +297,8 @@ def __init__(self,
         self._d2 = None if dp[8] is None else int(dp[8].strip('-: '))
         if self._y1 is None:
             raise BaseError("Invalid date format! " + str(value))
+
+        '''
```
tobiasschweizer, line 301: @subotic @lrosenth Should we improve this code or get rid of it?

If I understand correctly, this code was supposed to check for periods whose start date was after the end date (and then swap them if this is the case). So I do not know how hard it will be to make that work correctly with eras.

subotic (approved):
cool :-)

knora/dsplib/models/value.py
tobiasschweizer: @subotic @lrosenth What about this logic? Should we fix or simply remove it?

subotic: I don't understand. I thought you were fixing it?

tobiasschweizer: I just commented it out because it does some validation before sending the date to Knora. I think fixing it would take quite some time and I don't think it's worth the effort since Knora would reject an invalid date anyway.

tobiasschweizer: So far, the validation logic created a bug ;-)

tobiasschweizer: -> malidation

subotic (approved):

PR #52, opened 2021-04-07
chore: automate release process (DSP-1492)

BalduinLandolt:
Resolves: DSP-1492

Questions:
- [x] in previous version: why the strange dependencies installation?
- [x] under ubuntu-latest, should one call `python` or `python3`, and `pip`/`pip3` respectively?

### DIFF
--- a/.github/workflows/ckeck-pr-title.yml
+++ b/.github/workflows/ckeck-pr-title.yml
@@ -1,0 +1,20 @@
+name: PR-Title
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+
+jobs:
+  test:
+    name: Check PR Title
+    runs-on: ubuntu-latest
+    steps:
+      # check PR title
+      - uses: deepakputhraya/action-pr-title@master
+        with:
+          regex: '([a-z])+(\(([a-z\-_])+\))?!?: ([a-z])+' # Regex the title should match.
+          allowed_prefixes: 'fix,refactor,feat,docs,chore,style,test' # title should start with the given prefix
+          disallowed_prefixes: 'feature,hotfix,doc' # title should not start with the given prefix
+          prefix_case_sensitive: true # title prefix are case insensitive
+          min_length: 7 # Min length of the title
+          max_length: 60 # Max length of the title
--- a/.github/workflows/ckeck-pr-title.yml
+++ b/.github/workflows/ckeck-pr-title.yml
@@ -12,7 +12,7 @@ jobs:
       # check PR title
       - uses: deepakputhraya/action-pr-title@master
         with:
-          regex: '([a-z])+(\(([a-z\-_])+\))?!?: ([a-z ])+ \([A-Z]+-\d+\)' # Regex the title should match.
+          regex: '([a-z])+(\(([a-z\-_])+\))?!?: ([a-zA-Z ])+ \([A-Z]+-\d+\)' # Regex the title should match.
           allowed_prefixes: 'fix,refactor,feat,docs,chore,style,test' # title should start with the given prefix
           disallowed_prefixes: 'feature,hotfix,doc' # title should not start with the given prefix
           prefix_case_sensitive: true # title prefix are case insensitive
--- a/.github/workflows/ckeck-pr-title.yml
+++ b/.github/workflows/ckeck-pr-title.yml
@@ -12,7 +12,7 @@ jobs:
       # check PR title
       - uses: deepakputhraya/action-pr-title@master
         with:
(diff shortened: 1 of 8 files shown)


kilchenmann (approved):
LGTM

Since the single steps are in different files, I miss a bit the overview in `/actions`. But that's only my opinion. Let's see how it works...

tobiasschweizer:
crazy shit! I'll take a look now!

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,34 @@
...
+    name: Prepare next release
+    runs-on: ubuntu-latest
+    # release only if tests pass
+    if: ${{ github.event.workflow_run.conclusion == 'success'}}
+    steps:
+      - uses: GoogleCloudPlatform/release-please-action@v2
+        with:
+          token: ${{ secrets.GH_TOKEN }}
+          release-type: python
+          package-name: gh_action_test
```
kilchenmann, line 33: Package name is maybe wrong here. This name will be used as prefix in the release

BalduinLandolt:
I divided it into different files in hope the "separation of concerns" would make it simpler...

Also, should I add automated notification too?

kilchenmann:
> I divided it into different files in hope the "separation of concerns" would make it simpler...

No problem. That's fine.

> 
> Also, should I add automated notification too?

Yes, if you like. Why not?

BalduinLandolt:
> > I divided it into different files in hope the "separation of concerns" would make it simpler...
> 
> No problem. That's fine.
> 

It just poses the question, to what extent it should be unified all over the other repos. And if so, how it should be done everywhere.  
(I think the PR title test and my modifications to the PR template would add value everywhere. Splitting it into different files is probably a matter of taste.)

> > Also, should I add automated notification too?
> 
> Yes, if you like. Why not?

Done. I hope this will work like this, haven't tested that beforehand.

.github/workflows/ckeck-pr-title.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  test:
+    name: Check PR Title
+    runs-on: ubuntu-latest
+    steps:
+      # check PR title
+      - uses: deepakputhraya/action-pr-title@master
+        with:
+          regex: '([a-z])+(\(([a-z\-_])+\))?!?: ([a-z])+' # Regex the title should match.
```
tobiasschweizer, line 15: Would it make sense to add something like "\(DSP-\d+\)" to the end of the regex?
By convention, the PR's title should end with "(DSP-123)".

It's is actually missing from this PR's title :-)

kilchenmann: Not all PRs have a task on youtrack. So, this can be tricky. And for example, in another repo I had a task of type `PM-` and not `DSP-`.

BalduinLandolt: You're right that it's missing here in the PR title... ;-)  
But I also think it might become at some point problematic to enforce it. (There are also some tasks with `BS-123`. I'm not sure this will be the case in the future, so I think it's a bit risky to enforce this strictly.)

tobiasschweizer: maybe make it more general: expect something like: `\(.+-\d+\)`

tobiasschweizer: however, I leave this up to you

tobiasschweizer (approved):

.github/workflows/ckeck-pr-title.yml
BalduinLandolt: >maybe make it more general: expect something like: `\(.+-\d+\)`

good idea, I'll add that

BalduinLandolt: I went even stricter: `\([A-Z]+-\d+\)`

Do you think it's a problem that I allow only lower case letters in the PR title?  Should probably be extended to upper case too,  right? (e.g. for abbreviations that often use upper case.)

tobiasschweizer: > I went even stricter: `\([A-Z]+-\d+\)`

you are reading my thoughts :-)

> 
> Do you think it's a problem that I allow only lower case letters in the PR title? Should probably be extended to upper case too, right? (e.g. for abbreviations that often use upper case.)

no idea

BalduinLandolt: I'll just add upper case, just to be sure

tobiasschweizer: you can never be too paranoid

BalduinLandolt: we could argue about that ;-)

PR #57, opened 2021-04-09
chore: fix doc deployment (DSP-1492)

BalduinLandolt:
resolves DSP-1492


.github/workflows/release.yml
```diff
@@ -42,20 +42,11 @@ jobs:
...
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install -r requirements.txt
-      # release latest docs
-      - run: make publish-docs
+      - name: Deploy docs
+        uses: mhausenblas/mkdocs-deploy-gh-pages@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
```
kilchenmann, line 49: I think the "install requirements" is missing. You can add it to the `/mkdocs-deploy-gh-pages` action:

```
env:
          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
          REQUIREMENTS: requirements.txt
```

But I'm not sure about your list of requirements. I think in this repo they're not for the documentation only. Maybe it could give you a conflict. You should try to run the action on the next push. Otherwise we will see the action only in the next release process...

BalduinLandolt: in the [example usage](https://github.com/marketplace/actions/deploy-mkdocs#example-usage) they don't have it, so I assume it defaults to the `./requirements.txt`.  
Yes, I'll run it and we'll see if it works or if it needs separate requirements.

BalduinLandolt: But in order to be able to run it, I need to merge this PR, I think.

kilchenmann: You can activate this step on push.. Move it into a separate but temporary workflow script and to run it with next commit

BalduinLandolt: the requirements did indeed cause problems. I now reused the solution with the separate requirements file for the docs, and now it works.

kilchenmann (approved):

PR #59, opened 2021-04-14
fix: fix import ontology from salsah-export (DSP-1532)

lrosenth:
resolves DSP-1532


BalduinLandolt (approved):
could `knora/mylist.json` be removed and added to `.gitignore`? I don't think we need to have it here.

Apart from that, looks good.

PR #69, opened 2021-07-14
docs(dsp-tools-xmlupload): add Warning section (DSP-1693)

jnussbaum:
It is very important to know about this detail. Already two people ran into this, and it costed a lot of time to both of us. This warning is definitively needed.

### DIFF
--- a/docs/dsp-tools-xmlupload.md
+++ b/docs/dsp-tools-xmlupload.md
@@ -169,13 +169,18 @@ The `<resource>` element contains a property element (e.g. `<text-prop>`) for ea
 describing the resource. The property element itself contains one or several value elements (e.g. `<text>`) and must
 have an attribute `name` with the name of the property as defined in the project specific ontology.
 
-Example for a property element of type text (`<text-prop>`) with the value element `<text>`:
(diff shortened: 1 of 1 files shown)


docs/dsp-tools-xmlupload.md
```diff
@@ -169,13 +169,18 @@ The `<resource>` element contains a property element (e.g. `<text-prop>`) for ea
...
 ```xml
 <text-prop name=":hasTranslation">
    <text encoding="utf8">Dies ist eine Übersetzung</text>
+   <text encoding="utf8">Und eine andere Übersetzung</text>
 </text-prop>
 ```
+
+| ⚠ Look out  |
+|:----------|
+| In case of a cardinality 1-n, you must **NOT** create multiple `<text-prop>` tags, but multiple `<text>` tags in one single `<text-prop>`! |
```
irinaschubert, line 183:
```suggestion
| In case of a cardinality 1-n, don't use multiple `<text-prop>` tags, but multiple `<text>` tags inside `<text-prop>`.|
```

irinaschubert: I would put it more neutrally.

irinaschubert (approved):

PR #70, opened 2021-07-20
feat(verbose xml upload): use v option to print verbose output in XML upload (DSP-1797)

irinaschubert:
resolves DSP-1797

### DIFF
--- a/knora/dsp_tools.py
+++ b/knora/dsp_tools.py
@@ -19,7 +19,7 @@ sys.path.append(os.path.dirname(os.path.realpath(__file__)))
 
 def program(args: list) -> None:
     """
-    The program parses the command line arguments and calls the actions
+    The program parses the command line arguments and calls the requested action
 
     Args:
         args: list of arguments passed by the user from the command line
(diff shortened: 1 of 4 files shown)


knora/dsp_tools.py
```diff
@@ -32,8 +43,10 @@ def program(args):
     parser_create.add_argument("-s", "--server", type=str, default="http://0.0.0.0:3333", help="URL of the DSP server")
     parser_create.add_argument("-u", "--user", default="root@example.com", help="Username for DSP server")
     parser_create.add_argument("-p", "--password", default="test", help="The password for login")
-    parser_create.add_argument("-V", "--validate", action='store_true', help="Do only validation of JSON, no upload of the ontology")
-    parser_create.add_argument("-L", "--listfile", type=str, default="lists.json", help="Name of list node informationfile")
+    parser_create.add_argument("-V", "--validate", action='store_true',
+                               help="Do only validation of JSON, no upload of the ontology")
```
BalduinLandolt, line 47: I'm not a huge fan of these line breaks. I'm guessing PyCharm is following pep8 here, but I think in terms of line length, pep8 is very anachronistic. (On modern screens, you can read more than 80 characters per line.)  
So I tend to bump max line length to normally 130... but it's up to you, as long as we don't have any style guidelines. :)

BalduinLandolt (approved):
looks good

knora/dsp_tools.py
irinaschubert: totally agree, I changed it to 130 characters

PR #82, opened 2021-08-16
fix(import): fix import error when starting script directly (DSP-1869)

irinaschubert:
resolves DSP-1869

### DIFF
--- a/knora/dsplib/utils/expand_all_lists.py
+++ b/knora/dsplib/utils/expand_all_lists.py
@@ -1,6 +1,6 @@
-from typing import List, Dict
+from typing import Dict, List
 
-from knora.dsplib.utils.excel_to_json_lists import prepare_list_creation, make_json_list_from_excel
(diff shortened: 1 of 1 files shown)


BalduinLandolt (approved):
were you able to test it? From the code only, I find it impossible to say if it's going to work when published on PyPI...

irinaschubert:
> were you able to test it? From the code only, I find it impossible to say if it's going to work when published on PyPI...

I testet it on my laptop with `make run` which makes a distribution package and installs it from source. Johannes and I also testet it on his laptop. In both cases it worked well. Do you know any other test I could do?

BalduinLandolt:
> I testet it on my laptop with `make run` which makes a distribution package and installs it from source. Johannes and I also testet it on his laptop. In both cases it worked well. Do you know any other test I could do?

not really. have you run `make run` with a clean new virtual environment activated? if that worked, I guess that should be good enough.

PR #88, opened 2021-09-09
feat(excel-to-resources): create resources from excel (DSP-1576)

irinaschubert:
resolves DSP-1576


docs/dsp-tools-create.md
```diff
@@ -728,7 +729,9 @@ Represents a date. It's a string with the format `calendar:start:end`
 Please note that the DateValue is an extremely flexible data type. It can represent an exact date or a date with a given
 uncertainty, and the date can be given in several calendars (currently the Gregorian and the Julian calendars are supported, with
 the Jewish and Islamic coming soon). Internally, a date is always represented as a start and end date. If start and end date
-match, it's an exact date. A value like "1893" will automatically be expanded to a range from January 1st 1893 to December 31st 1893.
+match, it's an exact date. A value like "1893" will automatically be expanded to a range from January 1st 1893 to December 31st
+
+1893.
```
BalduinLandolt, line 734: why the line breaks?

docs/dsp-tools-excel.md
```diff
@@ -1,29 +1,50 @@
...
-the Excel file's file name after an underline and before the file extension, p.ex. `liste_de.xlsx` would be considered a list with 
-German (`de`) labels, `list_en.xlsx` a list with English (`en`) labels. The language has to be a valid ISO 639-1 or ISO
-639-2 language code.
+
+With dsp-tools a JSON list can be created from one or several Excel files. The list can then be inserted into a JSON ontology and
+uploaded to a DSP server. The expected format of the Excel files is described [here](./dsp-tools-create.md#lists-from-excel). It
+is possible to create multilingual lists. In this case, a separate Excel file has to be created for each language. The data has to
+be in the first worksheet of the Excel file(s). It is important that all the Excel lists have the same structure. So, the
+translation(s) of a label in one Excel sheet has to be in the exact same cell (i.e. with the same cell index) in its own Excel
+sheet.
```
BalduinLandolt, line 41: are those linebreak changes needed?  
they can't be hidden by github, so they clutter the diffs, without changing anything at the output.

BalduinLandolt: (this applies to multiple places, I won't comment it everywhere)

knora/dsplib/utils/knora-schema-resources-only.json
```diff
@@ -0,0 +1,132 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema-resources-only#",
+  "$id": "http://knora.org/pyknora/ontology/knora-schema-resources-only.json",
```
BalduinLandolt, line 3: what exactly is this? the link isn't pointing anywhere, and as a namespace it doesn't make much sense, as we don't use `pyknora` anywhere.

BalduinLandolt (approved):
I'm not so fond of some of the reformatting in the markdown... but that's cosmetics.

Apart from that, it looks good, thanks for your tireless work!

docs/dsp-tools-create.md
irinaschubert: I corrected this in the documentation PR.

knora/dsplib/utils/knora-schema-resources-only.json
irinaschubert: As far as I understood this is used as a unique identifier (https://json-schema.org/understanding-json-schema/basics.html#declaring-a-unique-identifier) and doesn't have to point anywhere. It seems to be good practice to use a URL and I've just copied the schema with pyknora etc. from other schemas used in dsp-tools.

However, I removed the `07` in the schema reference.

BalduinLandolt: fair enough, though the URL choice seems weird to me.

Are you sure you can leave out the draft number? also, `schema-resources-only` I have never seen before either. Does validation really work as it should, like this?

BalduinLandolt: the latest schema draft seems to be `"$schema": "https://json-schema.org/draft/2020-12/schema"`

irinaschubert: I see, I thought this was some sort of internal versioning from knora but I see now that this is the official json-schema versioning. So, you are right, it's correct to use the form it had before. I will change it back.

irinaschubert: (I use draft-07 instead of the newest schema, as I don't know what has changed since - I will create a new yt task to once check and update all the schemas written for dsp-tools)

BalduinLandolt: sounds good

PR #95, opened 2021-09-16
style(code-style): add code style configuration (DSP-1906)

irinaschubert:
resolves DSP-1906


.editorconfig
```diff
@@ -2,9 +2,3 @@
 [*]
 end_of_line = lf
 insert_final_newline = true
-
-[*.py]
-indent_style = space
-indent_size = 4
```
BalduinLandolt, line 8: we could leave those in under `[*]`, imo this should be correct pretty much everywhere. And if not, there can be further specification. But like this it would also apply to json, xml, etc.

BalduinLandolt (approved):
looks good

.editorconfig
irinaschubert: Good point

PR #101, opened 2021-09-22
feat(schema): add error codes for validation (DSP-1902)

irinaschubert:
resolves DSP-1902

### DIFF
--- a/knora/dsp_tools.py
+++ b/knora/dsp_tools.py
@@ -133,7 +133,10 @@ def program(user_args: list[str]) -> None:
                              dump=args.dump)
         else:
             if args.validate:
-                validate_ontology(args.datamodelfile)
+                if validate_ontology(args.datamodelfile):
+                    exit(0)
+                else:
+                    exit(1)
             else:
                 create_ontology(input_file=args.datamodelfile,
                                 lists_file=args.listfile,
--- a/knora/dsp_tools.py
+++ b/knora/dsp_tools.py
@@ -117,15 +117,12 @@ def program(user_args: list[str]) -> None:
 
     if not hasattr(args, 'action'):
         parser.print_help(sys.stderr)
-        exit(1)
+        exit(0)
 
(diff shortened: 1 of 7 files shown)


knora/dsp_tools.py
```diff
@@ -133,7 +133,10 @@ def program(user_args: list[str]) -> None:
                              dump=args.dump)
         else:
             if args.validate:
-                validate_ontology(args.datamodelfile)
+                if validate_ontology(args.datamodelfile):
+                    exit(0)
+                else:
+                    exit(1)
```
BalduinLandolt, line 139: makes sense like this.

Additionally, I think, similar handling should happen on:
- line 120 (exit code `1` not  `0`)
- line 125?

Is it correct that there is no `xml_upload` with validation only?

knora/dsplib/utils/onto_validate.py
```diff
@@ -45,7 +45,7 @@ def validate_ontology(input_file_or_json: Union[str, Dict, os.PathLike]) -> bool
     try:
         validate(instance=data_model, schema=schema)
     except jsonschema.exceptions.ValidationError as err:
-        print(err)
+        print('Data model did not pass validation. The error message is:', err.message)
         return False
```
BalduinLandolt, line 49: I think this could be simplified to just `validate(instance=data_model, schema=schema)`: Either it passes quetly, or it raises a validation error, which will cause it to exit with non-zero exit code

BalduinLandolt (approved):
note my two comments. If they make sense to you, please change it accordingly. Otherwise, looks good.

knora/dsp_tools.py
irinaschubert: Makes sense, I was only focussing on the ontology and xml, but you are right of course. I added it on both places.

For `xml_upload` it's implemented a bit differently with a separate parameter to the function called `validate_only`.

PR #108, opened 2021-10-26
feat(id-to-iri): extend xmlupload to allow references to existing resources (DEV-60)

irinaschubert:
resolves DEV-60


knora/dsplib/utils/id_to_iri.py
```diff
@@ -0,0 +1,78 @@
...
+    if not os.path.isfile(xml_file):
+        print(f"File {xml_file} could not be found.")
+        exit(1)
+
+    if not os.path.isfile(json_file):
+        print(f"File {json_file} could not be found.")
+        exit(1)
+
+    # load JSON from provided json file to dict
+    mapping = json.load(open(json_file))
```
BalduinLandolt, line 37: `open()` should allways be called from a `with` block, to ensure closing the file, etc. Also I'd specify the mode and the encoding explicitely, otherwise it might break on windows.

knora/dsplib/utils/id_to_iri.py
```diff
@@ -0,0 +1,78 @@
...
+
+    resource_elements = tree.xpath("/knora/resource/resptr-prop/resptr")
+    for resptr_prop in resource_elements:
+        try:
+            value_before = resptr_prop.text
+            value_after = mapping[resptr_prop.text]
+            resptr_prop.text = value_after
+            if verbose:
+                print(f"Replaced internal ID '{value_before}' with IRI '{value_after}'")
+        except KeyError:
```
BalduinLandolt, line 60: KeyErrors can usually be avoided by replacing `c = a[b]` with `c = a.get(b)` and then checking `if not c: ...`. It's a matter of taste, as long as the exception gets caught, so it's up to you. But I personally prefer to keep exceptions to a minimum where possible.

knora/dsplib/utils/xml_upload.py
```diff
@@ -603,21 +611,42 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+            res_iri_lookup[resource.id] = instance.iri
+            print("Created resource:", instance.label, "(", resource.id, ") with IRI", instance.iri)
+
+        except BaseError as err:
+            failed_uploads.append(resource.id)
+            print("ERROR while trying to upload", resource.label, "(", resource.id, "). The error message was:", err.message)
+
+        except Exception as exception:
+            failed_uploads.append(resource.id)
+            print("ERROR while trying to upload", resource.label, "(", resource.id, "). The error message was:", str(exception))
```
BalduinLandolt, line 639: when using an f-sting rather than multiple arguments to the print method, explicit string casting is not necessary, which would probably read nicer. (Also, fewer quotation marks and commas.)

requirements.txt
```diff
@@ -1,14 +1,61 @@
...
-mkdocs-material
-rdflib
-lxml
-validators
-requests
-jsonschema
-rfc3987
-openpyxl
-pystrict
+zipp==3.6.0
```
BalduinLandolt, line 61: do we really have so many requirements?

BalduinLandolt (approved):
looks good, just some minor comments

knora/dsplib/utils/xml_upload.py
```diff
@@ -603,21 +611,42 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+            res_iri_lookup[resource.id] = instance.iri
+            print(f"Created resource '{instance.label}' ({resource.id}) with IRI '{instance.iri}'")
+
+        except BaseError as err:
+            failed_uploads.append(resource.id)
+            print(f"ERROR while trying to upload '{resource.label}' ({resource.id}). The error message was: {err.message}")
+
+        except Exception as exception:
+            failed_uploads.append(resource.id)
+            print(f"ERROR while trying to upload '{resource.label}' ({resource.id}). The error message was: {str(exception)}")
```
BalduinLandolt, line 639:
```suggestion
            print(f"ERROR while trying to upload '{resource.label}' ({resource.id}). The error message was: {exception}")
```

f-strings automatically call str() on any non-string object

BalduinLandolt (approved):
LGTM

PR #111, opened 2021-11-01
feat(excel-to-json): allow comments in class and property definitions

mrivoal:
This PR updates `dsp-tools`' code to support `rdfs:comment` in classes and properties definition in order to provide a better documentation for the data. 
It also updates accordingly the Excel templates used to generate classes and properties.


irinaschubert (approved):
LGTM

PR #113, opened 2021-11-16
chore(documentation): add missing documentation for excel2resources (DEV-144)

irinaschubert:
resolves DEV-144


mkdocs.yml
```diff
@@ -24,9 +24,13 @@ theme:
     accent: 'deep blue'
   feature:
     tabs: false
+
 extra_css:
   - 'assets/style/theme.css'
 
+markdown_extensions:
+    - attr_list
```
BalduinLandolt, line 32: what does that do? :)

BalduinLandolt (approved):
LGTM, I'm just curious about the `attr_list`

mkdocs.yml
irinaschubert: It allows to resize the image inside the md file. Like so:
`{ width=50% }`

PR #117, opened 2021-11-18
fix: update dsp-tools to work with API version 16.0.0

irinaschubert:
resolves DEV-246

### DIFF
--- a/knora/dsplib/models/listnode.py
+++ b/knora/dsplib/models/listnode.py
@@ -57,7 +57,7 @@ def list_creator(con: Connection, project: Project, parent_node: 'ListNode', nod
             name=node["name"],
             parent=parent_node
         )
-        if node.get('nodes') is not None:
+        if node.get('nodes'):
             newnode.children = list_creator(con, project, newnode, node['nodes'])
         nodelist.append(newnode)
     return nodelist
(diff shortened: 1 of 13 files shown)


irinaschubert:
Hi @BalduinLandolt , apparently in the newest API version there is a comment needed for every single node in a list (not just the root node as until now). So, if there is a missing comment in the ontology JSON, an empty comment ("en", "") is created. I think this is more user friendly than force the user to invent a comment for every single node. What do you think?

knora/dsplib/models/listnode.py
```diff
@@ -184,7 +184,7 @@ def __init__(self,
         self._project = project.id if isinstance(project, Project) else str(project) if project is not None else None
         self._id = str(id) if id is not None else None
         self._label = LangString(label)
-        self._comments = LangString(comments)
+        self._comments = LangString(comments) if comments is not None else None
```
BalduinLandolt, line 187: any reason you don't just write `LangString(comments) if comments else None`?

BalduinLandolt: also, technically this violates typing here: `self._comments` should be `Optional[LangString]`, not just `LangString`, if it can be `None`

BalduinLandolt: and on that note `Union[str, None]` should probably be replaced with `Optional[str]`

BalduinLandolt (approved):
Looks good. Typing issues can be for some other time.

BalduinLandolt (approved):

PR #121, opened 2021-11-23
fix: inconsistencies in groups and projects (DEV-261)

irinaschubert:
resolves DEV-261

### DIFF
--- a/knora/dsplib/utils/onto_create_ontology.py
+++ b/knora/dsplib/utils/onto_create_ontology.py
@@ -161,14 +161,16 @@ def create_ontology(input_file: str,
             group_ids: Set[str] = set()
 
             user_groups = user.get("groups")
+            # if "groups" is provided, add user to the group(s)
             if user_groups:
-                # if "groups" is provided, add user to the group(s)
                 for group_name in user_groups:
                     if verbose:
                         print(f"Add user to group: {group_name}")
-                    # group_name has the form [proj_shortname]:group_name or "SystemAdmin"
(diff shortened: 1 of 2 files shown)


knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -148,73 +148,100 @@ def create_ontology(input_file: str,
...
+            if user_projects:
+                for full_project_name in user_projects:
+                    if verbose:
+                        print(f"Add user to project: {full_project_name}")
+                    # full_project_name has the form [project_name]:member or [project_name]:admin
+                    # if project_name is omitted, the user is added to the current project
+                    tmp = full_project_name.split(':')
+
+                    if not len(tmp) == 2:
+                        print(f"ERROR Provided project name {full_project_name} for user {user.username} is not valid.")
```
BalduinLandolt, line 211: here you should probably add a `continue`, otherwise if `len(tmp)` is not 2, it would still attempt to do stuff, but probably crash or create nonsense

BalduinLandolt (changes requested):
just a small thing, but I think it needs fixing. Otherwise it looks good

BalduinLandolt (approved):

knora/dsplib/utils/onto_create_ontology.py
irinaschubert: Yes, very important indeed ;) Thanks for the hint, added a `continue`.

PR #122, opened 2021-11-24
fix(schema): list root node needs a comments object (DEV-61)

jnussbaum:
resolves DEV-61


irinaschubert (approved):
I also updated the documentation and added a test case in the test ontology. Otherwise it looks good, thanks for your effort!

PR #124, opened 2021-11-25
docs: update fully fleshed-out example (DEV-265)

irinaschubert:
resolves DEV-265


irinaschubert:
will do in separate pr's

PR #127, opened 2021-11-29
fix(permissions): use permissions in xml upload (DEV-178)

irinaschubert:
resolves DEV-178


knora/dsplib/models/bitstream.py
```diff
@@ -0,0 +1,37 @@
...
+    def permissions(self) -> Optional[Permissions]:
+        return self._permissions
+
+    def toJsonLdObj(self, action: Actions) -> Dict[str, Any]:
+        tmp = {}
+        if action == Actions.Create:
+            tmp["knora-api:fileValueHasFilename"] = self._value
+            if self._permissions:
+                tmp["knora-api:hasPermissions"] = self.permissions.toJsonLdObj()
+        return tmp
```
BalduinLandolt, line 37: for these kinds of classes, we should probably at some point start using python's `dataclass`

testdata/test-data.xml
```diff
@@ -6,50 +6,45 @@
        shortcode="4123"
        default-ontology="testonto">
 
-    <!-- permissions: see https://docs.knora.org/03-apis/api-v2/reading-user-permissions/ -->
+    <!-- permissions: see https://docs-api.dasch.swiss/02-knora-ontologies/knora-base/#permissions -->
```
BalduinLandolt, line 9: I think we should link to docs.dasch.swiss, not to the repo-specific documentations, right?

BalduinLandolt (approved):
looks good

testdata/test-data.xml
irinaschubert: You are right. I didn't even notice that there were also repo specific documentation sites...

knora/dsplib/models/bitstream.py
irinaschubert: Thanks for this very good suggestion. I don't do it in this PR though as it would be better to check/refactor all classes systematically.

PR #130, opened 2021-11-30
fix(create-ontology): within an ontology, references to the ontology itself are not possible (DEV-135)

BalduinLandolt:
resolves DEV-135  
resolves DEV-278


knora/dsplib/models/helpers.py
```diff
@@ -180,14 +180,14 @@ def __init__(self, context: Optional[Dict[str, str]] = None):
                 if self._context.get(cc[0]) is None:
                     self._context[cc[0]] = cc[1]
         else:
```
musicEnfanthen, line 182: This else branch duplicates the default settings already checked in line 171-181. Would it make sense to move the checks in line 171-181 outside the if-branch, so that it will be checked for any given context (and remove the else branch instead)? Like:

```
self._context = ContextType({})

if context:
    <snip>
    for prefix, onto in cleaned_input.items():
        self._context[prefix] = OntoInfo(onto[:-1], True) if onto.endswith('#') else OntoInfo(onto, False)

# we always want the base ontologies/prefixes included in the context
#
for cc in self.base_ontologies.items():
    if self._context.get(cc[0]) is None:
        self._context[cc[0]] = cc[1]

#
# we always want the knora ontologies/prefixes included in the context
#
for cc in self.knora_ontologies.items():
    if self._context.get(cc[0]) is None:
        self._context[cc[0]] = cc[1]
```

BalduinLandolt: you're absolutely right, will change that. Thanks! (I tried to focus [with limited success] on the issue actually addressed in this PR - I'm sure there is a near infinite amount of stuff that could be cleaned up...)

knora/dsplib/models/helpers.py
```diff
@@ -160,34 +160,22 @@ def __init__(self, context: Optional[Dict[str, str]] = None):
...
-                "rdfs": OntoInfo("http://www.w3.org/2000/01/rdf-schema", True),
-                "owl": OntoInfo("http://www.w3.org/2002/07/owl", True),
-                "xsd": OntoInfo("http://www.w3.org/2001/XMLSchema", True),
-                "knora-api": OntoInfo("http://api.knora.org/ontology/knora-api/v2", True),
-                "salsah-gui": OntoInfo("http://api.knora.org/ontology/salsah-gui/v2", True)
-            }
+            for prefix, onto in context.items():
+                self._context[prefix] = OntoInfo(onto.removesuffix('#'), onto.endswith('#'))
+        
+        # we always want the base ontologies/prefixes included in the context
```
irinaschubert, line 169: I'm not very fond of these "we" comments. Maybe it can be changed to somewhat more neutral?

knora/dsplib/models/helpers.py
```diff
@@ -160,34 +160,22 @@ def __init__(self, context: Optional[Dict[str, str]] = None):
...
+        for k, v in self.base_ontologies.items():
+            if not self._context.get(k):
+                self._context[k] = v
+        
+        # we always want the DSP-API ontologies/prefixes included in the context
```
irinaschubert, line 174: I'm not very fond of these "we" comments. Maybe it can be changed to somewhat more neutral?

irinaschubert (approved):
LGTM, I just looked at the code, shall I also test it locally?

knora/dsplib/models/helpers.py
BalduinLandolt: you're right

irinaschubert (approved):

PR #136, opened 2021-12-21
fix: restrict the creation of classes without cardinalities (DEV-305)

irinaschubert:
resolves DEV-305

Note: Because of circular dependencies in the creation of properties and resource classes with dsp-tools (properties have fields `object`/`subject` which point to resource classes and resource classes point to properties), I decided to not allow the creation of resource classes without direct cardinalities with dsp-tools. It is questionable anyway whether this case will ever exist, so I considered it not to be a limitation of the user.


knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -301,151 +303,189 @@ def create_ontology(input_file: str,
...
+                                                                          superclasses=super_classes,
+                                                                          label=res_label,
+                                                                          comment=res_comment).create(
+                        last_modification_date)
+                except BaseError as err:
+                    print(
+                        f"ERROR while trying to create resource class {res_name}. The error message was {err.message}")
+                except Exception as exception:
+                    print(
+                        f"ERROR while trying to create resource class {res_name}. The error message was {exception}")
```
BalduinLandolt, line 373: again, this code doesn't stop in case of an exception. Is that intended?

knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -301,151 +303,189 @@ def create_ontology(input_file: str,
...
+            for prop_class in prop_classes:
+                prop_name = prop_class.get("name")
+                prop_label = LangString(prop_class.get("labels"))
+                '''
+                get the super-property/ies if defined, valid forms are:
+                  - "prefix:super-property" : fully qualified name of property in another ontology. The prefix has to be 
+                    defined in the prefixes part.
+                  - "super-property" : super-property defined in the knora-api ontology
+                  - if omitted, "knora-api:hasValue" is assumed
+                '''
```
BalduinLandolt, line 393: IMO, tripple quotes should be only for "real" docstrings, not for regular comments amidst the code

knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -301,151 +303,189 @@ def create_ontology(input_file: str,
...
+                if prop_class.get("super"):
+                    super_props = list(map(lambda a: a if ':' in a else knora_api_prefix + a, prop_class.get("super")))
```
BalduinLandolt, line 395: this could be rewirtten as a nice little list comprehention, instead of `list(map(lambda))`

knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -301,151 +303,189 @@ def create_ontology(input_file: str,
...
+                prop_subject = prop_class.get("subject")
+                gui_element = prop_class.get("gui_element")
+                gui_attributes = prop_class.get("gui_attributes")
+                if gui_attributes and gui_attributes.get("hlist"):
+                    gui_attributes["hlist"] = "<" + list_root_nodes[gui_attributes["hlist"]]["id"] + ">"
+                prop_comment = prop_class.get("comment")
+                if prop_comment:
+                    prop_comment = LangString(prop_comment)
+                else:
+                    prop_comment = "no comment given"
```
BalduinLandolt, line 424: does this mean, that each property that's created without a comment, gets the string `"no comment given"` added as a comment? if so, that seems a bit unelegant, right?

knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -301,151 +303,189 @@ def create_ontology(input_file: str,
...
+                                                                           gui_attributes=gui_attributes,
+                                                                           comment=prop_comment).create(
+                        last_modification_date)
+                except BaseError as err:
+                    print(
+                        f"ERROR while trying to create property class {prop_name}. The error message was: {err.message}"
+                    )
+                except Exception as exception:
+                    print(
+                        f"ERROR while trying to create property class {prop_name}. The error message was: {exception}")
```
BalduinLandolt, line 446: again, this continues even in case of an error

knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -301,151 +303,189 @@ def create_ontology(input_file: str,
...
+                            gui_order=card_info.get("gui_order"),
+                            last_modification_date=last_modification_date)
+                    except BaseError as err:
+                        print(
+                            f"ERROR while trying to add cardinality {prop_id} to resource class {res_class.get('name')}."
+                            f"The error message was {err.message}")
+                    except Exception as exception:
+                        print(
+                            f"ERROR while trying to add cardinality {prop_id} to resource class {res_class.get('name')}."
+                            f"The error message was {exception}")
```
BalduinLandolt, line 489: as above

BalduinLandolt (approved):
I'm not sure the error handling works as intended right now. Please verify this - and if so, it's good to merge. (I also commented on some stylistics, feel free to ignore those)

PR #140, opened 2022-01-05
chore: improve ontology schema and extend tests (DEV-313)

irinaschubert:
resolves DEV-313
resolves DEV-257

I removed anything-onto.json from testdata and tests because it was no longer valid (probably already some time ago). So, I extended test-onto.json a bit and used it instead. In that occasion I also extended the test for the `dsp-tools get` command. The actual task of DEV-313 was an extension to the ontology schema to validate the gui_element and gui_attributes properties and their values.

Besides that, this PR also resolves DEV-257: added the `subject` attribute to the documentation.


test/e2e/test_tools.py
```diff
@@ -29,62 +30,133 @@ def tearDown(self) -> None:
...
+            list_name = test_onto_list.get('name')
+            if list_name == 'testlist':
+                test_list_out = test_onto_list
             elif list_name == 'notUsedList':
-                not_used_list_out = anything_list
-            elif list_name == 'treelistroot':
-                tree_list_root_out = anything_list
+                not_used_list_out = test_onto_list
+            elif list_name == 'my-list-from-excel':
+                excel_list_out = test_onto_list
```
BalduinLandolt, line 149: this block is quite an eye sore... ;)

can I suggest the following:

```python

lists = anything_onto['project']['lists']
test_list = next((l for l in lists if l['name'] == 'testlist'), None)
not_used_list = next((l for l in lists if l['name'] == 'notUsedList'), None)
excel_list = next((l for l in lists if l['name'] == 'my-list-from-excel'), None)


lists_out = anything_onto_out['project']['lists']
test_list_out = next((l for l in lists_out if l['name'] == 'testlist'), None)
not_used_list_out = next((l for l in lists_out if l['name'] == 'notUsedList'), None)
excel_list_out = next((l for l in lists_out if l['name'] == 'my-list-from-excel'), None)

```

unless I'm overlooking something, it should do the exact same thing, and I think it's a bit easier on the eye

BalduinLandolt (approved):
looks good!  
one cosmetic suggestion, but nothing to complain

test/e2e/test_tools.py
irinaschubert: Nice suggestion, thanks!

BalduinLandolt (approved):
looks good

PR #145, opened 2022-01-14
docs: add isPartOf and seqnum properties to documentation (DEV-216)

irinaschubert:
resolves DEV-216

### DIFF
--- a/docs/dsp-tools-create-ontologies.md
+++ b/docs/dsp-tools-create-ontologies.md
@@ -635,11 +635,11 @@ Represents a node of a (possibly hierarchical) list
 
 - `Radio`: A GUI element for _ListValue_. A set of radio buttons. This works only with flat lists.
     - _gui_attributes_:
-        - `hlist=<list-name>` (required): The reference of a [list](#lists) root node
+        - `hlist=<list-name>` (required): The reference of a [list](./dsp-tools-create.md#lists) root node
 - `List`: A GUI element for _ListValue_. A list of values to select one from. This GUI element should be chosen for
   hierarchical lists or flat lists that could be expanded to hierarchical lists in the future.
     - _gui_attributes_:
-        - `hlist=<list-name>` (required): The reference of a [list](#lists) root node
+        - `hlist=<list-name>` (required): The reference of a [list](./dsp-tools-create.md#lists) root node
 
 *Example:*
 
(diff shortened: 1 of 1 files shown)


jnussbaum:
Thanks for your work. Here some questions/feedback: 
1. in rosetta, I was able to make 3 pages that belong to the same book, with the pagenums 2, 2, 3. And I made 1 page belonging to another book, with the pagenum 0. So it is possible to:
     - start with 0
     - skip 1
     - double a pagenum 
     - is this behaviour correct? 
 2. practical use of seqnum: I guess that the formulation "a client is able to leaf through the parts of a compound object" in the docs is rather theoretical than practical? I don't see a practical use.
 3. delete the following on line 672: "(how to derive a resource or property from another one is not part of this documentation)"
     - rationale: it is described in paragraphs #super and #super_1.
 4. in the following sentence, the word "required" is inaccurate. it is in fact possible to omit the seqnum: line 701: "In case of 
resources that are of type `StillImageRepresentation`, an additional property derived from `seqnum` with object `IntValue` is required."

irinaschubert:
> 

Hi Johannes, thanks for the review.
1. Yes, this is correct. It is (right now) the user's responsibility to set the right `pagenum`s (no double numbers, starting wherever he wants, skipping numbers...). If we want to expand that, it needs to go through product management for priorisation.
2. The practical use is: you have a book (compound object) with several pages (partOf objects). Defining pages as `isPartOf` the book, the book can be shown as one object and from there one can leaf through the book. You can try it out with the rosetta project and the `Book with Images` (book_obj_1) and its `ImagePartOfABook` parts/pages.
3. + 4.: I will adjust that

jnussbaum:
Thanks for your answers.
1. That's okay.
2. That's okay.
3. OK
4. It is the very last word of the sentence. The sentence says that the seqnum is required, but in fact, you can omit it.
     - line 701: "In case of resources that are of type StillImageRepresentation, an additional property derived from seqnum with object IntValue is **_required_**."

irinaschubert:
Ah, now I see/understand. Well, yes, it is not strictly required but without it the compound object is not displayed as such, so using it without `seqnum` doesn't make sense. So, I leave it like that.

jnussbaum (approved):

PR #155, opened 2022-02-11
refactor: excel to json list (DEV-431)

jnussbaum:
resolves DEV-431
resolves also DEV-137

### DIFF
--- a/knora/dsplib/utils/excel_to_json_lists.py
+++ b/knora/dsplib/utils/excel_to_json_lists.py
@@ -7,6 +7,7 @@ import re
 import unicodedata
 
 import jsonschema
+import openpyxl.worksheet.worksheet
 from jsonschema import validate
 from openpyxl import load_workbook
 
@@ -14,8 +15,14 @@ list_of_lists = []
 cell_names = []
 
 
-def get_values_from_excel(excelfiles: list[str], base_file: str, parentnode: {}, row: int, col: int,
-                          preval: list[str]) -> int:
+def get_values_from_excel(
+    excelfiles: dict[str, openpyxl.worksheet.worksheet.Worksheet],
+    base_file: dict[str, openpyxl.worksheet.worksheet.Worksheet],
+    parentnode: dict,
+    row: int,
+    col: int,
+    preval: list[str]
+) -> int:
     """
     This function calls itself recursively to go through the Excel files. It extracts the cell values and creates the JSON list
     file.
@@ -29,32 +36,35 @@ def get_values_from_excel(excelfiles: list[str], base_file: str, parentnode: {},
         preval: List of previous values, needed to check the consistency of the list hierarchy
 
     Returns:
-        int: Row index for the next loop (actual row index minus 1)
+        int: Row index for the next loop (current row index minus 1)
     """
     nodes = []
     currentnode = {}
-    wb = load_workbook(filename=base_file, read_only=True)
-    worksheet = wb.worksheets[0]
-    cell = worksheet.cell(column=col, row=row)
+    base_file_ws = list(base_file.values())[0]
+    cell = base_file_ws.cell(column=col, row=row)
 
     if col > 1:
         # append the cell value of the parent node (which is one value to the left of the actual cell) to the list of
         # previous values
-        preval.append(worksheet.cell(column=col - 1, row=row).value)
+        preval.append(base_file_ws.cell(column=col - 1, row=row).value)
 
     while cell.value:
         # check if all predecessors in row (values to the left) are consistent with the values in preval list
         for idx, val in enumerate(preval[:-1]):
-            if val != worksheet.cell(column=idx + 1, row=row).value:
+            if val != base_file_ws.cell(column=idx + 1, row=row).value:
                 print(
-                    f'Inconsistency in Excel list: {val} not equal to {worksheet.cell(column=idx + 1, row=row).value}')
+                    f'Inconsistency in Excel list: {val} not equal to {base_file_ws.cell(column=idx + 1, row=row).value}')
                 quit()
 
         # loop through the row until the last (furthest right) value is found
-        if worksheet.cell(column=col + 1, row=row).value:
-            row = get_values_from_excel(excelfiles=excelfiles, base_file=base_file, parentnode=currentnode, col=col + 1,
-                                        row=row,
-                                        preval=preval)
+        if base_file_ws.cell(column=col + 1, row=row).value:
+            row = get_values_from_excel(
+                excelfiles=excelfiles,
+                base_file=base_file,
+                parentnode=currentnode,
+                col=col + 1,
+                row=row,
+                preval=preval)
 
         # if value was last in row (no further values to the right), it's a node, continue here
         else:
@@ -81,11 +91,7 @@ def get_values_from_excel(excelfiles: list[str], base_file: str, parentnode: {},
             labels_dict = {}
 
             # read label values from the other Excel files (other languages)
-            for filename_other_lang in excelfiles:
-                wb_other_lang = load_workbook(filename=filename_other_lang, read_only=True)
-                ws_other_lang = wb_other_lang.worksheets[0]
-
-                lang = os.path.splitext(filename_other_lang)[0].split('_')[-1]
+            for lang, ws_other_lang in excelfiles.items():
                 labels_dict[lang] = ws_other_lang.cell(column=col, row=row).value
 
             # create current node from extracted cell values and append it to the nodes list
@@ -97,7 +103,7 @@ def get_values_from_excel(excelfiles: list[str], base_file: str, parentnode: {},
 
         # go one row down and repeat loop if there is a value
         row += 1
-        cell = worksheet.cell(column=col, row=row)
+        cell = base_file_ws.cell(column=col, row=row)
 
     if col > 1:
         preval.pop()
@@ -108,7 +114,7 @@ def get_values_from_excel(excelfiles: list[str], base_file: str, parentnode: {},
     return row - 1
 
 
-def make_json_list_from_excel(rootnode: {}, excelfiles: list[str]) -> None:
+def make_json_list_from_excel(rootnode: dict, excelfile_names: list[str]) -> None:
     """
     Reads Excel files and makes a JSON list file from them. The JSON can then be used in an ontology that is uploaded to the
     DaSCH Service Platform.
@@ -119,21 +125,28 @@ def make_json_list_from_excel(rootnode: {}, excelfiles: list[str]) -> None:
 
     Returns:
         None
-
     """
     # Define starting point in Excel file
     startrow = 1
     startcol = 1
 
-    # Check if English file is available and take it as base file, take last one from list of Excel files if English
-    # is not available. The node names are later  derived from the labels of the base file.
-    base_file = ''
-
-    for filename in excelfiles:
-        base_file = filename
+    # Check if English file is available and take it as base file. Take last one from list of Excel files if English
+    # is not available. The node names are later derived from the labels of the base file.
+    for filename in excelfile_names:
         if '_en.xlsx' in os.path.basename(filename):
-            base_file = filename
-            break
+            lang = 'en'
+            ws = load_workbook(filename, read_only=True).worksheets[0]
+            base_file = {lang: ws}
+    if not base_file:
+        lang = os.path.splitext(filename)[0].split('_')[-1]
+        ws = load_workbook(excelfile_names[-1], read_only=True).worksheets[0]
+        base_file = {lang: ws}
+
+    excelfiles = {}
+    for f in excelfile_names:
+        lang = os.path.splitext(f)[0].split('_')[-1]
+        ws = load_workbook(f, read_only=True).worksheets[0]
+        excelfiles[lang] = ws
 
     get_values_from_excel(excelfiles=excelfiles, base_file=base_file, parentnode=rootnode, row=startrow, col=startcol,
                           preval=[])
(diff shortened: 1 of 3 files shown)


jnussbaum:
Mir schien die Architektur von `excel_to_json_lists.py` nicht gut genug dokumentiert zu sein, und an einigen Stellen verbesserungswürdig.

Deshalb habe ich versucht, mehr Ordnung hineinzubringen, was mir auch einigermassen gut gelungen ist. Bloss eine Sache konnte ich nicht zufriedenstellend lösen: Der rekursive Aufruf von `get_values_from_excel()` dient (oberflächlich betrachtet) nur dazu, den Maximalwert von `row` zu ermitteln. Der zweite Rückgabewert `parentnode` wird ja verworfen. Doch in Tat und Wahrheit gibt es hier einen Seiteneffekt, den ich gerne eliminieren würde: 

In der Funktion `get_values_from_excel()` wird der mitgegebene Parameter `parentnode` verändert und dann zurückgegeben. Im rekursiven Funktionsaufruf wird also `currentnode` verändert, was aber verschleiert wird dadurch, dass der zweite Rückgabewert verworfen wird.

Um diese Funktionsweise expliziter zu machen, ohne effektiv eine Änderung im Programm hervorzurufen, möchte ich eigentlich folgende Änderung vornehmen:
 - Von `parentnode` wird eine Kopie verändert+zurückgegeben, und der Rückgabewert wird dann `currentnode` zugewiesen statt verworfen. 

Sollte semantisch dasselbe sein wie zuvor. Leider funktioniert es dann nicht mehr. Im letzten Commit (693eb11) ist sichtbar, was ich diese Verschlimmbesserung rückgängig machen musste. Was habe ich falsch gemacht? Wie könnte man es besser machen?

knora/dsplib/utils/excel_to_json_lists.py
```diff
@@ -216,36 +248,34 @@ def make_root_node_from_args(excelfiles: list[str], listname_from_args: str, com
...
-
-    # if an english list is available use its label as listname
-    if listname_en:
-        listname = listname_en
+            listname_en = lang_specific_listname
 
-    # if the user provided a listname use it
-    if listname_from_args:
+    # the listname is taken from the following sources, with descending priority
+    if isinstance(listname_from_args, str) and listname_from_args != '':
```
irinaschubert, line 271: You don't have to check if `listname_from_args` is a string as this is a method parameter defined as string. If you use `if listname_from_args` (as it was before) it returns False in case of an empty string.

knora/dsplib/utils/excel_to_json_lists.py
```diff
@@ -216,36 +248,34 @@ def make_root_node_from_args(excelfiles: list[str], listname_from_args: str, com
...
         listname = listname_from_args
+    elif listname_en != '':
```
irinaschubert, line 273:
```suggestion
    elif listname_en:
```

knora/dsplib/utils/excel_to_json_lists.py
```diff
@@ -259,65 +289,66 @@ def validate_list_with_schema(json_list: str) -> bool:
...
 
-def prepare_list_creation(excelfolder: str, listname: str, comments: dict):
+def prepare_list_creation(
+    excelfolder: str,
+    listname: Optional[str],
+    comments: dict[str, Any]
+) -> tuple[dict[str, Any], list[str]]:
     """
-    Gets the excelfolder parameter and checks the validity of the files. It then makes the root node for the list.
+    Create the list of Excel files that can be used to build a JSON list. Then, create the root node for the JSON list.
```
irinaschubert, line 312:
```suggestion
    Creates the list from Excel files that can be used to build a JSON list. Then, creates the root node for the JSON list.
```

knora/dsplib/utils/excel_to_json_lists.py
```diff
@@ -259,65 +289,66 @@ def validate_list_with_schema(json_list: str) -> bool:
...
 
     # create a list with all excel files from the path provided by the user
-    excel_files = [filename for filename in glob.iglob(f'{excelfolder}/*.xlsx') if
-                   not os.path.basename(filename).startswith('~$')]
+    excel_files = [filename for filename in glob.iglob(f'{excelfolder}/*.xlsx')
+                   if not os.path.basename(filename).startswith('~$')
+                   and os.path.isfile(filename)]
 
-    # check if all excel_files are actually files
+    # log the files that can be used
```
irinaschubert, line 340: IMO, "log" would mean writing it to a log file (which is not the case).

knora/dsplib/utils/excel_to_json_lists.py
```diff
@@ -1,145 +1,176 @@
...
-    wb = load_workbook(filename=base_file, read_only=True)
-    worksheet = wb.worksheets[0]
-    cell = worksheet.cell(column=col, row=row)
+    nodes: list[dict[str, Any]] = []
+    currentnode: dict[str, Any] = dict()
+    base_file_ws: Worksheet = list(base_file.values())[0]
+    cell: Cell = base_file_ws.cell(column=col, row=row)
+
+    for excelfile in excelfiles.values():
+        if any((not excelfile['A1'].value, excelfile['B1'].value)):
```
irinaschubert, line 50: I think it is a good idea to check if A1 has a value, but it seems somewhat random to check if there is also a vaue in B1 (what about the other cells in this row?). I don't know if it is easily possible to check if there is a value in any other cell besides A1. But that would certainly be more elegant.

irinaschubert (changes requested):
It's hard to tell what exactly went wrong in your attempt to remove the side effect as there is quite a lot of change in the whole code. I would need to invest more time to investigate the issue. It might have something to do with your method returning two values now instead of just `row`. This probably changes the whole behaviour of the method.

knora/dsplib/utils/excel_to_json_lists.py
jnussbaum: I see your point, but I think there is no alternative, because in openpyxl the row goes on until the maximum allowed in Excel, which is a lot. (In pandas, the row goes only as far as the longest row goes, so it is easy to check if all cells except the first are empty). 
In practice, I think that if B1 is empty, it is almost granted that the A1 is the only value in this row.

knora/dsplib/utils/excel_to_json_lists.py
```diff
@@ -216,36 +248,34 @@ def make_root_node_from_args(excelfiles: list[str], listname_from_args: str, com
...
-        # check if language code is valid
         if not check_language_code(lang_code):
-            print('Invalid language code is used. Only language codes from ISO 639-1 and ISO 639-2 are accepted.')
+            print(f'Invalid language code "{lang_code}" is used. Only language codes from ISO 639-1 ',
+                  f'and ISO 639-2 are accepted.')
             quit()
 
-        rootnode_labels_dict[lang_code] = label
-        listname = label
+        lang_code_2_listname[lang_code] = lang_specific_listname
```
jnussbaum, line 265:
```suggestion
        lang_code_2_listname[lang_code] = lang_specific_listname
        listname_from_lang_code[lang_code] = lang_specific_listname
```

knora/dsplib/utils/excel_to_json_lists.py
```diff
@@ -216,36 +248,34 @@ def make_root_node_from_args(excelfiles: list[str], listname_from_args: str, com
...
+    # the listname is taken from the following sources, with descending priority
+    if isinstance(listname_from_args, str) and listname_from_args != '':
         listname = listname_from_args
+    elif listname_en != '':
+        listname = listname_en
+    else:
+        listname = lang_specific_listname
 
-    rootnode = {'name': listname, 'labels': rootnode_labels_dict, 'comments': comments}
+    rootnode = {'name': listname, 'labels': lang_code_2_listname, 'comments': comments}
```
jnussbaum, line 278:
```suggestion
    rootnode = {'name': listname, 'labels': lang_specific_listname, 'comments': comments}
```

knora/dsplib/utils/excel_to_json_lists.py
jnussbaum: And what happens if `isinstance(listname_from_args, str)` is not defined? It's type is `Optional[str]`. But you're right, my version is not so good...

irinaschubert: Optional means it is either a string or `None`. If it is `None`, `if listname_from_args` would return `False`.
Does that answer your question?

knora/dsplib/utils/excel_to_json_lists.py
irinaschubert: OK, if it is not easily possible to check for all the other cells in the row, we don't need to do it. Your solution is fine than.

knora/dsplib/utils/excel_to_json_lists.py
jnussbaum: yes, thanks. I changed it back.

knora/dsplib/utils/excel_to_json_lists.py
jnussbaum, line 340:
```suggestion
    # print the files that can be used
```

knora/dsplib/utils/expand_all_lists.py
```diff
@@ -1,37 +1,38 @@
...
 from knora.dsplib.utils.excel_to_json_lists import make_json_list_from_excel, prepare_list_creation
 
 
-def expand_lists_from_excel(data_model: dict[str, Any]) -> list[str]:
+def expand_lists_from_excel(
+    data_model: dict[str, dict[str, list[dict[str, Union[str, dict[str, Any]]]]]]
+) -> list[dict[str, Any]]:
     """
-    Gets all list definitions from a data model and expands them to JSON if they are only referenced via an Excel file
+    Get all list definitions from a data model and expand them to JSON if they are only referenced via an Excel file
```
jnussbaum, line 10:
```suggestion
    Gets all list definitions from a data model and expands them to JSON if they are only referenced via an Excel file
```

knora/dsplib/utils/expand_all_lists.py
```diff
@@ -1,37 +1,38 @@
...
         data_model: The data model (JSON) the lists are read from
 
     Returns:
         A list of all expanded lists. It can be added to the root node of an ontology as lists section.
     """
-    # create and add lists from Excel references to the ontology
-    lists = data_model["project"].get("lists")
 
-    if not lists:
+    if 'project' not in data_model or 'lists' not in data_model['project']:
```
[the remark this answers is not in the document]
jnussbaum: Because I cannot be sure that `data_model` is a valid JSON data model, I want to do all checks beforehand. Afterwards, I can go on.

knora/dsplib/utils/expand_all_lists.py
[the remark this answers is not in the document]
jnussbaum: ok, changed it back

knora/dsplib/utils/expand_all_lists.py
```diff
@@ -1,37 +1,38 @@
...
         return []
 
+    lists = data_model['project']['lists']
```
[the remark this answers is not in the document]
jnussbaum: I see, but I like to have a clear separation between all tests (the extensive testing on line 19 you wondered about) and the straight access (line 22). 

If I use the construction with `.get()` thoroughly, it would need to be `lists = data_model.get('project').get('lists')`. But then, the first get could return `None`.

knora/dsplib/utils/expand_all_lists.py
```diff
@@ -1,37 +1,38 @@
...
     new_lists = []
+
     for rootnode in lists:
-        nodes = rootnode["nodes"]
+        nodes = rootnode['nodes']
         # check if the folder parameter is used
-        if rootnode.get("nodes") and isinstance(nodes, dict) and nodes.get("folder"):
+        if nodes and isinstance(nodes, dict) and nodes.get("folder"):
```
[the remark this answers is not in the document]
jnussbaum: thanks for the hint, it was really inconsistent. I improved it.

knora/dsplib/utils/expand_all_lists.py
```diff
@@ -1,37 +1,38 @@
...
             # get the Excel files from the folder and create the rootnode of the list
-            excel_folder = nodes["folder"]
-            rootnode, excel_files = prepare_list_creation(excel_folder, rootnode.get("name"), rootnode.get("comments"))
-
+            prepared_rootnode, excel_files = prepare_list_creation(
+                nodes['folder'], str(rootnode['name']), rootnode['comments']
```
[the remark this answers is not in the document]
jnussbaum: Yes, I expanded the checks. 

I had used `str()` to make Mypy happy. But I didn't do it in a good way. I think it's better now, after I implemented your feedback.

knora/dsplib/utils/excel_to_json_lists.py
jnussbaum: thanks for the hint, changed it back

knora/dsplib/utils/expand_all_lists.py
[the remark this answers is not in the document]
irinaschubert: OK, I don't really see why this is needed here. Shouldn't this be checked somewhere else as it has nothing to do with the lists?

knora/dsplib/utils/expand_all_lists.py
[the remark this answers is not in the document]
irinaschubert: Yes that's true, you would have to take it apart into two operations.

irinaschubert (approved):
LGTM

PR #158, opened 2022-02-18
feat: make comments optional for new properties and resources (DEV-502)

jnussbaum:
resolves DEV-502

### DIFF
--- a/testdata/test-onto.json
+++ b/testdata/test-onto.json
@@ -22,6 +22,10 @@
                 "labels": {
                     "en": "Testlist"
                 },
+                "comments": {
+                    "en": "no comment",
+                    "de": "kein Kommentar"
+                },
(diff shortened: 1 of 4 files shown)


testdata/test-onto.json
```diff
@@ -31,15 +31,12 @@
...
-                        },
-                        "comments": {
-                            "en": "no comment"
                         }
                     },
                     {
                         "name": "b",
                         "labels": {
-                            "en": "b_label with no comment"
+                            "en": "b_label with a comment"
```
irinaschubert, line 39: "no" was correct, wasn't it? This node doesn't have a comment.

irinaschubert (approved):
LGTM just one minor typo issue. I assume you've tested it with an example ontology both for resources and properties?

PR #164, opened 2022-02-24
feat: improve prefix handling in json ontos (DEV-572)

jnussbaum:
resolves DEV-572

### DIFF
--- a/docs/dsp-tools-create-ontologies.md
+++ b/docs/dsp-tools-create-ontologies.md
@@ -905,5 +905,5 @@ it is necessary to reference entities that are defined elsewhere. The following 
   [prefixes](dsp-tools-create.md#prefixes-object) and the prefix must be identical to the ontology's `name`.  
 - A different ontology defined in the same file: Within one data model file, multiple ontologies can be defined.
   These will be created in the exact order they appear in the `ontologies` array. Once an ontology has been created,
-  it can be referenced by the following ontologies, provided it is added in the
-  [prefixes](dsp-tools-create.md#prefixes-object).
+  it can be referenced by the following ontologies via its name: `first-onto:hasName`. It is not necessary to add 
+  `first-onto` to the prefixes.
--- a/docs/dsp-tools-create.md
+++ b/docs/dsp-tools-create.md
@@ -66,8 +66,8 @@ full qualified IRI each time it is used. So, instead of writing a property calle
 }
 ```
 
-Note that prefixes can be defined for the ontologies defined in this file, but this is only necessary if the ontology
-needs to be referred to explicitly by another ontology within the same file.
+It is not necessary to define prefixes for the ontologies that are defined in this file. Ontologies in the same
+file can refer to each other via their name. See also [here](./dsp-tools-create-ontologies.md#referencing-ontologies).
 
 ### "$schema" object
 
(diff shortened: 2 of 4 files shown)


jnussbaum:
Right now, the second step described in Jira works only because the rosetta prefix ist defined with `"rosetta": "http://0.0.0.0:3333/ontology/082E/rosetta/v2"`. But it should also be possible to define it as `"http://api.test.dasch.swiss/ontology/082E/rosetta/v2"`. This is not possible yet, and the API protests:

```ERROR while trying to create property class hasLinkToOtherOnto. The error message was: KNORA-ERROR: status code=400
Message:{"knora-api:error":"org.knora.webapi.exceptions.BadRequestException: Invalid knora-api:objectType: http://api.test.dasch.swiss/ontology/082E/rosetta/v2#Image2D","@context":{"knora-api":"http://api.knora.org/ontology/knora-api/v2#"}}```

irinaschubert:
> Right now, the second step described in Jira works only because the rosetta prefix ist defined with `"rosetta": "http://0.0.0.0:3333/ontology/082E/rosetta/v2"`. But it should also be possible to define it as `"http://api.test.dasch.swiss/ontology/082E/rosetta/v2"`. This is not possible yet, and the API protests:
> 
> ```
> ERROR while trying to create property class hasLinkToOtherOnto. The error message was: KNORA-ERROR: status code=400
> Message:{"knora-api:error":"org.knora.webapi.exceptions.BadRequestException: Invalid knora-api:objectType: http://api.test.dasch.swiss/ontology/082E/rosetta/v2#Image2D","@context":{"knora-api":"http://api.knora.org/ontology/knora-api/v2#"}}```
> ```

Hi Johannes, the reason why the API protests when trying to reference `http://api.test.dasch.swiss/ontology/082E/rosetta/v2` from within your local instance is that it does not know `api.test.dasch.swiss`. For the API only one API exists (itself). This makes sense (at least to me), as we don't want to mix ontologies from different servers.

You could upload rosetta3.json to `test.dasch.swiss` like this: 
`dsp-tools create -s https://api.test.dasch.swiss -u user@email.ch -p secretpassword rosetta3.json``

Does this answer your question?

irinaschubert (approved):

jnussbaum:
Thanks for the explanation, Irina. My intention was to make it working locally as well as on the test server. Is there no possibility for that?

irinaschubert:
> Thanks for the explanation, Irina. My intention was to make it working locally as well as on the test server. Is there no possibility for that?
> 
> Furthermore, I think we do want to allow ontologies from a different server, don't we? The other prefix `"dcterms": "http://purl.org/dc/terms/"` is also from a different server.

I don't think that there is a possibility right now for that. And I can't think of any use case for that either. What would be your use case to reference properties of different dsp-api servers?

The difference with `dcterms` is: it is not used in `object` (as your example) but in `super`. This is a huge difference - for `object` the API has to able to get information about that `object` (about the resource class `Image2D`, in your example) whereas it does not for `super` which just expresses a relationship. 

Actually, there is an issue for the support of external vocabularies in the actual sprint: https://dasch.atlassian.net/browse/DEV-512. I'm not sure what it is exactly about but it is assigned to me and I will clarify this next week ;)

PR #165, opened 2022-02-25
feat(xmlupload): allow circular references (DEV-577)

jnussbaum:
resolves DEV-577

### DIFF
--- a/knora/dsplib/models/resource.py
+++ b/knora/dsplib/models/resource.py
@@ -278,7 +278,6 @@ class ResourceInstance(Model):
     def create(self) -> 'ResourceInstance':
         jsonobj = self.toJsonLdObj(Actions.Create)
         jsondata = json.dumps(jsonobj, indent=4, separators=(',', ': '), cls=KnoraStandoffXmlEncoder)
-        # print("jsondata", jsondata)
         result = self._con.post(ResourceInstance.ROUTE, jsondata)
         newinstance = self.clone()
         newinstance._iri = result['@id']
--- a/knora/dsplib/models/resource.py
+++ b/knora/dsplib/models/resource.py
@@ -25,7 +25,7 @@ class KnoraStandoffXmlEncoder(json.JSONEncoder):
 
     def default(self, obj) -> str:
         if isinstance(obj, KnoraStandoffXml):
-            return '<?xml version="1.0" encoding="UTF-8"?>\n<text>' + obj.getXml() + '</text>'
+            return '<?xml version="1.0" encoding="UTF-8"?>\n<text>' + str(obj) + '</text>'
         elif isinstance(obj, OntoInfo):
             return obj.iri + "#" if obj.hashtag else ""
         return json.JSONEncoder.default(self, obj)
--- a/knora/dsplib/models/resource.py
+++ b/knora/dsplib/models/resource.py
@@ -275,7 +275,7 @@ class ResourceInstance(Model):
             tmp['@context'] = self.context
         return tmp
 
-    def create(self):
+    def create(self) -> 'ResourceInstance':
         jsonobj = self.toJsonLdObj(Actions.Create)
         jsondata = json.dumps(jsonobj, indent=4, separators=(',', ': '), cls=KnoraStandoffXmlEncoder)
         # print("jsondata", jsondata)
@@ -291,7 +291,8 @@ class ResourceInstance(Model):
         return self.fromJsonLdObj(con=self._con, jsonld_obj=result)
 
     def update(self):
-        pass
+        jsonobj = self.toJsonLdObj(Actions.Update)
+        ...
 
     def delete(self):
         pass
@@ -394,7 +395,7 @@ class ResourceInstanceFactory:
             return self._get_baseclass(gaga.superclasses)
         return None
 
-    def get_resclass(self, prefixedresclass: str) -> Type:
+    def get_resclass_type(self, prefixedresclass: str) -> Type:
         prefix, resclass_name = prefixedresclass.split(':')
         resclass = [x for x in self._ontologies[prefix].resource_classes if x.name == resclass_name][0]
         baseclass = self._get_baseclass(resclass.superclasses)
--- a/knora/dsplib/models/resource.py
+++ b/knora/dsplib/models/resource.py
@@ -290,8 +290,7 @@ class ResourceInstance(Model):
         return self.fromJsonLdObj(con=self._con, jsonld_obj=result)
 
     def update(self):
-        jsonobj = self.toJsonLdObj(Actions.Update)
-        ...
+        pass
 
     def delete(self):
         pass
--- a/knora/dsplib/models/sipi.py
+++ b/knora/dsplib/models/sipi.py
@@ -1,7 +1,6 @@
 import os
-
 import requests
-
+from typing import Any
 from .helpers import BaseError
 
 
@@ -30,7 +29,7 @@ class Sipi:
         self.sipi_server = sipi_server
         self.token = token
 
-    def upload_bitstream(self, filepath):
+    def upload_bitstream(self, filepath: str) -> dict[Any, Any]:
         """
         Uploads a bitstream to the Sipi server
 
@@ -45,5 +44,5 @@ class Sipi:
             req = requests.post(self.sipi_server + "/upload?token=" + self.token, files=files)
         on_api_error(req)
         print(f'Uploaded file {filepath}')
-        res = req.json()
+        res: dict[Any, Any] = req.json()
         return res
--- a/knora/dsplib/utils/onto_validate.py
+++ b/knora/dsplib/utils/onto_validate.py
@@ -117,7 +117,7 @@ def check_cardinalities_of_circular_references(data_model: dict[Any, Any]) -> bo
                         dependencies[resname][cardname].extend(targets)
 
     # iteratively purge dependencies from non-circular references
-    for i in range(30):
+    for _ in range(30):
         # remove targets that point to a resource that is not in dependencies,
         # remove cardinalities that have no targets
         for res, cards in dependencies.copy().items():
--- a/knora/dsplib/utils/onto_validate.py
+++ b/knora/dsplib/utils/onto_validate.py
@@ -7,7 +7,7 @@ import jsonpath_ng, jsonpath_ng.ext
 from ..utils.expand_all_lists import expand_lists_from_excel
 
 
-def validate_ontology(input_file_or_json: Union[str, dict[Any, Any], os.PathLike]) -> bool:
+def validate_ontology(input_file_or_json: Union[str, dict[Any, Any], 'os.PathLike[Any]']) -> bool:
     """
     Validates an ontology against the knora schema
 
@@ -17,8 +17,8 @@ def validate_ontology(input_file_or_json: Union[str, dict[Any, Any], os.PathLike
     Returns:
         True if ontology passed validation, False otherwise
     """
-    data_model: dict[Any, Any] = {}
 
+    data_model: dict[Any, Any] = {}
     if isinstance(input_file_or_json, dict):
         data_model = input_file_or_json
     elif os.path.isfile(input_file_or_json):
@@ -46,9 +46,25 @@ def validate_ontology(input_file_or_json: Union[str, dict[Any, Any], os.PathLike
               f'The error occurred at {err.json_path}')
         return False
 
-    # Check if there are properties derived from hasLinkTo that form a circular reference. If so, these
-    # properties must have the cardinality 0-1 or 0-n, because during the xmlupload process, these values
-    # are temporarily removed.
+    # cardinalities check for circular references
+    if check_cardinalities_of_circular_references(data_model):
+        print('Data model is syntactically correct and passed validation.')
+        return True
+    else:
+        return False
+
+
+def check_cardinalities_of_circular_references(data_model: dict[Any, Any]) -> bool:
+    """
+    Check if there are properties derived from hasLinkTo that form a circular reference. If so, these
+    properties must have the cardinality 0-1 or 0-n, because during the xmlupload process, these values
+    are temporarily removed.
+    """
+
+    # search the ontology for all properties that are derived from hasLinkTo. store them in a dict, and map
+    # them to their objects (i.e. the resource classes they point to)
+    # example: if the property 'rosetta:hasTextMedium' points to 'rosetta:Image2D':
+    # link_properties = {'rosetta:hasTextMedium': ['rosetta:Image2D'], ...}
     ontos = data_model['project']['ontologies']
     link_properties: dict[str, List[str]] = dict()
     for index, onto in enumerate(ontos):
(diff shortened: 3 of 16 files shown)


knora/dsplib/models/resource.py
```diff
@@ -394,7 +393,7 @@ def _get_baseclass(self, superclasses: list[str]) -> Union[str, None]:
             return self._get_baseclass(gaga.superclasses)
         return None
 
-    def get_resclass(self, prefixedresclass: str) -> Type:
+    def get_resclass_type(self, prefixedresclass: str) -> Type:
```
EricSommerhalder, line 396: What's the difference between a resource class and a resource class type?

knora/dsplib/models/sipi.py
```diff
@@ -45,5 +44,5 @@ def upload_bitstream(self, filepath):
             req = requests.post(self.sipi_server + "/upload?token=" + self.token, files=files)
         on_api_error(req)
         print(f'Uploaded file {filepath}')
-        res = req.json()
+        res: dict[Any, Any] = req.json()
```
EricSommerhalder, line 47: Seems weird to me to type this, when we use "Any" anyways.

knora/dsplib/utils/onto_validate.py
```diff
@@ -18,8 +17,8 @@ def validate_ontology(input_file_or_json: Union[str, Dict, os.PathLike]) -> bool
     Returns:
         True if ontology passed validation, False otherwise
     """
-    data_model = ''
 
+    data_model: dict[Any, Any] = {}
```
EricSommerhalder, line 21: Same comment as above about the "Any" typing

knora/dsplib/utils/onto_validate.py
```diff
@@ -38,15 +37,138 @@ def validate_ontology(input_file_or_json: Union[str, Dict, os.PathLike]) -> bool
...
+    Check if there are properties derived from hasLinkTo that form a circular reference. If so, these
+    properties must have the cardinality 0-1 or 0-n, because during the xmlupload process, these values
+    are temporarily removed.
+    """
+
+    # search the ontology for all properties that are derived from hasLinkTo. store them in a dict, and map
+    # them to their objects (i.e. the resource classes they point to)
+    # example: if the property 'rosetta:hasTextMedium' points to 'rosetta:Image2D':
+    # link_properties = {'rosetta:hasTextMedium': ['rosetta:Image2D'], ...}
+    ontos = data_model['project']['ontologies']
```
EricSommerhalder, line 68: Theoretically you could create this dict after identifying problematic cases so this dict doesn't get as large. Can't imagine this becoming a problem though.

EricSommerhalder (approved):
I had some small comments but overall it looks fine to me.

knora/dsplib/models/resource.py
jnussbaum: As far as I understand, there are resourceclass names (the parameter of this method, a string), and resourceclass types (the return value of this method, a https://docs.python.org/3.9/library/functions.html#type)

knora/dsplib/models/sipi.py
jnussbaum: This helps in xml_upload.py, where `img = sipi_server.upload_bitstream(os.path.join(imgdir, resource.bitstream.value))` is then automatically recognized as dict. It's still quite imprecise, but better than Any, in my opinion.

knora/dsplib/utils/onto_validate.py
jnussbaum: In order to identify the problematic cases, I need this dict first.

knora/dsplib/utils/onto_validate.py
```diff
@@ -38,15 +37,138 @@ def validate_ontology(input_file_or_json: Union[str, Dict, os.PathLike]) -> bool
...
+                    if resname not in dependencies:
+                        dependencies[resname] = dict()
+                        dependencies[resname][cardname] = targets
+                    elif cardname not in dependencies[resname]:
+                        dependencies[resname][cardname] = targets
+                    else:
+                        dependencies[resname][cardname].extend(targets)
+
+    # iteratively purge dependencies from non-circular references
+    for i in range(30):
```
irinaschubert, line 120: why is this 30?

knora/dsplib/utils/onto_validate.py
```diff
@@ -38,15 +37,138 @@ def validate_ontology(input_file_or_json: Union[str, Dict, os.PathLike]) -> bool
...
+        dependencies = {res: cards for res, cards in dependencies.items() if len(cards) > 0}
+        # remove resources that are not pointed to by any target
+        all_targets: Set[str] = set()
+        for cards in dependencies.values():
+            for trgt in cards.values():
+                all_targets = all_targets | set(trgt)
+        dependencies = {res: targets for res, targets in dependencies.items() if res in all_targets}
+
+    # check the remaining dependencies (which are only the circular ones) if they have all 0-1 or 0-n
+    okay_cardinalities = ['0-1', '0-n']
```
irinaschubert, line 138:
```suggestion
    ok_cardinalities = ['0-1', '0-n']
```

knora/dsplib/utils/onto_validate.py
```diff
@@ -38,15 +37,138 @@ def validate_ontology(input_file_or_json: Union[str, Dict, os.PathLike]) -> bool
...
+            if card_numbers not in okay_cardinalities:
+                if res not in notok_dependencies:
+                    notok_dependencies[res] = [card]
+                else:
+                    notok_dependencies[res].append(card)
+
+    if len(notok_dependencies) == 0:
+        return True
+    else:
+        print('ERROR: Your ontology contains properties derived from "hasLinkTo" that allow circular references '
```
irinaschubert, line 163: Here I would print out the name of the ontology (in case there are several ontologies).

knora/dsplib/utils/onto_validate.py
```diff
@@ -38,15 +37,138 @@ def validate_ontology(input_file_or_json: Union[str, Dict, os.PathLike]) -> bool
...
+              'between resources. This is not a problem in itself, but if you try to upload data that actually '
```
irinaschubert, line 164:
```suggestion
              'between resources. This is not a problem in itself, but if you try to upload data that '
```

knora/dsplib/utils/onto_validate.py
```diff
@@ -38,15 +37,138 @@ def validate_ontology(input_file_or_json: Union[str, Dict, os.PathLike]) -> bool
...
+              'contains circular references, these "hasLinkTo" cardinalities will be temporarily removed from the '
+              'problematic resources. To do so, it is important that the involved "hasLinkTo" cardinalities have a '
```
irinaschubert, line 166:
```suggestion
              'affected resources. Therefore, it is necessary that the involved "hasLinkTo" cardinalities have a '
```

knora/dsplib/utils/xml_upload.py
```diff
@@ -321,6 +338,22 @@ def print(self) -> None:
...
+        Get a list of all XMLProperties that have an outgoing link to another resource, be it a resptr-prop link
+        or a standoff link in a text.
+        """
+        link_properties: list[XMLProperty] = []
+        for prop in self._properties:
+            if prop.valtype == 'resptr':
+                link_properties.append(prop)
+            elif prop.valtype == 'text':
+                for value in prop.values:
+                    if value.resrefs is not None:
```
irinaschubert, line 352:
```suggestion
                    if value.resrefs:
```

knora/dsplib/utils/xml_upload.py
```diff
@@ -550,20 +594,57 @@ def do_sort_order(resources: list[XMLResource], verbose) -> list[XMLResource]:
...
-            print("ERROR Unable to resolve all resptr dependencies.")
-            for res in nok_resources:
-                unresolvable_resptrs = []
-                for resptr_id in res.get_resptrs():
-                    if resptr_id not in ok_res_ids:
-                        unresolvable_resptrs.append(resptr_id)
-                print(f"\tResource '{res.id}' has unresolvable resptrs to {unresolvable_resptrs}")
-            exit(1)
+            # there are circular references. go through all problematic resources, and stash the problematic references.
+            # there are two types, text and resptr, which need to be treated differently.
```
irinaschubert, line 598:
```suggestion
            # There are circular references. Go through all problematic resources, and stash the problematic references.
            # There are two types: text and resptr which need to be treated differently.
```

knora/dsplib/utils/xml_upload.py
```diff
@@ -550,20 +594,57 @@ def do_sort_order(resources: list[XMLResource], verbose) -> list[XMLResource]:
...
+                                    if link_prop not in stashed_resptr_props[res]:
+                                        stashed_resptr_props[res][link_prop] = [str(value.value)]
+                                    else:
+                                        stashed_resptr_props[res][link_prop].append(str(value.value))
+                                link_prop.values.remove(value)
+                    else:
+                        # This should never happen. If this code is ever executed, it proves that there is a
+                        # misconception in the underlying assumptions of the code logic
+                        print(f'Warning in remove_circular_references(): link_prop.valtype is neither text '
+                              f'nor resptr. Mistake in the underlying assumptions of the code logic.')
```
irinaschubert, line 632: This seems weird to me. If this should never happen, the code should be removed or at least an error returned / exception thrown.

knora/dsplib/utils/xml_upload.py
```diff
@@ -550,20 +594,57 @@ def do_sort_order(resources: list[XMLResource], verbose) -> list[XMLResource]:
...
+                                elif link_prop not in stashed_xml_texts[res]:
+                                    stashed_xml_texts[res][link_prop] = {value_hash: cast(KnoraStandoffXml, value.value)}
+                                else:
+                                    stashed_xml_texts[res][link_prop][value_hash] = cast(KnoraStandoffXml, value.value)
+                                value.value = KnoraStandoffXml(value_hash)
+                                value.resrefs = [_id for _id in value.resrefs if _id in ok_res_ids]
+                    elif link_prop.valtype == 'resptr':
+                        for value in link_prop.values.copy():
+                            if value.value not in ok_res_ids:
+                                # value.value is the id of the target resource. stash it, then delete it
```
irinaschubert, line 618:
```suggestion
                                # value.value is the id of the target resource. Stash it, then delete it.
```

knora/dsplib/utils/xml_upload.py
```diff
@@ -713,9 +872,9 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
         permissions_lookup[key] = perm.get_permission_instance()
 
     # create a dictionary to look up resource classes
-    res_classes: dict[str, type] = {}
+    resclass_name_2_type: dict[str, type] = {}
```
irinaschubert, line 875:
```suggestion
    resclass_name_to_type: dict[str, type] = {}
```

knora/dsplib/utils/xml_upload.py
```diff
@@ -730,35 +889,82 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
-                f"ERROR while trying to create resource '{resource.label}' ({resource.id}). The error message was: {err.message}")
+            print(f"ERROR while trying to create resource '{resource.label}' ({resource.id}). "
+                  f"The error message was: {err.message}")
             failed_uploads.append(resource.id)
             continue
-
         except Exception as exception:
-            print(
-                f"ERROR while trying to create resource '{resource.label}' ({resource.id}). The error message was: {exception}")
+            print(f"EXCEPTION while trying to create resource '{resource.label}' ({resource.id}). "
```
irinaschubert, line 917: I would leave the error message as it was. For the user an error is an error. Exception is only the technical term for an error.

knora/dsplib/utils/xml_upload.py
```diff
@@ -730,35 +889,82 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+                link_props=link_props,
+                res_iri_lookup=res_iri_lookup,
+                con=con,
+                verbose=verbose
+            )
+        except BaseError as err:
+            print(f'BaseError while updating an XML text of resource "{resource.id}": {err.message}')
+            continue
+        except Exception as exception:
+            print(f'Exception while updating an XML text of resource "{resource.id}": {exception}')
```
irinaschubert, line 944: Exception is too technical for the end user. Use "error" instead.

knora/dsplib/utils/xml_upload.py
```diff
@@ -730,35 +889,82 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+            update_xml_texts(
+                resource=resource,
+                res_iri=res_iri,
+                link_props=link_props,
+                res_iri_lookup=res_iri_lookup,
+                con=con,
+                verbose=verbose
+            )
+        except BaseError as err:
+            print(f'BaseError while updating an XML text of resource "{resource.id}": {err.message}')
```
irinaschubert, line 941: The user should not care about what sort of error it was. He just needs to know that there was an error. So, I would use "ERROR" (see other error messages for the format / formulation) instead of "BaseError".

knora/dsplib/utils/xml_upload.py
```diff
@@ -730,35 +889,82 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+                prop_2_resptrs=prop_2_resptrs,
+                res_iri_lookup=res_iri_lookup,
+                con=con,
+                verbose=verbose
+            )
+        except BaseError as err:
+            print(f'BaseError while updating an XML text of resource "{resource.id}": {err.message}')
+            continue
+        except Exception as exception:
+            print(f'Exception while updating an XML text of resource "{resource.id}": {exception}')
```
irinaschubert, line 966: same as above

knora/dsplib/utils/xml_upload.py
```diff
@@ -730,35 +889,82 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+            continue
+
+    # update the resources with the stashed resptrs
+    if len(stashed_resptr_props) > 0:
+        print('Update the stashed resptrs...')
+    for resource, prop_2_resptrs in stashed_resptr_props.items():
```
irinaschubert, line 950:
```suggestion
    for resource, prop_to_resptrs in stashed_resptr_props.items():
```

irinaschubert (approved):
Thanks for the work! 

I added quite a few commentaries / suggestions, feel free to take over whatever you see fit. 

I would test the code with unit tests so you can be sure that the code does what you expect it to do. There should never be tests inside the code itself to test the logic of the code. This should be done in unit tests.

Regarding the error messages: In my opinion (and experience), if an error happens, it should be called "error" (not exception). I recommend to stick to the pattern/formulation of error messages we already have in dsp-tools.

Regarding the security issue about the http, I declared it as safe.

The code is very complex (see code smell warnings from sonar cloud). Do you see any way to improve that? Could you take apart some of the code and make it less complex?

knora/dsplib/utils/onto_validate.py
jnussbaum: Random. It must be done a couple of times, until 'dependencies' reaches its final state and isn't changed any more by the for-loop. Perhaps it could be done more elegantly, with a while loop that checks after every pass if the pass has changed something. I tried it out quickly, but I don't see an easy solution.

knora/dsplib/utils/onto_validate.py
jnussbaum: the onto name is already prefixed to the resource names

knora/dsplib/utils/xml_upload.py
jnussbaum: ok, I agree. I raised a BaseError instead

knora/dsplib/utils/xml_upload.py
jnussbaum: I thought this difference would help us when debugging. If I get an error message, I like to copy it and search it in the source code. For this purpose, it helps to have a specific error message.

irinaschubert: I see, but "BaseError" doesn't tell you much more than "error", does it? The interesting information is in the error message itself. And, the general problem is that we don't have logging in dsp-tools. That would be the place to write out more information. The end user should IMO not be bothered with technical details.

PR #169, opened 2022-03-24
fix(onto creation): prevent sorting algorith from modifying original ontology

jnussbaum:
resolves DEV-626

### DIFF
--- a/knora/dsplib/utils/onto_create_ontology.py
+++ b/knora/dsplib/utils/onto_create_ontology.py
@@ -312,6 +312,7 @@ def sort_resources(unsorted_resources: list[dict[str, Any]], onto_name: str) -> 
         sorted list of resource classes
     """
 
+    unsorted_resources = unsorted_resources.copy()
     sorted_resources: list[dict[str, Any]] = list()
     ok_resource_names: list[str] = list()
     while len(unsorted_resources) > 0:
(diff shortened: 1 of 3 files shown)


knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -312,6 +312,7 @@ def sort_resources(unsorted_resources: list[dict[str, Any]], onto_name: str) ->
         sorted list of resource classes
     """
 
+    unsorted_resources = unsorted_resources.copy()
```
irinaschubert, line 315: I would call it differently, so it is clear which one is the original one and which one is the copy. When working on this in the future one can easily mix it up and be surprised by unexpected behaviour.

irinaschubert (changes requested):

knora/dsplib/utils/onto_create_ontology.py
jnussbaum: What would the new name then be? I can't think of a good name. And I think it's better to hide away the original as fast as possible, so that it isn't even possible to modify it any more. If the original `unsorted_resources` is still accessible, one could accidentally modify it.

irinaschubert: That's exactly my point. But with the same name, it's not clear which one you are referring to. Is there a reason why you don't you use a `for i in len(unsorted_resources)` construct and omit the while/remove and copy construct?

irinaschubert (approved):
LGTM

PR #170, opened 2022-03-25
feat: add support for external ontologies (dev-512)

irinaschubert:
resolves DEV-512

### DIFF
--- a/knora/dsplib/models/helpers.py
+++ b/knora/dsplib/models/helpers.py
@@ -16,8 +16,11 @@ from pystrict import strict
 @dataclass
 class OntoInfo:
     """
-    A small class thats holds an ontology IRI. The variable "hashtag" is True, if "#" is used s separate elements,
-    False if the element name is just appended
+    Holds an ontology IRI
+
+    Attributes:
+        iri: the ontology IRI
+        hashtag: True if "#" is used to separate elements, False if element name is appended after "/"
     """
     iri: str
     hashtag: bool
@@ -122,7 +125,10 @@ class Context:
         "skos": OntoInfo("http://www.w3.org/2004/02/skos/core", True),
         "bibtex": OntoInfo("http://purl.org/net/nknouf/ns/bibtex", True),
         "bibo": OntoInfo("http://purl.org/ontology/bibo/", False),
-        "cidoc": OntoInfo("http://purl.org/NET/cidoc-crm/core", True)
+        "cidoc": OntoInfo("http://purl.org/NET/cidoc-crm/core", True),
+        "schema": OntoInfo("https://schema.org/", False),
+        "edm": OntoInfo("http://www.europeana.eu/schemas/edm/", False),
+        "ebucore": OntoInfo("http://www.ebu.ch/metadata/ontologies/ebucore/ebucore", True)
     })
 
     knora_ontologies = ContextType({
@@ -242,8 +248,6 @@ class Context:
                 return
             if prefix in self.common_ontologies:
                 self._context[prefix] = self.common_ontologies[prefix]
-            else:
-                raise BaseError("The prefix '{}' is not known!".format(prefix))
         else:
             if iri.endswith("#"):
                 iri = iri[:-1]
@@ -303,15 +307,16 @@ class Context:
 
     def get_qualified_iri(self, val: Optional[str]) -> Optional[str]:
         """
-        We will return the full qualified IRI, if it is not yet a full qualified IRI. If
-        the IRI is already fully qualified, the we just return it.
+        Given an IRI, its fully qualified name is returned.
+        
+        Args:
+            val: The input IRI
 
-        :param val: The input short form
-        :return: the fully qualified IRI
+        Returns:
+            the fully qualified IRI
         """
-        if val is None:
+        if not val:
             return None
-        # if self.__is_iri(val):
         if IriTest.test(val):
             return val
         tmp = val.split(':')
@@ -326,7 +331,7 @@ class Context:
                 self._rcontext[entry[1].iri] = entry[0]
                 iri_info = entry[1]
             else:
-                raise BaseError("Ontology not known! Cannot generate full qualified IRI")
+                raise BaseError("Ontology not known! Cannot generate fully qualified IRI")
         if iri_info.hashtag:
             return iri_info.iri + '#' + tmp[1]
         else:
@@ -370,7 +375,7 @@ class Context:
                 prefix = entry[0]
             else:
                 raise BaseError(
-                    "Ontology {} not known! Cannot generate full qualified IRI: prefix={}".format(iri, prefix))
+                    "Ontology {} not known! Cannot generate fully qualified IRI: prefix={}".format(iri, prefix))
         return prefix + ':' + element
 
     def reduce_iri(self, iristr: str, ontoname: Optional[str] = None) -> str:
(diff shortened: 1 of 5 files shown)


knora/dsplib/models/helpers.py
```diff
@@ -339,66 +344,67 @@ def get_prefixed_iri(self, iri: Optional[str]) -> Optional[str]:
...
+        if split_point == -1:
+            split_point = iri.rfind('/')
+            onto_part = iri[:split_point + 1]
+            element = iri[split_point + 1:]
         else:
-            ontopart = iri[:splitpoint]
-            element = iri[splitpoint + 1:]
-        prefix = self._rcontext.get(ontopart)
+            onto_part = iri[:split_point]
+            element = iri[split_point + 1:]
```
BalduinLandolt, line 365: I'm pretty sure this could be simplified significantly, though I don't exactly know how because I have trouble understanding the whole thing (which is an indication that it should be simplified or commented).  
But I'd try something along the lines of:

```python
if '#' in iri:
    onto, elem = iri.split('#')
else:
    onto, elem = iri.rsplit('/', 1)
```

knora/dsplib/models/helpers.py
```diff
@@ -339,66 +344,67 @@ def get_prefixed_iri(self, iri: Optional[str]) -> Optional[str]:
...
+
+        prefix = self._rcontext.get(onto_part)
         if prefix is None:
-            entrylist = list(filter(lambda x: x[1].iri == ontopart, self.common_ontologies.items()))
-            if len(entrylist) == 1:
-                entry = entrylist[0]
+            entry_list = list(filter(lambda x: x[1].iri == onto_part, self.common_ontologies.items()))
```
BalduinLandolt, line 369: this could be simplified with a list comprehension

knora/dsplib/models/helpers.py
```diff
@@ -339,66 +344,67 @@ def get_prefixed_iri(self, iri: Optional[str]) -> Optional[str]:
...
-            iristr = self.get_prefixed_iri(iristr)
-        tmp = iristr.split(':')
+        if IriTest.test(iri_str):
+            if self.get_prefixed_iri(iri_str):
+                iri_str = self.get_prefixed_iri(iri_str)
+        tmp = iri_str.split(':')
         if tmp[0] == knora_api or tmp[0] == salsah_gui:
             return tmp[1]
-        elif ontoname is not None and tmp[0] == ontoname:
+        elif onto_name and tmp[0] == onto_name:
```
BalduinLandolt, line 404: I think this could even be simplified to `elif onto_name == tmp[0]` because we expect `tmp[0]` not to be `None` so the first check should be redundant, right?

knora/dsplib/models/helpers.py
```diff
@@ -410,7 +416,11 @@ def toJsonObj(self) -> dict[str, str]:
 
     def get_externals_used(self) -> dict[str, str]:
         exclude = ["rdf", "rdfs", "owl", "xsd", "knora-api", "salsah-gui"]
-        return {prefix: onto.iri for prefix, onto in self._context.items() if prefix not in exclude}
+        prefixes: dict[str, str] = {}
+        for prefix, onto in self._context.items():
+            if prefix not in exclude:
+                prefixes[prefix] = onto.iri
+        return prefixes
```
BalduinLandolt, line 423: why remove the elegant dict comprehension? :)

BalduinLandolt (approved):
looks good, just some minor comments

knora/dsplib/models/group.py
```diff
@@ -249,16 +249,14 @@ def getAllGroups(con: Connection) -> Optional[list[Group]]:
...
-                project_groups.append(group)
-        return list(map(lambda a: Group.fromJsonObj(con, a), project_groups))
+    def getAllGroupsForProject(con: Connection, proj_shortcode: str) -> Optional[list[Group]]:
+        all_groups: Optional[list[Group]] = Group.getAllGroups(con)
+        if all_groups:
+            project_groups = []
+            for group in all_groups:
+                if group.project == "http://rdfh.ch/projects/" + proj_shortcode:
+                    project_groups.append(group)
+            return project_groups
```
jnussbaum, line 259: Cool, you fixed DEV-786 even before the bug was reported :-) The e-mail exchange we had yesterday was about a researcher who encountered problems with exactly this :-)

knora/dsplib/models/helpers.py
```diff
@@ -16,8 +16,11 @@
 @dataclass
 class OntoInfo:
```
jnussbaum, line 17: I suggest renaming the class to a more concise name, e.g. OntoIRI

knora/dsplib/models/helpers.py
```diff
@@ -242,8 +248,6 @@ def add_context(self, prefix: str, iri: Optional[str] = None) -> None:
                 return
             if prefix in self.common_ontologies:
                 self._context[prefix] = self.common_ontologies[prefix]
-            else:
-                raise BaseError("The prefix '{}' is not known!".format(prefix))
```
jnussbaum, line 246: If I understand correctly, you basically did 3 things in helpers.py:
 1. tidy up without semantic changes
 2. remove error raisings
 3. add more external ontos

Like Balduin, I have troubles understanding the code, but I just wanted to ask if you are sure that you want to remove all the errors that were raised.

knora/dsplib/models/propertyclass.py
```diff
@@ -417,7 +417,7 @@ def delete(self, last_modification_date: LastModificationDate) -> LastModificati
 
     def createDefinitionFileObj(self, context: Context, shortname: str):
         """
-        Create an object that jsonfied can be used as input to "create_onto"
+        Create an object that can be used as input to create an ontology on a DSP server
```
jnussbaum, line 420: "create an ontology on a DSP server" is very general. I always find it useful if I get precise directives like a method name. Or is that considered bad style?

jnussbaum:
LGTM, just some comments. I also added the additional Jira Issue that you fixed in the comment field of the PR.

jnussbaum (approved):

knora/dsplib/models/group.py
irinaschubert: haha, true, I didn't even remember :D

knora/dsplib/models/helpers.py
irinaschubert: 1. correct
2. I just removed one error because we now allow all ontologies (it is not checked anymore if it is known or not)
3. correct

The point of still having some ontology IRIs is that when using `dsp-tools get` the IRIs of the external ontologies can be converted back to the notation with the prefix. If the ontology is not known, the fully qualified IRI is used.

knora/dsplib/models/helpers.py
irinaschubert: same as above

knora/dsplib/models/helpers.py
irinaschubert: Because I find them hard to read ;) I probably tried to understand and rewrote it. But I changed it back now.

knora/dsplib/models/propertyclass.py
BalduinLandolt: I don't have a strong opinion on that. I think clear instructions which method works with which can be helpful if input/output is very specific. In this case, where the output is according to our API specifications, it may make sense to keep it more generic.  
And I would normally keep the reference rather at the receiving method than on the producing method. But that may be another matter of taste.  
In any case, if we go for referencing methods directly, I suggest using back ticks to it's formatted as code, and maybe even write "`create_onto()`" to indicate that is's a method.

irinaschubert: Thanks for your thoughts, I now wrote `create_onto()`, so it's clear that it is a method.

PR #173, opened 2022-03-30
fix(requirements): correct typo in requirements.txt (DEV-763)

jnussbaum:
resolves DEV-763


irinaschubert (changes requested):
I fear that with "chore" in the PR title there will be no new release. According to https://docs.dasch.swiss/developers/dsp/contribution/ you need to use fix, refactor or feat. I think fix would be the right choice here.

irinaschubert (approved):

PR #177, opened 2022-03-31
chore: change to pipenv (DEV-764)

jnussbaum:
resolves DEV-764


README.md
```diff
@@ -1,74 +1,96 @@
...
 
 ## Information for developers
+There is a `Makefile` for all the following tasks (and more). Type `make` to print the available targets. 
+
+For a quick start, use: 
+```
+pip install pipenv
+pipenv install --dev
+make run
+```
```
jnussbaum, line 15: `make run` executes `pip3 install -e .`, instead of `pipenv install -e .` Is that good or bad?

README.md
```diff
@@ -1,74 +1,96 @@
...
-To install the requirements run:
+If a package is only needed for development, please install it with `pipenv install package --dev`,
+so it gets added to the `[dev-packages]` section of `Pipfile`.
 
-```bash
-pip3 install -r requirements.txt
-```
+For security reasons, the maintainer regularly executes
+ - `pipenv check` to get informed about vulnerabilities. 
+ - `pipenv lock` to update `Pipfile.lock` with the latest version of every package
```
jnussbaum, line 51: I'm a bit confused if this is correct. Does `pipenv lock` really search for updates of the packages? For me it would seem more logical to execute `pipenv update`, but according to the docs, `pipenv update` does something else, namely:
pipenv update: runs lock, then sync
pipenv lock: Generates Pipfile.lock.
pipenv sync: Installs all packages specified in Pipfile.lock.

Makefile
```diff
@@ -29,7 +29,7 @@ upload: ## upload distribution package to PyPI
 
 .PHONY: upgrade-dist-tools
 upgrade-dist-tools: ## upgrade packages necessary for testing, building, packaging and uploading to PyPI
-	python3 -m pip install --upgrade pip setuptools wheel tqdm twine pytest mkdocs
```
jnussbaum, line 32: I removed tqdm from "upgrade-dist-tools", because it's used nowhere. Is that correct?

jnussbaum:
In makefile, line 22, two arguments are passed to setup.py: sdist and bdist_wheel, but these are not used. Should we remove them?

Makefile
BalduinLandolt: tqdm is not part of dist tools, so no idea why this is there

README.md
BalduinLandolt: then it's not guaranteed to run in the virtual environment. I suggest changing it to `pipenv run make run`

README.md
BalduinLandolt: You could (or maybe even should) change `lock` to `update` here, so it also executes `sync`.

But the commands make sense, and yes: `lock` rebuilds the lock file according to the pipfile, and since we use `*` for all versions, it will grab the latest version and build the lock file accordingly, i.e. with the latest version.  
All that `sync` does, is updating the virtual environment according to the lock file.  
And then `updates` fist updates the lock file according to the pipfile and then installs this updated lockfile into your virtual environment.

Does that clarify it for you?

BalduinLandolt (approved):
See my comments for details, but overall it looks good

README.md
```diff
@@ -1,74 +1,96 @@
...
 ```
 
-## Install from source
+## Pipenv
+We use pipenv for our dependency management. There are two ways to get started:
+ - `pipenv install --dev` installs all dependencies, while giving them the opportunity to update themselves
+ - `pipenv install --ignore-pipfile` is used to get a deterministic build in production
 
-To install from source run:
+This works because there are two files defining the dependencies:
```
irinaschubert, line 31:
```suggestion
There are two files defining the dependencies:
```

README.md
```diff
@@ -1,74 +1,96 @@
...
+   - `[packages]` lists the dependencies used to run the software.
+   - `[dev-packages]` lists additional dependencies used for tests and deployment.
+ - `Pipfile.lock` enables deterministic builds, by exactly pinning the version of all (sub-) dependencies. 
+   This is done automatically, you must not edit `Pipfile.lock`.
 
-```bash
-python3 setup.py install
-```
+The diverse `requirements.txt` files in this repo are only present for backwards compatibility
+and for Github CI.
```
irinaschubert, line 39:
```suggestion
and for GitHub CI.
```

README.md
```diff
@@ -1,74 +1,96 @@
...
 
-## Install the requirements
+If you want to install a new package, install it with `pipenv install package`. This 
+ - installs the package (incl. sub-dependencies) in your virtual environment.
```
irinaschubert, line 42:
```suggestion
 - installs the package (incl. sub-dependencies) in your virtual environment
```

README.md
```diff
@@ -1,74 +1,96 @@
...
+ - adds the package to the section `[packages]` of `Pipfile`. By default, no versions are pinned.
```
irinaschubert, line 43:
```suggestion
 - adds the package to the section `[packages]` of `Pipfile`. By default, no versions are pinned
```

README.md
```diff
@@ -1,74 +1,96 @@
...
+ - adds the pinned versions of package and all sub-dependencies to `Pipfile.lock`.
```
irinaschubert, line 44:
```suggestion
 - adds the pinned versions of package and all sub-dependencies to `Pipfile.lock`
```

README.md
```diff
@@ -1,74 +1,96 @@
...
 
-To install the requirements run:
+If a package is only needed for development, please install it with `pipenv install package --dev`,
+so it gets added to the `[dev-packages]` section of `Pipfile`.
 
-```bash
-pip3 install -r requirements.txt
-```
+For security reasons, the maintainer regularly executes
+ - `pipenv check` to get informed about vulnerabilities.
```
irinaschubert, line 50:
```suggestion
 - `pipenv check` to get informed about vulnerabilities
```

README.md
irinaschubert: I would even consider to use `pipenv install -e` in the Makefile for `make run`. Is there anything that speaks against it?

irinaschubert (approved):
LGTM, just some minor details

README.md
jnussbaum: > I would even consider to use `pipenv install -e` in the Makefile for `make run`. Is there anything that speaks against it?

Not everyone has pipenv installed.

PR #181, opened 2022-04-07
feat(get): get more infos from user (DEV-641)

jnussbaum:
resolves DEV-641

### DIFF
--- a/knora/dsplib/models/user.py
+++ b/knora/dsplib/models/user.py
@@ -701,7 +701,7 @@ class User(Model):
                 if proj["id"] == "http://rdfh.ch/projects/" + proj_shortcode:
                     project_users.append(user)
 
-        # find out the 'groups' and 'projects' of every user
+        # get the 'groups' and 'projects' of every user
         for user in project_users:
             user_iri_esc = urllib.parse.quote_plus(user["id"])
             project_memberships      = con.get(f'/admin/users/iri/{user_iri_esc}/project-memberships')
@@ -722,9 +722,9 @@ class User(Model):
             user['groups'].reverse()
 
         # convert to User objects
-        res: list[User] = list(map(lambda a: User.fromJsonObj(con, a), project_users))
+        res: list[User] = [User.fromJsonObj(con, a) for a in project_users]
 
-        # add the informations about projects and groups to the User objects
+        # add the projects and groups to the User objects
         for project_user, res_user in zip(project_users, res):
             res_user._in_groups = res_user._in_groups | set(project_user['groups'])
             for proj in project_user['projects']:
--- a/knora/dsplib/models/user.py
+++ b/knora/dsplib/models/user.py
@@ -689,7 +689,9 @@ class User(Model):
         members = con.get(f'/admin/projects/shortcode/{proj_shortcode}/members')
         if members is None or len(members) < 1:
             return None
-        return [User.fromJsonObj(con, a) for a in members['members']]
+        res: list[User] = [User.fromJsonObj(con, a) for a in members['members']]
+        res.reverse()
+        return res
 
     def createDefinitionFileObj(
         self,
--- a/knora/dsplib/models/user.py
+++ b/knora/dsplib/models/user.py
@@ -476,18 +476,17 @@ class User(Model):
         in_groups: set[str] = set()
         if json_obj.get('permissions') is not None and json_obj['permissions'].get('groupsPerProject') is not None:
             sysadmin = False
-            project_groups = json_obj['permissions']['groupsPerProject']
-            for project in project_groups:
-                if project == Project.SYSTEM_PROJECT:
-                    if Group.PROJECT_SYSTEMADMIN_GROUP in project_groups[project]:
+            for project_iri, group_memberships in json_obj['permissions']['groupsPerProject'].items():
+                if project_iri == Project.SYSTEM_PROJECT:
+                    if Group.PROJECT_SYSTEMADMIN_GROUP in group_memberships:
                         sysadmin = True
                 else:
-                    for group in project_groups[project]:
+                    for group in group_memberships:
                         if group == Group.PROJECT_MEMBER_GROUP:
-                            if in_projects.get(project) is None:
-                                in_projects[project] = False
+                            if in_projects.get(project_iri) is None:
+                                in_projects[project_iri] = False
                         elif group == Group.PROJECT_ADMIN_GROUP:
-                            in_projects[project] = True
+                            in_projects[project_iri] = True
                         else:
                             in_groups.add(group)
         return cls(con=con,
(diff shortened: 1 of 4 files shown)


jnussbaum:
Open concerns:
 - user.py, line 725: The method User.fromJsonObj() wants the project and the groups to be inside the user[permissions] as dict. It ignores the fields "groups" and "projects" in the API-retrieved user-JSON-objects. I suspect that User.fromJsonObj() should be adapted, but I don't really understand how. So I decided to make a "hack": I add the "groups" and "projects" two times: first to the API-retrieved object, and then to the models.user.User objects.
 - In the original onto, the ontos of the current file are not listed in the "prefixes" section (because we made them optional some weeks ago). But in the retrieved one, they are. In the GET-process, I think it would make sense to eliminate all prefixes of the current onto file. What do you think?
 - In users>groups and users>projects, The prefixes could be omitted as well.

irinaschubert (commented):
A question: how are users treated if they are members of several projects and members of groups of other projects, are these projects and groups retrieved as well? If so, should this information be provided in with the `dsp-tools get` command?

About your questions:
- user.py, line 725: I'm not sure if I understand your question: The API route that is used on line 692 does not return the groups and projects, so you need to get the groups and users separately, like you did. What do you mean with "I add the groups and projects two times"? You only add them once to the user you retrieved from the API. Not sure if I miss something here...
- Prefixes of the acutal ontology: I would leave them there. It is correct and it's a useful information. Or is there a reason why it shouldn't be in the ontology? --> I don't really have a strong opinion on that - do what you prefer
- I would leave them there, to me they aren't disturbing. --> I don't really have a strong opinion on that - do what you prefer

knora/dsplib/models/user.py
```diff
@@ -699,17 +700,56 @@ def getAllUsersForProject(con: Connection, proj_shortcode: str) -> list[Any]:
...
+            for group in normal_group_memberships['groups']:
+                project_prefix = group['project']['shortname']
+                user['groups'].append(f"{project_prefix}:{group['name']}")
+
+            user['groups'].reverse()
+
+        # convert to User objects
+        res: list[User] = list(map(lambda a: User.fromJsonObj(con, a), project_users))
+
+        # add the informations about projects and groups to the User objects
```
irinaschubert, line 727:
```suggestion
        # add the projects and groups to the User objects
```

knora/dsplib/models/user.py
```diff
@@ -689,7 +690,7 @@ def getAllUsersForProject(con: Connection, proj_shortcode: str) -> list[Any]:
 
         result = con.get(User.ROUTE)
```
irinaschubert, line 691: Now that I think about it, there is a route to get all users for a specific project, it's:
`/admin/projects/shortcode/{{project_shortcode}}/members`

It's probably easier to use this for `getAllUsersForProject` directly. There are also groups and projects for the user.

knora/dsplib/models/user.py
```diff
@@ -699,17 +700,56 @@ def getAllUsersForProject(con: Connection, proj_shortcode: str) -> list[Any]:
...
+                    user['projects'].append(f"{project['shortname']}:member")
+
+            for group in normal_group_memberships['groups']:
+                project_prefix = group['project']['shortname']
+                user['groups'].append(f"{project_prefix}:{group['name']}")
+
+            user['groups'].reverse()
+
+        # convert to User objects
+        res: list[User] = list(map(lambda a: User.fromJsonObj(con, a), project_users))
```
BalduinLandolt, line 725:
```suggestion
        res: list[User] = [User.fromJsonObj(con, a) for a in project_users]
```

I personally don't like lambdas in python, nor its map implementation, for that matter. So I'd suggest using list comprehensions instead.

BalduinLandolt (approved):
I don't have much to add to this one, beyond the comments Irina has already made

knora/dsplib/models/user.py
```diff
@@ -699,17 +700,56 @@ def getAllUsersForProject(con: Connection, proj_shortcode: str) -> list[Any]:
...
+
+            for project in project_memberships['projects']:
+                if f"{project['shortname']}:admin" not in user['projects']:
+                    user['projects'].append(f"{project['shortname']}:member")
+
+            for group in normal_group_memberships['groups']:
+                project_prefix = group['project']['shortname']
+                user['groups'].append(f"{project_prefix}:{group['name']}")
+
+            user['groups'].reverse()
```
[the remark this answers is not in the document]
jnussbaum: The API returns them in the inversed order, so I need to reverse them again.

jnussbaum:
Thanks to Irina's comment (https://github.com/dasch-swiss/dsp-tools/pull/181#discussion_r847283917) I managed to considerably simplify the code and to resolve the concern I had in https://github.com/dasch-swiss/dsp-tools/pull/181#issuecomment-1091927269.

Besides, I implemented your feedback and I also did some refactorings where it seemed useful.

The PR has now changed considerably, so I ask you again for a review.

knora/dsplib/models/user.py
```diff
@@ -475,18 +476,17 @@ def fromJsonObj(cls, con: Connection, json_obj: Any):
...
-                    for group in project_groups[project]:
+                    for group in group_memberships:
                         if group == Group.PROJECT_MEMBER_GROUP:
-                            if in_projects.get(project) is None:
-                                in_projects[project] = False
+                            if in_projects.get(project_iri) is None:
+                                in_projects[project_iri] = False
                         elif group == Group.PROJECT_ADMIN_GROUP:
-                            in_projects[project] = True
+                            in_projects[project_iri] = True
```
jnussbaum, line 489: This is just a refactoring, without semantic modification.

knora/dsplib/models/user.py
```diff
@@ -675,41 +675,53 @@ def getAllUsers(con: Connection) -> list[Any]:
...
-                User.IRI + urllib.parse.quote_plus(user["id"], safe='') + '/project-memberships')
-            project = project_list["projects"]
-            for proj in project:
-                if proj["id"] == "http://rdfh.ch/projects/" + proj_shortcode:
-                    project_users.append(user)
-        return list(map(lambda a: User.fromJsonObj(con, a), project_users))
-
-    def createDefinitionFileObj(self):
-        user = {
+        members = con.get(f'/admin/projects/shortcode/{proj_shortcode}/members')
```
jnussbaum, line 689: With this route, the same task can be achieved with less code

knora/dsplib/models/user.py
```diff
@@ -475,18 +476,17 @@ def fromJsonObj(cls, con: Connection, json_obj: Any):
         in_groups: set[str] = set()
         if json_obj.get('permissions') is not None and json_obj['permissions'].get('groupsPerProject') is not None:
```
BalduinLandolt, line 477: generally, `if x is not None` can normally be rewritten as `if not x` (but that doesn't need to be done in this PR... I know it's all over the place in dsp-tools)

knora/dsplib/models/user.py
```diff
@@ -475,18 +476,17 @@ def fromJsonObj(cls, con: Connection, json_obj: Any):
...
             sysadmin = False
-            project_groups = json_obj['permissions']['groupsPerProject']
-            for project in project_groups:
-                if project == Project.SYSTEM_PROJECT:
-                    if Group.PROJECT_SYSTEMADMIN_GROUP in project_groups[project]:
+            for project_iri, group_memberships in json_obj['permissions']['groupsPerProject'].items():
+                if project_iri == Project.SYSTEM_PROJECT:
+                    if Group.PROJECT_SYSTEMADMIN_GROUP in group_memberships:
                         sysadmin = True
```
BalduinLandolt, line 482: this could be simplified even further (I think): `sysadmin = (Group.PROJECT_SYSTEMADMIN_GROUP in group_memberships)`

knora/dsplib/models/user.py
BalduinLandolt: nice!

knora/dsplib/models/user.py
```diff
@@ -675,41 +675,53 @@ def getAllUsers(con: Connection) -> list[Any]:
...
-            project = project_list["projects"]
-            for proj in project:
-                if proj["id"] == "http://rdfh.ch/projects/" + proj_shortcode:
-                    project_users.append(user)
-        return list(map(lambda a: User.fromJsonObj(con, a), project_users))
-
-    def createDefinitionFileObj(self):
-        user = {
+        members = con.get(f'/admin/projects/shortcode/{proj_shortcode}/members')
+        if members is None or len(members) < 1:
```
BalduinLandolt, line 690: this can be simplified to `if not members:` (both `None` and empty lists/dicts/sets evaluate as false in an if-clause)

knora/dsplib/models/user.py
```diff
@@ -675,41 +675,53 @@ def getAllUsers(con: Connection) -> list[Any]:
...
+            return None
+        res: list[User] = [User.fromJsonObj(con, a) for a in members['members']]
+        res.reverse()
```
BalduinLandolt, line 693: I'm not sure the order of the objects you get from the API is deterministic (at least if the cache is rebuilt), so this may not reliably reproduce the initial order. But if it works in the cases you tested, feel free to leave it in.

knora/dsplib/utils/onto_get.py
```diff
@@ -62,11 +62,14 @@ def get_ontology(project_identifier: str, outfile: str, server: str, user: str,
...
-        if verbose:
-            print(f"\tGot user '{user.username}'")
-    project_obj["users"] = users_obj
+    if users:
+        for usr in users:
+            users_obj.append(usr.createDefinitionFileObj(
+                con=con, proj_shortname=project.shortname, proj_shortcode=project.shortcode
+            ))
+            if verbose:
+                print(f"\tGot user '{usr.username}'")
```
BalduinLandolt, line 71: is `usr` intentional?

BalduinLandolt (approved):
one can always find more to change and improve... but I think this should be good for now. (I added some minor remarks nonetheless)

knora/dsplib/utils/onto_get.py
jnussbaum: no, rather random

irinaschubert (approved):
LGTM

PR #183, opened 2022-04-11
fix(groups): dsp-tools should not allow group creation if group name already in use (DEV-798)

irinaschubert:
resolves DEV-798


knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -126,10 +126,10 @@ def create_groups(con: Connection, groups: list[dict[str, str]], project: Projec
         group_exists: bool = False
         if all_groups:
             for group_item in all_groups:
-                if group_item.project == project.id and group_item.name == group_name:
+                if group_item.name == group_name:
                     group_exists = True
```
BalduinLandolt, line 130: This could be neatly rewritten as follows:
```python
group_exists: bool = any(group_item.name == group_name for group_item in all_groups)
```

PR #187, opened 2022-04-26
chore: update schema-files (DEV-449)

gNahcab:
resolves DEV-449


knora/dsplib/schemas/ontology.json
```diff
@@ -1178,7 +1178,6 @@
             "required": [
                 "shortcode",
                 "shortname",
-                "longname",
```
jnussbaum, line 966: Why should longname be optional? I think it's important to have it...

irinaschubert: @jnussbaum that came from me. I saw this already some time ago and asked @gNahcab to adjust it. From the API side it's not mandatory: https://docs.dasch.swiss/DSP-API/03-apis/api-admin/projects/#create-a-new-project
I think it is good if DSP-TOOLS has the same restrictions as the API. Otherwise we need to maintain two different systems / logics which is much harder to keep up to date.

jnussbaum (approved):
LGTM

knora/dsplib/schemas/ontology.json
jnussbaum: Thanks for explaining, Irina. I approved the PR, and you can merge, @gNahcab 👍

irinaschubert (approved):
LGTM

PR #192, opened 2022-05-16
fix(onto validation): correctly identify circular dependencies (DEV-769)

gNahcab:
resolves DEV-769


irinaschubert (approved):
LGTM, there are already tests that test the expected behaviour, right?

knora/dsplib/utils/onto_validate.py
```diff
@@ -111,64 +110,46 @@ def check_cardinalities_of_circular_references(data_model: dict[Any, Any]) -> bo
...
+    errors: set[str] = set()
+    circles = list(nx.simple_cycles(graph))
+    for circle in circles:
+        for index, resource in enumerate(circle):
+            target = circle[(index+1) % len(circle)]
+            for property, targets in dependencies[resource].items():
+                if target in targets:
+                    prop = property
+            if cardinalities[resource][prop] not in ['0-1', '0-n']:
+                errors.add(f'Resource "{resource}", property "{prop}"')
```
BalduinLandolt, line 138: this is a really neat solution for the problem, congratulations!

BalduinLandolt (approved):
Very nice!

I second Irina that this should be tested. If it already is, even better! Otherwise it would be good to have a couple of unit tests for the circular dependency method.

jnussbaum:
I added tests, and refactored the code to make it better testable. 
For the test: First I wanted to take the existing test-onto.json, and inject circularity into it inside the testing method. But then, I realised that this gets easily broken if someone modifies the test onto. So it's better to have a separate circular onto saved as JSON file (this way it is reusable).

knora/dsplib/utils/onto_validate.py
```diff
@@ -91,11 +124,19 @@ def check_cardinalities_of_circular_references(data_model: dict[Any, Any]) -> bo
...
-    # example: if 'rosetta:Text' has the property 'rosetta:hasTextMedium' that points to 'rosetta:Image2D':
-    # dependencies = {'rosetta:Text': {'rosetta:hasTextMedium': ['rosetta:Image2D'], ...}}
-    dependencies: dict[str, dict[str, List[str]]] = dict()
-    for onto in ontos:
+    return link_properties
+
+
+def identify_problematic_cardinalities(data_model: dict[Any, Any], link_properties: dict[str, list[str]]) -> list[tuple[str, str]]:
+    """
+    make an error list with all cardinalities that are part of a circle but have "1" or "1-n"
```
irinaschubert, line 132:
```suggestion
    make an error list with all cardinalities that are part of a circle but have a cardinality of "1" or "1-n"
```

PR #194, opened 2022-05-25
fix: xmlupload crashes without writing id2iri mapping (DEV-813)

jnussbaum:
resolves DEV-813


knora/dsplib/utils/xml_upload.py
```diff
@@ -909,7 +909,15 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
 
         resource_bitstream = None
         if resource.bitstream:
-            img = sipi_server.upload_bitstream(os.path.join(imgdir, resource.bitstream.value))
+            try:
+                img = sipi_server.upload_bitstream(os.path.join(imgdir, resource.bitstream.value))
+            except BaseException as err:
```
jnussbaum, line 914: I'm a bit unsure which type of error is best to catch. I tested it by restarting SIPI in the Docker Desktop window while an upload was ongoing. This lead to several error types:
- ConnectionError
- ConnectionResetError,
- BaseException, 
- http.client.RemoteDisconnected
- HTTPConnectionPool

irinaschubert (approved):
LGTM, thanks!

knora/dsplib/utils/xml_upload.py
irinaschubert: OK, now I see. 
For the Exception that is not caught: did you try `except Exception as err:`?

knora/dsplib/utils/xml_upload.py
```diff
@@ -897,8 +919,55 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
 
-    failed_uploads = []
+def try_sipi_upload(sipi_server: Sipi, filepath: str) -> dict[Any, Any]:
+    img = None
+    for _ in range(20):
+        try:
+            img = sipi_server.upload_bitstream(filepath)
+            break
+        except BaseException:
+            print(f'{datetime.now().isoformat()}: Try reconnecting to DSP server (SIPI)...')
```
irinaschubert, line 951:
```suggestion
            print(f'{datetime.now().isoformat()}: Try reconnecting to SIPI...')
```

irinaschubert: SIPI is its own component, so I would just call it like that.

knora/dsplib/utils/xml_upload.py
```diff
@@ -404,6 +405,9 @@ def get_propvals(
                         for iri_ref in iri_refs:
                             res_id = iri_ref.split(':')[1]
                             iri = resiri_lookup.get(res_id)
+                            if not iri:
+                                raise BaseError(f'Resource cannot be created, because it contains a salsah-Link to '
+                                                f'the following invalid resource: {res_id}.')
```
jnussbaum, line 410: This fixes a bug which appeared during my testing. In some cases, it was possible that `iri` was None, and then line 411 failed, which lead to a difficult stacktrace.

knora/dsplib/utils/xml_upload.py
```diff
@@ -843,65 +769,147 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+    """
+    Wrapper around sipi_server.upload_bitstream(filepath), that handles errors and retries in case of connectivity
+    issues.
+
+    Args:
+        sipi_server: Sipi instance
+        filepath: file to upload
+
+    Returns:
+        answer from knora
```
irinaschubert, line 862:
```suggestion
        response from DSP-API
```

irinaschubert: We don't use the name "knora" anymore.

knora/dsplib/utils/xml_upload.py
```diff
@@ -843,65 +769,147 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+    id2iri_mapping: dict[str, str],
+    con: Connection,
+    failed_uploads: list[str]
+) -> tuple[dict[str, str], list[str]]:
+    """
+    Iterates through all resources, uploads them to knora (incl. error handling)
+
+    Args:
+        verbose: bool
+        resources: list of XMLResources to upload to knora
```
irinaschubert, line 900:
```suggestion
        resources: list of XMLResources to upload to DSP
```

test/e2e/test_tools.py
```diff
@@ -176,38 +176,45 @@ def test_create_ontology(self) -> None:
...
+                       sipi=self.sipi,
+                       verbose=False,
+                       validate_only=False,
+                       incremental=False)
+        self.assertEqual(cm.exception.code, 0)
 
         mapping_file = ''
         for mapping in [x for x in os.scandir('.') if x.name.startswith('id2iri_test-data_mapping_')]:
             delta = datetime.datetime.now() - datetime.datetime.fromtimestamp(mapping.stat().st_mtime_ns / 1000000000)
             if delta.seconds < 15:
```
irinaschubert, line 194: Why is this delta needed?

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +918,340 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+        f.write('*********************************************\n')
+        for res, props in stashed_xml_texts.items():
+            f.write(f'\n{res.id}\n---------\n')
+            for prop, stashed_texts in props.items():
+                f.write(f'{prop.name}\n')
+                for i, standoff in enumerate(stashed_texts.values()):
+                    f.write(f'\t{i}. text: {standoff}')
+
+
+def write_stashed_respr_props(
```
irinaschubert, line 1233:
```suggestion
def write_stashed_resptr_props(
```

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +918,340 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+        con: connection to knora
+        stashed_resptr_props: all resptr props that have been stashed
+
+    Returns:
+        nonapplied_resptr_props: the resptr props that could not be updated
+    """
+
+    print('Update the stashed resptrs...')
+    for resource, prop_2_resptrs in stashed_resptr_props.copy().items():
         print(f'Update resptrs of resource "{resource.id}"...')
```
irinaschubert, line 1088: Shouldn't this be printed only if verbose=True? I'm not sure, but I feel that this could be very verbose in case of lots of resptrs..?

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +918,340 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+        for prop, xmldict in propdict.items():
+            if len(xmldict) > 0:
+                if res not in nonapplied_xml_texts:
+                    nonapplied_xml_texts[res] = {}
+                nonapplied_xml_texts[res][prop] = xmldict
+
+    return nonapplied_xml_texts
+
+
+def update_stashed_resptr_props(
```
irinaschubert, line 1067: At first glance the name of this method seems misleading to me because the stashed resptr props are not updated, instead the resources are updated with the resptr props. Or am I mistaken?

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +918,340 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+                        "knora-api:textValueHasMapping": {
+                            '@id': 'http://rdfh.ch/standoff/mappings/StandardMapping'
+                        }
+                    },
+                    "@context": existing_resource['@context']
+                }
+                jsondata = json.dumps(jsonobj, indent=4, separators=(',', ': '), cls=KnoraStandoffXmlEncoder)
+
+                # execute API call
+                answer = None
```
irinaschubert, line 1034: I would call this `response` instead of `answer`.

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +918,340 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+                if not old_xmltext:
+                    continue
+
+                # strip all xml tags from the old xmltext, so that the pure text itself remains
+                pure_text = re.sub(r'(<\?xml.+>\s*)?<text>\s*(.+)\s*<\/text>', r'\2', old_xmltext)
+
+                # if the pure text is a hash, the replacement must be made
+                if pure_text not in hash_to_value:
+                    continue
+                new_xmltext = hash_to_value[pure_text]
```
irinaschubert, line 1011: Where does this hash come from?

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +918,340 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+            except BaseError:
+                break
+
+        if not resclass_instance or not resclass_instance.iri:
+            print(f"ERROR while trying to create resource '{resource.label}' ({resource.id}). ")
             failed_uploads.append(resource.id)
-            continue
+        else:
+            id2iri_mapping[resource.id] = resclass_instance.iri
+            print(f"Created resource '{resclass_instance.label}' ({resource.id}) with IRI '{resclass_instance.iri}'")
```
irinaschubert, line 967: Should this always be printed or only in case of verbose=True? It could be very verbose in case of hundreds or thousands of resources.

knora/dsplib/utils/xml_upload.py
```diff
@@ -843,65 +769,147 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+            time.sleep(1)
+            continue
+        except RequestException:
+            print(f'{datetime.now().isoformat()}: Try reconnecting to DSP server (SIPI)...')
+            time.sleep(1)
+            continue
+    if img:
+        return img
+    else:
+        return sipi_server.upload_bitstream(filepath)
```
irinaschubert, line 881: It looks strange to me that you make the method call again here. You could store the response in a variable and then just use it here. There is assumably a pattern with which you could differentiate between the success/failure case.

test/e2e/test_tools.py
```diff
@@ -176,38 +176,45 @@ def test_create_ontology(self) -> None:
...
                 mapping_file = mapping.name
```
irinaschubert, line 195: When would you have a file that starts with "id2iri_test-data_mapping_" which is older than 15 seconds?

test/e2e/test_tools.py
```diff
@@ -176,38 +176,45 @@ def test_create_ontology(self) -> None:
...
+        self.assertNotEqual(mapping_file, '')
 
+        id2iri_replaced_xml_filename = 'testdata/tmp/_test-id2iri-replaced.xml'
         id_to_iri(xml_file='testdata/test-id2iri-data.xml',
```
irinaschubert, line 199: Could you move this test to a separate test? Ideally, they shouldn't depend on each other (unless you want to test this dependency).

irinaschubert (changes requested):
This loos like a big improvement on the stability of dsp-tools! I added some comments, most of them are minor. I would appreciate if you could check if the one case where you call a method as a return value could be improved.

Apart from that some general remarks/possible improvements:
- please don't use "knora" anywhere in new code. I know that we have it a lot everywhere. But we want to get rid of this name and use DSP instead. "knora" is deprecated.
- the whole file is pretty long now. I would consider splitting it up into smaller pieces if that makes sense.
- I haven't checked properly but it seems that there are still some methods that aren't tested. Also, I would recommend to test negative cases where you actually have an exception or an error. IMO, also the log files that are created should be checked (the content not just their existence).

knora/dsplib/utils/xml_upload.py
```diff
@@ -843,65 +769,147 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+    failed_uploads: list[str] = []
 
-    # get the project information and project ontology from the server
-    project = ResourceInstanceFactory(con, shortcode)
+    try:
+        id2iri_mapping, failed_uploads = upload_resources(verbose, resources, imgdir, sipi_server, permissions_lookup,
+                                                          resclass_name_2_type, id2iri_mapping, con, failed_uploads)
+    except BaseException as err:
+        handle_upload_error(err, input_file, id2iri_mapping, failed_uploads, stashed_xml_texts, stashed_resptr_props)
+        exit(1)
```
BalduinLandolt, line 814: I think here you should do `raise err` rather than `exit(1)`, otherwise the information contained by the exception is lost to the user. This is essentially "hiding" an exception, which you should only do when you have successfully handled it.

knora/dsplib/utils/xml_upload.py
```diff
@@ -843,65 +769,147 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+
+    # update the resources with the stashed XML texts
+    nonapplied_xml_texts = {}
+    if len(stashed_xml_texts) > 0:
+        try:
+            nonapplied_xml_texts = update_stashed_xml_texts(verbose, id2iri_mapping, con, stashed_xml_texts)
+        except BaseException as err:
+            handle_upload_error(err, input_file, id2iri_mapping, failed_uploads, stashed_xml_texts, stashed_resptr_props)
+            exit(1)
```
BalduinLandolt, line 823: as above. Also, `except BaseException` is (practically) equivalent and imo less clear (and less common) than `except Exception`

knora/dsplib/utils/xml_upload.py
```diff
@@ -843,65 +769,147 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
-        permissions_lookup[key] = perm.get_permission_instance()
+    if failed_uploads:
+        print(f"Could not upload the following resources: {failed_uploads}")
+        exit_code = 1
 
-    # create a dictionary to look up resource classes
-    resclass_name_2_type: dict[str, type] = {}
-    for res_class_name in project.get_resclass_names():
-        resclass_name_2_type[res_class_name] = project.get_resclass_type(res_class_name)
+    exit(exit_code)
```
BalduinLandolt, line 849: I'm not a fan of this: Like this, you force-end the program in any case. So it's no longer possible to call this method from any code that should resume after this has been executed.  
If something non-recoverable happens, it's legit to forcefully quit the program; but if all goes as it should, it should just run through and end by itself. Otherwise the code is no longer reusable within or outside of the codebase.

knora/dsplib/utils/xml_upload.py
```diff
@@ -843,65 +769,147 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
 
-    res_iri_lookup: dict[str, str] = {}
 
-    failed_uploads = []
+def try_sipi_upload(sipi_server: Sipi, filepath: str) -> dict[Any, Any]:
```
BalduinLandolt, line 852: this should typically be protected/private, I would say

knora/dsplib/utils/xml_upload.py
```diff
@@ -843,65 +769,147 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+    """
+
+    img = None
+    for _ in range(5):
+        try:
+            img = sipi_server.upload_bitstream(filepath)
+            break
+        except ConnectionError:
+            print(f'{datetime.now().isoformat()}: Try reconnecting to DSP server (SIPI)...')
+            time.sleep(1)
```
BalduinLandolt, line 872: might it maybe make more sense to sleep a bit longer? Now you sleep 5*1 seconds, and not necessarily all network issues recover within 5 seconds, right?

You could even have a growing timeout that uses the index as a loop:
```python
for t in range(5):
    ...
    sleep(2**i)
```
this would sleep for 1, 2, 4, 8 and 16 seconds (total 31 seconds)

knora/dsplib/utils/xml_upload.py
```diff
@@ -843,65 +769,147 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+
+    Returns:
+        answer from knora
+    """
+
+    img = None
+    for _ in range(5):
+        try:
+            img = sipi_server.upload_bitstream(filepath)
+            break
```
BalduinLandolt, line 869: if that succeeds, you could just return the image and not care for breaking and checking for the image

knora/dsplib/utils/xml_upload.py
BalduinLandolt: also, you have already done the retry thing above, so either it has worked, or something is really wrong (and if you return right away, when upload succeeds, already coming here means that the upload has failed for good)

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +918,340 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
-            internal_file_name_bitstream = img['uploadedFiles'][0]['internalFilename']
-            resource_bitstream = resource.get_bitstream(internal_file_name_bitstream, permissions_lookup)
+            try:
+                img = try_sipi_upload(sipi_server=sipi_server, filepath=os.path.join(imgdir, resource.bitstream.value))
+                internal_file_name_bitstream = img['uploadedFiles'][0]['internalFilename']
+                resource_bitstream = resource.get_bitstream(internal_file_name_bitstream, permissions_lookup)
+            except BaseError:
+                print(f"ERROR while trying to create resource '{resource.label}' ({resource.id}). ")
+                failed_uploads.append(resource.id)
+                continue
```
BalduinLandolt, line 931: here again, all exceptions are caught, but not propagated/handled, which is bad practice.

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +918,340 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
-            print(f"ERROR while trying to create resource '{resource.label}' ({resource.id}). "
-                  f"The error message was: {err.message}")
-            failed_uploads.append(resource.id)
-            continue
-        except Exception as exception:
-            print(f"EXCEPTION while trying to create resource '{resource.label}' ({resource.id}). "
-                  f"The exception message was: {exception}")
+        # create the resource in knora
+        resclass_instance: ResourceInstance = None
+        for _ in range(5):
```
BalduinLandolt, line 937: this whole bit is essentially the same as above, so this could probably be refactored to get rid of the duplication.

knora/dsplib/utils/xml_upload.py
BalduinLandolt: yes

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +918,340 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+                        print(f'{datetime.now().isoformat()}: Try reconnecting to DSP server...')
+                        time.sleep(1)
+                        continue
+                    except RequestException:
+                        print(f'{datetime.now().isoformat()}: Try reconnecting to DSP server...')
+                        time.sleep(1)
+                        continue
+                    except BaseError:
+                        print(f'ERROR while updating the xml text of "{link_prop.name}" of resource "{resource.id}"')
+                        break
```
BalduinLandolt, line 1050: again, blanco error catching and not handling it

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +918,340 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+                        print(f'{datetime.now().isoformat()}: Try reconnecting to DSP server...')
+                        time.sleep(1)
+                        continue
+                    except RequestException:
+                        print(f'{datetime.now().isoformat()}: Try reconnecting to DSP server...')
+                        time.sleep(1)
+                        continue
+                    except BaseError:
+                        print(f'ERROR while updating the resptr prop of "{link_prop.name}" of resource "{resource.id}"')
+                        break
```
BalduinLandolt, line 1121: as above

BalduinLandolt (changes requested):
I'm still not happy with the error handling part:  
`except BaseException` (or as I prefer `except Exception`) should ever only be used when you want to do something before you kill the program. And then you should normally raise the same error again, so that the stack trace doesn't get lost.  
You should only handle exceptions when you know how to handle them.  
Also, remember that it shoud be possible at any time, to do e.g. a keyboard interrupt.

Additionally, I think you could make more liberal use of the `finally` block: It wouldn't hurt to just write the IRI mapping in a finally at the end of every attempt to upload something.

Some more thoughts, more in the direction of refactoring the code to make it more maintainable:  
- I agree that the file should be split up in multiple files (ideally I'd put it in a submodule `xmlupload` so that everything is together that belongs together).
- I think you could go even further in splitting those long methods up into shorter methods (also Sonar gives you codesmells where some of those functions have a complexity score of 55 whereas no more than 15 is recommended; and some linting tools don't allow functions that have more than 15 lines - I wouldn't be so strict, but I still think that methods should be kept as simple as possible).
- relating to that: in python code, by convention, protected methods have a name starting with an underscore and private methods start with double-underscore. It's generally good practice to do that for every function that is specific for the logic of the present module, so that you don't expect it to be called from a different file than the one you're in. So ideally `xml_upload()` which is the public method that you expose, should be short and just call  bunch of protected/private methods that again delegate the work, until you get a nice granularity.

knora/dsplib/utils/xml_upload.py
jnussbaum: `Exception` is the base class for the non-system-exiting exceptions. But I want to catch the system-exiting ones too, so it's necessary to catch `BaseException`.

knora/dsplib/utils/xml_upload.py
jnussbaum: handle_upload_error() will now reraise the exception, instead of `exit(1)`

knora/dsplib/utils/xml_upload.py
jnussbaum: Thanks for the hint. I replaced it with 
```
if exit_code == 1:
    exit(1)
```

knora/dsplib/utils/xml_upload.py
jnussbaum: to be concrete: should I rename it to `def _try_sipi_upload()`?

knora/dsplib/utils/xml_upload.py
jnussbaum: Don't you think there is a danger that big uploads that take weeks under normal circumstances will become even longer? Initially I wanted to sleep longer, but then this concern was raised. I find it difficult to say if the concern is a real danger.

knora/dsplib/utils/xml_upload.py
BalduinLandolt: Ok, then I don't understand why you would want to catch system-exiting exceptions at this point (which in my opinion should never be done without re-raising them right away)

knora/dsplib/utils/xml_upload.py
BalduinLandolt: I would do it, but it's a matter of taste, so definitely no requirement for getting this merged :)

knora/dsplib/utils/xml_upload.py
BalduinLandolt: hard to say... but then, this should not delay the upload longer than the network outage lasts, right?

knora/dsplib/utils/xml_upload.py
jnussbaum: I wonder how I could refactor this... I don't see a simple solution.

knora/dsplib/utils/xml_upload.py
jnussbaum: This results in 1 line per resource: either success or error. This was present since the early days, and is actually very helpful to see how the xmlupload makes progress.

knora/dsplib/utils/xml_upload.py
jnussbaum: From `stash_circular_references()`, where all XML texts with a salsah-link that leads to circularity are replaced by their hash, and the original text is stashed. Now, the inverse is done: All XML texts that were stashed are taken out, and put to the place where they came from. For this, they are identified by their hash.

knora/dsplib/utils/xml_upload.py
BalduinLandolt: Ah... I'm sorry! I misread this as `BaseException`. Never mind my comment in this case. (But those two are named too similar to not mix them up...)

knora/dsplib/utils/xml_upload.py
BalduinLandolt: in short: by splitting everything up into smaller methods: then you could have a `retry()` method that takes a function as a parameter, which it then calls repeatedly.

But yes, it's not trivial - and certainly not necessary for this PR.

irinaschubert (approved):
LGTM, thanks!

knora/dsplib/utils/xml_upload.py
```diff
@@ -843,65 +770,111 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
-    project = ResourceInstanceFactory(con, shortcode)
+    try:
+        id2iri_mapping, failed_uploads = upload_resources(verbose, resources, imgdir, sipi_server, permissions_lookup,
+                                                          resclass_name_2_type, id2iri_mapping, con, failed_uploads)
+    except BaseException as err:
+        handle_upload_error(err, input_file, id2iri_mapping, failed_uploads, stashed_xml_texts, stashed_resptr_props)
+
+    # update the resources with the stashed XML texts
+    nonapplied_xml_texts = {}
+    if len(stashed_xml_texts) > 0:
```
BalduinLandolt, line 818: I think you could simply put `if stashed_xml_texts: ...`

BalduinLandolt: or even move that check inside the function, so you don't need the if clause here

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +883,435 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
         if resource.ark:
             resource_iri = convert_ark_v0_to_resource_iri(resource.ark)
 
+        # in case of a multimedia resource: upload the multimedia file
         resource_bitstream = None
         if resource.bitstream:
-            img = sipi_server.upload_bitstream(os.path.join(imgdir, resource.bitstream.value))
+            img: dict[Any, Any] = try_network_action(
```
BalduinLandolt, line 889: the typing here is far from ideal (as it is both unspecific and not correct, it could also be None), but that can be improved at a later date.

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +883,435 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+                object=sipi_server,
+                method='upload_bitstream',
+                args={'filepath': os.path.join(imgdir, resource.bitstream.value)},
```
BalduinLandolt, line 892: by convention, `args` is a list of arguments, and `kwargs` is a dictionary of keyword-arguments, so this should be called `kwargs`

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +883,435 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
         if resource.ark:
             resource_iri = convert_ark_v0_to_resource_iri(resource.ark)
 
+        # in case of a multimedia resource: upload the multimedia file
         resource_bitstream = None
         if resource.bitstream:
-            img = sipi_server.upload_bitstream(os.path.join(imgdir, resource.bitstream.value))
+            img: dict[Any, Any] = try_network_action(
+                object=sipi_server,
+                method='upload_bitstream',
```
BalduinLandolt, line 891: if you pass the function rather than a string here, you can call it directly, instead of having to find out, what function the string points to

BalduinLandolt: I see now, why you used the string. Doesn't it work, to just pass `sipi_server.upload_bitstream` as a callable?

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +883,435 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+    After all resources are uploaded, the stashed resptr props must be applied to their resources in DSP.
 
-    # update the resources with the stashed resptrs
-    if len(stashed_resptr_props) > 0:
-        print('Update the stashed resptrs...')
-    for resource, prop_2_resptrs in stashed_resptr_props.items():
-        print(f'Update resptrs of resource "{resource.id}"...')
-        res_iri = res_iri_lookup[resource.id]
+    Args:
+        verbose: bool
```
BalduinLandolt, line 1048: bool is a type, not a docstring description

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +883,435 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+            if object and isinstance(method, str):
+                if not args:
+                    return getattr(object, method)()
+                else:
+                    return getattr(object, method)(**args)
+            else:
+                if not args:
+                    return method()
+                else:
+                    return method(**args)
```
BalduinLandolt, line 1132: see my comment above regarding this: if it works to pass a class method as a function to this method, then you wouldn't need the `object` and distinction between string and callable, so this if-else could be removed.

And I think you can pass an empty dictionary as kwargs, so if you'd simply default it to `{}`, you could get rid of the inner if-else too.

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +883,435 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+                '(Not listed are the stripped resources that haven\'t been created in DSP yet.) \n')
+        for res, props in stashed_xml_texts.items():
+            f.write(f'\n{res.id}')
+            f.write('\n' + '=' * len(res.id))
+            for prop, stashed_texts in props.items():
+                if len(stashed_texts) > 0:
+                    f.write(f'\n{prop.name}')
+                    f.write('\n' + '-' * len(prop.name))
+                    for hash, standoff in stashed_texts.items():
+                        f.write(f'\ntext with hash {hash}:\n{str(standoff).strip()}\n')
```
BalduinLandolt, line 1281: is there a reason to write this to a .txt file?
If this was written to e.g. a JSON file, it would be much easier to get the information out again, especially in the case of a very big upload.

knora/dsplib/utils/xml_upload.py
```diff
@@ -910,94 +883,435 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+
+    Args:
+        stashed_resptr_props: all resptr props that have been stashed
+        timestamp_str: timestamp for log file identification
+
+    Returns:
+        None
+    """
+
+    filename = f'stashed_resptr_properties_{timestamp_str}.txt'
```
BalduinLandolt, line 1300: as above: why TXT not JSON?

PR #200, opened 2022-06-07
chore: unpin dependency versions (DEV-983)

jnussbaum:
resolves DEV-983


BalduinLandolt (approved):
I still think this should not be necessary... but as long as it works, I'm ok with this solution :)

PR #203, opened 2022-06-23
chore(xmlupload): refactor xmlupload, add unittests (DEV-1043)

jnussbaum:
resolves DEV-1043


BalduinLandolt (commented):
How would you feel about moving all the newly created and xml-related files in the models folder to a subfolder `xml`?

knora/dsplib/models/projectContext.py
```diff
@@ -0,0 +1,50 @@
...
+
+    _projects: list[Project]
+    _project_map: dict[str, str]  # dictionary of (project name:project IRI) pairs
+    _inv_project_map: dict[str, str]  # dictionary of (project IRI:project name) pairs
+    _groups: Optional[list[Group]]
+    _group_map: Optional[dict[str, str]]
+    _shortcode: Optional[str]
+    _project_name: Optional[str]
+
+    def __init__(self, con: Connection, shortcode: Optional[str] = None):
```
irinaschubert, line 20: docstring is missing, see the other classes

jnussbaum:
I thought about bringing some order into all these models, but I'm unsure about it. Intuitively, I don't like the idea of a `xml` subfolder. (The main reason why I am reluctant about reordering them, is because I don't know which classes belong together). So it seemed best to me to leave them all on top level.

BalduinLandolt:
> I thought about bringing some order into all these models, but I'm unsure about it. Intuitively, I don't like the idea of a `xml` subfolder. (The main reason why I am reluctant about reordering them, is because I don't know which classes belong together). So it seemed best to me to leave them all on top level.

fair enough

PR #210, opened 2022-08-05
fix: catch network interruptions during onto creation (DEV-1073)

jnussbaum:
resolves DEV-1073


docs/dsp-tools-usage.md
```diff
@@ -35,10 +35,10 @@ The following options are available:
 - `-s` | `--server` _server_: URL of the DSP server (default: 0.0.0.0:3333)
 - `-u` | `--user` _username_: username used for authentication with the DSP API (default: root@example.com)
 - `-p` | `--password` _password_: password used for authentication with the DSP API (default: test)
-- `-V` | `--validate`: If set, only the validation of the JSON file is performed.
-- `-l` | `--lists`: If set, only the lists are created using a [simplified schema](./dsp-tools-create.md#lists). Please
-  note that in this case the project must already exist.
-- `-v` | `--verbose`: If set, some information about the progress is printed to the console.
+- `-V` | `--validate-only`: If set, only the validation of the JSON file is performed.
+- `-l` | `--lists-only`: If set, only the lists are created. Please note that in this case the project must already exist.
```
jnussbaum, line 39: small change to make it more clear

docs/dsp-tools-usage.md
```diff
@@ -35,10 +35,10 @@ The following options are available:
...
+- `-v` | `--verbose`: If set, more information about the progress is printed to the console.
+- `-d` | `--dump`: If set, dump test files for DSP-API requests.
```
jnussbaum, line 41: This option already existed in the code, but was not documented

knora/dsp_tools.py
```diff
@@ -142,31 +140,27 @@ def program(user_args: list[str]) -> None:
...
                              dump=args.dump)
         else:
-            if args.validate:
-                if validate_ontology(args.datamodelfile):
+            if args.validate_only:
+                if validate_project(args.datamodelfile):
+                    print('Data model is syntactically correct and passed validation.')
                     exit(0)
-                else:
-                    exit(1)
```
jnussbaum, line 161: The behaviour of `validate_project()` changed: It returns True or raises an Error with detailed explanation. So, it's not necessary any more to `exit()`.

knora/dsplib/models/group.py
```diff
@@ -240,23 +240,18 @@ def delete(self):
         return Group.fromJsonObj(self._con, result['group'])
 
     @staticmethod
-    def getAllGroups(con: Connection) -> Optional[list[Group]]:
-        try:
-            result = con.get(Group.ROUTE)
-            return [Group.fromJsonObj(con, group_item) for group_item in result["groups"]]
-        except BaseError:
-            # return None if no groups are found or an error happened
-            return None
```
jnussbaum, line 249: I want the error to escalate, so that I can handle it in the higher-level methods

knora/dsplib/utils/excel_to_json_lists.py
```diff
@@ -11,19 +10,59 @@
...
 
-# Global variables used to ensure that there are no duplicate node names
 list_of_lists_of_previous_cell_values: list[list[str]] = []
+"""Module level variable used to ensure that there are no duplicate node names"""
+
 list_of_previous_node_names: list[str] = []
+"""Module level variable used to ensure that there are no duplicate node names"""
+
 
+def expand_lists_from_excel(
```
jnussbaum, line 22: I moved this method here, because it thematically belongs here. Before, it was in the file `expand_all_lists.py`, which contained nothing else than just this method. I think it makes more sense to move it here and delete the other file.

knora/dsplib/utils/excel_to_json_lists.py
```diff
@@ -11,19 +10,59 @@
...
+
+def _get_values_from_excel(
     excelfiles: dict[str, Worksheet],
     base_file: dict[str, Worksheet],
     parentnode: dict[str, Any],
     row: int,
     col: int,
-    preval: list[str]
+    preval: list[str],
+    verbose: bool = False
```
jnussbaum, line 65: This method is used 
 - by the `create` command, where it shouldn't print too much by default
 - during the `excel` command, where it should print every created node by default

The `verbose` switch seemed to be a good solution for this.

knora/dsplib/utils/excel_to_json_lists.py
```diff
@@ -277,66 +324,59 @@ def validate_list_with_schema(json_list: str) -> bool:
...
-    # create a list with all excel files from the path provided by the user
-    excel_files = [filename for filename in glob.iglob(f'{excelfolder}/*.xlsx')
-                   if not os.path.basename(filename).startswith('~$')
-                   and os.path.isfile(filename)]
+        raise BaseError(f"ERROR: {excelfolder} is not a directory.")
 
-    # print the files that can be used
-    print('Found the following files:')
-    for file in excel_files:
-        print(file)
```
jnussbaum, line 329: This will be printed in the higher-level methods, because depending on the caller, a different print is needed.

knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -13,598 +13,618 @@
...
 from knora.dsplib.models.user import User
-from knora.dsplib.utils.expand_all_lists import expand_lists_from_excel
+from knora.dsplib.utils.excel_to_json_lists import expand_lists_from_excel
 from knora.dsplib.utils.onto_create_lists import create_lists
-from knora.dsplib.utils.onto_validate import validate_ontology
+from knora.dsplib.utils.onto_validate import validate_project
+from knora.dsplib.utils.shared_methods import login, try_network_action
 
 
-def login(server: str, user: str, password: str) -> Connection:
```
jnussbaum, line 21: I moved the method "login()" to "shared_methods.py"

knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -13,598 +13,618 @@
...
-        group_status_bool = True
-        if isinstance(group_status, str):
-            group_status_bool = json.loads(group_status.lower())  # lower() converts string to boolean
-
-        # check if selfjoin is defined, set default value if not
-        group_selfjoin: Optional[str] = group.get("selfjoin")
-        group_selfjoin_bool = False
-        if isinstance(group_selfjoin, str):
-            group_selfjoin_bool = json.loads(group_selfjoin.lower())  # lower() converts string to boolean
-
```
jnussbaum, line 146: I simplified these two radically: They are now directly in the constructor call below, as one-liners

knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -13,598 +13,618 @@
...
+            print(f"\tWARNING: User '{username}' already exists on the DSP server. Skipping...")
+            overall_success = False
+            continue
         except BaseError:
             pass
-        if maybe_user:
-            print(f"WARN User '{username}' already exists. Skipping...")
-            continue
 
-        sysadmin = False
```
jnussbaum, line 197: From here on, GitHub's diff renderling is a bit hard to understand. In PyCharm, it was more understandable. Basically, there are 4 things done inside the big loop `for user in users:`
 -  skip the user if he already exists
 -  if "groups" is provided, add user to the group(s)
 - if "projects" is provided, add user to the project(s)
 - create the user

I tried to simplify these four steps, to minimize the indentation depth, and to improve code understandability.

The diff might be hard to understand, but the code should be better understandable than before.

knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -13,598 +13,618 @@
...
-    project = None
+    # if project exists, update it, otherwise create it
+    project_local = Project(con=con, shortcode=project_definition["project"]["shortcode"])
     try:
-        project = Project(con=con, shortcode=data_model["project"]["shortcode"]).read()
+        project_remote: Project = try_network_action(
+            object=project_local,
+            method="read",
+            error_message_on_failure=""
+        )
```
jnussbaum, line 440: If the project can't be read, it raises a BaseError without message --> create the project.
If the reading suceeds without error --> update project

knora/dsplib/utils/shared_methods.py
```diff
@@ -0,0 +1,99 @@
...
+            continue
+        except RequestException:
+            print(f'{datetime.now().isoformat()}: Try reconnecting to DSP server, next attempt in {2 ** i} seconds...')
+            time.sleep(2 ** i)
+            continue
+        except BaseError as err:
+            if re.search(r'try again later', err.message) or re.search(r'status code=5\d\d', err.message):
+                print(f'{datetime.now().isoformat()}: Try reconnecting to DSP server, next attempt in {2 ** i} seconds...')
+                time.sleep(2 ** i)
+                continue
```
jnussbaum, line 84: These lines are new: If the API's response says that it is a temporary error, it is worth retrying. All other base errors are permanent and will escalate.

knora/dsplib/utils/xml_upload.py
```diff
@@ -5,14 +5,12 @@
 import json
```
jnussbaum, line 5: Why did I change this file?
 - I moved `_try_network_action()` to `shared_methods.py` 
 - I changed the behaviour of `_try_network_action()`: If the connectivity issues remain, it won't return None anymore, but it will escalate the error that it received from the method that failed. For this reasons, I have to embed all usages of `_try_network_action()` into a try-except block

test/e2e/test_tools.py
```diff
@@ -21,8 +20,8 @@ class TestTools(unittest.TestCase):
     password = 'test'
     imgdir = '.'
     sipi = 'http://0.0.0.0:1024'
-    test_onto_file = 'testdata/test-onto.json'
-    test_list_file = 'testdata/test-list.json'
```
jnussbaum, line 25: I got rid of the lists.json output file

test/e2e/test_tools.py
```diff
@@ -68,23 +67,23 @@ def test_get(self) -> None:
         group_descriptions_received = []
         group_selfjoin_received = []
         group_status_received = []
-        for group in groups_expected:
+        for group in sorted(groups_expected, key=lambda x: x["name"]):
```
jnussbaum, line 70: I added some groups with more diverse characteristics. It is now necessary to sort the groups by name

test/e2e/test_tools.py
```diff
@@ -21,8 +20,8 @@ class TestTools(unittest.TestCase):
     password = 'test'
     imgdir = '.'
     sipi = 'http://0.0.0.0:1024'
-    test_onto_file = 'testdata/test-onto.json'
-    test_list_file = 'testdata/test-list.json'
+    test_project_file = 'testdata/test-project-systematic.json'
+    test_project_minimal_file = 'testdata/test-project-minimal.json'
```
jnussbaum, line 24: It is not only important to have a systematic test project, but also to have a minimal one. During my work, I accidentally spotted a bug which was never detected. With the new minimal test project, it would have been detected.

knora/dsplib/utils/onto_create_lists.py
```diff
@@ -1,141 +1,165 @@
...
+        comments=node.get("comments"),
+        name=node["name"],
+        parent=parent_node
+    )
+    try:
+        new_node = try_network_action(
+            object=new_node,
+            method="create",
+            error_message_on_failure=f"ERROR while trying to create list node '{node['name']}'."
+        )
```
BalduinLandolt, line 47: I think this method would be much cleaner if it would always accept only a callable. And in instances like this, where you can't simply pass a function to it, you'd be using a lambda.  
So this would look as follows:

```python
new_node = try_network_action(lambda: new_node.create(), f"ERROR ...")
```

knora/dsplib/utils/onto_create_ontology.py
```diff
@@ -13,598 +13,618 @@
...
+    overall_success = True
+    current_project_groups: dict[str, Group] = {}
+    try:
+        remote_groups: list[Group] = try_network_action(
+            object=Group,
+            method="getAllGroupsForProject",
+            kwargs={"con": con, "proj_shortcode": project.shortcode},
+            error_message_on_failure="WARNING: Unable to check if group names are already existing on DSP server, "
+                                     "because it is not possible to retrieve the remote groups from DSP server."
+        )
```
BalduinLandolt, line 105: another advantage of using lambdas here, would be that you don't need the kwargs anymore:

```python
remote_groups: list[Group] = try_network_action(
    method = lambda: Group.getAllGroupsForProject(con=con, proj_shortcode=project.shortcode),
    msg = "WARNING ..."
)
```

knora/dsplib/utils/shared_methods.py
```diff
@@ -0,0 +1,99 @@
...
+        error_message_on_failure="ERROR: Cannot login to DSP server"
+    )
+    return con
+
+
+def try_network_action(
+    error_message_on_failure: str,
+    method: Union[str, Callable[..., Any]],
+    object: Optional[Any] = None,
+    kwargs: Optional[dict[str, Any]] = None
```
BalduinLandolt, line 38: as mentioned above, I think this function would be much neater if it only accepted a callable and a message (which I would maybe call `failure_msg` to make function calls a bit slimmer); and in those "special" case where you need the object or kwargs, I'd solve that with lambdas in the function call.

knora/dsplib/utils/shared_methods.py
```diff
@@ -0,0 +1,99 @@
...
+        except RequestException:
+            print(f'{datetime.now().isoformat()}: Try reconnecting to DSP server, next attempt in {2 ** i} seconds...')
+            time.sleep(2 ** i)
+            continue
+        except BaseError as err:
+            if re.search(r'try again later', err.message) or re.search(r'status code=5\d\d', err.message):
+                print(f'{datetime.now().isoformat()}: Try reconnecting to DSP server, next attempt in {2 ** i} seconds...')
+                time.sleep(2 ** i)
+                continue
+            if hasattr(err, 'message'):
```
BalduinLandolt, line 85: what kind of exception does not have a message?

BalduinLandolt (approved):
Not an easy one to review... but I think you improved a lot. I hope I didn't miss anything important.

knora/dsplib/utils/onto_create_lists.py
jnussbaum: Thank you very much, this is a great improvement! I tried it out, and it works :-) You always have to use a lambda, so that the method is not executed too early. This is really a big improvement!

knora/dsplib/utils/shared_methods.py
jnussbaum: sometimes it's called differently... `msg` or so...

knora/dsplib/utils/onto_create_lists.py
BalduinLandolt: cool :)

PR #213, opened 2022-08-10
test: add all supported file extensions to testdata and docs (DEV-1187)

jnussbaum:
resolves DEV-1187


docs/dsp-tools-create-ontologies.md
```diff
@@ -797,13 +797,13 @@ The following base resources can be used as super-resource:
...
-- `StillImageRepresentation`: A resource representing an image
-- `TextRepresentation`: A resource representing a text
+- `ArchiveRepresentation`: A resource representing an archive file (e.g. ZIP)
 - `AudioRepresentation`: A resource representing an audio file
 - `DDDRepresentation`: A resource representing a 3-D representation (not yet implemented)
 - `DocumentRepresentation`: A resource representing an opaque document (e.g. a PDF)
 - `MovingImageRepresentation`: A resource representing a video
-- `ArchiveRepresentation`: A resource representing an archive file (e.g. ZIP)
+- `StillImageRepresentation`: A resource representing an image
+- `TextRepresentation`: A resource representing a text
```
BalduinLandolt, line 806: I see you ordered this alphabetically. Previously it was supposed to be (but wasn't really) ordered by importance/similarity, I think.  
E.g. it would make sense to me to group Document and Text representation; to have StillImage further up than Audio and Video; and especially to have 3D at the very bottom, as it's not implemented.  
But that's just a thought. (and I'm not saying the previous order was sensible...)

test/e2e/test_tools.py
```diff
@@ -197,7 +197,7 @@ def test_xml_upload(self) -> None:
         self.assertTrue(result)
 
         mapping_file = ''
-        for mapping in [x for x in os.scandir('.') if x.name.startswith('id2iri_test-data_mapping_')]:
+        for mapping in [x for x in os.scandir('.') if x.name.startswith('id2iri_test-data-systematic_mapping_')]:
```
BalduinLandolt, line 200: this could be done a bit more neatly with a `for mapping in glob.glob('id2iri_test-data-systematic_mapping_*'):` (or something along those lines)

BalduinLandolt (approved):
So far so good, thanks!

Some points: 
- I did not check all the bitstream files' contents. I trust you made sure they are all valid and contain the expected contents.
- I think you're still missing `.iso`, is that intentional?
- maybe it would be a good idea to add a `README.md` file to the `testdata/` folder, where you list all the files and mention that they are either property of the DaSCH, or if that's not the case, where they are from and how they are licensed.

PR #214, opened 2022-08-11
feat: add isSequenceOf (DEV-746)

jnussbaum:
resolves DEV-746


jnussbaum:
@BalduinLandolt: This is now ready for review. It was a surprisingly low-effort task, I thought there would be more to do.  
There is one decision that I took that I would like to double-check with you: In the schema and the docs, it is now prescribed that if a property has the object `InvervalValue`, it must have super `hasSequenceBounds`. This is because it doesn't make sense to me to have an interval value on another resource than an audio/video sequence. Is that correct?

BalduinLandolt:
@jnussbaum this is certainly not how `IntervalValue` was intended when it was introduced in the API. But personally I don't know of many use cases either, so I don't really mind this restriction. We'll have to see if it needs to be changed in the future...

docs/dsp-tools-create-ontologies.md
```diff
@@ -755,6 +833,81 @@ This property is actually very similar to a simple text field.
...
+
+*gui-elements/gui_attributes*:
+
+- `Searchbox`: The only GUI element for _isSequenceOf_. Allows searching resources by entering the target resource name.
+    - _gui_attributes_:
+        - `numprops=integer` (optional): Number of search results to be displayed
+
+*Example:*
+
+```json
```
BalduinLandolt, line 856: not sure why Git Diff makes things in this file red? Maybe you can have a quick look to make sure there's nothing wrong here?

BalduinLandolt (approved):
looks good to me, I hope I didn't miss anything

docs/dsp-tools-create-ontologies.md
jnussbaum: found the culprit. The JSON in the example missed a closing bracket.

PR #219, opened 2022-08-19
feat: add module csv2xml to convert tabular data to DSP-XML (DEV-134)

jnussbaum:
resolves DEV-134


test/unittests/test_excel2xml.py
```diff
@@ -0,0 +1,583 @@
...
+                returned = f.read()
+                self.assertEqual(returned, expected, msg=f"Failed with extension {ext}")
+        if os.path.isfile("excel2xml-output-data.xml"):
+            os.remove("excel2xml-output-data.xml")
+
+    @pytest.mark.filterwarnings("ignore")
+    def test_excel2xml_sample_script(self) -> None:
+        with open("docs/assets/templates/excel2xml_script.py") as f:
+            template_script = f.read()
+            exec(template_script, {})
```
jnussbaum, line 563: I thought it would be good to test the sample script offered for download. But could it be a security risk to execute code from another file?

knora/dsplib/models/value.py
```diff
@@ -279,10 +279,10 @@ def __init__(self,
         self._y1 = None if dp[2] is None else int(dp[2].strip('-: '))
         self._m1 = None if dp[3] is None else int(dp[3].strip('-: '))
         self._d1 = None if dp[4] is None else int(dp[4].strip('-: '))
-        self._e2 = 'CE' if dp[5] is None else dp[5].strip('-: ')
```
jnussbaum, line 282: The capturing groups in the above regex changed.

knora/dsplib/models/value.py
```diff
@@ -636,7 +636,8 @@ def __init__(self,
                  iri: Optional[str] = None,
                  ark_url: Optional[str] = None,
                  vark_url: Optional[str] = None):
-        m = re.match("^(http)s?://([\\w\\.\\-~]+)?(:\\d{,6})?(/[\\w\\-~]+)*(#[\\w\\-~]*)?", str(value))
```
jnussbaum, line 639: The old URI validation was quite far away from what an URI actually is, and from what the API allows

test/e2e/test_tools.py
```diff
@@ -24,16 +24,17 @@ class TestTools(unittest.TestCase):
     test_project_minimal_file = 'testdata/test-project-minimal.json'
     test_data_file = 'testdata/test-data-systematic.xml'
 
-    def setUp(self) -> None:
```
jnussbaum, line 27: This has nothing to do with the core of this PR, but I came across it. It's not necessary to do the setup and teardown before and after each method call.

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+    Args:
+        path: path to a valid file that will be uploaded
+        permissions: permissions string
+        calling_resource: the name of the parent resource (for better error messages)
+
+    Examples:
+        >>> resource = make_resource(...)
+        >>> resource.append(make_bitstream_prop("data/images/tree.jpg"))
+        >>> root.append(resource)
+
```
BalduinLandolt, line 633: return documentation missing (though I'm not sure it really needs it)

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+    See https://docs.dasch.swiss/latest/DSP-TOOLS/dsp-tools-xmlupload/#bitstream
+    """
+
+    if not os.path.isfile(path):
+        handle_warnings(f"The following is not the path to a valid file: {path} (resource '{calling_resource}')",
+                        stacklevel=3)
```
BalduinLandolt, line 639: should this even run through?

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+        handle_warnings(f"Both ARK and IRI were provided for resource '{label}' ({id}). The ARK will override the IRI.")
+
+    resource_ = etree.Element(
+        "{%s}resource" % (xml_namespace_map[None]),
+        **kwargs
+    )
+    return resource_
+
+
+def make_bitstream_prop(
```
BalduinLandolt, line 616: could all the `make_XX_prop()` functions be moved to a separate file? we're already at line 600 here...

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+        return "false"
+    elif unformatted in (True, "true", "1", 1, "yes"):
+        return "true"
+    else:
+        raise BaseError(f"Invalid boolean format for prop '{name}' in resource '{calling_resource}': '{unformatted}'")
+
+
+def make_boolean_prop(
+    name: str,
+    value: Union[PropertyElement, str, int, bool, Iterable[Union[PropertyElement, str, int, bool]]],
```
BalduinLandolt, line 659: to me, this type signature indicates that the function is too much overloaded, which indicates that the overall design should be improved:  
I would either always or never want this to be expecting a PropertyElement; and I would try to separate the iterable case out if possible at all.  
One option could be that you insist this to require a PropertyElement (and if you don't have one, you need to create one before being able to call this function), and that PropertyElement actually always contains a iterable as value, and if you only have one value, you simply pack it into a list.  
Because what you do here is to create an artificial choke point where you allow very different things to go into one variable by means of union types, just to then split that up again by type checking the same variable. That's not the cleanest design. (it would at least look tidy with the python 3.10 match case, but we shouldn't require 3.10 in DSP-Tools yet.)

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+            kwargs["comment"] = val.comment
+        value_ = etree.Element(
+            "{%s}color" % (xml_namespace_map[None]),
+            **kwargs,
+            nsmap=xml_namespace_map
+        )
+        value_.text = str(val.value).strip()
+        prop_.append(value_)
+
+    return prop_
```
BalduinLandolt, line 807: there is a lot of code repetition between those `make_XY_prop()` functions.

They always seem to do:
- perform `_check_and_prepare_values()`
- loop over the new values, apply a regex and throw an exception if the regex doesn't match
- create the prop_ et.Element
- loop over the values again, add the value subelements to the prop_
- return the prop_

I wonder if that repetition could be extracted elegantly

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+        name=name,
+        calling_resource=calling_resource
+    )
+
+    # check value type
+    for val in values_new:
+        # URI = scheme ":" ["//" host [":" port]] path ["?" query] ["#" fragment]
+        if not regex.search(
+            r"(?<scheme>[a-z][a-z0-9+.\-]*):(//(?<host>[\w_.\-\[\]:~]+)(?<port>:\d{0,6})?)(?<path>/[\p{L}_\-.~]*)*"
+            r"(?<query>\?[\p{L}_.\-=]+)*(?<fragment>#[\p{L}_/\-~:.]*)?", str(val.value), flags=regex.UNICODE):
```
BalduinLandolt, line 1628: seeing how many Regexes we have floating around, maybe they should all be collected in one file (and get pre-compiled). to ensure that touching a regex doesn't mean having to find all the places in the code where this regex appears, and changing it everywhere.

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+    Returns:
+        None
+    """
+    etree.indent(root, space="    ")
+    xml_string = etree.tostring(root, encoding="unicode", pretty_print=True)
+    xml_string = '<?xml version="1.0" encoding="UTF-8"?>\n' + xml_string
+    xml_string = xml_string.replace("&lt;", "<")
+    xml_string = xml_string.replace("&gt;", ">")
+    with open(filepath, "w", encoding="utf-8") as f:
+        f.write(xml_string)
```
BalduinLandolt, line 1960: why those string operations? to get the processing instructions in there?

BalduinLandolt: I'd have to look up how, but it can be solved directly in lxml so that you don't need to do any string juggling

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+
+            resource.append(make_prop_function(**kwargs_propfunc))
+
+    # append the resource of the very last iteration of the for loop
+    root.append(resource)
+
+    # write file
+    # ----------
+    write_xml(root, f"{default_ontology}-data.xml")
+    print(f"XML file successfully created at {default_ontology}-data.xml")
```
BalduinLandolt, line 2127: this file is definitely too long. It also does many things at once (data class, util functions, process functions, pipeline of processes to be called from the CLI tool), so it can easily be broken up into multiple files.

test/unittests/test_excel2xml.py
```diff
@@ -0,0 +1,612 @@
...
+def run_test(
+    testcase: unittest.TestCase,
+    prop: str,
+    method: Callable[..., etree._Element],
+    different_values: Sequence[Union[str, int, float, bool]],
+    invalid_values: Sequence[Union[str, int, float, bool]],
+    listname: Optional[str] = None
+) -> None:
+    """
+    XML-properties have always a similar structure, and all make_*_prop() methods have some similar things to test. This
```
BalduinLandolt, line 25: if this shared logic was extracted, then it could be tested a lot easier

test/unittests/test_excel2xml.py
BalduinLandolt: I don't think that's an issue

BalduinLandolt (approved):
Overall certainly a great improvement to DSP-Tools.

I ended up with quite a number of comments - but not all of them are equally relevant.  
I understand if you don't want to drag out that PR forever, so feel free to apply/ignore as much as you feel is right. But I really don't think any file should be 2k+ lines, so I'd like to see that broken up, if possible.

docs/assets/templates/excel2xml_script.py
```diff
@@ -0,0 +1,100 @@
...
+      resource.append(excel2xml.make_boolean_prop(name=":isComplete", value=row["Complete?"]))
+   if excel2xml.check_notna(row["Color"]):
+      resource.append(excel2xml.make_color_prop(":colorprop", row["Color"]))
+   if pd.notna(row["Date discovered"]):
+      potential_date = excel2xml.find_date_in_string(row["Date discovered"])
+      if potential_date:
+         resource.append(excel2xml.make_date_prop(":date", potential_date))
+      else:
+         excel2xml.handle_warnings(f"Error in row {index+2}: The column 'Date discovered' should contain a date, "
+                                   f"but no date was detected in the string {row['Date discovered']}")
```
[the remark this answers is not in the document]
jnussbaum: You're right. After some reflections, I decided to delete this function. Instead, I use `warnings.warn()`, and also added an example of this in the sample script.

knora/excel2xml.py
```diff
@@ -0,0 +1,2058 @@
...
+                    <boolean permissions="prop-restricted" comment="example">true</boolean>
+                </boolean-prop>
+        >>> make_boolean_prop(":testproperty", value=["false", "1"])
+                ERROR. Use this if some fields in your data source should be equal, but you cannot trust
+                your source. This method call will only work if all items of "value" are equivalent.
+
+    See https://docs.dasch.swiss/latest/DSP-TOOLS/dsp-tools-xmlupload/#boolean-prop
+    """
+
+    # check and validate input
```
irinaschubert, line 596:
```suggestion
    # validate input
```

knora/excel2xml.py
irinaschubert: I agree.

test/e2e/test_tools.py
irinaschubert: I remember that I've had problems without emptying the lists before and after the tests but if it works now without, then the better.

irinaschubert (approved):
LGTM, I have the feeling that the module is very hard to use for a custom user who does not do this work regularly. But I guess you clarified the use and need of this module beforehand?

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+##############################
+xml_namespace_map = {
+    None: "https://dasch.swiss/schema",
+    "xsi": "http://www.w3.org/2001/XMLSchema-instance"
+}
+__muted_warnings: list[str] = []
+__make_warnings_persistent = True
+
+
+class PropertyElement:
```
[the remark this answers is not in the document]
jnussbaum: Thanks for the hint, after watching a YT video that explained me how to do it, I successfully transformed this class into a dataclass 👍

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+    """
+
+    if not isinstance(string, str) or not check_notna(string):
+        raise BaseError(f"The string {string} cannot be made an xsd:id")
+
+    # if start of string is neither letter nor underscore, add an underscore
+    res = re.sub(r"^(?=[^A-Za-z_])", "_", string)
+
+    # add hash
+    _hash = hash(f"{res}{datetime.datetime.now()}")
```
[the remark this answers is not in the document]
jnussbaum: Thanks for the hint. Done.

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+
+    if startdate and enddate:
+        return f"GREGORIAN:CE:{startdate.isoformat()}:CE:{enddate.isoformat()}"
+    elif startyear and endyear:
+        return f"GREGORIAN:CE:{startyear}:CE:{endyear}"
+    else:
+        return None
+
+    # Fancy, because auto-extracts many date formats out of strings, but discouraged, produces too many false positives:
+    # datetime_obj = parser.parse(string, fuzzy = True, ignoretz = True)
```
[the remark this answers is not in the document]
jnussbaum: done

knora/excel2xml.py
jnussbaum: Yes, because in the development phase, RDU (Research Data Unit, formerly Client Services) members develop a script, but perhaps the bitstreams aren't available yet. In a later phase of development, the bitstreams are delivered. That's why I only raise a warning, not an error.

knora/excel2xml.py
```diff
@@ -0,0 +1,2127 @@
...
+    elif year_only:
+        startyear = int(year_only.group(0))
+        endyear = startyear
+
+    if startdate and enddate:
+        return f"GREGORIAN:CE:{startdate.isoformat()}:CE:{enddate.isoformat()}"
+    elif startyear and endyear:
+        return f"GREGORIAN:CE:{startyear}:CE:{endyear}"
+    else:
+        return None
```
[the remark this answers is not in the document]
jnussbaum: Yes, it was a lot of work. This method covers the use cases that had come across in our daily work in the Research Data Unit (formerly Client Services) over the last few monthes. So, this method grew over time.

 1. I have checked `dateparser`, and it would only do a part of the job, plus it has false positives.
 2. I experimented with splitting this up into multiple files. My try was to do this:
```
excel2xml
├── __init__.py
├── excel2xml.py
├── helper_functions.py
└── make_xml_element_functions.py
```
But there were 2 problems:
 1. In the above case, I splitted the methods into 3 groups that logically belong together, so that none of the 3 files was too long. But then I get circular import errors that I cannot resolve. The methods are too much entagled into each other.
 2. It seems odd to me to put only 1 method into a separate file, and leave everything else.

So I didn't succeed in splitting this file up. Have you got ideas what to do? Or, is it really so problematic to have a 2k+ lines long file? I don't see a problem in it...

knora/excel2xml.py
jnussbaum: thanks for pointing this out: I added it

knora/excel2xml.py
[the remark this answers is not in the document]
BalduinLandolt: ad 1.:  
with a quick search I found https://datefinder.readthedocs.io/en/latest/ and https://dateutil.readthedocs.io/en/stable/parser.html (from here: https://stackoverflow.com/questions/3276180/extracting-date-from-a-string-in-python). 
But I never used either of those, so I can't say if they're useful at all.

ad 2.:  
the fact that the methods are entangled and you get circular imports in and by itself indicates that the coupling is too high and it should be improved... but I'd have to give it more thought before I can say anything on how it should be done.  
I'd leave the call up to you: I can look into it, or we can decide that it's not a priority. Whatever you prefer is fine by me.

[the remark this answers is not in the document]
BalduinLandolt: (2k+ lines in and by itself is not a problem. But coupling is bad software design - and usually one has long files because one can't split it because the coupling is too high...)

knora/excel2xml.py
jnussbaum: I agree, and I read https://lxml.de/api/lxml.etree-module.html#tostring, but I don't suceed.

knora/excel2xml.py
jnussbaum: Noted for a next task.

knora/excel2xml.py
BalduinLandolt: I just had a quick look: When Lina faced the same issue in the eHKKA project, we used an `ElementTree` object, which can hold the processing instruction, and on which one can call `write()`.  
Here's some code I copied from there:

```python
    tree = et.ElementTree(root) # root is an `et._Element`
    # add processing instruction to the XML document
    pi = et.ProcessingInstruction('xml-model',
                                  'href="https://tei-c.org/release/xml/tei/custom/schema/xsd/tei_all.xsd" ' +
                                  'type="application/xml" ' +
                                  'schematypens="http://www.w3.org/2001/XMLSchema"')
    root.addprevious(pi)
    # write the XML data to the file
    tree.write(file_path, pretty_print=True, xml_declaration=True, encoding="UTF-8")
```

looking at it now, you don't even need the processing instruction, so you can ignore that part. But I think the relevant bits are `xml_declaration=True, encoding="UTF-8"` (which you could also do in `tostring()`). But I guess it's still more elegant to leave all the work to the ElementTree object.

So the function could literally be no more than:
```python
def write_xml(root: etree._Element, filepath: str) -> None:
    tree = et.ElementTree(root)
    tree.write(file_path, pretty_print=True, xml_declaration=True, encoding="UTF-8")
```
And that might just work ok.

docs/dsp-tools-excel2xml.md
```diff
@@ -0,0 +1,55 @@
+[![PyPI version](https://badge.fury.io/py/dsp-tools.svg)](https://badge.fury.io/py/dsp-tools)
+
+# EXCEL2XML: Convert a data source to XML
+dsp-tools assists you in converting a data source like CSV/XLSX to an XML file. Unlike the other features of dsp-tools,
```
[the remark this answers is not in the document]
jnussbaum: No, you're right. I improved the phrasing.

knora/excel2xml.py
jnussbaum: You're right, I allow too much in one variable. I have now removed the support for Iterables in the 'value' parameter of all make_*_prop() functions.

PR #226, opened 2022-09-09
chore: minor improvements of xmlupload (DEV-1141)

jnussbaum:
resolves DEV-1141

### DIFF
--- a/knora/dsplib/utils/xml_upload.py
+++ b/knora/dsplib/utils/xml_upload.py
@@ -22,7 +22,7 @@ from knora.dsplib.models.value import KnoraStandoffXml
 from knora.dsplib.models.xmlpermission import XmlPermission
 from knora.dsplib.models.xmlproperty import XMLProperty
 from knora.dsplib.models.xmlresource import XMLResource
-from knora.dsplib.utils.shared_methods import try_network_action
+from knora.dsplib.utils.shared_methods import try_network_action, validate_xml_against_schema
 
 
 def _remove_circular_references(resources: list[XMLResource], verbose: bool) -> \
@@ -147,29 +147,6 @@ def _stash_circular_references(
     return nok_resources, ok_res_ids, ok_resources, stashed_xml_texts, stashed_resptr_props
 
 
-def _validate_xml_against_schema(input_file: str, schema_file: str) -> bool:
-    """
-    Validates an XML file against an XSD schema
-
-    Args:
-        input_file: the XML file to be validated
-        schema_file: the schema against which the XML file should be validated
-
-    Returns:
-        True if the XML file is valid, False otherwise
-    """
-    xmlschema = etree.XMLSchema(etree.parse(schema_file))
-    doc = etree.parse(input_file)
-
-    if xmlschema.validate(doc):
-        return True
-    else:
-        print("The input data file cannot be uploaded due to the following validation error(s):")
-        for error in xmlschema.error_log:
-            print(f"  Line {error.line}: {error.message}")
-        return False
-
-
 def _convert_ark_v0_to_resource_iri(ark: str) -> str:
     """
     Converts an ARK URL from salsah.org (ARK version 0) of the form ark:/72163/080c-779b9990a0c3f-6e to a DSP resource
@@ -244,7 +221,7 @@ def _parse_xml_file(input_file: str) -> etree.ElementTree:
 
 
 def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: str, sipi: str, verbose: bool,
-               validate_only: bool, incremental: bool) -> bool:
+               incremental: bool) -> bool:
     """
     This function reads an XML file and imports the data described in it onto the DSP server.
 
@@ -256,7 +233,6 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
         imgdir: the image directory
         sipi: the sipi instance to be used
         verbose: verbose option for the command, if used more output is given to the user
-        validate_only: validation option to validate the XML data without the actual import of the data
         incremental: if set, IRIs instead of internal IDs are expected as resource pointers
 
     Returns:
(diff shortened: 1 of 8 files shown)


knora/dsplib/utils/xml_upload.py
```diff
@@ -267,17 +243,15 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
-        if validate_only:
-            exit(0)
-    else:
-        print("ERROR The input data file did not pass validation.")
-        exit(1)
+    try:
+        validate_xml_against_schema(input_file, schema_file)
+    except BaseError as err:
+        print(err)
+        quit(0)
```
jnussbaum, line 250: @Irina: Here I don't get the effect that I'd like to: I thought it would be a good design pattern to raise an error if the schema validation fails, and then to catch the error and print out the details. Plus, I want to hide the stacktrace from the user if validation fails. But with the code that I have now, I get the following output when the XML is invalid:

```
dsp-tools xmlupload testdata/test-data-systematic.xml --validate
  File "/Users/nusjoh00-adm/.local/share/virtualenvs/dsp-tools--UOnuDrn/bin/dsp-tools", line 8, in <module>
    sys.exit(main())
  File "/Users/nusjoh00-adm/Desktop/dsp-tools/knora/dsp_tools.py", line 215, in main
    program(sys.argv[1:])
  File "/Users/nusjoh00-adm/Desktop/dsp-tools/knora/dsp_tools.py", line 181, in program
    validate_xml_against_schema(input_file=args.xmlfile,
  File "/Users/nusjoh00-adm/Desktop/dsp-tools/knora/dsplib/utils/shared_methods.py", line 109, in validate_xml_against_schema
    raise BaseError(error_msg)
knora.dsplib.models.helpers.BaseError: ERROR: The input data file cannot be uploaded due to the following validation error(s):
  Line 386: Element '{https://dasch.swiss/schema}resource', attribute 'labell': The attribute 'labell' is not allowed.
  Line 386: Element '{https://dasch.swiss/schema}resource': The attribute 'label' is required but missing.
  Line 402: Element '{https://dasch.swiss/schema}text', attribute 'encoding': [facet 'enumeration'] The value 'öalkj' is not an element of the set {'utf8', 'xml'}.!

NoneType: None
```

I don't know why the error escalates into the console so that the user sees it. I thought that if I catch it, it is "neutralized".

Can you give me advice how to improve this? (Or if I can just leave it as it is?)

irinaschubert: Hm, interesting. I have no idea why it doesn't work. Did it work in other locations?

knora/dsplib/utils/xml_upload.py
```diff
@@ -391,12 +374,15 @@ def _upload_resources(
...
                     action=lambda: sipi_server.upload_bitstream(filepath=os.path.join(imgdir, resource.bitstream.value)),
-                    failure_msg=f'ERROR while trying to create resource "{resource.label}" ({resource.id}).'
+                    failure_msg=f'ERROR while trying to upload file "{resource.bitstream.value}" of resource '
+                                f'"{resource.label}" ({resource.id}).'
                 )
             except BaseError as err:
                 print(err.message)
                 failed_uploads.append(resource.id)
                 continue
+            bitstream_size_uploaded_mb += round(next(bitstream_all_sizes_iterator), 1)
```
irinaschubert, line 384: I'm just curious: why do you need this information? Is it worth the time it takes to calculate this information? Have you measured it?

knora/dsplib/utils/xml_upload.py
```diff
@@ -542,7 +529,8 @@ def _purge_stashed_xml_texts(
         a purged version of stashed_xml_text
     """
     # remove resources that couldn't be uploaded. If they don't exist in DSP, it's not worth caring about their xmltexts
-    if id2iri_mapping:
+    if id2iri_mapping is not None:
```
irinaschubert, line 532: If "XY is not None" is not the same as "XY", there seems to be something wrong. Could you maybe solve this problem at `id2iri_mapping` instead of here? I.e. make sure that `if id2iri_mapping` is the same as `if id2iri_mapping is not None` is actually the same?

irinaschubert (changes requested):
I suggest that you fix the `is not None` issue I commented. Also, the question you asked regarding the Error that isn't caught seems to be a code smell. I would try to fix it. Unfortunately I don't know how but I can have a look if you want.

knora/dsplib/utils/xml_upload.py
jnussbaum: Okay, I found the problem. `id2iri_mapping` should not be optional. It is in fact always available when this method is executed, so it is best to *always* check if the stashed elements are in `id2iri_mapping`. So I don't need the if-clause any more.

knora/dsplib/utils/xml_upload.py
jnussbaum: Actually, it's not worth the effort. I can just as well let the error escalate. The user will see the stack trace, but that doesn't hurt.

irinaschubert: I agree regarding the user experience. It's not too bad. But I worry about the source of the problem - this could mean that something "deeper" is wrong here. But I leave it up to you, of course, if you want to track this down or not.

jnussbaum: Thanks for the incentive to investigate. It was worth it: I found out that I had made a stupid mistake: If I print the entire `err` object, it's only logical that the stacktrace is printed. I have to print err.message instead :-)

knora/dsplib/utils/xml_upload.py
irinaschubert: cool, then it seems to be a nice feature for the user who uploads data

irinaschubert (approved):
Now it looks good, thanks!

PR #240, opened 2022-10-10
docs(xmlupload): improve examples, add documentation of geometry-prop JSON format

jnussbaum:
resolves DEV-1404


testdata/test-data-systematic.xml
```diff
@@ -693,6 +691,35 @@
         </color-prop>
     </region>
 
+    <region label="Circle: dysfunctional"
```
jnussbaum, line 694: It is still an open question how circles can be created. Here I just added a draft which is not functional yet. It doesn't affect the test runs, the only issue is that it isn't displayed in the APP (and I suppose that the API cannot handle it)

BalduinLandolt: who is deciding that? and how?  
Also, should it really be a circle, or rather an ellipsis? (in which case the radius and center would not work)

jnussbaum: I meant "draft" in the sense of a try. I think that this functionality is somewhere hidden in the API, but it is  documented nowhere. So I try to find the right syntax. Is there a method how I could find the syntax?

irinaschubert: In my opinion, if it doesn't work in DSP-APP it should not be in the docs of DSP-TOOLS. Otherwise we confuse users, don't you think so?
For circles, you would find an example in the incunabula test data: https://raw.githubusercontent.com/dasch-swiss/dsp-api/main/test_data/all_data/incunabula-data.ttl (open this page and search for "circle")

docs/dsp-tools-xmlupload.md
```diff
@@ -340,10 +335,10 @@ Attributes:
...
 - `comment`: a comment for this specific value (optional)
 
-A property with two color values would be defined as follows:
+Example of a property with a public and a hidden color value:
 ```xml
 <color-prop name=":hasColor">
-    <color>#00ff66</color>
+    <color permissions="prop-default">#00ff66</color>
     <color>#ff00ff</color>
 </color-prop>
```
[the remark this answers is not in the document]
jnussbaum: I can understand your point,  but it is very important for me to prevent users from creating a hidden value (`<color>#ff00ff</color>`). Unfortunately, the docs have only examples of hidden values until now. This is dangerous. So I wanted to find a way how I can grab the user's attention, and teach him this important difference.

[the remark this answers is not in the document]
jnussbaum: On the long run, I could perhaps disallow the construct `<color>#ff00ff</color>`, and force the user to write `<color permissions="None">#ff00ff</color>`. This would prevent mistakes, but it would be a breaking change.

docs/dsp-tools-xmlupload.md
```diff
@@ -854,8 +833,40 @@ Example:
...
+The underlying grid is a 0-1 normalized top left-anchored grid. The following coordinate system shows the three shapes
+that were defined above:
+![grid-for-geometry-prop](./assets/images/grid-for-geometry-prop.png)
+
+
+Technical notes: 
+ - A `<region>` is in fact a `<resource restype="Region">`. But it is mandatory to use the 
 shortcut, so that the XML file can be validated more precisely.
+ - In the SALSAH data, there is also a key named `original_index` in the JSON format of all three shapes, but it doesn't 
+   seem to have an influence on the shapes that TANGOH displays, so it can be omitted.
```
BalduinLandolt, line 868: "doesn't seem to ..." probably doesn't sound very confidence-inspiring to users

docs/dsp-tools-xmlupload.md
```diff
@@ -854,8 +833,40 @@ Example:
...
+
+The circle and polygon are created with the following syntax:
+```json
+{
+    "status": "active",
+    "type": "circle",
+    "lineColor": "#ff1100",
+    "lineWidth": 5,
+    "points": [{"x":0.5,"y":0.3}],
+    "radius": {"x":0.1,"y":0.1}
```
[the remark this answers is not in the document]
BalduinLandolt: ah, I see now, this can be an ellipsis. But wouldn't it be simpler to define the ellipsis by the bounding rectangle? (like e.g. in paint, when you "drag" a box with the mouse, and it "fills" it with an ellipsis)

BalduinLandolt (approved):
looks good, just some minor remarks, and/or stuff for later

docs/dsp-tools-xmlupload.md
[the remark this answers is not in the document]
jnussbaum: No, I don't think  that an ellipsis is possible. It's just that the creators of knora decided to define the radius as vector (x, y). It's a bit like the hypothenuse of a right-angled triangle. I added a comment line in the JSON example to make it more clear.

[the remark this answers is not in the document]
BalduinLandolt: ok, I understand, thanks for the explanation. Interesting design choice. ;-)  
Then my point remains, that in my opinion it _should_ be an ellipsis, not a circle... (but again, that's not scope of this PR, so let's just keep it in the back of our heads)

docs/dsp-tools-xmlupload.md
BalduinLandolt: fair enough

PR #243, opened 2022-10-22
fix(excel2xml): prevent writing empty text-prop, make text-prop validation less restrictive (DEV-1440)

jnussbaum:
resolves DEV-1440


knora/excel2xml.py
```diff
@@ -1189,9 +1189,12 @@ def make_text_prop(
 
     # check value type
     for val in values:
-        if not isinstance(val.value, str) or not check_notna(val.value):
+        if not isinstance(val.value, str) or len(val.value) < 1:
```
BalduinLandolt, line 1192:
```suggestion
        if not isinstance(val.value, str) or len(val.value) == 0:
```

I think that's slightly more readable (and you don't care about `len <= -1`, so its semantically the same)

BalduinLandolt (approved):
LGTM

PR #246, opened 2022-11-01
feat: startup API and APP with dsp-tools (DEV-126)

jnussbaum:
resolves DEV-126

### DIFF
--- a/docs/dsp-tools-usage.md
+++ b/docs/dsp-tools-usage.md
@@ -218,7 +218,6 @@ about it [here](./dsp-tools-excel2xml.md).
 
 
 
-
 ## Replace internal IDs with IRIs in XML file
 
 ```bash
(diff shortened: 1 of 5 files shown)


knora/dsplib/utils/start-api.sh
```diff
@@ -0,0 +1,42 @@
...
+cd ~
+mkdir -p .dsp-tools
+cd .dsp-tools
+if [[ ! -d dsp-api ]]; then
+    echo "git clone https://github.com/dasch-swiss/dsp-api.git" 2>&1 | tee -a "$logfile"
+    git clone https://github.com/dasch-swiss/dsp-api.git >>"$logfile" 2>&1
+fi
+cd dsp-api
+rm -f "$logfile"
+echo "make stack-down-delete-volumes..." 2>&1 | tee -a "$logfile"
```
BalduinLandolt, line 31: I don't think this is needed if you have `init-db-test-minimal`

knora/dsplib/utils/start-api.sh
```diff
@@ -0,0 +1,42 @@
...
+make stack-down-delete-volumes >>"$logfile" 2>&1
+if echo -e "GET http://google.com HTTP/1.0\n\n" | nc google.com 80 -w 10 > /dev/null 2>&1; then
+    # only pull if there is an internet connection
+    echo "git pull ..." 2>&1 | tee -a "$logfile"
+    git pull >>"$logfile" 2>&1
+fi
+echo "make init-db-test-minimal..." 2>&1 | tee -a "$logfile"
```
BalduinLandolt, line 38: are you sure you want `minimal` and not the one with test data included? (not a rhetorical question, just wondering)

BalduinLandolt (approved):

docs/dsp-tools-usage.md
```diff
@@ -238,3 +237,51 @@ Note that internal IDs and IRIs cannot be mixed. The input XML file has to be pr
...
+it finds an existing clone there, it runs `git pull` instead. If the API is already running, it shuts down the old 
+instance and starts a new one. If the dependencies are outdated or not installed, a warning is printed to the console.
+
+
+### Shut DSP-API down
+```
+dsp-tools stop-api
+```
+
+This command shuts DSP-API down, deletes all Docker volumes, and removes temporary files.
```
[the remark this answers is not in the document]
jnussbaum: I don't understand. I always use `make stack-down-delete-volumes` to shut down the  API. Is that not good?

By the way, thanks for pointing out that I mistakenly echoed `make init-db-test-minimal`. The command that I execute is without the "minimal".

knora/dsp_tools.py
```diff
@@ -204,6 +221,12 @@ def program(user_args: list[str]) -> None:
         excel2xml(datafile=args.datafile,
                   shortcode=args.shortcode,
                   default_ontology=args.default_ontology)
+    elif args.action == 'start-api' and not sys.platform.startswith('win'):
```
jnussbaum, line 224: This is probably a better way to check the OS, I guess?

docs/dsp-tools-usage.md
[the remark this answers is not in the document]
BalduinLandolt: not necessarily "not good" - but if you invest time to set up specific test data, then that will be gone if you delete the volumes (and you need to re-initialize the DB). So e.g. if you modify your datamodel in the App, and then you want to export it to json again using dsp-tools, delete-volumes would not be what you want.

knora/dsp_tools.py
BalduinLandolt: not sure how that would behave with WSL... but probably you shouldn't worry about that too much. So this looks good to me

PR #249, opened 2022-11-03
docs: improve docs (DEV-1478)

jnussbaum:
resolves DEV-1478


docs/dsp-tools-create.md
```diff
@@ -113,38 +115,6 @@ The following fields are optional (if one or more of these fields are not used,
 - groups
 - users
 
-A simple example definition of the `project` object looks like this:
```
jnussbaum, line 116: A complete example of a project is already given some lines above, so it doesn't make sense to repeat this here.

docs/dsp-tools-create.md
```diff
@@ -426,7 +396,7 @@ example, the list "colors" could be imported as follows:
                 "en": "A list with categories"
             },
             "nodes": [
-                ...
```
jnussbaum, line 429: This and other similar changes are just that PyCharm doesn't show that many problems. In the future, I'd like to look more frequently on the problems that are detected by PyCharm. But for this, I need to kick out the false positives.

docs/dsp-tools-xmlupload.md
```diff
@@ -11,7 +11,7 @@ The command to import an XML file on a DSP server is described [here](./dsp-tool
 
 The import file must start with the standard XML header:
 
-```xml
```
jnussbaum, line 14: If it's invalid XML, it's rather a pain that it is marked XML. (PyCharm detects this as a problem)

docs/dsp-tools-xmlupload.md
```diff
@@ -427,76 +428,78 @@ Example of a property with a public and a hidden decimal value:
 ### &lt;geometry-prop&gt;
```
jnussbaum, line 428: I reorganised the docs of <geometry-prop>. In principle I just moved information from one place to the other, with only few real deletions.

BalduinLandolt (approved):
See my comments - overall I'm not enthusiastic about some of these changes, but at the end of the day the code base has to work for you, not for my code review :)

PR #250, opened 2022-11-04
refactor(xmlupload): handling of upload errors (DEV-1505)

jnussbaum:
resolves DEV-1481


knora/dsplib/utils/xml_upload.py
```diff
@@ -351,42 +359,69 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+            server=server,
+            save_location=save_location
+        )
 
     # update the resources with the stashed XML texts
-    nonapplied_xml_texts = {}
     if len(stashed_xml_texts) > 0:
         try:
             nonapplied_xml_texts = _upload_stashed_xml_texts(verbose, id2iri_mapping, con, stashed_xml_texts)
-            nonapplied_xml_texts = _purge_stashed_xml_texts(nonapplied_xml_texts, id2iri_mapping)
```
jnussbaum, line 361: This line is redundant, because _upload_stashed_xml_texts() returns the xml texts already purged.

knora/dsplib/utils/xml_upload.py
```diff
@@ -351,42 +359,69 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
+            stashed_xml_texts=stashed_xml_texts,
+            stashed_resptr_props=stashed_resptr_props,
+            proj_shortcode=shortcode,
+            onto_name=default_ontology,
+            server=server,
+            save_location=save_location
+        )
 
     # update the resources with the stashed XML texts
-    nonapplied_xml_texts = {}
```
jnussbaum, line 357: It's not necessary any more to create an empty `nonapplied_xml_texts`, because the case cannot happen any more that `nonapplied_xml_texts` is used before being instantiated

knora/dsplib/utils/xml_upload.py
```diff
@@ -351,42 +359,69 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
...
 
     # write log files
     success = True
     timestamp_str = datetime.now().strftime("%Y%m%d-%H%M%S")
-    if len(nonapplied_xml_texts) > 0:
-        _write_stashed_xml_texts(nonapplied_xml_texts, timestamp_str)
-        success = False
-    if len(nonapplied_resptr_props) > 0:
-        _write_stashed_resptr_props(nonapplied_resptr_props, timestamp_str)
-        success = False
```
jnussbaum, line 382: This case cannot happen any more. If one of them is > 0, this part of the code is not reached.

knora/dsplib/utils/xml_upload.py
```diff
@@ -304,6 +305,13 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
               f"{err.message}")
         quit(0)
 
+    if sys.platform.startswith("darwin") or sys.platform.startswith("linux"):
+        save_location = f"{os.path.expanduser('~')}/.dsp-tools"
+    elif sys.platform.startswith("win"):
+        save_location = "."
+    else:
+        save_location = "."
```
BalduinLandolt, line 313: Pathlib has a home directory that works OS independently. And generally I would not use any `os.path` code anymore, but use Pathlib instead

knora/dsplib/utils/xml_upload.py
```diff
@@ -720,33 +757,62 @@ def _handle_upload_error(
...
+        r"0.0.0.0": "localhost"
+    }
+    for pattern, repl in server_substitutions.items():
+        server = re.sub(pattern, repl, server)
+
+    if save_location == ".":
+        save_location_full = f"xmluploads/{server}/{proj_shortcode}/{onto_name}"
+    else:
+        save_location_full = f"{save_location}/xmluploads/{server}/{proj_shortcode}/{onto_name}"
+    os.makedirs(save_location_full, exist_ok=True)
```
BalduinLandolt, line 790: again, a Path object would make this much nicer (plus, if you can eliminate the `.` case, this will get simpler

knora/dsplib/utils/xml_upload.py
```diff
@@ -720,33 +757,62 @@ def _handle_upload_error(
...
 
-    stashed_resptr_props_purged = _purge_stashed_resptr_props(stashed_resptr_props, id2iri_mapping)
-    if len(stashed_resptr_props_purged) > 0:
-        _write_stashed_resptr_props(stashed_resptr_props_purged, timestamp_str)
+    if id2iri_mapping:
+        id2iri_mapping_file = f"{save_location_full}/{timestamp_str}_id2iri_mapping.json"
+        with open(id2iri_mapping_file, "x") as f:
+            json.dump(id2iri_mapping, f, ensure_ascii=False, indent=4)
+        print(f"The mapping of internal IDs to IRIs was written to {id2iri_mapping_file}")
+    if stashed_xml_texts:
```
BalduinLandolt, line 801: since these are different if statements, I think it would be nicer to separate them with an empty line. Like this it looks as if it was an if-else

BalduinLandolt (approved):
Only cosmetics from my side. I generally like pathlib and would avoid writing paths as strings. But that is by no means pressing or important

knora/dsplib/utils/xml_upload.py
```diff
@@ -720,33 +757,62 @@ def _handle_upload_error(
 
     Args:
         err: error that was the cause of the abort
-        input_file: file name of the original XML file
         id2iri_mapping: mapping of ids from the XML file to IRIs in DSP (only successful uploads appear here)
         failed_uploads: resources that caused an error when uploading to DSP
         stashed_xml_texts: all xml texts that have been stashed
         stashed_resptr_props: all resptr props that have been stashed
+        proj_shortcode: shortcode of the project where the data is uploaded
```
irinaschubert, line 764:
```suggestion
        proj_shortcode: shortcode of the project the data belongs to
```

knora/dsplib/utils/xml_upload.py
```diff
@@ -720,33 +757,62 @@ def _handle_upload_error(
...
+        onto_name: name of the ontology into which the data is uploaded
```
irinaschubert, line 765:
```suggestion
        onto_name: name of the ontology the data references
```

knora/dsplib/utils/xml_upload.py
```diff
@@ -720,33 +757,62 @@ def _handle_upload_error(
...
+        server: the server onto which the data is uploaded
```
irinaschubert, line 766:
```suggestion
        server: the server which the data is uploaded onto
```

knora/dsplib/utils/xml_upload.py
```diff
@@ -720,33 +757,62 @@ def _handle_upload_error(
...
+        save_location: path to the directory where dsp-tools can save logs (OS dependent)
```
irinaschubert, line 767:
```suggestion
        save_location: path to the directory where dsp-tools should save logs (OS dependent)
```

knora/dsplib/utils/xml_upload.py
irinaschubert: I agree

irinaschubert (approved):
LGTM, have you thought about restricting the file size of the log file? If the file gets too large, some editors can't open it anymore, so it is usually an important use case - but at this point maybe not relevant to us..?

knora/dsplib/utils/xml_upload.py
jnussbaum: okay, it's noted with a TODO comment. For the moment, reading through the docs of Pathlib would be too time consuming. I want to take the time to do the transition to Pathlib properly, in order to really understand (and use) its advantages. 

In the current case, Pathlib wouldn't change anything, because I use the home directory only on Macs. On Windows, it' an entirely different story to obtain an app data storage location from the OS. 

But thanks for gently pushing me into using good libraries, I appreciate this endeavour!

PR #251, opened 2022-11-07
fix: improve start-api and start-app (DEV-1485)

jnussbaum:
resolves DEV-1485


Pipfile
```diff
@@ -15,6 +15,8 @@ networkx = "*"
 pandas = "*"
 regex = "*"
 xlrd = "*"
+pyyaml = "*"
+types-pyyaml = "*"
```
BalduinLandolt, line 19: I think `types-pyyaml` could be a dev dependency

docs/dsp-tools-usage.md
```diff
@@ -256,17 +256,26 @@ In order to upload data incrementally the procedure described [here](dsp-tools-x
...
 
-For testing purposes, it is sometimes necessary to have DSP-API and/or DSP-APP running on a local machine. The startup 
+For testing purposes, it is sometimes necessary to run DSP-API and DSP-APP on a local machine. But the startup 
 and shutdown of API and APP can be complicated: Both repos need to be cloned locally, a `git pull` has to be executed 
 from time to time to stay up to date, and then there are several commands for each repository to remember. 
-That's why dsp-tools offers some commands to facilitate the handling of API and APP. 
+
+Another challenge is the software that DSP depends upon: JDK, node, npm, Angular, etc. should be kept up to date. And
+it might happen that a dependency is replaced, e.g. JDK 11 Zulu by JDK 17 Temurin. An non-developer can quickly get lost
+in this jungle.
```
BalduinLandolt, line 265: the "jungle" is very colloquial

knora/dsp_tools.py
```diff
@@ -233,11 +237,28 @@ def program(user_args: list[str]) -> None:
...
+        try:
+            response = requests.get("https://raw.githubusercontent.com/dasch-swiss/dsp-api/main/.github/actions/preparation/action.yml")
+            action = yaml.safe_load(response.content)
+            for step in action.get("runs", {}).get("steps", {}):
+                if re.search("(JDK)|(Java)", step.get("name", "")):
+                    distribution = step.get("with", {}).get("distribution", "").lower()
+                    java_version = step.get("with", {}).get("java-version", "").lower()
+        except requests.ConnectionError:
+            distribution = "temurin"
+            java_version = "17"
```
BalduinLandolt, line 249: for my taste this goes too far: I understand that not everyone is a developer. But keeping up with the right Java version is not too much to ask. And I suspect we change the names of our Github Actions more frequently than our Java versions, so this script would break less if you'd just hard-code temurin 17 here

BalduinLandolt (approved):
I'm not really convinced that it should be the responsibility of dsp-tools to do all those things. I get the need for simplifying startup of the stack. But dependency management seems to go a bit too far for my taste. But at the end of the day that's not up to me, but product management and RDU to decide what dsp-tools should and shouldn't do.

Pipfile
jnussbaum: thanks. it's moved.

knora/dsp_tools.py
jnussbaum: I made sure it will never break by making the `except` clause bare (not recommended, I know. But it will serve the purpose here)

PR #252, opened 2022-11-10
fix: bugs in json schema (DEV-1142)

jnussbaum:
resolves DEV-1142


jnussbaum:
In this PR, I (exceptionally) managed to make meaningful commits. It probably makes reviewing easier if you go through the commits one after the other.

BalduinLandolt (approved):
Not an easy PR to review, but from what I saw, it seems to make sense

PR #253, opened 2022-11-10
fix: excel2json is not broken any more by hidden system files (DEV-1492)

jnussbaum:
resolves DEV-1492


knora/dsplib/utils/excel_to_json_project.py
```diff
@@ -56,7 +56,7 @@ def excel2json(
 
     listfolder = [x for x in folder if os.path.isdir(x) and x.name == "lists"]
     if listfolder:
-        listfolder_contents = list(os.scandir(listfolder[0]))
+        listfolder_contents = [x for x in os.scandir(listfolder[0]) if not re.search(r"^(\.|~\$).+", x.name)]
         if not all([re.search(r"(de|en|fr|it|rm).xlsx", file.name) for file in listfolder_contents]):
             raise BaseError(f"The only files allowed in '{data_model_files}/lists' are en.xlsx, de.xlsx, fr.xlsx, "
                             f"it.xlsx, rm.xlsx")
```
BalduinLandolt, line 62: I never understood why these are supposed to be different files, rather than different sheets in one excel file. I think that would be much less error prone.  
Also, this could probably be solved more elegantly with a glob picking the desired files and ignoring everything else.

BalduinLandolt (approved):
I'm pretty sure this could be solved more elegantly - but as a fix this is definitely good enough. And refactoring can be done later on.

PR #258, opened 2022-11-22
fix(start-app, start-api): make git handling more robust (DEV-1520)

jnussbaum:
resolves DEV-1520


BalduinLandolt (approved):
I'm not a specialist in this, but I think that should be good

PR #261, opened 2022-11-29
feat: use docker commands for stack handling (DEV-1530)

jnussbaum:
resolves DEV-1530


knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,52 @@
...
+        exit(1)
+    time.sleep(5)
+
+    # create the "knora-test" repository
+    url_prefix = "https://raw.githubusercontent.com/dasch-swiss/dsp-api/main"
+    repo_template_response = requests.get(f"{url_prefix}/webapi/scripts/fuseki-repository-config.ttl.template")
+    if not repo_template_response or repo_template_response.status_code != 200:
+        exit(1)
+    repo_template = re.sub(r"@REPOSITORY@", "knora-test", repo_template_response.text)
+    response = requests.post(url="http://0.0.0.0:3030/$/datasets", files={"file": ("file.ttl", repo_template)},
```
jnussbaum, line 24: @seakayone Here I am stuck. I get a 500 status and the message `[line: 1, col: 1 ] Out of place: [MINUS]`. 
This line tries to imitate https://github.com/dasch-swiss/dsp-api/blob/main/webapi/scripts/fuseki-functions.sh#L70
Like in line 41 below, I try to imitate that a file is sent instead of a string, as documented here: https://requests.readthedocs.io/en/latest/user/quickstart/#post-a-multipart-encoded-file.

Balduin already tried to help me, but we both don't know what the problem is. Could you help out?

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,67 @@
...
+    # start up the fuseki database
+    completed_process = subprocess.run("docker compose up db -d", shell=True, cwd="knora/dsplib/docker")
+    if not completed_process or completed_process.returncode != 0:
+        raise BaseError("Cannot start the API: Error while executing 'docker compose up db -d'")
+    time.sleep(5)
+
+    # inside fuseki, create the "knora-test" repository
+    url_prefix = "https://raw.githubusercontent.com/dasch-swiss/dsp-api/main"
+    repo_template = requests.get(f"{url_prefix}/webapi/scripts/fuseki-repository-config.ttl.template").text
+    repo_template = re.sub(r"@REPOSITORY@", "knora-test", repo_template)
```
BalduinLandolt, line 23: I don't think you need a regex for this simple string replacement, as it only has to fit an exact literal match.  
(something like `repo_template = repo_template.replace("@REPOSITORY@", "knora-test")`?)

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,67 @@
...
+    ttl_files = [
+        (f"{url_prefix}/knora-ontologies/knora-admin.ttl", f"{graph_prefix}http://www.knora.org/ontology/knora-admin"),
+        (f"{url_prefix}/knora-ontologies/knora-base.ttl", f"{graph_prefix}http://www.knora.org/ontology/knora-base"),
+        (f"{url_prefix}/knora-ontologies/standoff-onto.ttl", f"{graph_prefix}http://www.knora.org/ontology/standoff"),
+        (f"{url_prefix}/knora-ontologies/standoff-data.ttl", f"{graph_prefix}http://www.knora.org/data/standoff"),
+        (f"{url_prefix}/knora-ontologies/salsah-gui.ttl", f"{graph_prefix}http://www.knora.org/ontology/salsah-gui"),
+        (f"{url_prefix}/test_data/all_data/admin-data-minimal.ttl", f"{graph_prefix}http://www.knora.org/data/admin"),
+        (f"{url_prefix}/test_data/all_data/permissions-data-minimal.ttl",
+         f"{graph_prefix}http://www.knora.org/data/permissions")
+    ]
```
BalduinLandolt, line 43: this would look a lot slimmer if you would only store the actual string, and then prepend the prefix in the loop. (only cosmetics...)

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,67 @@
...
+            files={"file": ("file.ttl", ttl_text, "text/turtle; charset: utf-8")},
+            auth=("admin", "test"),
+        )
+        if not response.ok:
+            raise BaseError(f"Cannot start the API: Error when creating graph '{graph}'")
+
+    # get sipi.docker-config.lua
+    docker_config_lua_text = requests.get(f"{url_prefix}/sipi/config/sipi.docker-config.lua").text
+    with open("knora/dsplib/docker/sipi.docker-config.lua", "w") as f:
+        f.write(docker_config_lua_text)
```
BalduinLandolt, line 57: what is this file saved/used for?

docs/dsp-tools-usage.md
```diff
@@ -254,61 +254,45 @@ In order to upload data incrementally the procedure described [here](dsp-tools-x
...
-
-Please note that these commands were developed for DaSCH-internal use only. They only work on Macs that have the 
-required software installed that makes it possible to run the API and APP. We don't offer support or troubleshooting 
-for these commands.
-
+from time to time to stay up to date, and then there are several commands for each repository to remember. And your 
+local clone of the API might get cluttered with old files over time.
 
-### Start DSP-API
+Another challenge is the software that DSP depends upon, and that should be kept up to date: JDK, node, npm, Angular,
```
BalduinLandolt, line 267:
```suggestion
Another challenge is the software that DSP depends on, and that should be kept up to date: JDK, node, npm, Angular, 
```

docs/dsp-tools-usage.md
```diff
@@ -254,61 +254,45 @@ In order to upload data incrementally the procedure described [here](dsp-tools-x
...
 
-### Start DSP-API
+Another challenge is the software that DSP depends upon, and that should be kept up to date: JDK, node, npm, Angular, 
+etc. And it might happen that a dependency is replaced, e.g. JDK 11 Zulu by JDK 17 Temurin. A non-developer can quickly 
+get lost in this jungle. 
 
+That's why dsp-tools offers a command to facilitate the handling of API and APP:
 ```
-dsp-tools start-api
+dsp-tools start-stack
```
BalduinLandolt, line 273: maybe I'm professionally deformed by our make targets... but to me `dsp-tools stack-up` (or even shorter `dsp-tools stack`) and `stack-down` would sound more familiar/intuitive.

docs/dsp-tools-usage.md
```diff
@@ -254,61 +254,45 @@ In order to upload data incrementally the procedure described [here](dsp-tools-x
...
-it finds an existing clone there, it runs `git pull` instead. If the API is already running, it shuts down the old 
-instance, deletes all data that was in it, and starts a new one. If the dependencies are outdated or not installed, a 
-warning is printed to the console.
-
-
-### Shut DSP-API down
+This calls Docker commands to start up the latest deployed versions of API and the APP, i.e. the versions that are 
+running on [https://admin.dasch.swiss](https://admin.dasch.swiss). It's no longer necessary to have two local clones, 
+nor to execute `make` commands inside them. The only requirement for this command is that Docker must be running. It 
+can even be executed on a Windows machine without any of the software dependencies.
```
BalduinLandolt, line 279: maybe I just overlooked it... but you'd still have docker installed&running as a requirement - so that should definitely be explicitly mentioned.

BalduinLandolt: also, speaking of requirements: do we write anywhere that python (and which python version) must be installed? just came to my mind...

BalduinLandolt (approved):
LGTM, I think that's a nice improvement for everyone

knora/dsplib/utils/stack_handling.py
jnussbaum: It's referenced in docker-compose.yml, and without it, SIPI crashes immediately. It's actually a config file that SIPI needs. But if you know better, I would be glad to remove it. The simpler, the better.

docs/dsp-tools-usage.md
jnussbaum: I wanted to find a terminology that prevents confusion with the make-targets, otherwise I confuse them. In the initial phase, when the two were similar, I accidentally typed things like `dsp-tools stack-down-delete-volumes`.

docs/dsp-tools-usage.md
jnussbaum: dsp-tools is a pip package, so it's difficult to have it without Python. setup.py says `python_requires=">=3.9.0"`, so we already have it pinned there. But actually, I don't think that dsp-tools really requires a certain Python version, does it?

knora/dsplib/utils/stack_handling.py
BalduinLandolt: not at all, I was just curious (didn't see it in the docker-compose). Thanks for the explanation

docs/dsp-tools-usage.md
BalduinLandolt: yes... don't ask me why... but I remember Eric at some point having an issue which turned out to be him having 3.8 installed. So probably it would be good to specify that somewhere as I don't think pip will stop you from installing it on a lower python version (not sure though). But that's by no means urgent.

docs/dsp-tools-usage.md
```diff
@@ -254,61 +254,45 @@ In order to upload data incrementally the procedure described [here](dsp-tools-x
...
-required software installed that makes it possible to run the API and APP. We don't offer support or troubleshooting 
-for these commands.
-
+from time to time to stay up to date, and then there are several commands for each repository to remember. And your 
+local clone of the API might get cluttered with old files over time.
 
-### Start DSP-API
+Another challenge is the software that DSP depends on, and that should be kept up to date: JDK, node, npm, Angular, 
+etc. And it might happen that a dependency is replaced, e.g. JDK 11 Zulu by JDK 17 Temurin. A non-developer can quickly 
+get lost in this jungle.
```
irinaschubert, line 269: I recommend to remove this whole section from the documentation. It is not the right place to inform the user about decisions that were made regarding the setup. I assume the user just wants to know what to do - as briefly and simple as possible. 
We also have the problem in DSP-API that different "users" are targeted by the same documentation (end-users, internal/external software developers). This resulted in a messy documentation over time. We can brainstorm about this topic, if you like. It's not easy to set this up alone.

docs/dsp-tools-usage.md
```diff
@@ -254,61 +254,45 @@ In order to upload data incrementally the procedure described [here](dsp-tools-x
...
 ```
 
-This command makes a clone of the [DSP-API repository](https://github.com/dasch-swiss/dsp-api) into `~/.dsp-tools`. If
-it finds an existing clone there, it runs `git pull` instead. If the API is already running, it shuts down the old 
-instance, deletes all data that was in it, and starts a new one. If the dependencies are outdated or not installed, a 
-warning is printed to the console.
-
-
-### Shut DSP-API down
+This calls Docker commands to start up the latest deployed versions of API and the APP, i.e. the versions that are
```
irinaschubert, line 276:
```suggestion
This calls Docker commands to start up the latest versions of DSP-API and DSP-APP, i.e. the versions that are 
```

docs/dsp-tools-usage.md
```diff
@@ -254,61 +254,45 @@ In order to upload data incrementally the procedure described [here](dsp-tools-x
...
 
-This command makes a clone of the [DSP-API repository](https://github.com/dasch-swiss/dsp-api) into `~/.dsp-tools`. If
-it finds an existing clone there, it runs `git pull` instead. If the API is already running, it shuts down the old 
-instance, deletes all data that was in it, and starts a new one. If the dependencies are outdated or not installed, a 
-warning is printed to the console.
-
-
-### Shut DSP-API down
+This calls Docker commands to start up the latest deployed versions of API and the APP, i.e. the versions that are 
+running on [https://admin.dasch.swiss](https://admin.dasch.swiss). It's no longer necessary to have two local clones,
```
irinaschubert, line 277:
```suggestion
running on [https://admin.dasch.swiss](https://admin.dasch.swiss).
```

docs/dsp-tools-usage.md
```diff
@@ -254,61 +254,45 @@ In order to upload data incrementally the procedure described [here](dsp-tools-x
...
-warning is printed to the console.
-
-
-### Shut DSP-API down
+This calls Docker commands to start up the latest deployed versions of API and the APP, i.e. the versions that are 
+running on [https://admin.dasch.swiss](https://admin.dasch.swiss). It's no longer necessary to have two local clones, 
+nor to execute `make` commands inside them. The only requirement for this command is that Docker must be running. It 
+can even be executed on a Windows machine without any of the software dependencies.
 
+When your work is done, shut down API and APP with
```
irinaschubert, line 281:
```suggestion
Shut down DSP-API and DSP-APP:
```

docs/dsp-tools-usage.md
```diff
@@ -254,61 +254,45 @@ In order to upload data incrementally the procedure described [here](dsp-tools-x
...
 ```
-dsp-tools stop-api
+dsp-tools stop-stack
 ```
 
-This command shuts DSP-API down, deletes all Docker volumes, and removes temporary files.
-
-
-### Start DSP-APP
+This deletes all Docker volumes, and removes all data that was in the database.
```
irinaschubert, line 286:
```suggestion
This command deletes all Docker volumes, and removes all data that was in the database.
```

docs/dsp-tools-usage.md
```diff
@@ -254,61 +254,45 @@ In order to upload data incrementally the procedure described [here](dsp-tools-x
...
-```
-dsp-tools start-app
-```
+Some notes:
 
-This command makes a clone of the [DSP-APP repository](https://github.com/dasch-swiss/dsp-app) into `~/.dsp-tools`. If
-it finds an existing clone there, it runs `git pull` instead. Then, it installs the `npm` dependencies and runs DSP-APP.
-You must keep the terminal window open as long as you work with the APP. Then, you can press `Ctrl` + `C` to stop DSP-APP.
+ - As long as you want to keep the data in the database, don't execute `dsp-tools stop-stack`. It is possible to 
+   leave the API up for a long time. If you want to save power, you can pause Docker. When you resume it, the API
```
irinaschubert, line 291:
```suggestion
   leave the API up for a long time. If you want to save resources, you can pause Docker. When you resume it, the API 
```

docs/dsp-tools-usage.md
```diff
@@ -254,61 +254,45 @@ In order to upload data incrementally the procedure described [here](dsp-tools-x
...
 
-This command makes a clone of the [DSP-APP repository](https://github.com/dasch-swiss/dsp-app) into `~/.dsp-tools`. If
-it finds an existing clone there, it runs `git pull` instead. Then, it installs the `npm` dependencies and runs DSP-APP.
-You must keep the terminal window open as long as you work with the APP. Then, you can press `Ctrl` + `C` to stop DSP-APP.
+ - As long as you want to keep the data in the database, don't execute `dsp-tools stop-stack`. It is possible to 
+   leave the API up for a long time. If you want to save power, you can pause Docker. When you resume it, the API 
+   will still be running, in the state how you left it.
+ - You can also send your computer to sleep while the DSP stack is running. For this, you don't even need to pause 
+   Docker.
+ - This command provides you with the latest released & deployed version of API and APP. If you want to run a
```
irinaschubert, line 295:
```suggestion
 - This command provides you with the latest versions of DSP-API and DSP-APP. If you want to run a 
```

knora/dsp_tools.py
```diff
@@ -149,19 +145,14 @@ def program(user_args: list[str]) -> None:
     parser_excel2xml.add_argument('shortcode', help='Shortcode of the project that this data belongs to')
     parser_excel2xml.add_argument('default_ontology', help='Name of the ontology that this data belongs to')
 
-    # startup DSP-API
-    parser_stackup = subparsers.add_parser('start-api', help='Startup a local instance of DSP-API')
-    parser_stackup.set_defaults(action='start-api')
+    # startup DSP stack
+    parser_stackup = subparsers.add_parser('start-stack', help='Startup a local instance of the DSP stack (API and APP)')
```
irinaschubert, line 149:
```suggestion
    parser_stackup = subparsers.add_parser('start-stack', help='Startup a local instance of the DSP stack (DSP-API and DSP-APP)')
```

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,67 @@
...
+
+def start_stack() -> None:
+    """
+    Start the Docker containers of DSP-API, and load some basic data models and data.
+    """
+    # start up the fuseki database
+    completed_process = subprocess.run("docker compose up db -d", shell=True, cwd="knora/dsplib/docker")
+    if not completed_process or completed_process.returncode != 0:
+        raise BaseError("Cannot start the API: Error while executing 'docker compose up db -d'")
+    time.sleep(5)
```
irinaschubert, line 18: Why is this sleep needed?

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,67 @@
...
+    repo_template = re.sub(r"@REPOSITORY@", "knora-test", repo_template)
+    response = requests.post(
+        url="http://0.0.0.0:3030/$/datasets",
+        files={"file": ("file.ttl", repo_template, "text/turtle; charset=utf8")},
+        auth=("admin", "test")
+    )
+    if not response.ok:
+        raise BaseError("Cannot start the API: Error when creating the 'knora-test' repository")
+
+    # load some basic ontos and data into the repository
```
irinaschubert, line 32: Sometimes you use "onto", sometimes "ontology", sometimes "data model". I would always use the same term for the same thing.

irinaschubert (approved):
LGTM, I think this is a great improvement! 
In general (doesn't need to be part of this PR):
- documentation: we should think about the structure and content (target) of the documentation
- ~~code organization: you added all scripts under `knora.dsplib.utils` but I would keep these things bundled and separated from the other code in their own package / module~~ --> as discussed, you can ignore that, I saw now that you actually deleted all the previous shell scripts
- naming: I would consistently call our components by their name, i.e. `DSP-API` instead of `API` and `DSP-APP` instead of `APP`, especially when using capital letters, because that's their product name. If using `api` or `app` only, I would use lower-case to reference the thing they are, which is an api / app like any other api / app.

knora/dsplib/utils/stack_handling.py
jnussbaum: @BalduinLandolt and/or @seakayone told me to do it, when we were pair programming. I'm happy to remove it if it is not necessary. What do you think?

docs/dsp-tools-usage.md
jnussbaum: I principally agree with you, but there is one reason why I am reluctant: It is now a years-old practice to install all this software on the machines of new employees, and to execute commands in the repositories. For me it was an "aha moment" to learn the distinction between building from the repositories (needs dependencies) versus running Docker containers (no dependencies needed). That's the reason why I feel the need to clearly document this change. If I just communicate it internally in a chat, it will be forgotten after some days/weeks. 

But I'm open for brainstorming about better options.

knora/dsplib/utils/stack_handling.py
BalduinLandolt: that was to avoid needing the `wait-for-db` script. Not sure if it's needed.

docs/index.md
```diff
@@ -2,17 +2,23 @@
 
 # DSP-TOOLS documentation
 
-dsp-tools is a command line tool that helps you to interact with a DaSCH Service Platform (DSP) server.
+DSP-TOOLS is a Python package with a command line tool that helps you to interact with a DaSCH Service Platform (DSP)
```
irinaschubert, line 5: same as above, I don't repeat it here

docs/index.md
```diff
@@ -2,17 +2,23 @@
...
-In order to archive your data on the DaSCH Service Platform, you need a data model (ontology) that describes your data.
+**Create a project with its data model(s), described in a JSON file, on a DSP server**  
+In order to archive your data on the DaSCH Service Platform, you need a data model that describes your data.
 The data model is defined in a JSON project definition file which has to be transmitted to the DSP server. If the DSP 
 server is aware of the data model for your project, conforming data can be uploaded into the DSP repository.
 
-Often, data is initially added in large quantities. Therefore, dsp-tools allows you to perform bulk imports of your
-data. In order to do so, the data has to be described in an XML file. dsp-tools is able to read the XML file and upload
+**Upload data, described in an XML file, to a DSP server that has a project with a matching data model**  
+Often, data is initially added in large quantities. Therefore, DSP-TOOLS allows you to perform bulk imports of your
```
irinaschubert, line 15:
```suggestion
Sometimes, data is added in large quantities. Therefore, DSP-TOOLS allows you to perform bulk imports of your
```

docs/index.md
```diff
@@ -38,5 +44,5 @@ dsp-tools helps you with the following tasks:
 - [`dsp-tools id2iri`](./dsp-tools-usage.md#replace-internal-ids-with-iris-in-xml-file)
   takes an XML file for bulk data import and replaces referenced internal IDs with IRIs. The mapping has to be provided
   with a JSON file.
-- [`dsp-tools start-api / stop-api / start-app`](./dsp-tools-usage.md#start-a-dsp-stack-on-your-local-machine-for-dasch-internal-use-only)
+- [`dsp-tools start-stack / stop-stack`](./dsp-tools-usage.md#start-a-dsp-stack-on-your-local-machine)
   assist you in running a DSP software stack on your local machine.
```
irinaschubert, line 48:
```suggestion
  assist you in running a DSP stack on your local machine.
```

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,122 @@
...
+from knora.dsplib.models.helpers import BaseError
+
+# relative path to "knora/dsplib/docker", to make it accessible when dsp-tools is called from another working directory
+docker_path = Path(__file__).parent / Path("../docker")
+
+
+def start_stack(
+    max_file_size: Optional[int] = None,
+    enforce_docker_system_prune: bool = False,
+    suppress_docker_system_prune: bool = False
```
irinaschubert, line 19: they are mutually exclusive, couldn't they just be represented in one parameter instead of two?

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,122 @@
...
+        suppress_docker_system_prune: if True, don't prune Docker (and don't ask)
+    """
+    if enforce_docker_system_prune and suppress_docker_system_prune:
+        raise BaseError('The arguments "--prune" and "--no-prune" are mutually exclusive')
+
+    # get sipi.docker-config.lua
+    latest_release = json.loads(requests.get("https://api.github.com/repos/dasch-swiss/dsp-api/releases").text)[0]
+    url_prefix = f"https://github.com/dasch-swiss/dsp-api/raw/{latest_release['target_commitish']}/"
+    docker_config_lua_text = requests.get(f"{url_prefix}sipi/config/sipi.docker-config.lua").text
+    if max_file_size is not None:
```
irinaschubert, line 37:
```suggestion
    if max_file_size:
```

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,122 @@
...
+    # inside fuseki, create the "knora-test" repository
+    repo_template = requests.get(f"{url_prefix}webapi/scripts/fuseki-repository-config.ttl.template").text
+    repo_template = repo_template.replace("@REPOSITORY@", "knora-test")
+    response = requests.post(
+        url="http://0.0.0.0:3030/$/datasets",
+        files={"file": ("file.ttl", repo_template, "text/turtle; charset=utf8")},
+        auth=("admin", "test")
+    )
+    if not response.ok:
+        raise BaseError("Cannot start DSP-API: Error when creating the 'knora-test' repository. Is there perhaps "
```
irinaschubert, line 71:
```suggestion
        raise BaseError("Cannot start DSP-API: Error when creating the 'knora-test' repository. Is DSP-API perhaps already running? "
```

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,122 @@
...
+    # docker system prune
+    if enforce_docker_system_prune:
+        prune_docker = "y"
+    elif suppress_docker_system_prune:
+        prune_docker = "N"
+    else:
+        prune_docker = None
+        while prune_docker not in ["y", "N"]:
+            prune_docker = subprocess.run(
+                "read -p \"Allow us executing a docker system prune? This is necessary to keep your Docker clean. If "
```
irinaschubert, line 108:
```suggestion
                "read -p \"Allow dsp-tools to execute a docker system prune? This is necessary to keep your Docker clean. If "
```

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,122 @@
...
+                stdout=subprocess.PIPE,
+                text=True
+            ).stdout.replace("\n", "")
+    if prune_docker == "y":
+        subprocess.run("docker system prune -f", shell=True, cwd=docker_path)
+
+
+def stop_stack() -> None:
+    """
+    Shut down the Docker containers of DSP-API and delete all data that is in them.
```
irinaschubert, line 120:
```suggestion
    Shut down the Docker containers of your local DSP stack and delete all data that is in it.
```

irinaschubert (approved):
LGTM

knora/dsplib/utils/stack_handling.py
jnussbaum: Theoretically yes. There are 3 cases:
 - default: ask user
 - `--prune`: prune
 - `--no-prune`: don't prune

This could theoretically be translated into one parameter:
 - default --> `prune: Optional[bool] = None`
 - `--prune` --> `prune: Optional[bool] = True`
 - `--no-prune` --> `prune: Optional[bool] = False`

This single parameter could then be evaluated with 
 - `if prune==None`
 - `if prune==True`
 - `if prune==False`

But I don't particulary like this construct... :-)

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,122 @@
...
+        while prune_docker not in ["y", "N"]:
+            prune_docker = subprocess.run(
+                "read -p \"Allow dsp-tools to execute a docker system prune? This is necessary to keep your Docker clean. If "
+                "you are unsure what that means, just type y and press Enter. [y/N]\" response; echo $response",
+                shell=True,
+                stdout=subprocess.PIPE,
+                text=True
+            ).stdout.replace("\n", "")
+    if prune_docker == "y":
+        subprocess.run("docker system prune -f", shell=True, cwd=docker_path)
```
BalduinLandolt, line 115: this seems overly complicated to me: why delegate the asking to the docker process? you could simply prompt the user yourself wether or not they want "prune" (in a separate function that returns `true` if they typed "y" or "Y", `false` if "n" or "N", and takes care of re-asking in any other case, maybe `_do_prune()`) and then do `if enforce_docker_system_prune or _do_prune(): ...`

But more generally: Wouldn't it be simpler to define a default behaviour (either prune or no prune, that's up to you), and then only have a flag for the other case.

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,122 @@
...
+
+import requests
+
+from knora.dsplib.models.helpers import BaseError
+
+# relative path to "knora/dsplib/docker", to make it accessible when dsp-tools is called from another working directory
+docker_path = Path(__file__).parent / Path("../docker")
+
+
+def start_stack(
```
BalduinLandolt, line 16: a remark purely on code style: a function should never be 100 lines long (including docstring), it should never "do more than one thing" and it should normally not have to operate on different levels of abstraction. All three of those are the case for this function.  
If you find yourself writing that kind of code (which is normal, that's how one writes code), then you should afterwards refactor it by extracting all sub-tasks and lower-level stuff into separate private functions.  
(I'm freely quoting Bob Martin's "Clean Architecture" here - but it's a good rule of thumb to follow.)

BalduinLandolt (approved):
I added some remarks on what I think could be improved - but I don't want to keep you from rolling the feature out. So feel free to consider/address them in a separate task/PR

knora/dsplib/utils/stack_handling.py
```diff
@@ -0,0 +1,122 @@
...
+    # get sipi.docker-config.lua
+    latest_release = json.loads(requests.get("https://api.github.com/repos/dasch-swiss/dsp-api/releases").text)[0]
+    url_prefix = f"https://github.com/dasch-swiss/dsp-api/raw/{latest_release['target_commitish']}/"
+    docker_config_lua_text = requests.get(f"{url_prefix}sipi/config/sipi.docker-config.lua").text
+    if max_file_size is not None:
+        if not 1 <= max_file_size <= 100_000:
+            raise BaseError("max_file_size must be between 1 and 100000")
+        max_post_size_regex = r"max_post_size ?= ?[\'\"]\d+M[\'\"]"
+        if not re.search(max_post_size_regex, docker_config_lua_text):
+            raise BaseError("sipi.docker-config.lua doesn't contain a variable max_post_size that could be replaced")
```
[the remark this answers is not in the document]
jnussbaum: Thanks for the hints. One of my long-term ideas is to raise different kinds of exceptions in the code, and then catching the user-feedback-exceptions in `dsp_tools.py`, print the message, and then `exit()` without stack trace. Only the "real errors" would escalate uncaught and lead to a stack trace. Do you think that would be a good design pattern?

Or even easier: if an error occurs which is the user's fault, print message and `exit()`. Use BaseError only if an internal-caused error occurs.

I will do that in a separate PR, but what do you think about it?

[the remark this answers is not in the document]
BalduinLandolt: That's definitely a reasonable and common approach. You can then create an inheritance hierarchy of exceptions for different categories of errors, and things like that.  
And it will definitely be a great improvement to how it is currently.

Personally I'm not the biggest fan of exceptions, but in Python it's not exactly easy to find better solutions. To me, exceptions always feel like "blowing up stuff", and there is nothing very "exceptional" about a user entering a wrong parameter... But as I said, there is no super-neat pythonic way of a controlled and typed error channel like we have in ZIO.

knora/dsplib/utils/stack_handling.py
BalduinLandolt: fair enough - and I agree that pruning by default would be risky.  
Then I still think it would be simpler in code and more user friendly not to have this "ask" case, where you pipe stuff back and forth between DSP-Tools and docker, but rather you ask them `input("do you want to call 'docker prune'? [Y/N ] ")`

BalduinLandolt: oops... writing at the same time... and now I understand what's happening :).
yes, that will streamline it a bit.

knora/dsplib/utils/stack_handling.py
jnussbaum: Thanks for the hint, it's important that I learn such stuff about architecture. I think I won't do it in this PR. But a general question concerning this rule of thumb: Doesn't it make the code more complicated if I have a lot of private methods that are so specific that are only called once?

BalduinLandolt: Bob Martin (and I) would say: no, your code should get more readable and easier to understand.

You have to get some things right though:  
- a method name should be descriptive of what the method does
- a method must only operate on one abstraction level, namely calling methods of the next lower abstraction level

(and there are more finesses of course)

If you follow those rules, all your methods should turn out pretty much self explanatory: the name says what it does and the (maybe 3 to 10) function calls within it are self-explaining steps that need to be taken, to achieve this.  
And if you are interested in further detail, you can jump into each of those steps, where the same thing will be true: the function name is what it does, and the contents are the steps taken to do it.

Does that make sense?  
If you're interested, there is - among tons of other material - a several hour long talk series by "Uncle Bob" named "Clean Code" on Youtube, which is definitely worth watching.

PR #266, opened 2022-12-12
fix(xmlupload): improve URL recognition (DEV-1557)

jnussbaum:
resolves DEV-1557


knora/excel2xml.py
```diff
@@ -1339,10 +1340,9 @@ def make_uri_prop(
...
     # check value type
     for val in values:
-        # URI = scheme ":" ["//" host [":" port]] path ["?" query] ["#" fragment]
-        if not regex.search(
-            r"(?<scheme>[a-z][a-z0-9+.\-]*):(//(?<host>[\w_.\-\[\]:~]+)(?<port>:\d{0,6})?)(?<path>/[\p{L}%()_\-.~]*)*"
-            r"(?<query>\?[\p{L}_.\-=]+)*(?<fragment>#[\p{L}_/\-~:.]*)?", str(val.value), flags=regex.UNICODE):
+        try:
+            UriValue(str(val.value))
+        except BaseError:
             raise BaseError(f"Failed validation in resource '{calling_resource}', property '{name}': "
```
BalduinLandolt, line 1346: this seems like a bit of an anti-pattern: Raising an exception just to discard it again and raising the same exception again. If you moved the nicer error message to the `UriValue` object, you would not need to do the `try` or the `raise` here at all.

BalduinLandolt (approved):
LGTM, I remarked only one minor thing

irinaschubert (approved):

PR #273, opened 2022-12-20
fix(start-stack): use TTL files from DSP-API v24.0.8 (DEV-1580)

jnussbaum:
resolves DEV-1580


irinaschubert (changes requested):
As discussed, it would probably make sense to not support Windows right now, as there are no known users at the moment who would use dsp-tools on an Windows machine. In case this is mentioned in the documentation, it should be deleted.

irinaschubert (approved):

PR #279, opened 2023-01-10
chore: use start-stack command for e2e tests, replace Makefile by poetry-exec-plugin (DEV-1597)

jnussbaum:
resolves DEV-1597

- make docs/index.md more suitable for PyPI landing page
- include "anything" onto+data in start-stack command, so that this command can be used for the e2e tests
- replace Makefile by poetry-exec-plugin
- first time using Semantic Line Breaks (https://sembr.org/) 
- remove daily tests, because now, the e2e tests don't run any more with the daily-changing main branch of DSP-API, but always with the same Docker images
- tidy up all YML GitHub Actions files, modify/remove some of the names


docs/index.md
```diff
@@ -2,46 +2,50 @@
 
 # DSP-TOOLS documentation
 
-DSP-TOOLS is a Python package with a command line interface that helps you interact with a DSP server. The DSP server
```
jnussbaum, line 5: The changes in this file attempt to make the landing page on https://pypi.org/project/dsp-tools/ more readable. From line 22 on, it was for example necessary to put the full links to the docs pages, instead of relative links.

src/dsp_tools/utils/stack_handling.py
```diff
@@ -91,7 +93,9 @@ def start_stack(
         ("knora-ontologies/standoff-data.ttl", "http://www.knora.org/data/standoff"),
         ("knora-ontologies/salsah-gui.ttl", "http://www.knora.org/ontology/salsah-gui"),
         ("test_data/all_data/admin-data-minimal.ttl", "http://www.knora.org/data/admin"),
-        ("test_data/all_data/permissions-data-minimal.ttl", "http://www.knora.org/data/permissions")
+        ("test_data/all_data/permissions-data-minimal.ttl", "http://www.knora.org/data/permissions"),
```
jnussbaum, line 96: The e2e tests rely on `anything`, so I had to add `anything` here.

.github/workflows/release.yml
```diff
@@ -1,38 +1,35 @@
-name: Release
+name: Release to PyPI
```
irinaschubert, line 1: Not sure, but this formulation sounds strange to me. You release a version. And you publish this version on a platform (PyPI). But I wouldn't say that you release a version to a platform.

.github/workflows/release.yml
```diff
@@ -1,38 +1,35 @@
...
 env:
   TOKEN: ${{ secrets.GH_TOKEN }}
   POETRY_HTTP_BASIC_PYPI_USERNAME: ${{ secrets.PYPI_USER }}
   POETRY_HTTP_BASIC_PYPI_PASSWORD: ${{ secrets.PYPI_PW }}
 
 jobs:
-  # release to PyPI
   release-pypi:
-    name: Release to PyPI
+    name: release-to-pypi
```
irinaschubert, line 14: Is there a reason why you changed the name here to kebab case?

src/dsp_tools/utils/stack_handling.py
irinaschubert: I'm not sure how the test setup is now, but I wouldn't add this dependency to DSP-TOOLS. It might break if we change this dataset in DSP-API.
Would it be possible to create new tests that don't rely on the "anything" project?

irinaschubert (approved):
LGTM

src/dsp_tools/utils/stack_handling.py
jnussbaum: Well, until now it worked quite well... If it breaks some day, I'm of course eager to fix it (or rewrite the tests). But rewriting the tests just because it **might** break one day would be a bit demotivating. Because it's quite a lot of testing code that uses the anything data...

.github/workflows/release.yml
jnussbaum: good point. Changed it to publish.

.github/workflows/release.yml
jnussbaum: I tried to harmonize the naming convention across the different YML files. But motivated by your comment, I had a 2nd look at the names and found out that some of them aren't even necessary, and can be removed, because the name is already defined by the YAML-key.

Makefile
```diff
@@ -44,25 +17,24 @@ install: ## install Poetry, which in turn installs the dependencies and makes an
 	poetry install
 
 .PHONY: test
-test: dsp-stack ## run all tests located in the "test" folder (intended for local usage)
-	-poetry run pytest test/	# ignore errors, continue anyway with stack-down (see https://www.gnu.org/software/make/manual/make.html#Errors)
-	$(MAKE) stack-down
+test: ## run all tests located in the "test" folder
+	poetry run dsp-tools start-stack --no-prune
+	-poetry run pytest test/	# ignore errors, continue anyway with stop-stack (see https://www.gnu.org/software/make/manual/make.html#Errors)
+	poetry run dsp-tools stop-stack
```
BalduinLandolt, line 23: you could also look into the [poetry exec plugin](https://github.com/keattang/poetry-exec-plugin) to automate poetry workflows, which might spare you even more of the makefile

src/dsp_tools/utils/stack_handling.py
```diff
@@ -91,7 +93,9 @@ def start_stack(
         ("knora-ontologies/standoff-data.ttl", "http://www.knora.org/data/standoff"),
         ("knora-ontologies/salsah-gui.ttl", "http://www.knora.org/ontology/salsah-gui"),
         ("test_data/all_data/admin-data-minimal.ttl", "http://www.knora.org/data/admin"),
```
BalduinLandolt, line 95: as mentioned in chat, this should not be minimal, so that the tests pass.

```suggestion
        ("test_data/all_data/admin-data.ttl", "http://www.knora.org/data/admin"),
```

src/dsp_tools/utils/stack_handling.py
BalduinLandolt: I agree that this dependency is not ideal, but I also think it's not urgent. Over time it would be good to work towards having less e2e tests (only as many as it really needs) and more unit tests. Then you would also have less tests depending on the `anything` project.

BalduinLandolt (approved):
looks good over all, just some minor suggestions.  
To fix the failing tests, I think you can just use the full `admin` (and maybe `permissions`) data turtle files, instead of the `minimal` ones.

Makefile
jnussbaum: Good idea, it would be cool to get rid of the Makefile, esp. because there are only 2 commands left in it, each a 1-liner. But I'm a bit reluctant using a plugin which is still pre-1.0... What do you think?

BalduinLandolt: I think it would be worth trying, even if it's only to see how well it works. In any case, it would be easy enough to revert back to having the make file (and if it's only two one-liners, you could just as well put them into the readme and execute them manually).  
Speaking of those two commands: in the `make clean` command, are most of them still needed? I would guess that both mkdocs (for the docs) and poetry (for building) take care of removing old files pretty well.

jnussbaum: OK, done :-) 
There is only one exec-script remaining, and that one is only used by me. This allowed me to keep the documentation of the poetry-exec-plugin minimal.

README.md
```diff
@@ -87,14 +87,22 @@ Please ensure you have only one pull request per feature.
 
 ## Testing
 
-Please note that testing requires launching the complete DSP API stack which is based on docker images. 
-Therefore, we recommend installing the [docker desktop client](https://www.docker.com/products).  
-To run the complete test suite:
+The tests of this repository
```
jnussbaum, line 90: Here I use Semantic Line Breaks (https://sembr.org/) for the first time. It sounds convincing, but I'm curious to hear your feedback.

One of the big advantages, which is not even described in the specification: we will get cleaner git diffs, because changing a single word in the middle of a paragraph won't force us any more to change all line breaks of that paragraph.

PR #280, opened 2023-01-13
refactor: project creation (DEV-1165)

jnussbaum:
resolves DEV-1165


irinaschubert (approved):
LGTM

src/dsp_tools/utils/project_create.py
```diff
@@ -6,72 +6,142 @@
...
 
-def _create_project(con: Connection, project_definition: dict[str, Any]) -> Project:
+def _create_project(
+        input_file: str,
+        server: str,
+        user_mail: str,
+        password: str,
+        verbose: bool,
+        dump: bool
+) -> Tuple[Connection, Context, dict[str, Any], Project, bool]:
```
BalduinLandolt, line 29: these kind of return types normally indicate that you might want to consider using a DTO (data transfer object), which could be a simple dataclass like this

```python
@dataclass(frozen=True)
class createProjectResponse:  # or come up with a better name
    connection: Connection
    context: Context
    json: dict[str, Any]
    project: Project
    success: bool
```

so, essentially it's really just a named tuple, but it's far more expressive than a simple Tuple.  
For now you can absolutely leave it like this. But in the future, it may be something to consider

src/dsp_tools/utils/project_create.py
```diff
@@ -6,72 +6,142 @@
...
+    updated. Returns the updated project (or the unchanged project if not successful) and a boolean saying if the update
+    was successful or not. If the update was not successful, an error message is printed to stdout.
 
     Args:
         project: the project to be updated (must exist on the DSP server)
         project_definition: a parsed JSON project file with the same shortname and shortcode than the existing project
 
     Returns:
-        updated project
+        tuple of (updated project/None, success status)
```
BalduinLandolt, line 128: in what case would project be `None`? Also, the return type does not reflect that

BalduinLandolt (approved):
Maybe not the most thorough review ever, from my side. But looks good

src/dsp_tools/utils/project_create.py
jnussbaum: cool, I note that for later

src/dsp_tools/utils/project_create.py
jnussbaum: That's a leftover that I forgot to tidy up. Thanks for spotting it

PR #288, opened 2023-01-20
chore: access data files correctly (DEV-1618)

jnussbaum:
resolves DEV-1618


docs/developers/user-data.md
```diff
@@ -1,11 +1,98 @@
...
+```python
+import os
+data_path = os.path.join(os.path.dirname(__file__), 'schemas', 'data.xsd')
+with open(data_path) as data_file:
+     ...
+```
+
+However, this manipulation isn’t compatible with PEP 302-based import hooks, 
+including importing from zip files and Python Eggs.
+
```
irinaschubert, line 81: Once more, I don't think that explanations how things were in earlier time should be included in documentation. I would prefer software documentation to be consise.

docs/developers/user-data.md
```diff
@@ -1,11 +1,98 @@
...
+# address "schemas" directory in module syntax: needs __init__py
+data_text = files('dsp_tools.schemas').joinpath('data.xsd').read_text()
+# avoid module syntax when addressing "schemas" directory: no __init__.py necessary
+data_text = files('dsp_tools').joinpath('schemas').joinpath('data.xsd').read_text()
+```
+
+Note that depending on how the directory is addressed, 
+an `__init__.py` file is necessary or can be omitted.
+
+The information on this page is mainly based upon:
```
irinaschubert, line 95: I would add a date, something like "last checked on January 24, 2023". Because content behind links can quickly change.

src/dsp_tools/utils/excel_to_json_lists.py
```diff
@@ -253,9 +254,8 @@ def validate_lists_section_with_schema(
     """
     if bool(path_to_json_project_file) == bool(lists_section):
         raise BaseError("Validation of the 'lists' section works only if exactly one of the two arguments is given.")
-    current_dir = os.path.dirname(os.path.realpath(__file__))
-    with open(os.path.join(current_dir, "../schemas/lists-only.json")) as schema:
-        lists_schema = json.load(schema)
+    with importlib.resources.files("dsp_tools").joinpath("schemas").joinpath("lists-only.json").open() as schema_file:
```
irinaschubert, line 257:
```suggestion
    with files("dsp_tools").joinpath("schemas").joinpath("lists-only.json").open() as schema_file:
```
or is the module name needed here?

irinaschubert (approved):

docs/developers/user-data.md
jnussbaum: I thought that the "developers" part of the docs are here for exactly this... ? I created this part of the docs to write all the stuff there that I would not write into the normal documentation. I thought that the "developers" part can be more reflective, with historical things, with reasons why we do some things... Or did I get that wrong?

irinaschubert: Of course, it very much depends on the goal of the documentation and the intended audience. But to me, documentation should only be about _how_ to do things. The _why_ should go into decision records. If I would need to read through paragraphs about how things were done previously - it would cost me time and effort without getting to the relevant information that I actually need to do something.

Also, the _why_ is important for big decisions (e.g. architecture) and not so much small changes. From this perspective, the way to open a file is an implementation detail. If you would describe every change in this level of detail, you would spend a lot of time writing documentation :) But maybe I don't fully understand you point of view: so, I would be curious to hear what you think. Why does a developer need to know how files were opened previously?

PR #289, opened 2023-01-24
chore: update links according to new structure of docs (DEV-1648)

jnussbaum:
resolves DEV-1648


jnussbaum:
This PR must be merged even though the tests fail. This is because it contains links to docs.dasch.swiss that are invalid  now. They will only become valid once docs.dasch.swiss is updated, which happens via this PR.

PR #293, opened 2023-02-07
feat(excel2json): add support for gui_order, make templates better understandable (DEV-1711)

jnussbaum:
resolves DEV-1711


src/dsp_tools/utils/excel_to_json_properties.py
```diff
@@ -1,14 +1,25 @@
...
 
+# Pandas relies on openpyxl to parse XLSX files.
+# A strange behaviour of openpyxl prevents pandas from opening files with some formatting properties
+# (unclear which formatting properties exactly).
+# Apparently, the excel2json template files have one of the unsupported formatting properties.
+# The following two lines of code help out.
+# Credits: https://stackoverflow.com/a/70537454/14414188
+p = mock.patch('openpyxl.styles.fonts.Font.family.max', new=100)
+p.start()
+
```
jnussbaum, line 22: This is obscure, but the only way to make it work. If you have a better solution, I'm open for feedback!

irinaschubert: Would it be possible to find out which of the formatting properties cause the problem and get rid of it?

src/dsp_tools/utils/excel_to_json_resources.py
```diff
@@ -56,12 +70,35 @@ def _row2resource(row: pd.Series, excelfile: str) -> dict[str, Any]:
         location_of_sheet=f"Sheet '{name}' in file '{excelfile}'"
     )
 
+    # validation
+    # 4 cases:
+    #  - column gui_order absent
+    #  - column gui_order empty
+    #  - column gui_order present but not properly filled in (missing values / not integers)
+    #  - column gui_order present and properly filled in
+    all_gui_order_cells = [x for x in details_df.get("gui_order", []) if x]
```
irinaschubert, line 79: I'm wondering if `gui_order` should be an attribute that can be set by the user. I think it would be more user-friendly to just take the order given in the Excel sheet (which is already the case). I can't imagine a use case where a user actually _wants_ to set the gui order him/herself other than the order in the Excel sheet.

BalduinLandolt (approved):
I really don't like the monkey-patching to read strange excel files, but I wouldn't know of a better way to deal with it.  
Unless that's very common, I personally would rather address it as data-cleanup rather than in code, but that's up to you.

src/dsp_tools/utils/excel_to_json_properties.py
jnussbaum: Unfortunately, the Excel file is not exotic. I just took the existing template, and added comments to it. Nothing special. 
After reading the SO answers, I tried to reduce the number of used fonts, but to no avail: The error still appears. It's odd that in some SO answers, the comment boxes seemed to be the root of the problem, but others reported that the number of fonts was the problem.
I think it's a strange behavior of openpyxl that I should perhaps report.

One important reason why I would like to leave the monkey-patching in place: I think it's very probable that the error will also occur on client side.

src/dsp_tools/utils/excel_to_json_resources.py
jnussbaum: This feature was requested by Noémi, she would find it useful.

src/dsp_tools/utils/excel_to_json_resources.py
```diff
@@ -44,24 +46,62 @@ def _row2resource(row: pd.Series, excelfile: str) -> dict[str, Any]:
...
-    labels = {lang: row[lang] for lang in languages if row.get(lang)}
+    labels = {lang: row[f"label_{lang}"] for lang in languages if row.get(f"label_{lang}")}
+    if not labels:
+        labels = {lang: row[lang] for lang in languages if row.get(lang)}
     comments = {lang: row[f"comment_{lang}"] for lang in languages if row.get(f"comment_{lang}")}
     supers = [s.strip() for s in row["super"].split(",")]
 
     # load the cardinalities of this resource
-    details_df = pd.read_excel(excelfile, sheet_name=name)
+    try:
```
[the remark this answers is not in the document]
irinaschubert: If you didn't find another solution, I'm totally OK with it (also the way you did it before). I'm not sure if adding a try/catch makes it better ;)
One other way I could think of is removing all formatting from the Excel file before loading it, but you've probably already tried this?

irinaschubert (approved):

src/dsp_tools/utils/excel_to_json_resources.py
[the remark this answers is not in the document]
jnussbaum: Is it technically possible to modify the formatting **before** parsing the file? The problem is that openpyxl is not able to open the file, so I would need another library that opens it for me, deletes the formatting, writes it back to disk, so that openpyxl can then open it...

(Or did you mean something else?)

src/dsp_tools/utils/excel_to_json_properties.py
BalduinLandolt: with what program did you do that? some of the comments mentioned WPS causing this issue.

In any case, you should definitely report this problem, if you feel it's a bug in the library.

src/dsp_tools/utils/excel_to_json_resources.py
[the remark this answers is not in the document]
BalduinLandolt: ideally one would use a formatting agnostic format like CSV. but I don't think we should over-complicate it at this point. It works, which is the most important thing. :)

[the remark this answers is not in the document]
irinaschubert: A short google search brought me to xlwings: https://docs.xlwings.org/en/stable/api.html
Maybe it is worth a try..?

src/dsp_tools/utils/excel_to_json_properties.py
BalduinLandolt: well... ok then. In that case it's probably really the fault of openpyxl

PR #296, opened 2023-02-13
fix(excel2xml): proper handling of special characters in text properties, depending on the encoding (DEV-1719)

jnussbaum:
resolves DEV-1719


src/dsp_tools/utils/project_create.py
```diff
@@ -744,11 +744,6 @@ def _add_cardinalities_to_resource_classes(
             if ":" in card_info["propname"]:
                 prefix, prop = card_info["propname"].split(":")
                 qualified_propname = card_info["propname"] if prefix else f"{ontology_remote.name}:{prop}"
-                if not remote_prop_classes.get(ontology_remote.id + "#" + prop):
```
jnussbaum, line 747: This validation is misleading and unnecessary.
Until now, it worked because of a peculiarity of the test onto. But now that I modified the test onto, this validation became a handicap: `remote_prop_classes` doesn't contain properties from other ontos from the same JSON file. 
I realised that the validation is unnecessary, because some lines further down, an attempt is made to add the cardinality to the resource class, and if the attempt is not successful, the same warning will be printed.

src/dsp_tools/excel2xml.py
```diff
@@ -1684,6 +1685,25 @@ def _name_label_mapper_iterator(json_subset: list[dict[str, Any]], language_labe
             # the actual values of the name and the label
 
 
+def _unescape_html_tags(xml_string: str) -> str:
```
jnussbaum, line 1688: It is still an open question whether it should be allowed to have > and < in text properties of an XML file. By default, they are replaced by the escape sequences &gt; and &lt; by lxml. Until now, I replaced the escape sequences by the original tags in write_xml(), but this can lead to an error some lines further down in the same method, in the validation: If a text property contains for example: My text … <Fantasy> my text… , this is invalid (this actually happened to me).

The problem is non-trivial: In most cases, the tags should be escaped to prevent validation errors. But in XML-standoff formatted texts, the tags must be unescaped.

The only solution that I found is the present one. I took the list of tags from https://github.com/dasch-swiss/salsah-export/blob/master/salsah2xml.py#L74, and hope it's exhaustive.

BalduinLandolt: regarding these replacements: I think the way it should work is
- you have an input string which may or may not be XML, depending on if its rich or simple text
- in case of simple text, `<>` should be escaped because they don't bear special meaning in a plain text string, as is done by lxml
- in case of richtext, we expect it to be an XML string, so `<>` bear special meaning and must not be escaped. In this case, the string should be parsed to an XML DOM by lxml (in which case it wouldn't escape characters), and be added as child node to the `<text>` element.

in that case, the whole un-escaping should not be needed anymore

BalduinLandolt (commented):
See my comments here and in chat

src/dsp_tools/excel2xml.py
```diff
@@ -444,7 +445,7 @@ def make_bitstream_prop(
                       stacklevel=2)
     prop_ = etree.Element("{%s}bitstream" % (xml_namespace_map[None]), permissions=permissions,
```
irinaschubert, line 446: It's a detail, but I'm curious: why do you append a `_` to `prop_`? Is prop a reserved name in Python?

irinaschubert (approved):
As discussed in chat, this whole topic about text values needs further discussion with PM.

docs/excel2xml-module.md
```diff
@@ -258,23 +258,11 @@ else:
     excel2xml.make_boolean_prop(":hasBoolean", False)
 ```
 
-#### Supported text values
```
jnussbaum, line 261: I think that text values " " and "-" are only a theoretical concern. I don't need to bother the end user with such subtleties. (The way how excel2xml handles this remains unchanged. I just kicked it out of the docs).

test/unittests/test_xmlupload.py
```diff
@@ -71,7 +71,7 @@ def test_remove_circular_references(self) -> None:
                     '                <a class="salsah-link" href="IRI:zip_thing_1:IRI">zip_thing_1</a>\n'
                     '                <a class="salsah-link" href="IRI:audio_thing_1:IRI">audio_thing_1</a>\n'
                     '                <a class="salsah-link" href="IRI:test_thing_2:IRI">test_thing_2</a>\n'
-                    '            \n            '
```
jnussbaum, line 74: Don't ask me why I had to remove these whitespaces. I don't know!

testdata/excel2xml-expected-output.xml
```diff
@@ -112,7 +112,8 @@
             <text permissions="prop-restricted" encoding="utf8">Nochmals ein einfacher Text</text>
         </text-prop>
         <text-prop name=":hasRichtext">
-            <text permissions="prop-default" encoding="xml">This is <em>bold and <strong>strong</strong></em> text! It contains links to all resources:
+            <text permissions="prop-default" encoding="xml">This is <em>bold and <strong>strong</strong>
+                </em> text! It contains links to all resources:
```
jnussbaum, line 116: I admit that the pretty-printing is not optimal, but that's how it is.

testdata/test-data-systematic.xml
```diff
@@ -59,17 +59,28 @@
...
-            <text encoding="utf8">!</text>
-            <text encoding="utf8">?</text>
-            <text encoding="utf8">-</text>
-            <text encoding="utf8">œ</text>
+        <text-prop name=":hasSimpleText">
+            <text encoding="utf8">&lt;</text>
+            <text encoding="utf8">&amp;lt;</text>
+            <text encoding="utf8">&amp;</text>
+            <text encoding="utf8">&amp;amp;</text>
+            <!-- <text encoding="utf8">&lt;em&gt;text&lt;/em&gt;</text>
```
jnussbaum, line 67: These two are possible, but discouraged.

src/dsp_tools/excel2xml.py
```diff
@@ -1217,7 +1217,19 @@ def make_text_prop(
...
         )
-        value_.text = val.value
+        if kwargs["encoding"] == "utf8":
+            # write the text into the tag, without validation
+            value_.text = str(val.value)
+        else:
+            # enforce that the text is well-formed XML: serialize tag, insert text, and parse it again
+            value_.text = "replace-me_8(1-)replace-me"
+            content = etree.tostring(value_, encoding="unicode")
+            content = content.replace("replace-me_8(1-)replace-me", str(val.value))
```
BalduinLandolt, line 1227: Can you elaborate what's happening here? Without fully understanding it, I am inclined to say that there has to be a way to achieve this without inserting and replacing some placeholder

src/dsp_tools/excel2xml.py
```diff
@@ -1217,7 +1217,19 @@ def make_text_prop(
...
+            try:
+                value_ = etree.fromstring(content)
+            except etree.XMLSyntaxError:
+                raise BaseError("The XML tags contained in a richtext property (encoding=xml) must be well-formed. "
+                                "The special characters <, > and & are only allowed to construct a tag.") from None
```
BalduinLandolt, line 1232: just curious: what does `from None` do here?

src/dsp_tools/utils/shared.py
```diff
@@ -96,12 +97,34 @@ def validate_xml_against_schema(input_file: str) -> bool:
...
+    
+    # make sure there are no XML tags in simple texts
+    lines_with_illegal_xml_tags = list()
+    for text in doc_without_namespace.findall(path="resource/text-prop/text"):
+        if text.attrib["encoding"] == "utf8":
+            if regex.search(r'<([a-zA-Z/"]+|\S.*\S)>', str(text.text)) or len(text.getchildren()) > 0:
+                lines_with_illegal_xml_tags.append(text.sourceline)
+    if lines_with_illegal_xml_tags:
+        raise BaseError(f"XML-tags are not allowed in text properties with encoding=utf8. "
+                        f"The following lines of your XML file are affected: {lines_with_illegal_xml_tags}")
```
BalduinLandolt, line 127: could this check be delegated to the schema?

BalduinLandolt (commented):
I don't like the replace-placeholder solution in one place, see comment below. Maybe we can find a nicer way to do this?

src/dsp_tools/excel2xml.py
jnussbaum: That's an exciting new thing that I just learned. In fact, there are many cases in the DSP-TOOLS codebase where I try to beautify an error with 

```
try:
    something()
except DifficultErrorWithComplicatedStacktrace:
    raise BaseError("Nice Message with simpler stacktrace")
```

Now I just learnt that this doesn't work as intended. In fact, the result is even uglier than what I tried to prevent: In addition to the complicated stacktrace, I get the following: `During handling of the above exception, another exception occurred`.

The solution is to add `from None` behind the `raise` statement. It is explained here: https://stackoverflow.com/questions/52725278/#comment103220039_52725410.

In a separate PR, I will add `from None` to all occurrences in the codebase.

BalduinLandolt: I see, nice

src/dsp_tools/utils/shared.py
jnussbaum: I don't think so. I haven't found a solution in XML Schema

BalduinLandolt: ok, I guess this will have to do, then

src/dsp_tools/excel2xml.py
jnussbaum: Better now?

src/dsp_tools/excel2xml.py
```diff
@@ -1697,10 +1710,8 @@ def write_xml(root: etree.Element, filepath: str) -> None:
         None
     """
     etree.indent(root, space="    ")
-    xml_string = etree.tostring(root, encoding="unicode", pretty_print=True)
-    xml_string = '<?xml version="1.0" encoding="UTF-8"?>\n' + xml_string
-    xml_string = xml_string.replace("&lt;", "<")
-    xml_string = xml_string.replace("&gt;", ">")
+    xml_string = etree.tostring(root, encoding="unicode", pretty_print=True, doctype='<?xml version="1.0" encoding="UTF-8"?>')
+    xml_string = xml_string.replace("\'", "'")
```
jnussbaum, line 1714: One new problem popped up: 
```
>>> etree.tostring(excel2xml.make_text_prop(":test", "don't"), encoding="unicode")
'<text-prop name=":test"><text>don\'t</text></text-prop>'
```
If I upload such an XML to DSP, DSP-APP displays the backslash as is. 
If I want to get rid of the backslash, I only see the solution to replace it before writing the file. Do you see a better option?

src/dsp_tools/excel2xml.py
```diff
@@ -1217,7 +1217,20 @@ def make_text_prop(
...
+            # write the text into the tag, without validation
+            value_.text = str(val.value)
+        else:
+            # enforce that the text is well-formed XML: serialize tag ...
+            content = etree.tostring(value_, encoding="unicode")
+            # ... insert text at the very end of the string, and add an ending tag to the previously single <text/> tag ...
+            content = re.sub(r"/>$", f">{val.value}</text>", content)
+            # ... try to parse it again
+            try:
+                value_ = etree.fromstring(content)
```
BalduinLandolt, line 1230: it looks much neater now (and I can of course live with that). But it still acts in a complicated way:
1. create an etree Element
2. serialize that element to a string
3. add the provided value to the string by means of replacing
4. parse the modified string to an etree Element.

IMO it should be:
1. create an Element
2. parse the provided content to an etree Element structure
3. add this structure as children to the Element

If that proves tricky, then go ahead and merge it like this... but I think the other way would be cleaner.

BalduinLandolt (approved):
See my comment. Feel free to try to adapt my suggestion, or leave it as is for now

PR #297, opened 2023-02-13
fix(create): improve validation to prevent endless loop if properties are undefined (DEV-1720)

jnussbaum:
resolves DEV-1720

### DIFF
--- a/src/dsp_tools/utils/project_validate.py
+++ b/src/dsp_tools/utils/project_validate.py
@@ -13,9 +13,100 @@ from dsp_tools.models.helpers import BaseError
 from dsp_tools.utils.excel_to_json_lists import expand_lists_from_excel
 
 
-def check_for_undefined_cardinalities(project_definition: dict[str, Any]):
-    for onto in project_definition["project"].get("ontologies"):
+def  check_for_undefined_superproperty(project_definition: dict[str, Any]) -> bool:
+    """
+    Check if the properties referenced in "super" of each property actually point to a property 
+    (DSP base properties and properties from other ontologies are not considered.)
+
+    Args:
+        project_definition: parsed JSON project definition
+
+    Raises:
+        BaseError: detailed error message
+
+    Returns:
+        True if the superproperties are valid
+    """
+    errors: dict[str, list[str]] = dict()
+    for onto in project_definition["project"]["ontologies"]:
+        ontoname = onto["name"]
+        propnames = [p["name"] for p in onto["properties"]]
+        for prop in onto["properties"]:
+            supers = prop["super"]
+            # form of supers:
+            #  - isSequenceOf  # DSP base property
+            #  - other:prop    # other onto
+            #  - same:prop     # same onto
+            #  - :prop         # same onto (short form)
+            
+            # filter out DSP base properties
+            supers = [s for s in supers if ":" in s]
+            # extend short form
+            supers = [regex.sub(r"^:", f"{ontoname}:", s) for s in supers]
+            # filter out other ontos
+            supers = [s for s in supers if regex.search(f"^{ontoname}:", s)]
+            # convert to short form
+            supers = [regex.sub(ontoname, "", s) for s in supers]
+
+            invalid_references = [s for s in supers if regex.sub(":", "", s) not in propnames]
+            if invalid_references:
+                errors[f"Ontology '{ontoname}', property '{prop['name']}'"] = invalid_references
+    
+    if errors:
+        err_msg = "Your data model contains properties that are derived from an invalid super-property:"
+        for loc, invalids in errors.items():
+            err_msg = err_msg + f"\n - {loc}: {invalids}"
+        raise BaseError(err_msg)
+    else:
+        return True
+
+
+def check_for_undefined_cardinalities(project_definition: dict[str, Any]) -> bool:
+    """
+    Check if the propnames that are used in the cardinalities of each resource are defined in the "properties" 
+    section. (DSP base properties and properties from other ontologies are not considered.)
+
+    Args:
+        project_definition: parsed JSON project definition
+
+    Raises:
+        BaseError: detailed error message
+
+    Returns:
+        True if all cardinalities are defined in the "properties" section
+    """
+    errors: dict[str, list[str]] = dict()
+    for onto in project_definition["project"]["ontologies"]:
+        ontoname = onto["name"]
         propnames = [prop["name"] for prop in onto["properties"]]
+        for res in onto["resources"]:
+            cardnames = [card["propname"] for card in res["cardinalities"]]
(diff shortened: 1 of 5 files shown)


src/dsp_tools/utils/project_validate.py
```diff
@@ -13,6 +13,102 @@
...
+            err_msg = err_msg + f"\n - {loc}: {invalids}"
+        raise BaseError(err_msg)
+    else:
+        return True
+
+
+def check_for_undefined_cardinalities(project_definition: dict[str, Any]) -> bool:
+    """
+    Check if the propnames that are used in the cardinalities of each resource are defined in the "properties" 
+    section. (DSP base properties and properties from other ontologies are not considered.)
```
irinaschubert, line 67:
```suggestion
    section. (knora-base properties and properties from other ontologies are not considered.)
```

`knora-base` still exists as ontology. So I would call it like that.

src/dsp_tools/utils/project_validate.py
```diff
@@ -13,6 +13,102 @@
...
+        True if all cardinalities are defined in the "properties" section
+    """
+    errors: dict[str, list[str]] = dict()
+    for onto in project_definition["project"]["ontologies"]:
+        ontoname = onto["name"]
+        propnames = [prop["name"] for prop in onto["properties"]]
+        for res in onto["resources"]:
+            cardnames = [card["propname"] for card in res["cardinalities"]]
+            # form of the cardnames:
+            #  - isSequenceOf  # DSP base property
```
irinaschubert, line 85:
```suggestion
            #  - isSequenceOf  # knora-base property
```

src/dsp_tools/utils/project_validate.py
```diff
@@ -13,6 +13,102 @@
...
+            #  - other:prop    # other onto
+            #  - same:prop     # same onto
+            #  - :prop         # same onto (short form)
+            
+            # filter out DSP base properties
```
irinaschubert, line 90:
```suggestion
            # filter out knora-base properties
```

src/dsp_tools/utils/project_validate.py
```diff
@@ -13,6 +13,102 @@
 from dsp_tools.utils.excel_to_json_lists import expand_lists_from_excel
 
 
+def  check_for_undefined_superproperty(project_definition: dict[str, Any]) -> bool:
+    """
+    Check if the properties referenced in "super" of each property actually point to a property 
+    (DSP base properties and properties from other ontologies are not considered.)
```
irinaschubert, line 19:
```suggestion
    (knora-base properties and properties from other ontologies are not considered.)
```

src/dsp_tools/utils/project_validate.py
```diff
@@ -13,6 +13,102 @@
...
+        True if the superproperties are valid
+    """
+    errors: dict[str, list[str]] = dict()
+    for onto in project_definition["project"]["ontologies"]:
+        ontoname = onto["name"]
+        propnames = [p["name"] for p in onto["properties"]]
+        for prop in onto["properties"]:
+            supers = prop["super"]
+            # form of supers:
+            #  - isSequenceOf  # DSP base property
```
irinaschubert, line 37:
```suggestion
            #  - isSequenceOf  # knora-base property
```

src/dsp_tools/utils/project_validate.py
```diff
@@ -13,6 +13,102 @@
...
+            #  - other:prop    # other onto
+            #  - same:prop     # same onto
+            #  - :prop         # same onto (short form)
+            
+            # filter out DSP base properties
```
irinaschubert, line 42:
```suggestion
            # filter out knora-base properties
```

test/unittests/test_create_project.py
```diff
@@ -72,5 +72,27 @@ def test_circular_references_in_onto(self) -> None:
         self.assertListEqual(sorted(errors), sorted(expected_errors))
 
 
+    def test_check_for_undefined_cardinalities(self) -> None:
```
irinaschubert, line 75: As far as I understand, you're only testing the negative case. It would be good to also test the positive case where the method returns `True`.

irinaschubert (approved):
LGTM

src/dsp_tools/utils/project_validate.py
```diff
@@ -13,6 +13,102 @@
...
+            cardnames = [regex.sub(ontoname, "", card) for card in cardnames]
+            
+            invalid_cardnames = [card for card in cardnames if regex.sub(":", "", card) not in propnames]
+            if invalid_cardnames:
+                errors[f"Ontology '{ontoname}', resource '{res['name']}'"] = invalid_cardnames
+    
+    if errors:
+        err_msg = "Your data model contains cardinalities with invalid propnames:"
+        for loc, invalids in errors.items():
+            err_msg = err_msg + f"\n - {loc}: {invalids}"
```
BalduinLandolt, line 106:
```suggestion
        err_msg += "\n".join(f"\n - {loc}: {invalids}" for loc, invalids in errors.items())
```
I think something like this would work too (and probably be more efficient)

src/dsp_tools/utils/project_validate.py
```diff
@@ -13,6 +13,102 @@
...
+        raise BaseError(err_msg)
+    else:
+        return True
```
BalduinLandolt, line 109:
```suggestion
    return True
```
if you check for an early exit condition, you don't need an else

BalduinLandolt (approved):
LGTM, I only remarked a bit of cosmetics

src/dsp_tools/utils/project_validate.py
jnussbaum: They are called like this
 - here: https://docs.dasch.swiss/2023.02.01/DSP-TOOLS/file-formats/json-project/caveats/#dsp-base-resources-and-base-properties-to-be-used-directly-in-the-xml-file
 - and here: https://docs.dasch.swiss/2023.02.01/DSP-TOOLS/file-formats/xml-data-file/#dsp-base-resources-and-base-properties-to-be-used-directly-in-the-xml-file

I think it's user friendlier to say DSP base property. Otherwise, I would have to rename the headings in the docs.

test/unittests/test_create_project.py
jnussbaum: thanks for the hint, I have now added such a case

src/dsp_tools/utils/project_validate.py
jnussbaum: Wow, much more elegant. I like that!

src/dsp_tools/utils/project_validate.py
jnussbaum: 👍

PR #307, opened 2023-02-24
chore: new error handling strategy (DEV-1773)

jnussbaum:
resolves DEV-1773


src/dsp_tools/dsp_tools.py
```diff
@@ -166,77 +166,82 @@ def program(user_args: list[str]) -> None:
...
+                              enforce_docker_system_prune=args.prune,
+                              suppress_docker_system_prune=args.no_prune)
     elif args.action == "stop-stack":
-        stop_stack()
+        success = stop_stack()
+    else:
+        success = False
 
+    if not success: 
+        exit(1)
```
jnussbaum, line 244: At a later point of time, I'd like to test if the CLI commands of dsp_tools.py can be executed. But for this, it's necessary to distinguish between exit code 0 and 1.

src/dsp_tools/utils/excel_to_json_lists.py
```diff
@@ -24,7 +24,7 @@
 
 def expand_lists_from_excel(
     lists_section: list[dict[str, Union[str, dict[str, Any]]]]
-) -> Tuple[list[dict[str, Any]], bool]:
+) -> list[dict[str, Any]]:
```
jnussbaum, line 27: having a success status is most important in the methods called by dsp_tools.py. Here, in the deeper levels, it's enough that the methods runs through without errors, and then it's clear that it ran through successfully.

src/dsp_tools/dsp_tools.py
```diff
@@ -166,77 +166,82 @@ def program(user_args: list[str]) -> None:
...
     elif args.action == "excel2json":
-        excel2json(data_model_files=args.data_model_files,
-                   path_to_output_file=args.outfile)
+        success = excel2json(data_model_files=args.data_model_files,
+                             path_to_output_file=args.outfile)
     elif args.action == "excel2lists":
-        excel2lists(excelfolder=args.excelfolder,
-                    path_to_output_file=args.outfile,
-                    verbose=args.verbose)
+        _, success = excel2lists(excelfolder=args.excelfolder,
```
BalduinLandolt, line 216: is there a reason for the first return value, if you discard it anyways?

src/dsp_tools/utils/project_create.py
```diff
@@ -489,6 +490,10 @@ def _create_ontologies(
         project_remote: representation of the project on the DSP server
         verbose: verbose switch
 
+    Raises:
+        BaseError if an error occurs during the creation of an ontology. 
+        All other errors are printed, the process continues, but the success status will be false.
```
BalduinLandolt, line 495: This raises a question for me: Do all errors necessarily mean that the overall run was not successful? E.g. if the project already exists, but not the ontology, wouldn't there be an error because the project could not be created; but that doesn't have to mean that it was an unsuccessful run, if the user wanted to create the ontology and knows that the project already exists. What do you think?

BalduinLandolt (approved):
LGTM, just some minor comments

src/dsp_tools/dsp_tools.py
jnussbaum: Yes, the method is used elsewhere, where the first return value is used

src/dsp_tools/utils/excel_to_json_lists.py
```diff
@@ -34,20 +34,21 @@ def expand_lists_from_excel(
...
-    overall_success = True
     new_lists = []
     for _list in lists_section:
         if "folder" not in _list["nodes"]:
             # this list is a JSON list: return it as it is
             new_lists.append(_list)
         else:
             # this is a reference to a folder with Excel files
-            foldername = _list["nodes"]["folder"]
+            foldername = _list["nodes"]["folder"]   # type: ignore
```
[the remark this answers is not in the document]
jnussbaum: Thanks for the idea, but unfortunately the error lies on a deeper level.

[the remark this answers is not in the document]
BalduinLandolt: too bad... maybe it would be interesting to model the whole domain entities as objects (dataclass, pydantic, ...) instead of using dictionaries everywhere. That would be a lot of work, but I think it would improve both development experience as well as code quality

src/dsp_tools/utils/shared.py
```diff
@@ -162,11 +172,11 @@ def prepare_dataframe(df: pd.DataFrame, required_columns: list[str], location_of
     # delete rows that don't have the required columns
     for req in required_columns:
         if req not in new_df:
-            raise ValueError(f"{location_of_sheet} requires a column named '{req}'")
+            raise BaseError(f"{location_of_sheet} requires a column named '{req}'")
```
irinaschubert, line 175: Is the idea to only use BaseError in the future? Or do you plan to have other Errors as well?

irinaschubert (approved):
LGTM, just some minor remarks

src/dsp_tools/dsp_tools.py
```diff
@@ -166,77 +166,82 @@ def program(user_args: list[str]) -> None:
...
-                validate_lists_section_with_schema(path_to_json_project_file=args.projectfile)
+                success = validate_lists_section_with_schema(path_to_json_project_file=args.projectfile)
                 print("'Lists' section of the JSON project file is syntactically correct and passed validation.")
             else:
-                create_lists(project_file_as_path_or_parsed=args.projectfile,
-                             server=args.server,
-                             user=args.user,
-                             password=args.password,
-                             dump=args.dump)
+                _, success = create_lists(project_file_as_path_or_parsed=args.projectfile,
```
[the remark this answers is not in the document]
irinaschubert: I think you've answered this already further down. So I resolve the issue.

src/dsp_tools/utils/project_create.py
jnussbaum: From testing perspective, it would count as unsuccessful.
And from user perpective: At the end, the following warning is printed:
```
f"WARNING: The project '{project_definition['project']['shortname']}' ({project_definition['project']['shortcode']}) "
              f"with its ontologies could be created, but during the creation process, some problems occurred. "
              f"Please carefully check the console output.")
```

The warning informs the user that some unexpexted things occured, and asks him to check the console output. I think that's a good compromise: We never know what the user expects, so I think that a gentle warning is a sound compromise.

Or should I reformulate the warning? Perhaps "unexpected things occurred" instead of "some problems occurred"?

src/dsp_tools/excel2xml.py
```diff
@@ -1737,8 +1783,11 @@ def excel2xml(datafile: str, shortcode: str, default_ontology: str) -> None:
         shortcode: shortcode of the project that this data belongs to
         default_ontology: name of the ontology that this data belongs to
 
+    Raises:
+        BaseError if something went wrong
+    
     Returns:
-        None
+        True if everything went well, False otherwise
```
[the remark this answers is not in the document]
jnussbaum: thank you for spotting that. Now, there is a case when it's false

src/dsp_tools/utils/excel_to_json_lists.py
```diff
@@ -59,12 +60,11 @@ def expand_lists_from_excel(
                 print(f"\tThe list '{_list['name']}' contains a reference to the folder '{foldername}'. The Excel "
                       f"files therein have been temporarily expanded into the 'lists' section of your project.")
             except BaseError as err:
-                print(f"\tWARNING: The list '{_list['name']}' contains a reference to the folder '{foldername}', but a "
+                raise BaseError(f"\tWARNING: The list '{_list['name']}' contains a reference to the folder '{foldername}', but a "
                       f"problem occurred while trying to expand the Excel files therein into the 'lists' section of "
-                      f"your project: {err.message}")
-                overall_success = False
+                      f"your project: {err.message}") from None
```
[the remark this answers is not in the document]
jnussbaum: That's an interesting detail that I learned only recently: If I catch an error, and as part of the error handling, I raise a new error, then the default of Python is to print both stack traces: First the old one that is caught already, then "During handling of this exception, another exception occurred" (or something like that), and then the second stack trace. It's described here: https://stackoverflow.com/questions/52725278/#comment103220039_52725410

But that's not what I want in most cases. That's why I made a PR some time ago where I added the "from None" everywhere: https://linear.app/dasch/issue/DEV-1766

src/dsp_tools/utils/project_create.py
BalduinLandolt: From a user perspective, this is absolutely ok, with this warning. But in terms of exit code (which may be relevant, if users have a sequence of instructions, which will not continue, if one of the commands has a non-zero exit code), it is a bit strange that when wanting to upload an ontology to an already existing project, I have to expect an error. It's probably not a big deal, but I still don't think it's ideal.

PR #311, opened 2023-03-01
chore: allow input of xmlupload to be parsed (DEV-1780)

jnussbaum:
resolves DEV-1780

### DIFF
--- a/src/dsp_tools/dsp_tools.py
+++ b/src/dsp_tools/dsp_tools.py
@@ -200,7 +200,7 @@ def program(user_args: list[str]) -> None:
         if args.validate:
             success = validate_xml_against_schema(xml_file_as_path_or_parsed=args.xmlfile)
         else:
-            success = xml_upload(xml_file_as_path_or_parsed=args.xmlfile,
+            success = xml_upload(input_file=args.xmlfile,
                                  server=args.server,
                                  user=args.user,
                                  password=args.password,
--- a/src/dsp_tools/dsp_tools.py
+++ b/src/dsp_tools/dsp_tools.py
@@ -198,7 +198,7 @@ def program(user_args: list[str]) -> None:
                               verbose=args.verbose)
     elif args.action == "xmlupload":
         if args.validate:
-            success = validate_xml_against_schema(xml_file_as_path_or_parsed=args.xmlfile)
+            success = validate_xml_against_schema(input_file=args.xmlfile)
         else:
             success = xml_upload(input_file=args.xmlfile,
                                  server=args.server,
--- a/src/dsp_tools/dsp_tools.py
+++ b/src/dsp_tools/dsp_tools.py
@@ -198,9 +198,9 @@ def program(user_args: list[str]) -> None:
                               verbose=args.verbose)
     elif args.action == "xmlupload":
         if args.validate:
-            success = validate_xml_against_schema(input_file=args.xmlfile)
+            success = validate_xml_against_schema(xml_file_as_path_or_parsed=args.xmlfile)
         else:
-            success = xml_upload(input_file=args.xmlfile,
+            success = xml_upload(xml_file_as_path_or_parsed=args.xmlfile,
                                  server=args.server,
                                  user=args.user,
                                  password=args.password,
--- a/src/dsp_tools/utils/shared.py
+++ b/src/dsp_tools/utils/shared.py
@@ -99,7 +99,7 @@ def try_network_action(
     raise BaseError(failure_msg)
 
 
-def validate_xml_against_schema(xml_file_as_path_or_parsed: Union[str, etree.ElementTree]) -> bool:  # type: ignore
+def validate_xml_against_schema(xml_file_as_path_or_parsed: Union[str, etree._ElementTree[Any]]) -> bool:
     """
     Validates an XML file against the DSP XSD schema.
 
--- a/src/dsp_tools/utils/shared.py
+++ b/src/dsp_tools/utils/shared.py
@@ -122,13 +122,20 @@ def validate_xml_against_schema(xml_file_as_path_or_parsed: Union[str, etree.Ele
     else:
         doc = xml_file_as_path_or_parsed
 
-    # remove namespaces
+    if not xmlschema.validate(doc):
+        error_msg = "The XML file cannot be uploaded due to the following validation error(s):"
+        for error in xmlschema.error_log:
+            error_msg = error_msg + f"\n  Line {error.line}: {error.message}"
+        raise BaseError(error_msg)
+    
+    # make sure there are no XML tags in simple texts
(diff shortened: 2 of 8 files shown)


src/dsp_tools/utils/xml_upload.py
```diff
@@ -202,23 +204,26 @@ def _convert_ark_v0_to_resource_iri(ark: str) -> str:
...
-def _parse_xml_file(input_file: str) -> etree.ElementTree:
+def _parse_xml_file(input_file: Union[str, etree.ElementTree]) -> etree.ElementTree:  # type: ignore
     """
-    Parse an XML file with DSP-conform data, remove namespace URI from the elements' names, and transform the special
-    tags <annotation>, <region>, and <link> to their technically correct form <resource restype="Annotation">,
-    <resource restype="Region">, and <resource restype="LinkObj">.
+    Parse an XML file with DSP-conform data, 
+    remove namespace URI from the elements' names, 
+    and transform the special tags <annotation>, <region>, and <link> 
+    to their technically correct form
```
irinaschubert, line 212:
```suggestion
    and transform the special tags <annotation>, <region>, and <link> to their technically correct form 
```

src/dsp_tools/utils/xml_upload.py
```diff
@@ -320,12 +334,14 @@ def xml_upload(input_file: str, server: str, user: str, password: str, imgdir: s
         uploaded because there is an error in it
     """
 
+    # Validate the input XML file
+    validate_xml_against_schema(xml_file_as_path_or_parsed=xml_file_as_path_or_parsed)
```
irinaschubert, line 338:
```suggestion
    validate_xml_against_schema(xml_file_as_path_or_parsed)
```

irinaschubert (approved):
LGTM

src/dsp_tools/dsp_tools.py
```diff
@@ -198,9 +198,9 @@ def program(user_args: list[str]) -> None:
                               verbose=args.verbose)
     elif args.action == "xmlupload":
         if args.validate:
-            success = validate_xml_against_schema(input_file=args.xmlfile)
+            success = validate_xml_against_schema(xml_file_as_path_or_parsed=args.xmlfile)
         else:
-            success = xml_upload(input_file=args.xmlfile,
+            success = xml_upload(xml_file_as_path_or_parsed=args.xmlfile,
```
BalduinLandolt, line 203: Maybe I'm lacking context here, as I'm not involved in the mass upload project. But to me, "parsed" is not self explanatory, so the renaming of those variables doesn't add any clarity for me, but makes it a lot harder to read.  
And in terms of the code context here: this is a function of the highest abstraction level, called from the program's entry point. For that reason I don't think that the parameter name here should have to reflect the format of the input. So IMO this renaming of the parameter is not an improvement.

src/dsp_tools/utils/shared.py
```diff
@@ -98,12 +99,12 @@ def try_network_action(
     raise BaseError(failure_msg)
 
 
-def validate_xml_against_schema(input_file: str) -> bool:
+def validate_xml_against_schema(xml_file_as_path_or_parsed: Union[str, etree.ElementTree]) -> bool:  # type: ignore
```
BalduinLandolt, line 102: why does this need the type ignore? Because of the ElementTree? If I recall correctly, the actual type returned by `etree.ElementTree` is `etree._ElementTree` or something like that. It is definitely possible to use types correctly with lxml, and if type hints are not used correctly, they are better not used at all, otherwise it's only confusing for programmers, users and tools.

And content-wise: Does it really make sense to introduce this change? It violates the principle of single responsibility and makes both the function signature and the function body more complex. I think it would be much cleaner to have two different functions to take parsed xml or the path, and the path-function would read the file, parse it and then call the other function.

src/dsp_tools/utils/xml_upload.py
```diff
@@ -202,23 +204,26 @@ def _convert_ark_v0_to_resource_iri(ark: str) -> str:
     return "http://rdfh.ch/" + project_id + "/" + dsp_uuid
 
 
-def _parse_xml_file(input_file: str) -> etree.ElementTree:
+def _parse_xml_file(input_file: Union[str, etree.ElementTree]) -> etree.ElementTree:  # type: ignore
```
BalduinLandolt, line 207: same as above

BalduinLandolt (commented):
I don't like the changes introduced in this PR: It adds complexity to functions that are already too long in my opinion. I understand that it is in the best interest of the project, to get this merged as soon as possible, so I don't oppose it. But I think it's introducing complexity and coupling, so it's adding technical debt, and I would really encourage to modularize and decouple the code, rather than adding content on top of the existing structures.

src/dsp_tools/dsp_tools.py
jnussbaum: Thanks for pointing this out. I reverted this change

src/dsp_tools/utils/xml_upload.py
jnussbaum: I have a general question: I thought it would be better style to always use keyword arguments if possible ("explicit is better than implicit"). Would you say that this is not a good rule?

src/dsp_tools/utils/shared.py
jnussbaum: I always struggle with type annotations of lxml, even though I have types-lxml installed. types-lxml doesn't even seem to help me. I struggle that much that I seriously consider switching to the stdlib https://docs.python.org/3/library/xml.html. 

`etree._ElementTree` doesn't work: `Missing type parameters for generic type "_ElementTree"  [type-arg]mypy(error)`

So I tried `etree._ElementTree[Any]`, and this seems to work. At least, the tests run without problem, so it's probably a good solution, even though I'm a bit sceptic about the `[Any]`

src/dsp_tools/utils/xml_upload.py
irinaschubert: In this case I would say that explicity doesn't add any value but complexity instead. It think it is becuase of the long parameter name. As the parameter name and the parameter value are called exactly the same, it can as well be left out. What do you think?

jnussbaum: I have shortened the parameter name (per recommendation of Balduin), but left in place the keyword. Now it looks fine :-) Thanks for your thoughts.

src/dsp_tools/utils/shared.py
BalduinLandolt: Have you tried `etree._ElementTree[etree._Element]`?  
(and maybe for the sanity of your eyes, define a type alias for that...)

And yes, if lxml is hampering you, it would make sense to switch. I'm not sure though, if you can switch entirely - I don't think the standard library includes schema validation. You'd have to check.

PR #313, opened 2023-03-02
fix(excel2xml): make_xsd_id_compatible() does now replace all special characters (DEV-1792)

jnussbaum:
resolves DEV-1792


src/dsp_tools/excel2xml.py
```diff
@@ -57,7 +57,7 @@ def make_xsd_id_compatible(string: str) -> str:
     res = f"{res}_{_uuid}"
 
     # replace all illegal characters by underscore
-    res = re.sub(r"[^\d\w_\-.]", "_", res)
+    res = re.sub(r"[^A-Za-z0-9_\-.]", "_", res)
```
jnussbaum, line 60: I don't know what the semantic difference between the old and the new regex are. Apparently, there must be a difference... 🤷‍♂️

BalduinLandolt: From the [python docs](https://docs.python.org/3/library/re.html):
> \w
> For Unicode (str) patterns:
> Matches Unicode word characters; this includes alphanumeric characters (as defined by [str.isalnum()](https://docs.python.org/3/library/stdtypes.html#str.isalnum)) as well as the underscore (_). If the [ASCII](https://docs.python.org/3/library/re.html#re.ASCII) flag is used, only [a-zA-Z0-9_] is matched.

So you'd need the ASCII flag, but then the digits and underscore part of the old regex would even be redundant. (And maybe we should just stick to the definition of `\W`, so not allow hyphen or dot, then it would be a very simple regex.)

BalduinLandolt (approved):
Maybe you can try the ASCII regex flag? If that doesn't work (easily), fair enough, but it would be a bit more simple, I suppose.

irinaschubert (approved):
I agree with Balduin's comment. Apart from that LGTM.

PR #320, opened 2023-03-09
docs: use single quotes around passwords to prevent problems with special characters (DEV-1825)

jnussbaum:
resolves DEV-1825


BalduinLandolt (approved):
I assume you ensured that argparse is smart enough to strip those away, so that they don't get included?

jnussbaum:
> I assume you ensured that argparse is smart enough to strip those away, so that they don't get included?

Yes. Since a long time, I always have to use single quotes around my passwords, because they contain special characters.

PR #322, opened 2023-03-09
chore: merge dublette class into one class (DEV-1831)

jnussbaum:
resolves DEV-1831


jnussbaum:
For me it is suspect that there are two classes with the same name. I think it's cleaner to have only one of these. I must admit though that I don't understand all implications that this PR might have.

PR #325, opened 2023-03-13
chore: refactor try network action (DEV-1844)

jnussbaum:
resolves DEV-1844

### DIFF
--- a/src/dsp_tools/utils/project_create.py
+++ b/src/dsp_tools/utils/project_create.py
@@ -5,7 +5,7 @@ from typing import Any, Union, cast
 
 from dsp_tools.models.connection import Connection
 from dsp_tools.models.group import Group
-from dsp_tools.models.exceptions import BaseError
+from dsp_tools.models.exceptions import BaseError, UserError
 from dsp_tools.models.helpers import Cardinality, Context, DateTimeStamp
 from dsp_tools.models.langstring import LangString
 from dsp_tools.models.ontology import Ontology
@@ -34,7 +34,7 @@ def _create_project_on_server(
         verbose: verbose switch
 
     Raises:
-        BaseError: if the project cannot be created on the DSP server
+        UserError: if the project cannot be created on the DSP server
 
     Returns:
         a tuple of the remote project and the success status (True if everything went smoothly, False otherwise)
@@ -42,36 +42,36 @@ def _create_project_on_server(
     try:
         # the normal, expected case is that this try block fails
         project_local = Project(con=con, shortcode=project_definition["project"]["shortcode"])
-        project_remote: Project = try_network_action(
-            action=lambda: project_local.read(),
-            failure_msg=""
-        )
+        project_remote: Project = try_network_action(lambda: project_local.read())
         print(f"\tWARNING: Project '{project_remote.shortname}' ({project_remote.shortcode}) already exists on the DSP "
               f"server. Updating it...")
-        success = False
         # try to update the basic info
         project_remote, _ = _update_basic_info_of_project(project=project_remote, project_definition=project_definition, verbose=verbose)
-        # It doesn't matter if the update is successful or not: continue anyway, because success is already false. 
+        # It doesn't matter if the update is successful or not: continue anyway, because success is anyways false. 
         # There are other things from this file that can be created on the server, e.g. the groups and users, so the process must continue.
+        return project_remote, False
     except BaseError:
-        success = True
-        project_local = Project(
-            con=con,
-            shortcode=project_definition["project"]["shortcode"],
-            shortname=project_definition["project"]["shortname"],
-            longname=project_definition["project"]["longname"],
-            description=LangString(project_definition["project"].get("descriptions")),
-            keywords=set(project_definition["project"].get("keywords")),
-            selfjoin=False,
-            status=True
-        )
-        project_remote = try_network_action(
-            action=lambda: project_local.create(),
-            failure_msg=f"ERROR: Cannot create project '{project_definition['project']['shortname']}' "
-                        f"({project_definition['project']['shortcode']}) on DSP server."
-        )
-        print(f"\tCreated project '{project_remote.shortname}' ({project_remote.shortcode}).")
-
+        pass
+    
+    success = True
+    project_local = Project(
+        con=con,
+        shortcode=project_definition["project"]["shortcode"],
+        shortname=project_definition["project"]["shortname"],
+        longname=project_definition["project"]["longname"],
+        description=LangString(project_definition["project"].get("descriptions")),
+        keywords=set(project_definition["project"].get("keywords")),
+        selfjoin=False,
+        status=True
+    )
+    try:
+        project_remote = try_network_action(lambda: project_local.create())
+    except BaseError:
+        raise UserError(
+            f"ERROR: Cannot create project '{project_definition['project']['shortname']}' "
+            f"({project_definition['project']['shortcode']}) on DSP server."
+        ) from None
+    print(f"\tCreated project '{project_remote.shortname}' ({project_remote.shortcode}).")
     return project_remote, success
 
 
@@ -96,16 +96,12 @@ def _update_basic_info_of_project(
     project.description = project_definition["project"].get("descriptions")
     project.keywords = project_definition["project"].get("keywords")
     try:
-        project_remote: Project = try_network_action(
-            action=lambda: project.update(),
-            failure_msg=f"WARNING: Could not update project '{project_definition['project']['shortname']}' "
-                        f"({project_definition['project']['shortcode']})."
-        )
+        project_remote: Project = try_network_action(lambda: project.update())
         if verbose:
             print(f"\tUpdated project '{project_definition['project']['shortname']}' ({project_definition['project']['shortcode']}).")
         return project_remote, True
-    except BaseError as err:
-        print(err.message)
+    except BaseError:
+        print(f"WARNING: Could not update project '{project_definition['project']['shortname']}' ({project_definition['project']['shortcode']}).")
         return project, False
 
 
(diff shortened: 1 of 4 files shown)


src/dsp_tools/utils/xml_upload.py
```diff
@@ -251,7 +251,7 @@ def _stash_circular_references(
                                 stashed_resptr_props[res][link_prop].append(str(value.value))
                         link_prop.values.remove(value)
             else:
-                logger.exception("ERROR in remove_circular_references(): link_prop.valtype is neither text nor resptr.")
+                logger.error("ERROR in remove_circular_references(): link_prop.valtype is neither text nor resptr.")
```
jnussbaum, line 254: logging.exception() is a shorthand for logging.error(exc_info=True), which should only be called inside an except block.

src/dsp_tools/utils/xml_upload.py
```diff
@@ -475,10 +475,12 @@ def xml_upload(
...
-    con = Connection(server)
-    try_network_action(failure_msg="Unable to login to DSP server", action=lambda: con.login(user, password))
-    proj_context = try_network_action(failure_msg="Unable to retrieve project context from DSP server",
-                                      action=lambda: ProjectContext(con=con))
+    con = login(server=server, user=user, password=password)
+    try:
+        proj_context = try_network_action(lambda: ProjectContext(con=con))
+    except BaseError:
+        logger.error("Unable to retrieve project context from DSP server", exc_info=True)
+        raise UserError("Unable to retrieve project context from DSP server") from None
```
jnussbaum, line 483: Until now, the error "Unable to retrieve project context from DSP server" escalated uncontrolled. Now, it is logged incl. stacktrace, and a user-friendly UserError is raised instead

src/dsp_tools/utils/xml_upload.py
```diff
@@ -475,10 +475,12 @@ def xml_upload(
     preparation_start = datetime.now()
 
     # Connect to the DaSCH Service Platform API and get the project context
-    con = Connection(server)
-    try_network_action(failure_msg="Unable to login to DSP server", action=lambda: con.login(user, password))
-    proj_context = try_network_action(failure_msg="Unable to retrieve project context from DSP server",
-                                      action=lambda: ProjectContext(con=con))
+    con = login(server=server, user=user, password=password)
```
jnussbaum, line 478: shared.login() is a convenience method that creates a Connection object and makes the login

jnussbaum:
@BalduinLandolt One thing that could be problematic is the manner how multiple-line log messages are written. It looks like this:
```
2023-03-21 13:47:12,248   shared.py            ERROR    The XML file contains the following syntax error: Opening and ending tag mismatch: xml line 69 and text, line 69, column 109
  File "/Users/nussbaum/Desktop/dsp-tools/src/dsp_tools/utils/shared.py", line 115, in validate_xml_against_schema
    doc = etree.parse(source=input_file)
  File "src/lxml/etree.pyx", line 3541, in lxml.etree.parse
  File "src/lxml/parser.pxi", line 1879, in lxml.etree._parseDocument
  File "src/lxml/parser.pxi", line 1905, in lxml.etree._parseDocumentFromURL
  File "src/lxml/parser.pxi", line 1808, in lxml.etree._parseDocFromFile
  File "src/lxml/parser.pxi", line 1180, in lxml.etree._BaseParser._parseDocFromFile
  File "src/lxml/parser.pxi", line 618, in lxml.etree._ParserContext._handleParseResultDoc
  File "src/lxml/parser.pxi", line 728, in lxml.etree._handleParseResult
  File "src/lxml/parser.pxi", line 657, in lxml.etree._raiseParseError
  File "testdata/xml-data/test-data-systematic.xml", line 69
lxml.etree.XMLSyntaxError: Opening and ending tag mismatch: xml line 69 and text, line 69, column 109
2023-03-21 13:47:41,226   xml_upload.py        INFO     Method call xml_upload(input_file='testdata/xml-data/test-data-systematic.xml', server='http://0.0.0.0:3333', user='root@example.com', imgdir='.', sipi='http://0.0.0.0:1024', verbose=False, incremental=False, save_metrics=False)
2023-03-21 13:47:46,894   shared.py            ERROR    XML-tags are not allowed in text properties with encoding=utf8. The following resources of your XML file violate this rule:
 - line 69: resource 'INV39A4413ADDA3452AAA8A1AC58B88D3D4', property ':hasId'
NoneType: None
```

By accident, I saw this today, and liked it much more than my solution:
```
2023-03-22 17:36:42 [INFO]   ---------------------------------------------
2023-03-22 17:36:42 [INFO]   --- GitHub Actions Multi Language Linter ----
2023-03-22 17:36:42 [INFO]    - Image Creation Date:[2023-01-04T06:59:26Z]
2023-03-22 17:36:42 [INFO]    - Image Revision:[985ef206aaca4d560cb9ee2af2b42ba44adc1d55]
2023-03-22 17:36:42 [INFO]    - Image Version:[985ef206aaca4d560cb9ee2af2b42ba44adc1d55]
2023-03-22 17:36:42 [INFO]   ---------------------------------------------
2023-03-22 17:36:42 [INFO]   ---------------------------------------------
2023-03-22 17:36:42 [INFO]   The Super-Linter source code can be found at:
2023-03-22 17:36:42 [INFO]    - https://github.com/github/super-linter
2023-03-22 17:36:42 [INFO]   ---------------------------------------------
```

I'm motivated to look for a solution how multi-line messages can be split up, and every line logged separately, so that the layout is more beautiful. What do you think about that?

src/dsp_tools/utils/project_create.py
```diff
@@ -35,51 +39,51 @@ def _create_project_on_server(
...
         # the normal, expected case is that this try block fails
         project_local = Project(con=con, shortcode=project_definition["project"]["shortcode"])
-        project_remote: Project = try_network_action(
-            action=lambda: project_local.read(),
-            failure_msg=""
-        )
-        print(f"\tWARNING: Project '{project_remote.shortname}' ({project_remote.shortcode}) already exists on the DSP "
-              f"server. Updating it...")
-        success = False
+        project_remote: Project = try_network_action(lambda: project_local.read())
```
BalduinLandolt, line 50:
```suggestion
        project_remote: Project = try_network_action(project_local.read)
```
This is beside the topic, but just because it caught my eye: I'm not 100% confident, but I think you don't need the lambda here, if you just pass the method as a function

src/dsp_tools/utils/project_create.py
```diff
@@ -35,51 +39,51 @@ def _create_project_on_server(
...
+        print(f"\tWARNING: Project '{project_remote.shortname}' ({project_remote.shortcode}) already exists on the DSP server. Updating it...")
+        logger.warning(f"Project '{project_remote.shortname}' ({project_remote.shortcode}) already exists on the DSP server. Updating it...")
```
BalduinLandolt, line 52:
```suggestion
        msg = f"Project '{project_remote.shortname}' ({project_remote.shortcode}) already exists on the DSP server. Updating it..."
        print(f"\tWARNING: {msg}")
        logger.warning(msg)
```

just a thought, how to make this duplication a bit nicer

src/dsp_tools/utils/project_create.py
```diff
@@ -35,51 +39,51 @@ def _create_project_on_server(
...
+        description=LangString(project_definition["project"].get("descriptions")),
+        keywords=set(project_definition["project"].get("keywords")),
+        selfjoin=False,
+        status=True
+    )
+    try:
+        project_remote = try_network_action(lambda: project_local.create())
+    except BaseError:
+        err_msg = f"Cannot create project '{project_definition['project']['shortname']}' " \
+                  f"({project_definition['project']['shortcode']}) on DSP server."
```
BalduinLandolt, line 76: I would personally put things like `{project_definition['project']['shortname']}` in a separate variable. It adds some extra lines but I think the code would still be more readable.

src/dsp_tools/utils/shared.py
```diff
@@ -24,46 +24,43 @@
...
+    and returns the active connection.
 
     Args:
         server: URL of the DSP server to connect to
         user: Username (e-mail)
         password: Password of the user
 
     Raises:
-        BaseError if the login fails
+        UserError if the login fails permanently
```
BalduinLandolt, line 37: what do you mean by "fails permanently"? I guess you want to stress that it failed despite multiple retries. But I feel just "fails" is clear enough

BalduinLandolt (approved):
Overall looking good. I remarked some minor things.  
I personally don't like the duplication of the error messages all over the place. I think I would prefer the pattern

```python
msg = "..."
print(msg)
log.error(msg)
```

Regarding how your logs look. I would definitely leave it as is! This is exactly how multi-line logs should be looking (even though I understand your dislike for the aesthetics of it).  
But when you have something like the counter example you showed, that's IMO actually bad practice, because there you have multiple log messages for one bit of information, some of them even containing only `--------------`. Imagine a service like DataDog that parses your logs for you - there you would then have to look at multiple log messages, to see what's going on.

src/dsp_tools/utils/project_create.py
jnussbaum: When I introduced this, I made experiments and I found out that the lambda is necessary. It also makes sense from a theoretical point of view: 
`try_network_action(project_local.read)` passes **the result** of `read()` to the method, which is exactly what we **don't** want.
On the other hand, `try_network_action(lambda: project_local.read())` passes the `read()` method itself to the method, and that's what we want

BalduinLandolt: no, as far as I know, the difference between  
`try_network_action(project_local.read)` and   
`try_network_action(project_local.read())`  
should be that one passes the function and the other the evaluation of the function - and the lambda is also just a function pointing to `project_local.read`.

jnussbaum: Ah, now I see, you have omitted the `()` in your first comment. Sorry, I wasn't attentive enough.

That's actually a good point. You're probably right and I could try that out. Now, I already merged, but I will keep it in mind.

PR #328, opened 2023-03-14
fix: save location is not computed correctly (DEV-1856)

jnussbaum:
resolves DEV-1856

### DIFF
--- a/src/dsp_tools/utils/xml_upload.py
+++ b/src/dsp_tools/utils/xml_upload.py
@@ -54,6 +54,18 @@ def _transform_server_to_foldername(server: str) -> str:
     return server
 
 
+def _initialize_logs(
+    server: str,
+    proj_shortcode: str,
+    onto_name: str
+) -> tuple[Path, str, str]:
+    server_as_foldername = _transform_server_to_foldername(server)
+    timestamp_str = datetime.now().strftime("%Y-%m-%d_%H%M%S")
+    save_location_full = Path.home() / Path(".dsp-tools") / "xmluploads" / server_as_foldername / proj_shortcode / onto_name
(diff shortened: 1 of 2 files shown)


src/dsp_tools/utils/xml_upload.py
```diff
@@ -32,6 +32,103 @@
...
 
+def _transform_server_url_to_foldername(server: str) -> str:
+    """
+    Take the server URL and transform it so that it can be used as foldername.
+
+    Args:
+        server: server, e.g. "https://api.test.dasch.swiss/" or "http://0.0.0.0:3333"
+
+    Returns:
+        simplified version, e.g. "test.dasch.swiss" or "localhost"
```
BalduinLandolt, line 43: I don't really like dots in folder names

src/dsp_tools/utils/xml_upload.py
```diff
@@ -32,6 +32,103 @@
...
+        r":\d{2,5}/?$": "",
+        r"/$": "",
+        r"0.0.0.0": "localhost"
+    }
+    for pattern, repl in server_substitutions.items():
+        server = re.sub(pattern, repl, server)
+    return server
+
+
+def _determine_save_location_of_logs(
```
BalduinLandolt, line 57: I think calling those "log" is actually a bit misleading, because this isn't what one would normally expect to be logs

BalduinLandolt (approved):
Maybe I'm just missing context here... but this looks like a very complicated solution for a simple problem. Anything that even remotely resembles a log, I would expect to be in a straight forward place that doesn't depend on anything of the workflow I've been doing. But I don't want to block your work, so I'd say, merging it like this is ok for now.

PR #332, opened 2023-03-14
feat: check out rosetta and upload it on local stack (DEV-1842)

jnussbaum:
resolves DEV-1842

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -66,3 +66,6 @@ DSP-TOOLS provides the following functionalities:
   The mapping has to be provided with a JSON file.
 - [`dsp-tools start-stack / stop-stack`](https://docs.dasch.swiss/latest/DSP-TOOLS/cli-commands/#start-stack)
   assist you in running a DSP stack on your local machine.
+- [`dsp-tools rosetta`](https://docs.dasch.swiss/latest/DSP-TOOLS/cli-commands/#rosetta)
+  clones the most up to data rosetta repository,
+  creates the data model and uploads the data.
--- a/src/dsp_tools/utils/rosetta.py
+++ b/src/dsp_tools/utils/rosetta.py
@@ -1,10 +1,11 @@
-from pathlib import Path
 import shutil
 import subprocess
+from pathlib import Path
 
 from dsp_tools.models.exceptions import UserError
 from dsp_tools.utils.project_create import create_project
 from dsp_tools.utils.xml_upload import xml_upload
+
 
 def upload_rosetta() -> bool:
     """
--- a/src/dsp_tools/utils/rosetta.py
+++ b/src/dsp_tools/utils/rosetta.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+import shutil
 import subprocess
 
 from dsp_tools.models.exceptions import UserError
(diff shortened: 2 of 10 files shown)


docs/index.md
```diff
@@ -66,3 +66,6 @@ DSP-TOOLS provides the following functionalities:
   The mapping has to be provided with a JSON file.
 - [`dsp-tools start-stack / stop-stack`](https://docs.dasch.swiss/latest/DSP-TOOLS/cli-commands/#start-stack)
   assist you in running a DSP stack on your local machine.
+- [`dsp-tools rosetta`](https://docs.dasch.swiss/latest/DSP-TOOLS/cli-commands/#rosetta)
```
BalduinLandolt, line 69: off topic, but because I'm seeing it right now: is there a reason for using absolute links here?

src/dsp_tools/utils/shared.py
```diff
@@ -99,7 +99,7 @@ def try_network_action(
     raise BaseError(failure_msg)
 
 
-def validate_xml_against_schema(input_file: Union[str, etree._ElementTree[Any]]) -> bool:
+def validate_xml_against_schema(input_file: Union[str, Path, etree._ElementTree[Any]]) -> bool:
```
BalduinLandolt, line 102: again, off topic: I'm not a fan of having these union types - IMO if you find yourself in these situations, it's best to have a method each, and e.g. the `Path` one should read the file, parse it, and then call the `ElementTree` one.  
To re-iterate, a method should to exactly one thing; and when you have these union types, you always do multiple things: find out, which of the types you got; transforming one thing into another; and finally doing the actual job at hand.

src/dsp_tools/utils/rosetta.py
```diff
@@ -0,0 +1,67 @@
...
+        print(f"Make 'git pull' in {rosetta_folder}...")
+        completed_process = subprocess.run("git pull", shell=True, cwd=rosetta_folder)
+        if not completed_process or completed_process.returncode != 0:
+            print(f"'git pull' failed. Remove '{rosetta_folder}'...")
+            shutil.rmtree(rosetta_folder, ignore_errors=True)
+            cloning_necessary = True
+    else:
+        cloning_necessary = True
+    
+    if cloning_necessary:
```
BalduinLandolt, line 38: the `cloning_necessary` is redundant. You can just do `if rosetta_folder.is_dir(): git-pull; else: git-clone`

src/dsp_tools/utils/rosetta.py
```diff
@@ -0,0 +1,67 @@
+from pathlib import Path
+import shutil
+import subprocess
+
+from dsp_tools.models.exceptions import UserError
+from dsp_tools.utils.project_create import create_project
+from dsp_tools.utils.xml_upload import xml_upload
+
+def upload_rosetta() -> bool:
```
BalduinLandolt, line 9: this method does multiple things... could be split up into something like: 

```python
def upload_rosetta() -> bool:
    if _require_clone():
        _clone_rosetta()
    else:
        _pull_rosetta()
    _upload_rosetta()
```

then it remains in one abstraction level (here, the highest), and can actually be read like natural language.

BalduinLandolt (approved):
Nice, thinks! I just noted some details, but looks good

docs/index.md
jnussbaum: yes. relative links wouldn't work on https://pypi.org/project/dsp-tools/

BalduinLandolt: makes sense. Doesn't that normally use the Readme? How did you override that behaviour?

src/dsp_tools/utils/rosetta.py
jnussbaum: Strictly speaking, it is not redundant: This mechanism allows that if `git pull` fails, it falls back to cloning the repo again. This adds more robustness (e.g. if the old clone is damaged). 
But from the perspective of requirements analysis, we can of course discuss if this mechanism is necessary.

docs/index.md
jnussbaum: That's defined here: https://github.com/dasch-swiss/dsp-tools/blob/main/pyproject.toml#L8

For me, the readme and the pypi landing page have two different goals:
 - readme: teach a developer how to get started if he wants to code
 - pypi: teach a user how he can use the package

That's why I take docs/index.md: For me, it has the same purpose than the pypi landing page

BalduinLandolt: yes, makes sense.

src/dsp_tools/utils/shared.py
jnussbaum: I understand your concerns, but I'm not a fan of having two slightly different methods... I wish I had good old Java method overloading at hand. I have googled method overloading in Python, and found the following possibilities:
 - https://docs.python.org/3/library/functools.html#functools.singledispatch
 - https://pypi.org/project/overloading/
 - https://pypi.org/project/multipledispatch/
 - https://pypi.org/project/multimethod/
 - https://pypi.org/project/multimethods/
 
 I have tried out the stdlib singledispatch and pushed it. How do you like it?

P.S. It doesn't work yet, because it doesn't support union types in Python 3.9. I would have to upgrade to Python 3.11 to use union types. If you think that singledispatch is generally a good idea, I would upgrade to 3.11.

P.P.S. I would use the union type for `Union[str, Path]`, because these two are very similar. While it's definitely worth writing two methods for str/Path and _ElementTree, it's NOT worth writing two methods for str and Path.

BalduinLandolt: I think the `singledispatch` is still better than the initial union implementation, but if there is a downside to tooling, then that's still a downside. And I don't see what the issue is with having two similar methods - at the end of the day, overloading is also just having similar methods in terms of the signature, they just happen to have the same name. So I would still prefer two different methods that don't impact tooling, describe in their name what they really do (`validate_xml_file()` vs. `validate_xml_tree()`), and one happens to call the other.

As for `str` vs. `Path`, I see your point, but the clean way would be to go for one (namely `Path`) and refactor the instances that call the function with `str` to use `Path` instead. Adding complexity to be able not to reduce inconsistency seems like a bad tradeoff, doesn't it? :)

PS: When you have informative method and parameter names, you don't really need docstrings either. When you have a method like this:

```python
validate_xml_file(path: Path) -> bool:
    """
    Validates an XML file against the DSP XML Schema.

    Args:
        path: the path pointing to the XML file

    Returns:
        True if the validation was successful, False otherwise
    """"
    ...
```

The docstrings do not add any value, but make the code harder to read and harder to maintain (having to change two places in code, if functionally one thing changes).

PR #333, opened 2023-03-15
feat: new command to generate template repo, move schema files (DEV-1839)

jnussbaum:
resolves DEV-1839


jnussbaum:
@BalduinLandolt If you like the idea to have a "resources" folder, I will move "docker" and "schemas" in there, too. So that in the future, we have a single folder where all resources (=non-Python files that must be distributed) are located.

src/dsp_tools/resources/4124-template-repo/template.json
BalduinLandolt, line 1: Personally, I would prefer not to have placeholders/dummies in there, but instead just empty values (`"shortcode": ""` etc.), then VS Code will actually shout at me, telling me what I have to enter, and it wouldn't happen that dummy values crawl into the DB because I forget to replace them. But this solution will work for me too, so if you think that's considerably better for some reason I'm not seeing, then I trust you on that.

src/dsp_tools/resources/4124-template-repo/template.xml
```diff
@@ -0,0 +1,28 @@
...
+        <allow group="UnknownUser">V</allow>
+        <allow group="KnownUser">V</allow>
+        <allow group="Creator">D</allow>
+        <allow group="ProjectAdmin">CR</allow>
+    </permissions>
+
+    <resource label="The only resource" restype=":minimalResource" id="res_0" permissions="res-default">
+        <text-prop name=":hasText">
+            <text encoding="utf8" permissions="prop-default">Hello!</text>
+        </text-prop>
```
BalduinLandolt, line 25: here, maybe you could have one for each possible `<xxx-prop>` type, so it's quicker to copy-paste around?

BalduinLandolt (approved):
See my comments on the details. I think the resources folder is definitely a good idea, and other stuff should be moved there too

src/dsp_tools/resources/4124-template-repo/template.xml
jnussbaum: And also in the ontology? My initial thought was to create two files that are compliant, and can be used as they are. If I add different properties, I'm inclined to add them in the onto as well...

BalduinLandolt: that relates to what you think about my other comment... that would require you to add things like `hasText`, `hasInt`, etc., which I'm not necessarily too fond of, as I would find myself most of the time just deleting them again.  
For me it doesn't have to be compliant, because if I want to upload something I don't have to modify first, I can now do `dsp-tools rosetta` :)

PR #335, opened 2023-03-16
chore: implement logging system (DEV-132)

jnussbaum:
resolves DEV-132

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "knora/dsplib/import_scripts"]
 	path = src/dsp_tools/import_scripts
-	url = https://github.com/dasch-swiss/0123-import-scripts
+	url = https://github.com/dasch-swiss/00A1-import-scripts
--- a/docs/developers/git-submodules.md
+++ b/docs/developers/git-submodules.md
@@ -2,20 +2,20 @@
 
 # Git submodules
 
-This repository embeds [https://github.com/dasch-swiss/0123-import-scripts](https://github.com/dasch-swiss/0123-import-scripts) 
+This repository embeds [https://github.com/dasch-swiss/00A1-import-scripts](https://github.com/dasch-swiss/00A1-import-scripts) 
 as a Git submodule in `src/dsp_tools/import_scripts`. That means that `src/dsp_tools/import_scripts` has no contents, but
-only a reference to a certain commit in the main branch of `0123-import-scripts`. When you clone DSP-TOOLS from GitHub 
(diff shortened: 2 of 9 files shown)


src/dsp_tools/utils/shared.py
```diff
@@ -82,20 +86,13 @@ def try_network_action(
...
-                exc_message = exc.message
-            else:
-                exc_message = str(exc).replace('\n', ' ')
-                exc_message = exc_message[:150] if len(exc_message) > 150 else exc_message
-            raise UserError(f"{failure_msg}.\nOriginal error message for diagnostic purposes:\n{exc_message}") from None
+            logger.exception(failure_msg)
+            raise UserError(f"{failure_msg}") from None
+        except Exception:
+            logger.exception(failure_msg)
+            raise UserError(f"{failure_msg}") from None
```
jnussbaum, line 93: `logger.exception()` is a handy shortcut that logs the stack trace with the original error message. So the entire handling of the error message that I had done by hand until now, is not necessary any more.

src/dsp_tools/utils/xml_upload.py
```diff
@@ -244,8 +251,8 @@ def _stash_circular_references(
                                 stashed_resptr_props[res][link_prop].append(str(value.value))
                         link_prop.values.remove(value)
             else:
-                raise BaseError(f'ERROR in remove_circular_references(): link_prop.valtype is '
-                                f'neither text nor resptr.')
+                logger.exception("ERROR in remove_circular_references(): link_prop.valtype is neither text nor resptr.")
+                raise BaseError("ERROR in remove_circular_references(): link_prop.valtype is neither text nor resptr.")
```
BalduinLandolt, line 255: is it necessary to log this here? since you raise the exception, you should be able to log it when you catch it again (and there's no point in logging the same exception twice)

BalduinLandolt (approved):
over all I like it, just one remark: From what I can see, a single error will be logged multiple times because you commonly catch an exception, log it and then raise another exception, which will be logged again. This will lead to quite cluttered logs, especially because each time, the stack trace gets logged. Maybe this could be streamlined over time, so that each error gets logged exactly once.

src/dsp_tools/utils/xml_upload.py
jnussbaum: I quickly checked it, and it seems that raising an Error with `raise BaseError()` does NOT automatically get logged. I explicitly have to log the error before raising it. So I don't think there will be a doubling of the stack traces.

Or do you think that this architecture (log error, then raise it) is suboptimal?

PR #341, opened 2023-03-20
fix(xmlupload): legitimate html-escapes in utf8-texts don't raise a validation error any more (DEV-1936)

jnussbaum:
resolves DEV-1936


BalduinLandolt (approved):
LGTM

PR #345, opened 2023-03-24
chore: JSON file: enforce resource names and property names to be unique for each ontology (DEV-1971)

jnussbaum:
resolves DEV-1971


src/dsp_tools/utils/project_validate.py
```diff
@@ -13,7 +13,56 @@
 from dsp_tools.utils.excel_to_json_lists import expand_lists_from_excel
 
 
-def check_for_undefined_super_resource(project_definition: dict[str, Any]) -> bool:
+def _check_for_duplicate_names(project_definition: dict[str, Any]) -> bool:
```
BalduinLandolt, line 16: there seems to be a lot of similar stuff going on between `excel_to_json_properties.py`, `excel_to_json_resources.py` and here. Could this be unified somehow?

BalduinLandolt: also, the method seems to only use `project_definition["project"]["ontologies"]`, could you pass in only that? As a rule of thumb, it's always best to give a function only access to the things it needs, so that only changes to that can break the function.

src/dsp_tools/utils/project_validate.py
```diff
@@ -13,7 +13,56 @@
...
+    propnames_duplicates: dict[str, set[str]] = dict()
+    for onto in project_definition["project"]["ontologies"]:
+        resnames = [r["name"] for r in onto["resources"]]
+        if len(set(resnames)) != len(resnames):
+            for elem in resnames:
+                if resnames.count(elem) > 1:
+                    if not resnames_duplicates.get(onto["name"]):
+                        resnames_duplicates[onto["name"]] = {elem,}
+                    else:
+                        resnames_duplicates[onto["name"]].add(elem)
```
BalduinLandolt, line 39: this block is repeated right below, so this should be extracted to a function `_get_duplicates(list)`

BalduinLandolt, line 39:
```suggestion
                    resnames_duplicates[onto["name"]] = resnames_duplicates.get(onto["name"], set()).add(elem)
```
this could be rewritten to something like this, by using default values on get you don't need to check if it's there or not

PR #347, opened 2023-03-24
test(xmlupload): add e2e test for invalid ontoname and invalid shortcode (DEV-1969)

jnussbaum:
resolves DEV-1969

### DIFF
--- a/test/e2e/test_xmlupload.py
+++ b/test/e2e/test_xmlupload.py
@@ -1,0 +1,46 @@
+import unittest
+
+from dsp_tools.models.exceptions import UserError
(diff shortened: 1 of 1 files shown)


irinaschubert (approved):
LGTM, just one remark: the test ist called "test_xml_upload" but actually does only test two negative cases. Shouldn't it be called "test_xml_upload_fails" or something like this?

BalduinLandolt (approved):
I second @irinaschubert's remark that it only checks negative cases, so the name doesn't really fit. (should it check the positive case?)

PR #348, opened 2023-03-28
fix: display error when server is addressed on admin subdomain (DEV-1979)

jnussbaum:
resolves DEV-1979

### DIFF
--- a/test/e2e/test_cli.py
+++ b/test/e2e/test_cli.py
@@ -1,7 +1,6 @@
 import subprocess
 import unittest
 
-from dsp_tools.models.exceptions import UserError
 
 class TestCLI(unittest.TestCase):
 
@@ -14,10 +13,12 @@ class TestCLI(unittest.TestCase):
             "http://admin.082e-test-server.dasch.swiss"
         ]
         for server in wrong_servers:
-            # TODO: doesn't work, because UserError is caught and printed. Watch stderr/stdout instead
-            with self.assertRaisesRegex(
-                UserError, 
-                r"You try to address the subdomain 'admin' of a DSP server, "
-                r"but DSP servers must be addressed on their 'api' subdomain, e\.g\. https://api\.dasch\.swiss\."
-            ):
-                subprocess.run(f"dsp-tools create -s {server} -u hans@muster.ch -p 1234 data_model.json", shell=True)
+            completed_process = subprocess.run(f"dsp-tools create -s {server} -u hans@muster.ch -p 1234 data_model.json", shell=True, capture_output=True)
+            stdout = completed_process.stdout.decode("utf-8")
+            self.assertRegex(
+                text=stdout,
+                expected_regex=
+                    r"You try to address the subdomain 'admin' of a DSP server, "
+                    r"but DSP servers must be addressed on their 'api' subdomain, e\.g\. https://api\.dasch\.swiss\."
+            )
+                


test/e2e/test_cli.py
jnussbaum, line 1: A comment concerning the names of the test files: As long as I don't have a good idea how to restructure the entire tests, I try to stick with the current naming conventions. My feeling is that the current names are thematical, i.e. the files are named according to the subject they are about. 

So I named this one "test_cli.py", even though it's not (yet) a thorough test of the entire CLI. But it is like a container that has space for more (future) tests related to CLI.

BalduinLandolt (commented):
Couldn't DSP-Tools just be forgiving, and assume `api.xxx.dasch.swiss` when it gets `admin.xxx.dasch.swiss`?

irinaschubert (approved):
LGTM

jnussbaum:
> Couldn't DSP-Tools just be forgiving, and assume `api.xxx.dasch.swiss` when it gets `admin.xxx.dasch.swiss`?

Good point. It depends in what direction we'd like to go:

1. Either user-friendliness: Don't care about subdomains at all, and just rely on the convention that admin=DSP-APP, api=DSP-API, iiif=SIPI. We could then get rid of this complicated paragraph that makes user's lives difficult: https://docs.dasch.swiss/2023.03.02/DSP-TOOLS/cli-commands/#before-starting-have-in-mind-the-subdomains-of-a-dsp-server. 
I could then entirely get rid of the ` -S https://iiif.dasch.swiss` flag, and the entire RDU team would love me for that.
(I would then close this PR and start a new one)

2. Or we tend towards purism: Conventions can change, so the subdomains must be passed to the CLI in the correct manner. (state as is, I would then merge this PR)

3. Or I find a perfect solution that queries the server to find out what subdomains it has. Some time ago, I tried to do that (https://linear.app/dasch/issue/DEV-958/), but it seems to be difficult.


My personal favorite is solution number 1. If both of you agree, I would close this PR and start over with solution 1. What do you think?

BalduinLandolt:
> My personal favorite is solution number 1. If both of you agree, I would close this PR and start over with solution 1. What do you think?

I generally like 1. too, however you would have to consider how that works locally. There we don't have subdomains but ports, which works somewhat differently.

irinaschubert:
> > Couldn't DSP-Tools just be forgiving, and assume `api.xxx.dasch.swiss` when it gets `admin.xxx.dasch.swiss`?
> 
> Good point. It depends in what direction we'd like to go:
> 
> 1. Either user-friendliness: Don't care about subdomains at all, and just rely on the convention that admin=DSP-APP, api=DSP-API, iiif=SIPI. We could then get rid of this complicated paragraph that makes user's lives difficult: https://docs.dasch.swiss/2023.03.02/DSP-TOOLS/cli-commands/#before-starting-have-in-mind-the-subdomains-of-a-dsp-server.
>    I could then entirely get rid of the ` -S https://iiif.dasch.swiss` flag, and the entire RDU team would love me for that.
>    (I would then close this PR and start a new one)
> 2. Or we tend towards purism: Conventions can change, so the subdomains must be passed to the CLI in the correct manner. (state as is, I would then merge this PR)
> 3. Or I find a perfect solution that queries the server to find out what subdomains it has. Some time ago, I tried to do that (https://linear.app/dasch/issue/DEV-958/), but it seems to be difficult.
> 
> My personal favorite is solution number 1. If both of you agree, I would close this PR and start over with solution 1. What do you think?

I also like 1. As Balduin already mentioned, you would need to make sure though that it is still possible to choose whatever server (project server, staging, prod, localhost etc.). As long as this is possible, I'm fine with 1. I also like 2. When I was working with dsp-tools it didn't bother me too much to type out the servername with `api` in front of it... But you know more about the requirements for RDU.

PR #349, opened 2023-03-29
fix(xmlupload): print reason of error if a resource cannot be created, remodel BaseError (DEV-1977)

jnussbaum:
resolves DEV-1977

### DIFF
--- a/src/dsp_tools/models/connection.py
+++ b/src/dsp_tools/models/connection.py
@@ -19,7 +19,7 @@ def check_for_api_error(response: requests.Response) -> None:
     """
     if response.status_code != 200:
         raise BaseError(
-            message="DSP-ERROR: status code=" + str(response.status_code) + "\nMessage:" + response.text,
+            message="KNORA-ERROR: status code=" + str(response.status_code) + "\nMessage:" + response.text,
             status_code_of_api_response=response.status_code,
             json_content_of_api_response=response.text,
             reason_for_failure_from_api_response=response.reason,
--- a/src/dsp_tools/models/connection.py
+++ b/src/dsp_tools/models/connection.py
@@ -100,10 +100,16 @@ class Connection:
         """
 
         if res.status_code != 200:
-            raise BaseError("KNORA-ERROR: status code=" + str(res.status_code) + "\nMessage:" + res.text)
+            raise BaseError(
+                message="DSP-ERROR: status code=" + str(res.status_code) + "\nMessage:" + res.text,
+                status_code_of_api_response=res.status_code,
+                json_content_of_api_response=res.content,
+                reason_for_failure_from_api_response=res.reason,
+                api_route=res.url
+            )
 
         if 'error' in res:
-            raise BaseError("KNORA-ERROR: API error: " + res.error)
+            raise BaseError("DSP-ERROR: API error: " + res.error)
 
     def post(self, path: str, jsondata: Optional[str] = None):
         """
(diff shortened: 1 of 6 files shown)


jnussbaum:
@BalduinLandolt @irinaschubert For more context, see the Linear ticket. Generally, I plan to write more informative Linear tickets in the future. That means for you that you can read the ticket before reviewing, and then you know the rationale.

irinaschubert (dismissed):
I like the intention of this PR very much, if the output is more informative this is certainly a nice support for the user! But I agree with all points mentioned by Balduin and have nothing more to add :)

src/dsp_tools/models/connection.py
```diff
@@ -7,6 +7,25 @@
...
+    """
+    if response.status_code != 200:
+        raise BaseError(
+            message="DSP-ERROR: status code=" + str(response.status_code) + "\nMessage:" + response.text,
+            status_code_of_api_response=response.status_code,
+            json_content_of_api_response=response.text,
+            reason_for_failure_from_api_response=response.reason,
+            api_route=response.url
+        )
+
```
jnussbaum, line 28: This code block is not new, I just moved and renamed `on_api_error()`

src/dsp_tools/models/connection.py
```diff
@@ -119,22 +126,28 @@ def post(self, path: str, jsondata: Optional[str] = None):
         if jsondata is None:
             if self._token is not None:
                 headers = {'Authorization': 'Bearer ' + self._token}
```
jnussbaum, line 128: The rest of this file, from here on, contains only refactorings: I wanted to rename the response of each request from `req` to `response`. But by doing so, I had to redo the entire indentation.

src/dsp_tools/models/sipi.py
```diff
@@ -3,25 +3,7 @@
...
-    Returns:
-        Knora Error that is being raised
-    """
-
-    if res.status_code != 200:
-        raise BaseError("SIPI-ERROR: status code=" + str(res.status_code) + "\nMessage:" + res.text)
-
-    if 'error' in res:
-        raise BaseError("SIPI-ERROR: API error: " + res.error)
+from dsp_tools.models.connection import check_for_api_error
```
jnussbaum, line 6: I thought it is not necessary that the `Sipi` class contains a duplicate of `check_for_api_error()`. It is the same logic, just that it's a SIPI error, not an API error.

src/dsp_tools/utils/shared.py
```diff
@@ -75,7 +75,7 @@ def try_network_action(action: Callable[..., Any]) -> Any:
             time.sleep(2 ** i)
             continue
         except BaseError as err:
-            if regex.search(r"try again later", err.message) or regex.search(r"status code=5\d\d", err.message):
+            if regex.search(r"try again later", err.message) or regex.search(r"5\d\d", str(err.status_code_of_api_response)):
```
jnussbaum, line 78: perhaps not the most elegant way. But if I wanted to analyze `err.status_code_of_api_response` as integer, 
 - I first have to check if it is not None, 
 - then check that it's > 500
 - then check that it's < 600

jnussbaum:
@irinaschubert @BalduinLandolt As suggested by you, I have remodeled `BaseError`. I think it was worth the effort, I very much like the idea of storing fine-grained information in the class `BaseError`.

BalduinLandolt (approved):
Definitely a step in the right direction. I added two minor comments, but over all looking good.

PR #356, opened 2023-05-03
fix(get): write proper tests for get command, fix multiple bugs (DEV-2123)

jnussbaum:
resolves DEV-2123


src/dsp_tools/models/resourceclass.py
```diff
@@ -788,16 +788,25 @@ def createDefinitionFileObj(self, context: Context, shortname: str, skiplist: li
             else:
                 superclasses = context.reduce_iri(self._superclasses[0], shortname)
             resource["super"] = superclasses
+        resource["labels"] = self._label.createDefinitionFileObj()
+        if self._comment:
+            resource["comments"] = self._comment.createDefinitionFileObj()
```
jnussbaum, line 793: lines 782-793 is just a new order of things, nothing new

test/e2e/test_tools.py
```diff
@@ -128,117 +129,132 @@ def test_create_project(self) -> None:
 
 
     def test_get_ontology(self) -> None:
-        with open(self.test_project_systematic_file) as f:
```
jnussbaum, line 131: The old testing strategy has a major disadvantage: Instead of comparing the **entire** original with the **entire** returned file, it only compares selected elements. This is because the original and the returned cannot be compared directly, there are many details where the order/syntax slightly differs.

The new testing strategy tackles this challenge by modifying both original and returned in such a way that they become directly comparable.

test/e2e/test_tools.py
```diff
@@ -128,117 +129,132 @@ def test_create_project(self) -> None:
 
 
     def test_get_ontology(self) -> None:
```
BalduinLandolt, line 131: this method does way more than testing. Could you extract the json modification into a/some helper method(s)?

test/e2e/test_tools.py
BalduinLandolt: generally, exhaustiveness in testing of course a good thing. However, with this new testing strategy, it will be really hard to see what's different, if this test fails.  
I wonder if something along the lines of the following wouldn't be nicer, both for reading the test and interpreting the test results:

```python
def test_get_ontology():
    response = get_project_from_api()
    expected = get_project_expected()
    # by checking that the dicts are of equal length, we can be sure that no additional field was returned, or none is missing
    assertTrue(compare_number_of_elements_returned(response, expected))
    # this comparator knows how the expected and returned users may be different, and takes care of a sensible comparison
    assertTrue(compare_users(response.get('users'), expected.get('users')))
    # etc.
```

This is of course only pseudo-code, but you get my meaning, I think.  
The advantage of this is firstly readability, secondly reusability (you could test the same with different projects and use the same comparators), and thirdly traceability (if the users aren't equal, you can tell, because you know which assertion failed).

PS: probably the method should also be called `test_get_project()` not `ontology`

BalduinLandolt (commented):
See my comment on the new testing strategy. I'm all for improving tests, but I think the new approach has some downsides

irinaschubert (commented):
It looks good to me, just wondering: have you tested the e2e test by providing two different projects for every comparison that you do? Just to be sure that the behaviour is as intended.

jnussbaum:
> It looks good to me, just wondering: have you tested the e2e test by providing two different projects for every comparison that you do? Just to be sure that the behaviour is as intended.

Yes, I tested it manually, to see if I get the error message that I expected.

irinaschubert (approved):

PR #357, opened 2023-05-05
chore: update mypy, resolve all mypy+pylint issues in "unittests", unignore "unittests" in CI pipline (DEV-2131)

jnussbaum:
resolves DEV-2131


test/unittests/test_excel_to_json_lists.py
```diff
@@ -140,9 +144,9 @@ def test_excel2lists(self) -> None:
 
         # make sure that the invalid lists raise an Error
         with self.assertRaisesRegex(BaseError, r"Found duplicate in column 2, row 9"):
-            e2l.excel2lists(excelfolder="testdata/invalid-testdata/excel2json/lists-invalid-1", path_to_output_file=outfile)
+            e2l.excel2lists(excelfolder="testdata/invalid-testdata/excel2json/lists-invalid-1")
         with self.assertRaisesRegex(BaseError, r"The Excel file with the language code 'de' should have a value in row 10, column 2"):
-            e2l.excel2lists(excelfolder="testdata/invalid-testdata/excel2json/lists-invalid-2", path_to_output_file=outfile)
```
jnussbaum, line 145: `path_to_output_file=outfile` raised a pylint error. After examining, I realized that the parameter `path_to_output_file` is optional, and that its presence/absence doesn't influence the outcome of the test.

src/dsp_tools/excel2xml.py
```diff
@@ -18,7 +18,12 @@
 
 from dsp_tools.models.exceptions import BaseError
 from dsp_tools.models.helpers import DateTimeStamp
-from dsp_tools.models.propertyelement import PropertyElement
+# explicitly export PropertyElement, so that API users can import it from this module 
+# (see https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport)
+# doing this requires silencing the corresponding pylint warning 
+# (see https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/useless-import-alias.html)
```
BalduinLandolt, line 24: I'm probably just ignorant of the issue at hand. But can't this be solved in the `__init__.py` file? I'm assuming no one has ever looked into making importing modules from DSP-Tools nice by adding contents to the init files.

BalduinLandolt (approved):
LGTM

src/dsp_tools/excel2xml.py
jnussbaum: The 2 links mention different approaches, but not `__init__.py`. A quick search in the internet doesn't tell me what to write into `__init__.py` in order to make a value exportable.

However, there is an other approach, which would perhaps be more elegant:
```
from dsp_tools.models.propertyelement import PropertyElement
__all__ = ['PropertyElement']
```

This wouldn't require me to silence a pylint warning. But I think that modifying a dunder variable is not elegant, too. What do you think?

BalduinLandolt: this seems misleading because `__all__` suggests "all", which wouldn't be the case then.

I'm no expert in this... but if in `src/dsp_tools/__init__.py` you add the one line

```
from dsp_tools.models.propertyelement import PropertyElement
```

can't you then use it like this?
```
from dsp_tools import PropertyElement
```

my assumption has always been that this is the exact usecase for having content in the `__init__.py` file. But again, I could be wrong.

What I don't understand: Can't users just use the actual import `from dsp_tools.models.propertyelement import PropertyElement` and be happy with it?

BalduinLandolt: and thinking about it some more: shouldn't the module structure just be more intuitive and useable, so that this is not an issue in the first place?

Probably, users would expect to have imports like this:

```python
from dsp_tools import create_project, xml_upload, excel2json
from dsp_tools.upload import upload_resource
from dsp_tools.excel2xml import PropertyElement # if that is even the place for it
# etc.
```

I mentioned this before - but now I think it becomes obvious how counter-intuitive (for users) and impractical (for maintaining a clean architecture) the base structure `dsp_tools.models` vs. `dsp_tools.utils` is.

BalduinLandolt: (but that is of course for a different PR)

BalduinLandolt: Fair enough (though any IDE helps with auto-importing). But then it seems to me all the more that `PropertyElement` shouldn't be located in `dsp_tools.models.propertyelement` but where it logically belongs, namely in a package `dsp_tools.excel2xml` (or `dsp_tools.excel`?), with all other Excel-related code. And all we're discussing here is just work-around for the core issue at hand.

I'm absolutely ok with merging this as is, but in the long term, solving causes is more sustainable than fighting symptoms.

PR #365, opened 2023-05-12
chore: bump dependencies, add docker, update readme, remove unused poetry exec script (DEV-2159)

jnussbaum:
resolves DEV-2159


pyproject.toml
```diff
@@ -61,9 +61,6 @@ dsp-tools = "dsp_tools.dsp_tools:main"
 # plugin (https://github.com/keattang/poetry-exec-plugin) to define commands available for the developers
 # e.g. `poetry exec check-links`
 check-links = "markdown-link-validator ./docs -i \\.\\/assets\\/.+ -i .+github\\.com\\/dasch\\-swiss\\/dsp-tools\\/settings"
-# `mkdocs serve` can normally be executed without custom port. This is only necessary on machines where the standard
-# port is taken already
-docs-serve = "mkdocs serve --dev-addr=localhost:7979"
```
jnussbaum, line 66: This was only necessary on my old machine. Now I can execute `mkdocs serve` like everyone else, so this script is not necessary any more.

src/dsp_tools/utils/project_validate.py
```diff
@@ -401,7 +401,7 @@ def _identify_problematic_cardinalities(
 
     # find elements of circles that have a cardinality of "1" or "1-n"
     errors: set[tuple[str, str]] = set()
-    circles = list(nx.cycles.simple_cycles(graph))
+    circles = list(nx.algorithms.cycles.simple_cycles(graph))
```
jnussbaum, line 404: apparently, networkx changed this in their new version

pyproject.toml
```diff
@@ -28,28 +28,28 @@ python = ">=3.9,<3.12"
...
-openpyxl = "^3.0.10"
-networkx = "^2.8.8"
-pandas = "^1.5.2"
-xlrd = ">=1.0.0"            # without this, pandas raises an ImportError: Missing optional dependency 'xlrd'
-regex = "^2022.10.31"
+openpyxl = "^3.1.2"
+networkx = "^3.1.0"
+pandas = { version = "^2.0.1", extras = ["excel"] }  # extra package that contains xlrd that is necessary for reading old .xls Excel files
+regex = "^2023.5.5"
+docker = "^6.1.2"
```
jnussbaum, line 37: Docker is necessary in the Fast xmlupload PR.
I added docker already now, so that I can remove all changes of `pyproject.toml` and `poetry.lock` from the Fast xmlupload PR.
This way, I won't get any merge conflicts.

BalduinLandolt: just wondering: if you have this docker dependency here - could this also be used for the docker based DSP-Tools commands like starting the stack?

BalduinLandolt (approved):

pyproject.toml
jnussbaum: After a quick look into `src/dsp_tools/utils/stack_handling.py`, I only see one possibility: There are 4 lines of code where a docker command is executed via `subprocess.run("docker do something")`. This could perhaps be replaced by the docker library.
But apart from that, I don't see a significant potential for this library.
But thank you for giving me the hint - it's always good to reflect about potential improvements :-)

PR #366, opened 2023-05-15
chore: lint e2e tests (DEV-2160)

jnussbaum:
resolves DEV-2160


src/dsp_tools/utils/excel_to_json_resources.py
```diff
@@ -1,3 +1,5 @@
+# pylint: disable=duplicate-code
+
```
BalduinLandolt, line 2: what duplicates would it complain about? could they be resolved?

src/dsp_tools/utils/rosetta.py
```diff
@@ -1,3 +1,5 @@
+# pylint: disable=duplicate-code
+
```
BalduinLandolt, line 2: as above

test/e2e/test_connection.py
```diff
@@ -35,7 +38,6 @@ def test_get(self) -> None:
         self.assertRaises(BaseError, self.con.get, "/doesNotExist")
 
     def test_put(self) -> None:
-        # TODO add test
         pass
```
BalduinLandolt, line 41: should this exist?

test/e2e/test_connection.py
```diff
@@ -86,7 +88,6 @@ def test_post(self) -> None:
         self.assertEqual(res['knora-api:result'], 'Resource erased')
 
     def test_delete(self) -> None:
-        # TODO add test
         pass
```
BalduinLandolt, line 91: as above

test/e2e/test_listnode.py
```diff
@@ -48,8 +48,8 @@ def test_ListNode_create(self) -> None:
         ).create()
         self.assertIsNotNone(node.id)
         self.assertEqual(node.project, self.project)
-        self.assertEqual(node.label['de'], "root node 1")
-        self.assertEqual(node.comments['de'], "first root node")
+        self.assertEqual(node.label['de'], "root node 1")               # type: ignore
```
BalduinLandolt, line 51: why does it report type errors on these? seems like a red flag, doesn't it?

BalduinLandolt (approved):
over all it's just linting, so I don't want to complain. But I'm a bit sceptical about putting tools in place and having to add so many ignore-comments to it. To me this seems like it's either the wrong tool or the wrong rule set, or the code isn't ready (i.e. the warnings should be resolved by fixing the code first).  
If no linting is done at all, it can easily be activated; if some rules are disabled globally, they can be re-enabled. If ignore statements are spread all over the code, it's a pain to remove them again, just to find the issues, to be able to fix them.

All in all I think it makes sense to have some rules deactivated for test code (like duplications, missing docstrings, etc.; but others, like type errors should normally not be necessary). But I think I would prefer it, if this could be done on a global config level, or at least on file basis not on line basis.

... but those are just some thoughts, nothing of importance

PR #369, opened 2023-05-15
chore: speed up CI tests with caching (DEV-2165)

jnussbaum:
resolves DEV-2165


BalduinLandolt (approved):
I'm no expert in github actions, but looks good from what I can tell

PR #380, opened 2023-05-24
chore: don't configure logging statically (DEV-2183)

jnussbaum:
resolves DEV-2183

### DIFF
--- a/src/dsp_tools/dsp_tools.py
+++ b/src/dsp_tools/dsp_tools.py
@@ -38,7 +38,7 @@ def _initiate_logger(parsed_args: argparse.Namespace) -> logging.Logger:
     Create a logger instance, 
     set its level to INFO,
     and configure it to write to a file in the user's home directory.
-    Then, a first log message is written with the arguments how the CLI was called.
+    Then, the CLI-arguments are logged as a first log message.
 
     This function should be called only once, 
     in the dsp_tools.py module, and not in any other module.
@@ -52,8 +52,8 @@ def _initiate_logger(parsed_args: argparse.Namespace) -> logging.Logger:
     Returns:
         the logger instance
     """
-    _logger = logging.getLogger(parsed_args.action)
-    _logger.setLevel(logging.INFO)
+    logger = logging.getLogger(parsed_args.action)
+    logger.setLevel(logging.INFO)
     formatter = logging.Formatter(
         fmt="{asctime} {filename: <20} {levelname: <8} {message}",
         style="{"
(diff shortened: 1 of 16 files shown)


src/dsp_tools/dsp_tools.py
```diff
@@ -269,6 +359,7 @@ def call_requested_action(
                     server=args.server,
                     user=args.user,
                     password=args.password,
+                    logger_instance=logger,
                     dump=args.dump
```
BalduinLandolt, line 363: The way I normally have seen logging, you have a logger instance per source code file (which tells you which file you're in, which is incredibly useful to get information from the logs). Is there a reason to not doing it that way and instead passing the logger instance around?

src/dsp_tools/fast_xmlupload/process_files.py
```diff
@@ -19,7 +19,10 @@
 
 from dsp_tools.models.exceptions import BaseError
 
+# default unconfigured logger created when the module is initialized (e.g. for testing, or when the module is imported)
+# if this module is used as part of the dsp-tools CLI, this default logger will be replaced by the one configured in dsp_tools.py
 logger = logging.getLogger(__name__)
```
BalduinLandolt, line 24: ok, I see this now... but I'm still not sure I like it. To me it would seem more streamlined to have a logger instance per file, and ensuring the configuration by means of a simple log util

BalduinLandolt (changes requested):
I don't like these changes. To me it seems like a step in the wrong direction; and the linear issue doesn't give me any indication of the nature of the problem that the previous behaviour caused, so I also can't really give any useful feedback.

Can you either elaborate on the problem you're solving, or see if my comments make sense to you?

src/dsp_tools/fast_xmlupload/process_files.py
jnussbaum: Is that possible? My first approach was to make one method in `shared.py`, where a logger is created and configured to log to `~/.dsp-tools`. And then every module would call that method, when it is initialized. 

But I ran into the following problem: One single call to a DSP-TOOLS CLI will initialize **ALL** modules, so the log file is opened a dozen times, so every log call is logged a dozen times...

src/dsp_tools/dsp_tools.py
```diff
@@ -228,17 +305,38 @@ def make_parser() -> argparse.ArgumentParser:
...
+        user_args: command line arguments
+
+    Returns:
+        parsed arguments
+    """
+
+    args = parser.parse_args(user_args)
+    if not hasattr(args, "action"):
+        parser.print_help(sys.stderr)
+        sys.exit(1)
```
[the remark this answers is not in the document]
jnussbaum: I quickly tried it out, and it is necessary

src/dsp_tools/fast_xmlupload/process_files.py
BalduinLandolt: [test-logging.zip](https://github.com/dasch-swiss/dsp-tools/files/11567456/test-logging.zip)

Here's a little test project I created. to me it seems to work - maybe you can try it out. Unless I'm misunderstanding, I think this would solve the problem you wanted to address, and still be to my taste. :)  
(but maybe I'm also misunderstanding)

One case in which I don't know what would happen, is if two instances of DSP-Tools are running at the same time? Would they log to the same file? Would that work without issues?  
But I suppose that's a question that applies to your solution too, right?

PR #385, opened 2023-05-26
chore: don't configure logging statically (DEV-2183)

jnussbaum:
resolves DEV-2183


BalduinLandolt (approved):
Code looks good. I would suggest keeping the log util in a separate file, though.

PR #395, opened 2023-06-02
fix(fast xmlupload): make process-files more robust (DEV-2235)

jnussbaum:
resolves DEV-2235

### DIFF
--- a/src/dsp_tools/fast_xmlupload/process_files.py
+++ b/src/dsp_tools/fast_xmlupload/process_files.py
@@ -585,6 +585,9 @@ def _process_file(
 
     # convert file (create derivative) and create sidecar file based on category (image, video or other)
     file_category = _get_file_category_from_extension(in_file)
+    if not file_category:
+        return in_file, None
+    
     if file_category == "OTHER":
         result = _process_other_file(
             in_file=in_file,
--- a/src/dsp_tools/fast_xmlupload/process_files.py
+++ b/src/dsp_tools/fast_xmlupload/process_files.py
@@ -796,6 +796,7 @@ def process_files(
         logger.error(f"An error occurred while writing the result to the pickle file. The result was: {result}")
 
     # remove the SIPI container
-    _stop_and_remove_sipi_container()
+    if success:
+        _stop_and_remove_sipi_container()
 
     return success
--- a/src/dsp_tools/fast_xmlupload/process_files.py
+++ b/src/dsp_tools/fast_xmlupload/process_files.py
@@ -102,7 +102,28 @@ def _process_files_in_parallel(
 
     orig_filepath_2_uuid: list[tuple[Path, Optional[Path]]] = []
     for processed in as_completed(processing_jobs):
-        orig_filepath_2_uuid.append(processed.result())
+        try:
+            orig_filepath_2_uuid.append(processed.result())
+        except docker.errors.APIError:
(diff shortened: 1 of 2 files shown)


src/dsp_tools/fast_xmlupload/process_files.py
```diff
@@ -102,7 +103,28 @@ def _process_files_in_parallel(
...
+            print(f"{datetime.now()}: ERROR: A Docker exception occurred. Restart SIPI and resume processing...")
+            logger.error("A Docker exception occurred. Restart SIPI and resume processing...", exc_info=True)
+            # cancel all running jobs
+            for job in processing_jobs:
+                job.cancel()
+            # restart sipi container
+            _stop_and_remove_sipi_container()
+            _start_sipi_container_and_mount_volumes(input_dir, output_dir)
+            global sipi_container
+            sipi_container = _get_sipi_container()
```
BalduinLandolt, line 118: perfect candidate to extract into a method `restart_sipi_container()` - then you don't even need the comment

src/dsp_tools/fast_xmlupload/process_files.py
```diff
@@ -102,7 +103,28 @@ def _process_files_in_parallel(
...
+            # restart processing
+            processed_paths = [x[0] for x in orig_filepath_2_uuid]
+            unprocessed_paths = [x for x in paths if x not in processed_paths]
+            orig_filepath_2_uuid += _process_files_in_parallel(
+                paths=unprocessed_paths,
+                input_dir=input_dir,
+                output_dir=output_dir,
+                nthreads=nthreads
+            )
```
BalduinLandolt, line 127: doing a recursive call like this can potentially be a cause for errors, if the max recursion depth gets reached. In general, it would probably be more robust to move calling this function again one "level out", so it's flat, not recursive.  
But, to be honest, I'm not sure how big the danger is of this actually happening

src/dsp_tools/fast_xmlupload/process_files.py
```diff
@@ -102,7 +103,28 @@ def _process_files_in_parallel(
 
     orig_filepath_2_uuid: list[tuple[Path, Optional[Path]]] = []
     for processed in as_completed(processing_jobs):
-        orig_filepath_2_uuid.append(processed.result())
+        try:
+            orig_filepath_2_uuid.append(processed.result())
```
BalduinLandolt, line 107: maybe, instead of having this as a one-liner, you could first get it, then log some indication of what happened, and finally append it?

BalduinLandolt (approved):

src/dsp_tools/fast_xmlupload/process_files.py
jnussbaum: To be safe, I moved calling this function again one level out. Can you check that there's no logical flaw in it?

src/dsp_tools/fast_xmlupload/process_files.py
jnussbaum: I refactored these methods, because they were a bit spurious anyway.

src/dsp_tools/fast_xmlupload/process_files.py
BalduinLandolt: I didn't look thoroughly, but at first glance it seems good

PR #400, opened 2023-06-09
chore: lint "models", part one (DEV-2265)

jnussbaum:
resolves DEV-2265


src/dsp_tools/models/helpers.py
```diff
@@ -25,11 +24,7 @@ class OntoIri:
     hashtag: bool
 
 
-ContextType = NewType("ContextType", dict[str, OntoIri])
-
-
-def LINE() -> int:
-    return sys._getframe(1).f_lineno
+ContextType = dict[str, OntoIri]
```
jnussbaum, line 27: - pylint doesn't support NewType (see https://github.com/pylint-dev/pylint/issues/3162#issuecomment-539970727), so I went for a type alias instead.
- `LINE()` is nowhere used in the codebase, so I can get rid of this protected-member-access (`sys._getframe`)

src/dsp_tools/models/helpers.py
```diff
@@ -119,17 +114,6 @@ class Context:
...
-        """
-        Test if string has a valid IRI pattern
-
-        :param val: Input string
-        :return: True, if val corresponds to IRI pattern
-        """
-
-        m = self._exp.match(val)
-        return m.span()[1] == len(val) if m else False
-
```
jnussbaum, line 132: unused code

src/dsp_tools/models/group.py
```diff
@@ -1,16 +1,3 @@
...
-from typing import Any, Optional, Union
-from urllib.parse import quote_plus
-
-from dsp_tools.models.connection import Connection
-from dsp_tools.models.exceptions import BaseError
-from dsp_tools.models.helpers import Actions
-from dsp_tools.models.langstring import LangString
-from dsp_tools.models.model import Model
-from dsp_tools.models.project import Project
-
```
jnussbaum, line 13: module docstring must stand before the imports

src/dsp_tools/models/group.py
```diff
@@ -19,7 +6,7 @@
     * Call the ``create``-method on the instance
 
 READ:
-    * Instantiate a new object with ``id``(IRI of group) given
+    * Instantiate a new object with ``iri``(IRI of group) given
```
jnussbaum, line 9: `id` would shadow built-in

src/dsp_tools/models/ontology.py
```diff
@@ -210,23 +210,19 @@ def fromJsonObj(cls, con: Connection, json_obj: Any) -> "Ontology":
...
         # evaluate the JSON-LD context to get the proper prefixes
         #
         context = Context(json_obj.get("@context"))
-        onto_name = id.split("/")[-2]
-        context.add_context(onto_name, id + "#")
-        rdf = context.prefix_from_iri("http://www.w3.org/1999/02/22-rdf-syntax-ns#")
+        onto_name = iri.split("/")[-2]
+        context.add_context(onto_name, iri + "#")
         rdfs = context.prefix_from_iri("http://www.w3.org/2000/01/rdf-schema#")
-        owl = context.prefix_from_iri("http://www.w3.org/2002/07/owl#")
```
jnussbaum, line 225: unused variables
(in fact, all lines that I deleted without comment were unused)

src/dsp_tools/models/ontology.py
```diff
@@ -379,7 +354,7 @@ def toJsonObj(self, action: Actions) -> Any:
                 tmp[rdfs + ":comment"] = self._comment
         return tmp
 
-    def create(self, dumpjson: Optional[str] = None) -> "Ontology":
+    def create(self) -> "Ontology":
```
jnussbaum, line 357: unused parameter

src/dsp_tools/models/group.py
```diff
@@ -1,16 +1,3 @@
...
-
-from dsp_tools.models.connection import Connection
-from dsp_tools.models.exceptions import BaseError
-from dsp_tools.models.helpers import Actions
-from dsp_tools.models.langstring import LangString
-from dsp_tools.models.model import Model
-from dsp_tools.models.project import Project
-
 """
 This module implements the handling (CRUD) of Knora groups.
```
BalduinLandolt, line 2:
```suggestion
This module implements the handling (CRUD) of DSP groups.
```

src/dsp_tools/models/langstring.py
```diff
@@ -95,7 +95,7 @@ def __getitem__(self, key: Optional[Union[Languages, str]] = None) -> str:
             else:
                 try:
                     return list(self._langstrs)[0]
-                except:
+                except:  # pylint: disable=bare-except
                     return None
```
BalduinLandolt, line 99: If returning `None` is possible, then the type hint is not correct. Also, this code is really bad and "low hinging fruit" to improve:

```python
            if self._simplestring is not None: # or even `if self._simplestring: return`
                return self._simplestring
            return list(self._langstrs)[0] if list(self._langstrs) else None # or probably even `... if self._langstrs else None`
```

src/dsp_tools/models/listnode.py
```diff
@@ -150,40 +150,44 @@ def __init__(
...
+            isRootNode: Is True if the ListNode is a root node of a list Cannot be set [read].
+            children: list of children nodes. Only available for nodes that have been read by getAllNodes()
+            rootNodeIri: IRI of the root node. Only available for nodes that have been read by the method getAllNodes()
         """
 
         super().__init__(con)
 
         self._project = project.id if isinstance(project, Project) else str(project) if project else None
-        self._id = str(id) if id else None
+        self._id = str(iri) if iri else None
```
BalduinLandolt, line 185: why is this operation needed? it gets an optional string and converts it into an optional string?

src/dsp_tools/models/ontology.py
BalduinLandolt: I'm not 100% sure about this... but couldn't it possibly be nice to have those in the context? like this you can use `owl:SomethingFromOwl` in your requests/responses, without ever having to worry about wether or not `owl` is defined in your context.  
Not sure about the use case, though.

Also, this kind of thing should live in a separate json-ld-context util file for reuseability

BalduinLandolt (approved):
Over all looking good. I commented some minor improvements that spontaneously struck me as "low hanging" - but it's of course totally arbitrary, so feel free to ignore them if you don't want to mix lintintg with refactoring

src/dsp_tools/models/ontology.py
jnussbaum: AFAIU, the deleted LOC doesn't have side effects, it just retrieves a value. Some of the values that are retrieved are used later, but some (like the deleted LOC), weren't used. (`owl` was an unused variable).

So I don't think that we lose something if I delete them. Or did I miss something here?

BalduinLandolt: no, probably you're right (and if not, the code is in version control, so it's brought back easily enough)

PR #404, opened 2023-06-12
chore: loosen python requirement, add py.typed marker (DEV-2278)

jnussbaum:
resolves DEV-2278

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ exclude = ["src/dsp_tools/import_scripts/*"]
 
 
 [tool.poetry.dependencies]
-python = ">=3.9,<3.12"
+python = "^3.9"
 jsonpath-ng = "^1.5.3"
 argparse = "^1.4.0"
 lxml = "^4.9.2"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -15,12 +15,15 @@ classifiers = [
...
+include = [
+    "src/dsp_tools/resources/*",
+    "src/dsp_tools/py.typed"  # see https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages
+]
 exclude = ["src/dsp_tools/import_scripts/*"]
 
 
 [tool.poetry.dependencies]
-python = ">=3.9,<3.12"
+python = "^3.9"
```
jnussbaum, line 26: Having "<3.12" is destructive. For example, I cannot install dsp-tools in our permissions-repo:

```
(dsp-permissions-scripts-py3.11) dsp-permissions-scripts % poetry add dsp-tools
...
The current project's Python requirement (>=3.11,<4.0) is not compatible with some of the required packages Python requirement:
  - dsp-tools requires Python >=3.9,<3.12, so it will not be satisfied for Python >=3.12,<4.0
```

I don't think it makes sense to strictly forbid python 3.12 and higher for dsp-tools.

BalduinLandolt (approved):

PR #406, opened 2023-06-15
chore: lint "models", part two (DEV-2287)

jnussbaum:
resolves DEV-2287

### DIFF
--- a/src/dsp_tools/models/group.py
+++ b/src/dsp_tools/models/group.py
@@ -94,7 +94,7 @@ class Group(Model):
         self._name = str(name) if name is not None else None
         self._descriptions = LangString(descriptions)
         if project is not None and isinstance(project, Project):
-            self._project = project.id
+            self._project = project.iri
         else:
             self._project = str(project) if project is not None else None
         self._selfjoin = bool(selfjoin) if selfjoin is not None else None
--- a/src/dsp_tools/models/group.py
+++ b/src/dsp_tools/models/group.py
@@ -152,12 +152,6 @@ class Group(Model):
         self._status = value
         self._changed.add("status")
 
-    def has_changed(self) -> bool:
-        if self._changed:
-            return True
-        else:
-            return False
-
     @classmethod
     def fromJsonObj(cls, con: Connection, json_obj: Any):
         group_id = json_obj.get("id")
--- a/src/dsp_tools/models/permission.py
+++ b/src/dsp_tools/models/permission.py
@@ -22,26 +22,6 @@ class PermissionValue(Enum):
         return tmp[self.value]
 
 
-class PermissionsIterator:
-    _permissions: "Permissions"
-    _group: list[str]
-    _index: int
-
-    def __init__(self, permissions: "Permissions"):
-        self._permissions = permissions
-        self._index = 0
-
-    def __next__(self):
-        if len(self._permissions.permissions) == 0 and self._index == 0:
-            return None, None
-        elif self._index < len(self._permissions.permissions):
-            tmp = self._prefixes[self._index]
-            self._index += 1
-            return tmp, self._permissions.permissions[tmp]
-        else:
-            raise StopIteration
-
-
 class Permissions:
     _permissions: Union[dict[PermissionValue, list[str]], None]
 
@@ -62,9 +42,6 @@ class Permissions:
 
     def __missing__(self, key: PermissionValue) -> None:
         return None
-
-    def __iter__(self) -> PermissionsIterator:
-        return PermissionsIterator(self)
 
     def __contains__(self, key: PermissionValue) -> bool:
         return key in self._permissions
--- a/src/dsp_tools/models/project.py
+++ b/src/dsp_tools/models/project.py
@@ -1,17 +1,3 @@
-from __future__ import annotations
-
-import json
-from pprint import pprint
-from typing import Any, Optional, Union
-from urllib.parse import quote_plus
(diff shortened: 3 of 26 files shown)


src/dsp_tools/models/group.py
```diff
@@ -152,12 +152,6 @@ def status(self, value: bool) -> None:
         self._status = value
         self._changed.add("status")
 
-    def has_changed(self) -> bool:
-        if self._changed:
-            return True
-        else:
-            return False
-
```
jnussbaum, line 160: unused code

src/dsp_tools/models/project.py
```diff
@@ -1,17 +1,3 @@
-from __future__ import annotations
```
jnussbaum, line 1: module docstring must be before imports

src/dsp_tools/models/project.py
```diff
@@ -472,44 +471,9 @@ def delete(self) -> Project:
...
-                }
-            ],
-        }
-        jsondata = json.dumps(permobj, indent=4)
-        print(jsondata)
-        result = self._con.post("/admin/permissions/ap", jsondata)
-        pprint(result)
-
-        return
-        permobj = {
```
jnussbaum, line 496: unreachable code after `return`. Then I realized the entire function is never used. 

(Might be interesting because it handles permissions. But if we implement permissions handling in dsp-tools, I prefer starting over from zero, rather than working with such pirate's code...)

src/dsp_tools/models/propertyclass.py
```diff
@@ -16,12 +16,12 @@ class PropertyClass(Model):
     ROUTE: str = "/v2/ontologies/properties"
 
     _context: Context
-    _id: str
+    _iri: str
     _name: str
     _ontology_id: str
     _superproperties: list[str]
-    _object: str
```
jnussbaum, line 23: `object` is reserved in python. But when renaming it, I had to be consistent and rename `_object` and `_subject`, too.

src/dsp_tools/models/resource.py
```diff
@@ -38,14 +38,14 @@
 class KnoraStandoffXmlEncoder(json.JSONEncoder):
     """Classes used as wrapper for DSP standoff-XML"""
 
-    def default(self, obj) -> str:
```
jnussbaum, line 41: the superclass defines `default(self, o)`, so it's not recommended to rename `o` to `obj`

src/dsp_tools/models/sipi.py
```diff
@@ -27,7 +27,11 @@ def upload_bitstream(self, filepath: str) -> dict[Any, Any]:
             files = {
                 "file": (os.path.basename(filepath), bitstream_file),
             }
-            response = requests.post(self.sipi_server + "/upload?token=" + self.token, files=files)
+            response = requests.post(
+                self.sipi_server + "/upload?token=" + self.token,
+                files=files,
+                timeout=5 * 60,
```
jnussbaum, line 33: do you think that 5 minutes are a good timeout for uploading big multimedia files?

src/dsp_tools/models/user.py
```diff
@@ -608,60 +596,52 @@ def update(self, requesterPassword: Optional[str] = None) -> Any:
         jsonobj = self.toJsonObj(Actions.Update)
         if jsonobj:
             jsondata = json.dumps(jsonobj)
-            result = self._con.put(User.IRI + quote_plus(self.id) + "/BasicUserInformation", jsondata)
```
jnussbaum, line 611: here and in the following occurrences, the result was not used. So it's better not to store it in a variable

src/dsp_tools/models/permission.py
```diff
@@ -22,26 +22,6 @@ def __str__(self):
...
-
-    def __init__(self, permissions: "Permissions"):
-        self._permissions = permissions
-        self._index = 0
-
-    def __next__(self):
-        if len(self._permissions.permissions) == 0 and self._index == 0:
-            return None, None
-        elif self._index < len(self._permissions.permissions):
-            tmp = self._prefixes[self._index]
```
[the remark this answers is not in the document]
BalduinLandolt: If I were to guess, the intention of this code was to make `Permissions` iterable... but I'm not even sure this would have worked like that. And if it's not used, it should definitely be removed.

src/dsp_tools/models/project.py
BalduinLandolt: understandable ^^

but I think this gives a good hint for one option of where permissions could be added. So maybe it's worth keeping a permalink to this version of the file, so that you can look at it again, when/if you implement permissions in dsp-tools

src/dsp_tools/models/propertyclass.py
BalduinLandolt: I sometimes use `obj` and `subj` to avoid this problem and keep the variables short. But that's good too

src/dsp_tools/models/sipi.py
BalduinLandolt: hard to say... ideally, this wouldn't be hard coded but configurable. But for now, 5 minutes can't be all bad and it can always be changed if needed.

BalduinLandolt (approved):
LGTM, definitely a step in the right direction

src/dsp_tools/models/project.py
jnussbaum: Good idea 👍

PR #411, opened 2023-06-16
chore(xmlupload): improve error handling of Timeout Error (DEV-2299)

jnussbaum:
resolves DEV-


src/dsp_tools/utils/xml_upload.py
```diff
@@ -1041,9 +1041,12 @@ def _handle_upload_error(
...
     """
-
-    print("\n==========================================\nxmlupload must be aborted because of an error")
-    logger.info("xmlupload must be aborted because of an error")
+    print(
+        f"\n==========================================\n"
+        f"xmlupload must be aborted because of an error.\n"
+        f"Error message: '{err}'\n"
+    )
+    logger.error("xmlupload must be aborted because of an error", exc_info=err)
```
jnussbaum, line 1049: print error message here, instead of re-raising the error at the end. That's much more practical.

src/dsp_tools/utils/shared.py
```diff
@@ -86,9 +86,13 @@ def try_network_action(
                 return action(*args, **kwargs)
             else:
                 return action()
+        except (TimeoutError, ReadTimeout):
+            logger.error(f"Exceeded timeout while executing {action_as_str}")
+            raise BaseError(f"Exceeded timeout while executing {action_as_str}") from None
```
jnussbaum, line 91: This is based on the assumption that it is not worth retrying after a timeout. (Other types of errors lead to a retry, and also a timeout error led to a retry until now.)

BalduinLandolt: I haven't given this much thought, just now. But is this indeed an assumption that holds? I don't know in which scenarios you would actually get a timeout, but if there is one, where the network/server could recover afterwards, then you might want to retry after all

src/dsp_tools/utils/shared.py
```diff
@@ -99,7 +103,7 @@ def try_network_action(
             if try_again_later or in_500_range:
                 msg = f"Try reconnecting to DSP server, next attempt in {2 ** i} seconds..."
                 print(f"{datetime.now().isoformat()}: {msg}")
-                logger.error(msg, exc_info=True)
+                logger.error(f"{msg} {action_as_str}", exc_info=True)
```
BalduinLandolt, line 106: you may want to add `i` here too, so it's more immediately clear on which attempt one is when this log happens

BalduinLandolt (approved):
LGTM

PR #447, opened 2023-07-27
fix(start-stack): fix outdated TTL paths, open config file correctly (DEV-2497)

jnussbaum:


BalduinLandolt (approved):
sorry, I forgot to inform you about this change

PR #452, opened 2023-07-28
chore: replace gui_element Slider by Spinbox, remove gui_element SimpleText for intervals, et al. (DEV-2501)

jnussbaum:


README.md
BalduinLandolt, line 1: do we really need those batches?

docs/file-formats/json-project/ontologies.md
```diff
@@ -264,7 +264,7 @@ These three are related as follows:
 | isPartOf                    | (resourceclass)                                                     | Searchbox                            |
 | seqnum                      | IntValue                                                            | Spinbox, <br>SimpleText              |
 | isSequenceOf                | (AudioRepresentation,<br>MovingImageRepresentation,<br>or subclass) | Searchbox                            |
-| hasSequenceBounds           | IntervalValue                                                       | Interval                             |
+| hasSequenceBounds           | IntervalValue                                                       | Interval, <br>SimpleText             |
```
BalduinLandolt, line 267: I'm having a hard time imagine an IntervalValue working with a SimpleText text input. Is that supported by the APP? And if not, should TOOLS be able to do anything that the APP doesn't support?

BalduinLandolt (approved):

README.md
jnussbaum: I thought it would be a good practice to show as many badges as possible, so that people see with a quick glance what kind of technology we use.
But of course, this is not strictly necessary. So if you think it's confusing or bad practice, I can remove them. 

What are your general thoughts about badges?

BalduinLandolt: I personally don't look at them too closely, so to me they are not too important; and my thought was that having too many of them again makes it harder to see each single batch. But if you deem them useful I have nothing against them.

docs/file-formats/json-project/ontologies.md
jnussbaum: Good point. It was in the docs (further down), so I added it here to make it consistent.

But now that I think about it, it makes absolutely no sense. I removed support for it.

PR #453, opened 2023-07-28
chore: use darglint to check if the docstrings match the actual implementation (DEV-2530)

jnussbaum:


src/dsp_tools/excel2xml.py
```diff
@@ -49,7 +49,7 @@ def make_xsd_id_compatible(string: str) -> str:
         BaseError: if the input cannot be transformed to an xsd:ID
 
     Returns:
-        an xsd:ID based on the input string
+        an xsd ID based on the input string
```
BalduinLandolt, line 52: IMO not necessarily an improvement, as the colon here suggests namespacing... but if the linter (or you, for that matter 🙂 ) prefers it, sure

BalduinLandolt (approved):

PR #460, opened 2023-08-08
feat(create): in property definitions of JSON files, accept references to list labels (DEV-2247)

jnussbaum:


BalduinLandolt (approved):
looks good, as fart as I can tel (though I never really used lists, so I'm not too familiar with this part)

PR #461, opened 2023-08-09
chore: move to Python 3.11 (DEV-2524)

jnussbaum:


BalduinLandolt (approved):
Good! Now too long if-elif listings can be replaced with pattern matching :)

PR #463, opened 2023-08-09
chore!: harmonize CLI flags, rename CLI entry point and add tests for the CLI (DEV-2525)

jnussbaum:


test/unittests/test_excel_to_json_lists.py
```diff
@@ -47,21 +47,14 @@ def test_expand_lists_from_excel(self) -> None:
         lists_without_excel_reference_output = e2l.expand_lists_from_excel(lists_without_excel_reference)
         self.assertListEqual(lists_without_excel_reference, lists_without_excel_reference_output)
 
-    def test_make_json_lists_from_excel(self) -> None:
-        paths = [f"testdata/excel2json/lists-multilingual/{lang}.xlsx" for lang in ["de", "en", "fr"]]
-        lists_multilingual_output = e2l._make_json_lists_from_excel(paths)  # pylint: disable=protected-access
-        with open("testdata/excel2json/lists-multilingual-output-expected.json", encoding="utf-8") as f:
-            lists_multilingual_output_expected = json.load(f)
-        self.assertListEqual(lists_multilingual_output, lists_multilingual_output_expected)
-
```
jnussbaum, line 56: not necessary any more: is tested in `test_tools.py`

test/e2e/test_tools.py
jnussbaum, line 1: After your review, I will rename this file to `test_cli.py`, because that's more telling.

src/dsp_tools/cli_entry_point.py
```diff
@@ -114,7 +114,7 @@ def _make_parser(
         help="The links in the XML file point to IRIs (on the server) instead of IDs (in the same XML file).",
     )
     parser_upload.add_argument(
-        "-V", "--validate", action="store_true", help="validate the XML file without uploading it"
+        "-V", "--validate-only", action="store_true", help="validate the XML file without uploading it"
```
jnussbaum, line 117: This is a breaking change. I did it for consistency reasons.

src/dsp_tools/utils/id_to_iri.py
```diff
@@ -84,7 +84,7 @@ def id_to_iri(
         out_file = file_name + "_replaced_" + timestamp_str + ".xml"
 
     et = etree.ElementTree(tree.getroot())
-    et.write(out_file, pretty_print=True)
+    et.write(out_file, pretty_print=True, xml_declaration=True, encoding="utf-8")
```
jnussbaum, line 87: all other DSP XML files have a declaration, so it's probably better to add one here, too.

test/e2e/test_tools.py
```diff
@@ -191,123 +194,161 @@ def test_get_project(self) -> None:
...
-        self.assertTrue(validate_xml_against_schema(input_file=self.test_data_systematic_file))
+        subprocess.run(
+            f"poetry run dsp-tools xmlupload --validate-only --verbose {self.test_data_systematic_file.absolute()}",
+            check=True,
+            shell=True,
+            capture_output=True,
+            cwd=self.cwd,
+        )
 
     def test_xml_upload(self) -> None:
```
jnussbaum, line 205: Here, the diff is a bit confusing. In principle, it's the same than before, just expressed differently.

I split the method `test_xml_upload()` in two: 
The first uploads the the minimal data file (as before).
The second uploads the systematic data file (as before), then reads the mapping and calls the `id2iri` command, to make an incremental xmlupload (as before).

test/e2e/test_tools.py
```diff
@@ -191,123 +194,161 @@ def test_get_project(self) -> None:
...
         with open("testdata/excel2json/excel2json-expected-output.json", encoding="utf-8") as f:
             output_expected = json.load(f)
-        with open("testdata/tmp/_out_project.json", encoding="utf-8") as f:
+        with open(out_file, encoding="utf-8") as f:
             output = json.load(f)
         self.assertDictEqual(output, output_expected)
-        os.remove("testdata/tmp/_out_project.json")
+        out_file.unlink()
 
     def test_excel_to_json_list(self) -> None:
```
jnussbaum, line 263: Starting from here, I substantially improve the tests: Instead of testing if the output file exists, I test its exact content.

test/unittests/test_excel_to_json_lists.py
```diff
@@ -47,21 +47,14 @@ def test_expand_lists_from_excel(self) -> None:
...
     def test_validate_lists_section_with_schema(self) -> None:
         with open("testdata/excel2json/lists-multilingual-output-expected.json", encoding="utf-8") as f:
             lists_section_valid = json.load(f)
 
-        # validate the valid "lists" section in a correct way
+        # validate the valid "lists" section: should not raise an error
```
jnussbaum, line 54: From here on, I just improve the readability of this file, without substantial changes

testdata/id2iri/test-id2iri-data.xml
jnussbaum, line 1: The changes in this file are just for readability, without substantially changing the things that matter.

src/dsp_tools/excel2xml.py
```diff
@@ -1267,14 +1267,14 @@ def make_text_prop(
...
+            kwargs["comment"] = val.comment  # type: ignore[assignment]
         if check_notna(val.encoding):
-            kwargs["encoding"] = val.encoding  # type: ignore
+            kwargs["encoding"] = val.encoding  # type: ignore[assignment]
         else:
             kwargs["encoding"] = "utf8"
         value_ = etree.Element(
             "{%s}text" % xml_namespace_map[None],
-            **kwargs,  # type: ignore
+            **kwargs,  # type: ignore[arg-type]
```
jnussbaum, line 1277: Changing 1 line at the bottom of this file triggered pre-commit-hook, whose dogs barked at me that ignore-comments should always have an exact error code they are ignoring :-)

src/dsp_tools/cli_entry_point.py
BalduinLandolt: makes sense. Note, though, that the  PR name does not have the exclamation mark for breaking change, so unless you have a different release-please setup here, this will not appear as a major version bump

src/dsp_tools/utils/id_to_iri.py
BalduinLandolt: I'm surprised this isn't done by default. It's always better to add the declaration

src/dsp_tools/cli_entry_point.py
BalduinLandolt, line 1: personally I prefer module names to be short and to the point, ideally not having to resort to underscore. How would you feel about `cli.py`?

pyproject.toml
```diff
@@ -55,7 +55,7 @@ types-pyyaml = "^6.0.12.11"
 
 
 [tool.poetry.scripts]
-dsp-tools = "dsp_tools.dsp_tools:main"    # definition of the CLI entry point
+dsp-tools = "dsp_tools.cli_entry_point:main"    # definition of the CLI entry point
```
BalduinLandolt, line 58: see my note below

BalduinLandolt: and in terms of useability: how would you feel about the CLI command being `dsp` instead of `dsp-tools`? (for now you could even support both for a while, maybe with a deprecation warning)

BalduinLandolt (approved):
See my comments - but over all looking good

pyproject.toml
jnussbaum: typing `dsp-tools ...` is probably quite deeply internalized in the users' brains, so I'm a bit reluctant of changing it. And also, the tool is called "DSP-TOOLS", and installed with `pip install dsp-tools`, so it is most natural to type `dsp-tools ...` in the Terminal.

I think that there must be a really strong reason to change something like this. Do you think that there is such a compelling reason? 

(I think the difference in the length is too small to justify such a change...)

BalduinLandolt: for your first point, I would argue: people can learn - I'm sure this change would be internalised rather quickly.  
for your second part, I would argue: no. 😉  there are many CLI tools that have a simpler entry point than their full /unique name (`brew` for Homebrew, `http` for HTTPie, etc.)

I personally think it would be nicer, not only because of it being shorter, but also because of it not needing the hyphen character (I'm most used to CLI tools being [a-z] only), and to me it would sound more like thought from the user perspective: they want to create a datamodel on the DSP, and do not necessarily care that they use DSP-TOOLS for that, so as a statement `dsp create` is closer to what they actually want than `dsp-tools create`.  
But no, I don't think there is an urgent reason for the change. So if you really deem it that radical, then it's not justified.

PR #464, opened 2023-08-11
fix(upload-files): handle server error, document how to circumvent resource leak (DEV-2528, DEV-2527)

jnussbaum:


src/dsp_tools/fast_xmlupload/upload_files.py
```diff
@@ -151,51 +152,93 @@ def _check_upload_candidates(
...
+        sipi_url: URL of the SIPI server
+        con: connection to the DSP server
+        err_msg: message to print and log
+
+    Returns:
+        return value of _upload_without_processing()
+    """
+    print(f"{datetime.now()}: ERROR: {err_msg}")
+    logger.error(err_msg, exc_info=True)
+    sleep(2)
```
Vijeinath, line 181: Why is this sleeping needed?

Vijeinath (approved):
LGTM

src/dsp_tools/fast_xmlupload/upload_files.py
jnussbaum: In case the error has something to do with server overload, the server might need some time to recover. But I admit that this is hypothetical.

PR #469, opened 2023-08-11
feat: add --dump flag to "xmlupload" and "get" commands, improve dumping mechanism (DEV-2534)

jnussbaum:


pyproject.toml
```diff
@@ -84,12 +84,34 @@ addopts = ["--import-mode=importlib"]
...
+    "src/dsp_tools/models/resourceclass.py",      # TODO: activate this
+    "src/dsp_tools/models/set_encoder.py",        # TODO: activate this
+    "src/dsp_tools/models/user.py    ",           # TODO: activate this
+    "src/dsp_tools/models/value.py",              # TODO: activate this
+    "src/dsp_tools/models/xmlallow.py",           # TODO: activate this
+    "src/dsp_tools/models/xmlbitstream.py",       # TODO: activate this
+    "src/dsp_tools/models/xmlpermission.py",      # TODO: activate this
+    "src/dsp_tools/models/xmlproperty.py",        # TODO: activate this
+    "src/dsp_tools/models/xmlresource.py",        # TODO: activate this
+    "src/dsp_tools/models/xmlvalue.py",           # TODO: activate this
```
jnussbaum, line 114: Instead of ignoring the entire "src/dsp_tools/models" folder, I want to ignore only the files that I haven't refactored yet. Like this, I make sure that the refactored ones are checked.

src/dsp_tools/cli.py
```diff
@@ -123,6 +124,7 @@ def _make_parser(
     )
     parser_upload.add_argument("-v", "--verbose", action="store_true", help=verbose_text)
     parser_upload.add_argument("-m", "--metrics", action="store_true", help="write metrics into a 'metrics' folder")
+    parser_upload.add_argument("-d", "--dump", action="store_true", help="write every request to DSP-API into a file")
```
BalduinLandolt, line 127: I would assume it dumps requests to SIPI too, so the description is technically not entirely correct. (though maybe from a pragmatic point of view clear enough)

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
...
-    _log: bool
-
-    def __init__(self, server: str, prefixes: dict[str, str] = None):
-        """
-        Constructor requiring the server address, the user and password of DSP
-        :param server: Address of the server, e.g https://api.dasch.swiss
-        :param prefixes: Ontology prefixes used
-        """
-
-        self._server = regex.sub(r"\/$", "", server)
```
BalduinLandolt, line 52: do you have any idea why this was here?

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
         )
 
 
+@dataclass
```
BalduinLandolt, line 32: I love it! I wonder though, if you could make it immutable by changing some small things in the connection functionality. I'll add some thoughts below.

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
...
-        self._prefixes = prefixes
-        self._token = None
-        self._log = False
+    server: str
+    prefixes: Optional[dict[str, str]] = None
+    token: Optional[str] = None
+    log: bool = False
+    log_directory: Path = Path("HTTP requests")
 
     def login(self, email: str, password: str) -> None:
```
BalduinLandolt, line 51: Is there a use case for a non-logged-in connection? Otherwise this could be moved to the `__post_init__()` method so it doesn't conflict with immutability

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
...
         """Start writing every API call to a file"""
-        self._log = True
+        self.log = True
+        self.log_directory.mkdir(exist_ok=True)
 
-    def stop_logging(self):
+    def stop_logging(self) -> None:
         """Stop writing every API call to a file"""
-        self._log = False
+        self.log = False
```
BalduinLandolt, line 76: I would say a connection either logs or not, this doesn't need to change during runtime. Or is there a good reason for it?

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
...
             response = requests.delete(
-                self._server + "/v2/authentication",
-                headers={"Authorization": "Bearer " + self._token},
+                self.server + "/v2/authentication",
+                headers={"Authorization": "Bearer " + self.token},
                 timeout=20,
             )
             check_for_api_error(response)
-            self._token = None
-
```
BalduinLandolt, line 112: could logout be removed entirely?

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
...
-    def __del__(self):
-        pass
-        # self.logout()
-
-    def post(self, path: str, jsondata: Optional[str] = None):
+            self.token = None
+
+    def write_log_file(
```
BalduinLandolt, line 89: I think I mentioned this before: to me the term "log" suggests something narrower than just writing information to a file, namely a certain line-by-line format with timestamp, loglevel, etc.  
How about `write_http_request`?

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
...
+
+        Args:
+            method: HTTP method (POST, GET, PUT, DELETE)
+            route: route of DSP-API that was called
+            headers: headers of the HTTP request
+            jsondata: data sent to the server
+            response: response of the server
         """
+        if not self.log:
+            return
```
BalduinLandolt, line 109: I would check this condition outside of the function: like this, the function does two things (and should be renamed to `write_log_file_or_maybe_not()` 😉  ), but it should only do one thing.

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
...
+            "method": method,
+            "headers": headers,
+            "params": params,
+            "body": json.loads(jsondata) if jsondata else None,
+            "return-headers": dict(response.headers),
+            "return": _return,
+        }
+        filename = f"{datetime.now().strftime('%Y-%m-%d %H.%M.%S.%f')} {method} {route.replace('/', '_')}.json"
+        with open(self.log_directory / filename, "w", encoding="utf-8") as f:
+            json.dump(logobj, f, indent=4)
```
BalduinLandolt, line 126: I was wondering two things here:

1. Should this really be one file per request (potentially very many files), or would it be better to have a single file per run, containing all requests (potentially a huge file; plus you'd need  to be able to append efficiently). I'm not sure to be honest.
2. is JSON the best format for this? Generally, JSON is nice most times - but the most relevant parts of this, request and response, are both JSON themselves; so you end up with JSON in JSON which greatly reduces readability.  
My first intuition was to suggest simply using a .txt file, but that's not nice either.  
Then I thought, you could try to emulate a .http file (seems to be adopted quite widely by now, see [here](https://marketplace.visualstudio.com/items?itemName=humao.rest-client), [here](https://www.jetbrains.com/help/idea/http-client-in-product-code-editor.html), [here](https://learn.microsoft.com/en-us/aspnet/core/test/http-files?view=aspnetcore-8.0), etc.). The nice thing would be that you get syntax highlighting, if properly done, the requests can even be executed, and the next request can simply be appended after a divider. The downside is that it would be quite a bit more complex to implement.

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
...
+
+    def post(
+        self,
+        path: str,
+        jsondata: Optional[str] = None,
```
BalduinLandolt, line 131: how can this be none?

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
...
+            response from server
+        """
         # timeout must be None,
         # otherwise the client can get a timeout error while the API is still processing the request
         # in that case, the client's retry will fail, and the response of the original API call is lost
         timeout = None
-
-        if path[0] != "/":
+        if not path.startswith("/"):
             path = "/" + path
```
BalduinLandolt, line 148: I'm not sure under what condition this can happen. But could it be fixed on the caller side?

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
...
+            route=path,
+            headers=headers,
+            jsondata=None,
+            params=None,
+            response=response,
+        )
         check_for_api_error(response)
-        json_response = response.json()
+        json_response = cast(dict[str, Any], response.json())
         return json_response
```
BalduinLandolt, line 206: maybe I'm going a bit over board now... but this is almost identical for all HTTP methods; could the shared logic be extracted to a separate method?

src/dsp_tools/models/connection.py
```diff
@@ -27,240 +29,260 @@ def check_for_api_error(response: requests.Response) -> None:
...
-        return result
+        json_response = cast(dict[str, Any], response.json())
+        return json_response
 
-    def delete(self, path: str, params: Optional[any] = None):
-        """
-        Send a delete request using the HTTP DELETE request
-        :param path: Path of RESTful route
-        :return: Response from server
+    def delete(
```
BalduinLandolt, line 252: does DSP-Tools ever do a delete request? if not, this could be removed

src/dsp_tools/models/sipi.py
BalduinLandolt, line 1: pretty much the same remarks as with the connection apply here

src/dsp_tools/utils/project_get.py
```diff
@@ -30,16 +32,18 @@ def get_project(
...
     Returns:
         True if the process finishes without errors
     """
-    con = Connection(server)
-    if user and password:
-        con.login(user, password)
+    # establish connection to DSP server
+    con = login(server=server, user=user, password=password)
+    if dump:
+        con.start_logging()
```
BalduinLandolt, line 46: If you imagine Connection to be immutable, this would change to 
```python
con = Connection(server, user, password, dump)
```

BalduinLandolt (approved):
I think this is great improvement.

I added quite a number of suggestions, how this could go even further - feel free to apply or ignore them, or keep them in the back hand for later. :)

src/dsp_tools/models/connection.py
jnussbaum: perhaps in the past, the final slash caused a problem? I just tried to remove it, and since the tests pass, I assume that the final slash is now okay...

src/dsp_tools/models/connection.py
jnussbaum: In the codebase, there is no example of a non-logged-in connection. But are there use cases where this should be possible? Should it perhaps be possible to get a project definition from a server with the `get` command, as UnknownUser?

On the other hand, that would be a privacy issue, because the users of a project are also retrieved by `get`.

src/dsp_tools/models/connection.py
BalduinLandolt: fair enough :)

src/dsp_tools/models/connection.py
BalduinLandolt: This is in fact something I've been confused about before, that get requires credentials (I thought I even opened an issue for it at some point).  
And, yes, that is potentially a privacy issue, but it's API side: It's not the client's fault to retrieve information that the API exposes publicly.

But for that use case, maybe this can just be a Connection where the token is `None`? Then the `GET` command could simply not provide the bearer header in this case.

jnussbaum: I tried to get rosetta from the test server, without credentials, and when retrieving the users, I get `dsp.errors.ForbiddenException: SystemAdmin or ProjectAdmin permissions are required`. If I catch this error, the rest of the project (ontologies, groups, lists, ...) can be retrieved correctly.

I think I should add support for this (would be a new feature: get a project without login)

BalduinLandolt: ah, all the better! Yes, that would be perfect. I suppose the users step should just be skipped if no credentials are provided

src/dsp_tools/models/connection.py
jnussbaum: no :-) `dump` is now a parameter that is passed when a `Connection` is created, so it's immutable.

src/dsp_tools/models/connection.py
jnussbaum: done

src/dsp_tools/models/connection.py
jnussbaum: good point. done

src/dsp_tools/models/connection.py
jnussbaum: There are quite some examples in the codebase where it is none, e.g. here: https://github.com/dasch-swiss/dsp-tools/blob/wip/dev-2534-refactor-models/src/dsp_tools/models/user.py#L543

src/dsp_tools/models/connection.py
jnussbaum: That's a simple precaution measure, if one of the (many!) callers forgets the slash. Of course I could remove it, and fix all the callers. But isn't it a nice little feature that makes usage easier?

src/dsp_tools/models/connection.py
jnussbaum: yes, there are delete requests

src/dsp_tools/models/connection.py
jnussbaum: I tried it out with the `create` and `xmlupload` commands, and I found it convenient with 1 file/request, and also the JSON format was well readable. If we are going to use this feature a lot, then I can think about .http files, which is quite cool, but not worth the effort right now.

src/dsp_tools/models/connection.py
BalduinLandolt: oof... bad API design from our side! 🙈

src/dsp_tools/models/connection.py
BalduinLandolt: let me re-phrase this:  
is any of the delete methods ever used? I had a quick look and didn't find any instances.  
I know it could be used by someone using DSP-Tools as a module for scripts... but do people use delete in these cases?

src/dsp_tools/models/connection.py
BalduinLandolt: fair enough. But then maybe some mechanism needs to be in place for large uploads (say Vij's bulk uploads or something of that size) to not swamp the file system with ten-thousands of files? or do you think that would not be an issue?

src/dsp_tools/models/connection.py
jnussbaum: To be sure that I understand you correctly: Connection.delete() is used by Group.delete(), which is only used in the e2e tests. 

If all usages of Connection.delete() are only used in the e2e tests, I could remove all usages, and modify the e2e tests. Is that what you mean?

src/dsp_tools/models/connection.py
jnussbaum: This is really an issue, but I'd like to put the burden on the user. Because I cannot think of an elegant way how to prevent too many files to be written. 
1. Detect if there are too many: in addition to sending requests and writing dump files, this class would have to measure how many MB are already in the dump directory. This measurement would take place before EVERY dump, which slows down the main program.
2. If the directory is full, what to do? Stop writing files? Deleting the old files?

I tend to be lazy here and appeal to the user's responsibility.

BalduinLandolt: As the user has to actively activate the dumping, I suppose the lazy approach is ok.

If you still wanted to go for it, here are some thoughts:
- I don't think disk storage space will be the biggest issue, instead what would be problematic much faster, is having too many small files in one directory.
- so the simple solution would be to create sub-directories
- a first step would be to have a subdirectory per run of DSP-Tools, so that the number of files does not accumulate
- and to go even further, you could even divide single runs up
- here, I don't think, creating datamodels would cause any issues, it's probably only the data upload.
- and for the data upload you parse the XML first, so you should know the number of resources to be created
- and you should be able to make a rough guess how many requests per resource there may be
- then you could say: if its less than x (I don't know... maybe 1000?) resources, no further subfolders are needed
- and if it's more, you could have subfolders `0`, `1`, etc. for every x resources
I don't think that would have big performance implications... but it's probably not worth the effort.

PR #483, opened 2023-08-22
chore: some small bug fixes and improvements (DEV-2564)

jnussbaum:


src/dsp_tools/resources/schema/project.json
```diff
@@ -16,18 +16,15 @@
         },
         "ncname": {
             "type": "string",
-            "pattern": "^[a-zA-Z_][\\w.-]*$",
-            "additionalProperties": false
```
jnussbaum, line 20: Here I'm unsure if "additionalProperties" should be kept or removed. (And also in the similar cases below).

src/dsp_tools/resources/schema/project.json
```diff
@@ -769,30 +765,46 @@
...
+                                            ]
+                                        },
+                                        {
+                                            "$ref": "#/definitions/prefixedname_or_uri"
+                                        }
+                                    ]
+                                }
+                            },
                             "gui_element": {
-                                "super": {
```
jnussbaum, line 773: This here was quite certainly a bug. `super` should not be subordinate of `gui_element`.

src/dsp_tools/utils/xml_upload.py
```diff
@@ -770,7 +770,11 @@ def _upload_resources(
         except BaseError as err:
             err_msg = err.orig_err_msg_from_api or err.message
             print(f"WARNING: Unable to create resource '{resource.label}' ({resource.id}): {err_msg}")
-            logger.warning(f"Unable to create resource '{resource.label}' ({resource.id}).", exc_info=True)
+            logger.warning(
+                f"Unable to create resource '{resource.label}' ({resource.id})\n"
+                f"Resource details:\n{vars(resource)}",
+                exc_info=True,
+            )
```
jnussbaum, line 777: with `vars(object)`, you can print detailed info about the object, especially all class attributes. That's very useful for debugging.

test/unittests/test_xmlupload.py
```diff
@@ -148,8 +148,7 @@ def test_remove_circular_references(self) -> None:
         stashed_xml_texts_expected = {
             "test_thing_1": {
                 "testonto:hasRichtext": [
-                    "\n                This is <em>bold and <strong>strong</strong></em> text! "
-                    "It contains links to all resources:\n"
+                    "\n                This text contains links to all resources:\n"
```
jnussbaum, line 151: Here, I separated 2 things: 

- "test_xmlupload.py" should use a text value that contains only some links, because that's the thing to be tested in this test.
- the `<em>` tags etc. are now moved into another property (see below), so that if something changes there, this test here doesn't have to be adapted.

testdata/xml-data/test-data-systematic.xml
```diff
@@ -84,8 +84,9 @@
             <text encoding="xml">&amp;</text>
             <text encoding="xml"><em>text</em></text>
             <text encoding="xml">text &lt;not a tag&gt; text</text>
-            <text permissions="prop-default" encoding="xml">
-                This is <em>bold and <strong>strong</strong></em> text! It contains links to all resources:
+            <text encoding="xml">This is <em>italicized and <strong>bold</strong></em> text! </text>
+            <text encoding="xml" permissions="prop-default">
+                This text contains links to all resources:
                 <a class="salsah-link" href="IRI:test_thing_0:IRI">test_thing_0</a>
```
jnussbaum, line 90: This is the change that I mentioned above: I splitted up 1 text value into 2 values.

test/unittests/test_xmlupload.py
Nora-Olivia-Ammann, line 151: Little confused by "should use a text value that contains only some links" because the first line (new 151) is plain text.

jnussbaum: ah okay, I should have omitted the word "only" :-)

Nora-Olivia-Ammann (approved):

PR #484, opened 2023-08-23
fix(xmlupload): sanitize textvalues: remove whitespaces and newlines (DEV-2569)

jnussbaum:


testdata/xml-data/test-data-systematic.xml
```diff
@@ -200,9 +199,6 @@
             <text permissions="prop-default" encoding="xml">
                 Another text without salsah-links
             </text>
-            <text permissions="prop-default" encoding="xml">
-                Another text without salsah-links
-            </text>
```
jnussbaum, line 205: Duplicate values for the same property are not allowed. Apparently, there were some whitespaces that had coverted this up until now.

testdata/xml-data/test-data-systematic.xml
```diff
@@ -174,7 +174,6 @@
         <text-prop name=":hasSimpleText">
             <text permissions="prop-default" encoding="utf8">Dies ist ein einfacher Text ohne Markup</text>
             <text permissions="prop-restricted" encoding="xml">Ein XML-formatierter Text <strong>mit Markup</strong></text>
-            <text encoding="utf8"> </text>
```
jnussbaum, line 177: this doesn't work any more

src/dsp_tools/models/xmlvalue.py
```diff
@@ -23,14 +24,82 @@ def __init__(
...
+        return xmlstr
+
+    def _cleanup_unformatted_text(self, string_orig: str) -> str:
+        """
+        In a utf8-encoded text value from the XML file,
+        there may be non-text characters that must be removed.
+        This method:
+            - removes the <text> tags
+            - replaces multiple spaces or tabstops by a single space
+            - replaces 3+ line breaks by 2 line breaks (=max. 1 empty line)
```
jnussbaum, line 83: Is this a good idea, or shall I leave all line breaks in place?

Nora-Olivia-Ammann: Well in my mind unformatted text is a plain string **without any** formatting including line breaks, tabs etc.

src/dsp_tools/models/xmlvalue.py
```diff
@@ -23,14 +24,82 @@ def __init__(
...
+        """
+        # remove the <text> tags
+        string = regex.sub("<text.*?>", "", string_orig)
+        string = regex.sub("</text>", "", string)
+
+        # replace multiple spaces or tabstops by a single space
+        string = regex.sub(r" {2,}|\t+", " ", string)
+
+        # replace 3+ line breaks by 2 line breaks
+        string = regex.sub(r"(\n ?){3,}", "\n\n", string)
```
BalduinLandolt, line 99:
```suggestion
        string = regex.sub(r"(\n *){3,}", "\n\n", string)
```
I think you need this to be sure to match double newlines followed by multiple spaces

BalduinLandolt (approved):
Thanks! See my comments below

src/dsp_tools/models/xmlvalue.py
Nora-Olivia-Ammann: Ok, makes sense if that is the data we get.

src/dsp_tools/models/xmlvalue.py
```diff
@@ -23,14 +24,82 @@ def __init__(
...
+
+        Args:
+            xmlstr_orig: original string from the XML file
+
+        Returns:
+            purged string, suitable to be sent to DSP-API
+        """
+        # remove the <text> tags
+        xmlstr = regex.sub("<text.*?>", "", xmlstr_orig)
+        xmlstr = regex.sub("</text>", "", xmlstr)
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I am currently working on something like that, and joining elements from different tags from an existing etree is quite complicated, especially if the structure with the amount of children etc is unknown. Additionally Gregor advised not to change a current etree but build a new one from an existing as it leads in his experience to less mistakes. I have a feeling that in such a case a string operation might be simpler.

src/dsp_tools/models/xmlvalue.py
```diff
@@ -23,14 +24,82 @@ def __init__(
...
+        # replace (multiple) line breaks by a space
+        xmlstr = regex.sub("\n+", " ", xmlstr)
+
+        # replace multiple spaces or tabstops by a single space (except within <code> tags)
+        # the regex selects all spaces/tabstops not followed by </code> without <code in between.
+        # credits: https://stackoverflow.com/a/46937770/14414188
+        xmlstr = regex.sub("( {2,}|\t+)(?!(.(?!<code))*</code>)", " ", xmlstr)
+
+        # remove spaces after <br/> tags (except within <code> tags)
+        xmlstr = regex.sub("((?<=<br/?>) )(?!(.(?!<code))*</code>)", "", xmlstr)
```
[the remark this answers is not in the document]
jnussbaum: thanks for the hint. Done

docs/file-formats/xml-data-file.md
```diff
@@ -688,11 +688,22 @@ contains a link to  the resource `http://rdfh.ch/4123/nyOODvYySV2nJ5RWRdmOdQ`, w
...
+Handling of pretty-print whitespaces and newlines:
+
+- encoding `utf8`: 
+    - leading and trailing whitespaces are removed
+    - multiple spaces are replaced by a single space
+    - (multiple) tabstops are replaced by a single space
+    - 3+ line breaks are replaced by 2 line breaks (=max. 1 empty line)
+- encoding `xml`:
+    - leading and trailing whitespaces are removed
+    - whitespaces and newlines don't have a meaning in XML, so they are removed. Use `<br/>` instead.
```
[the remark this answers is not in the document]
jnussbaum: Done

src/dsp_tools/models/xmlvalue.py
```diff
@@ -23,14 +24,82 @@ def __init__(
...
+
+        # remove leading and trailing spaces
+        xmlstr = xmlstr.strip()
+
+        return xmlstr
```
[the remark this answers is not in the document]
jnussbaum: I feel the same than Nora: It would perhaps be possible to do this with lxml, but I would be super scared to miss edge cases. I will never now if the `etree.Element` has children or not, and I think that the logic would become super complicated if I try to do the same with lxml. Perhaps it's just that I have more confidence in my regex skills than in my lxml skills 😄 .

src/dsp_tools/models/xmlvalue.py
jnussbaum: I just decided to leave all line breaks in place. I think that sanitation should be conservative, not too much messing up with what the user might want. If users want more than 2 line breaks, why should I hinder them to do so...

src/dsp_tools/models/xmlvalue.py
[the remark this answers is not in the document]
BalduinLandolt: fair enough :)

PR #487, opened 2023-08-24
refactor: excel2json (DEV-2547)

Nora-Olivia-Ammann:


src/dsp_tools/utils/excel_to_json/excel_to_json_properties.py
```diff
@@ -0,0 +1,470 @@
...
+    the string to the corresponding data type. If it is not a float or integer, it returns the string as is.
+
+    Args:
+        value_str: The value which is checked and may be converted
+
+    Returns:
+        A int if the string was an integer, float if the string was a float or str if it was neither
+    """
+    if regex.search(r"^\d+\.\d+$", value_str):
+        return float(value_str)
```
BalduinLandolt, line 107: this would miss `"1e+3"` which is a valid representation of `1000.0` in python

src/dsp_tools/utils/excel_to_json/excel_to_json_properties.py
```diff
@@ -0,0 +1,470 @@
...
+        attribute_str: A string containing the attributes
+
+    Returns:
+        A dictionary with the attribute name as a key and the attribute as value.
+
+    Raises:
+        IndexError if the attributes are not formatted correctly
+    """
+    attribute_dict = _unpack_gui_attributes(gui_str=attribute_str)
+    return {attrib: _search_convert_numbers(value_str=val) for attrib, val in attribute_dict.items()}
```
BalduinLandolt, line 157: I like this styli with ultra-short functions and speaking names; but if code is written like that, such an extent of documentation is somewhat redundant and actually makes the code harder to follow

BalduinLandolt (approved):
Generally looks good!  
I found it hard to review this PR reasonably, though, because the git diff was a bit all over the place (because stuff was moved around) and I'm not really familiar with this code

src/dsp_tools/cli.py
```diff
@@ -13,10 +13,10 @@
...
 from dsp_tools.fast_xmlupload.upload_xml import fast_xmlupload
 from dsp_tools.models.exceptions import UserError
-from dsp_tools.utils.excel_to_json_lists import excel2lists, validate_lists_section_with_schema
-from dsp_tools.utils.excel_to_json_project import excel2json
-from dsp_tools.utils.excel_to_json_properties import excel2properties
-from dsp_tools.utils.excel_to_json_resources import excel2resources
+from dsp_tools.utils.excel_to_json.excel_to_json_lists import excel2lists, validate_lists_section_with_schema
+from dsp_tools.utils.excel_to_json.excel_to_json_project import excel2json
+from dsp_tools.utils.excel_to_json.excel_to_json_properties import excel2properties
+from dsp_tools.utils.excel_to_json.excel_to_json_resources import excel2resources
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Good point! I'll change that

src/dsp_tools/utils/excel_to_json/excel_to_json_properties.py
Nora-Olivia-Ammann: Thanks I did not think of that. I will resolve this issue separately.

src/dsp_tools/utils/excel_to_json/excel_to_json_properties.py
Nora-Olivia-Ammann: Thanks! I have shortened it some.

jnussbaum:
@Nora-Olivia-Ammann: You can (should 😃 ) delete the comment at the very top of the PR, which is surrounded by `===REMOVE===`

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
+# pylint: disable=wrong-import-order
```
jnussbaum, line 1: It is a good practice to sort the imports:

1. python stdlib
2. third-party-packages from pip
3. dsp-tools internal packages

In VSCode, the `isort` extension offers auto-fix support for that.

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+from unittest import mock
+
+from dsp_tools.models.exceptions import UserError
+
+languages = ["en", "de", "fr", "it", "rm"]
+
+
+def read_excel_file(excel_filename: str) -> pd.DataFrame:
+    """
+    This function reads an Excel file, if there is a ValueError then it patches the openpyxl part that creates the
```
jnussbaum, line 18: I'm a big fan of [Semantic Line Breaks](https://sembr.org/). It has some huge advantages:

1) If you change something later, the diff is cleaner.
2) If you change something later, and you insert 2-3 words in the first line, you don't have to reorganize the entire paragraph.
3) For a writer, semantic line breaks allow the physical structure of text to reflect the logical structure of the thoughts that produce them.

That's why I use Semantic Line Breaks nearly everywhere. But of course, this is just a recommendation.

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+    error and opens it with that patch. It cleans and then returns the pd.DataFrame.
+
+    Args:
+        excel_filename: The name of the Excel file
+
+    Returns:
+        A pd.DataFrame
+    """
+    # load file
```
jnussbaum, line 27: Is this comment still necessary?

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+    try:
+        read_df: pd.DataFrame = pd.read_excel(excel_filename)
+    except ValueError:
+        # Pandas relies on openpyxl to parse XLSX files.
+        # A strange behaviour of openpyxl prevents pandas from opening files with some formatting properties
+        # (unclear which formatting properties exactly).
+        # Apparently, the excel2json test files have one of the unsupported formatting properties.
+        # Credits: https://stackoverflow.com/a/70537454/14414188
```
jnussbaum, line 35: Would it perhaps be more elegant to move this very long comment to the docstring?

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+        with mock.patch("openpyxl.styles.fonts.Font.family.max", new=100):
+            read_df = pd.read_excel(excel_filename)
+    read_df = clean_data_frame(unclean_df=read_df)
+    return read_df
+
+
+def clean_data_frame(unclean_df: pd.DataFrame) -> pd.DataFrame:
```
jnussbaum, line 42: Will this function eventually replace `prepare_dataframe()`? Now we have 2 functions that do similar things. 

Related question: `prepare_dataframe()` is only used in `excel_to_json/resources.py`. I realized that `excel_to_json/resources.py` is now outdated, and uses old patterns, that are improved in `properties.py`. Perhaps, you could make a follow-up PR that assimilates the `excel_to_json/resources.py` to `excel_to_json/properties.py`? I mean only the most important stuff, only the overall-structure. You don't have to go too far into the details, because there are other things that are more important.

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+
+    Args:
+        unclean_df: The pd.DataFrame that is to be cleaned
+
+    Returns:
+        pd.DataFrame which has the above-mentioned removed
+    """
+    # Remove leading and trailing blanks in column names and make them lower case
+    cleaned_df = unclean_df.rename(columns=lambda x: x.strip().lower())
+    # Remove the values of all cells that do not at least contain one character of any known language
```
jnussbaum, line 59: This function does 4 things. All 4 are described in the docstring, but there are only 3 operations in the code. It took me a while to realize that the 2nd operation does 2 things at a time. 
Would you perhaps like to write in the comment (line 59) that the following line of code also removes leading and trailing spaces from each cell?

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
+import importlib.resources
+import json
+import warnings
+from typing import Any, Optional
+import jsonpath_ng.ext
+import jsonschema
+import regex
+import pandas as pd
+import numpy as np
```
jnussbaum, line 9: Here, isort could add an empty line before pandas, and then sort the imports in each group alphabetically. (Okay, not the most important thing to be done, but I just mention it 😄 )

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
...
+    """
+    with importlib.resources.files("dsp_tools").joinpath("resources/schema/properties-only.json").open(
+        encoding="utf-8"
+    ) as schema_file:
+        properties_schema = json.load(schema_file)
+    try:
+        jsonschema.validate(instance=properties_list, schema=properties_schema)
+    except jsonschema.ValidationError as err:
+        err_msg = _search_validation_error(
+            list_properties=properties_list, excel_filename=excelfile, validation_error=err
```
jnussbaum, line 88: I tend to leave variable names as they are. Then I can search for `list_properties` in the entire file, and I find every occurrence. But if they are renamed to `list_properties` inside this function, I miss the ones inside this functions.

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
...
+languages = ["en", "de", "fr", "it", "rm"]
+language_label_col = ["label_en", "label_de", "label_fr", "label_it", "label_rm"]
+mandatory_properties = ["name", "object", "gui_element"]
+
+
+def _search_validation_error(
+    list_properties: list[dict[str, Any]], excel_filename: str, validation_error: jsonschema.ValidationError
+) -> str:
+    """
+    This function takes a list of properties, which are formatted as a dictionary and an error which was raised by the
```
jnussbaum, line 23: Here, I think that the docstring doesn't give a good description of what actually happens. 

- The properties are not "formatted" as dictionary. Rather, they have been transformed from Excel-structure into JSON-structure. So "structured" is perhaps better than "formatted".
- "error which was raised by the calling function" -->" error from the JSON schema validation" would be more telling
- "which exact location of the property caused the error". Perhaps better: "which location in the Excel file caused the error"

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
...
+
+import dsp_tools.utils.excel_to_json.utils as utl
+from dsp_tools.models.exceptions import UserError
+
+languages = ["en", "de", "fr", "it", "rm"]
+language_label_col = ["label_en", "label_de", "label_fr", "label_it", "label_rm"]
+mandatory_properties = ["name", "object", "gui_element"]
+
+
+def _search_validation_error(
```
jnussbaum, line 19: If I read the function name, I don't understand what this function does. After reading the code, I understand that it transforms a JSON validation error into a user-friendly error message. So what do you think about something like `_prepare_err_msg_from_json_validation_error()` as function name?

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
...
+    calling function. It then searches which exact location of the property caused the error. The results are added
+    to a string which is used in the error that is raised in the calling function.
+
+    Args:
+        list_properties: List of properties
+        excel_filename: Name of the Excel file
+        validation_error: The error from the calling function
+
+    Returns:
+        A string which is used in the Error message that contains detailed information about the problem
```
jnussbaum, line 33:
```suggestion
        User-friendly version of the original JSON validation error message
```

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+        read_df: pd.DataFrame = pd.read_excel(excel_filename)
+    except ValueError:
+        # Pandas relies on openpyxl to parse XLSX files.
+        # A strange behaviour of openpyxl prevents pandas from opening files with some formatting properties
+        # (unclear which formatting properties exactly).
+        # Apparently, the excel2json test files have one of the unsupported formatting properties.
+        # Credits: https://stackoverflow.com/a/70537454/14414188
+        with mock.patch("openpyxl.styles.fonts.Font.family.max", new=100):
+            read_df = pd.read_excel(excel_filename)
+    read_df = clean_data_frame(unclean_df=read_df)
```
jnussbaum, line 38: There is a rule in Software Architecture that says "Do one thing, and do it well". It is confusing that the function `read_excel_file()` reads the Excel file and cleans it. Better: take the cleaning out, and call it separately, like:

```
# properties.py:
property_df = utl.read_excel_file(excel_filename=excelfile)
property_df = clean_data_frame(unclean_df=property_df)
```

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+
+    Args:
+        check_df: pd.DataFrame that is checked for duplicates
+        duplicate_column: Name of the column that must not contain duplicates
+
+    Raises:
+        UserError: if there are duplicates in the column
+    """
+    # This checks if there are any duplicate values in a column,
+    # pd.NA values also count as duplicates if there are several empty cells.
```
jnussbaum, line 101: This inline comment is superfluous (already mentioned in the docstring)

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+    if not required_columns.issubset(set(check_df.columns)):
+        raise UserError(
+            f"The following columns are missing in the excel: " f"{required_columns.difference(set(check_df.columns))}"
+        )
+
+
+def check_duplicate_raise_error(check_df: pd.DataFrame, duplicate_column: str) -> None:
+    """
+    This function checks if a specified column contains duplicate values. Empty cells (pd.NA) also count as duplicates.
+    If there are any duplicate values, it creates a string with the duplicates which are displayed in the error message.
```
jnussbaum, line 91: Does it make sense to raise an error if 2 cells in a column are both pd.NA? It seems counter-intuitive to me.

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+        UserError: if there are required columns missing
+    """
+    # This checks if the required columns are in the Dataframe. Other columns are also permitted.
+    if not required_columns.issubset(set(check_df.columns)):
+        raise UserError(
+            f"The following columns are missing in the excel: " f"{required_columns.difference(set(check_df.columns))}"
+        )
+
+
+def check_duplicate_raise_error(check_df: pd.DataFrame, duplicate_column: str) -> None:
```
jnussbaum, line 88: If I try to describe what this function does, I would say it checks a column for duplicates. So I would name it `check_column_for_duplicates()`

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+    """
+    This function checks if a specified column contains duplicate values. Empty cells (pd.NA) also count as duplicates.
+    If there are any duplicate values, it creates a string with the duplicates which are displayed in the error message.
+
+    Args:
+        check_df: pd.DataFrame that is checked for duplicates
+        duplicate_column: Name of the column that must not contain duplicates
```
jnussbaum, line 95: I find `duplicate_column` confusing, because the column itself is not a duplicate. What do you think about `column_to_check`?

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+    # Remove the values of all cells that do not at least contain one character of any known language
+    cleaned_df = cleaned_df.applymap(
+        lambda x: str(x).strip() if pd.notna(x) and regex.search(r"[\w\p{L}]", str(x), flags=regex.U) else pd.NA
+    )
+    # drop all the rows that are entirely empty
+    cleaned_df.dropna(axis=0, how="all", inplace=True)
+    return cleaned_df
+
+
+def check_required_columns_raise_error(check_df: pd.DataFrame, required_columns: set[str]) -> None:
```
jnussbaum, line 68: If I try to describe what this function does, I would say it checks a dataframe if it contains the required columns. So I would name it `check_that_df_contains_required_columns()`

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+    """
+    This function takes a pd.DataFrame and a set of required column names. It checks if all the columns from the set
+    are in the pd.DataFrame. Additional columns to the ones in the set are allowed. It raises an error if any columns
+    are missing.
+
+    Args:
+        check_df: pd.DataFrame that is checked
```
jnussbaum, line 75: I would rename `check_df` to `df_to_be_checked` (or similar)

BalduinLandolt: I would argue to simply call it `df`: the variable has a lifetime of only 4 lines of code and it's the first argument of a function called `check_XY`. Similarly, `required_columns` could just be `columns`, if the function is called `check_required_columns(df, columns)` it's still perfectly clear what's going on

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
...
+    # This checks if the label columns have at least one value per row
+    missing_labels = utl.find_one_full_cell_in_cols(check_df=to_check_df, required_columns=language_label_col)
+    # If everything is ok, we get None, otherwise we update the dict
+    if missing_labels is not None:
+        missing_dict.update({"label": missing_labels})
+    # Some gui_element require a gui_attributes and others must not have one
+    missing_gui_attributes = _check_gui_attributes(check_df=to_check_df)
+    if missing_gui_attributes is not None:
+        missing_dict.update(missing_gui_attributes)
+    if missing_dict != {}:
```
jnussbaum, line 285: can be shortened to `if missing_dict`

(Sometimes I'm unsure about when I can apply this short way. To test, open a Python REPL and type `bool(xyz)`. This tests if you can use the shorthand `if xyz`, because `if xyz` is a shorthand for `if bool(xyz)`. 

P.S. While typing these lines, I was unsure if I were right. But Bing AI supports my view:

Screenshot 2023-08-29 at 11 49 21

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
...
+    else:
+        final_series = no_attribute_check
+    # The boolean series is returned
+    return {"wrong gui_attributes": final_series}
+
+
+def _check_missing_values_in_row_raise_error(to_check_df: pd.DataFrame, excel_filename: str) -> None:
+    """
+    This function checks if all the required values are in the df. If all the checks are ok, the function ends without
+    any effect. If any of the checks fail, the called functions return a dictionary with the column name as key,
```
jnussbaum, line 261:
```suggestion
    any effect. If any of the checks fail, the called functions return a dictionary with the column name as key.
```

jnussbaum, line 261: If I use this function, I'm not interested in internals like this dictionary. The user of this function will never be in contact with this dict, so it's just confusing.

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
...
+    After all the checks are done, the a UserError is raised which contains the information in which column and row
+    there are problems.
```
jnussbaum, line 263: "After all the checks are done, the a UserError is raised":

- "the a" must be a typo
- this sounds as if the error was always raised after all checks are done, which is not true

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
...
+    if missing_labels is not None:
+        missing_dict.update({"label": missing_labels})
+    # Some gui_element require a gui_attributes and others must not have one
+    missing_gui_attributes = _check_gui_attributes(check_df=to_check_df)
+    if missing_gui_attributes is not None:
+        missing_dict.update(missing_gui_attributes)
+    if missing_dict != {}:
+        # Get the row numbers from the boolean series
+        missing_dict = utl.get_wrong_row_numbers(wrong_row_dict=missing_dict, true_remains=True)
+        error_str = "\n".join([f"Column Name: {k} Row Number: {v}" for k, v in missing_dict.items()])
```
jnussbaum, line 288: If you want a nicely formatted bullet point list:
```suggestion
        error_str = "\n".join([f" - Column Name: {k} Row Number: {v}" for k, v in missing_dict.items()])
```

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
...
+        raise UserError(
+            f"The file '{excel_filename}' is missing values in some rows. See below for more information:\n"
```
jnussbaum, line 290: "below" sounds like scrolling down very far. 

```suggestion
            f"The file '{excel_filename}' is missing values in some rows:\n"
```

src/dsp_tools/utils/excel_to_json/utils.py
jnussbaum: But thinking again about this topic, I wonder if we should care about import sorting at all. If you want, we can just ignore the "wrong-import-order" globally (in `pyproject.toml`), and then never care again about import sorting. In the past, I got used to thinking that sorting imports would be good practice, but I wonder if it is worth spending time with it.

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,289 @@
...
+            f"The column '{duplicate_column}' may not contain any duplicate values. "
+            f"The following values appeared multiple times '{duplicate_values}'."
+        )
+
+
+def check_required_values(check_df: pd.DataFrame, required_values_columns: list[str]) -> dict[str, pd.Series]:
+    """
+    This function takes a pd.Dataframe and a list of column names which may not contain empty cells. If there are any
+    empty cells in the column, it adds the column name and a boolean pd.Series to the dictionary. If there are no empty
+    cells, then it is not included in the dictionary. If no column has any empty cells, then it returns an empty
```
jnussbaum, line 115: A general feedback to your docstrings: I think that you should make a conceptual distinction between inline comments and docstrings:

- Inline comments explain implementation details. They are intended at other developers. 
- Docstring are 1 abstraction level higher: They are intended at users of the function. A docstring should enable the user to decide if he needs this function or not. It should explain to the outside world what tasks can be accomplished with this function.

jnussbaum (changes requested):
This PR is a big improvement of the code. I especially like the modular approach. 

However, there are some practical things, some "handwerkliche Details", that should be improved, especially the naming of the functions. In my comments, I gave some examples how the functions could be named. 

Can you in a first step improve the naming of the functions? In a second step, I could then review again. Because at the moment, i find it hard to review, because I find it difficult to understand what the functions do.

src/dsp_tools/utils/excel_to_json/properties.py
Nora-Olivia-Ammann: I understand this point but in some cases this caused bugs for me that were really hard to find as a modification in a variable in a called function also changed the variable in the calling function. I started to do this as an easy way to prevent such mistakes. Only variables which have the same value everywhere were called the same. So in the excel case I'll rename it. In the other I will have a think...

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
...
+        final_series = pd.Series(np.logical_or(mandatory_check, no_attribute_check))
+    elif mandatory_check is not None:
+        final_series = mandatory_check
+    else:
+        final_series = no_attribute_check
+    # The boolean series is returned
+    return {"wrong gui_attributes": final_series}
+
+
+def _check_missing_values_in_row_raise_error(to_check_df: pd.DataFrame, excel_filename: str) -> None:
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: What is it here that is unclear?

src/dsp_tools/utils/excel_to_json/properties.py
jnussbaum: Are you talking about unwanted variable modifications like this?

```
def hidden_list_changer(some_list: list[int]) -> None:
    some_list.append(4)
    return None

my_list = [1,2,3]
hidden_list_changer(some_list=my_list)
print(my_list)
[1, 2, 3, 4]
```

In this case, variable names don't protect you! 

Or do you mean something else?

src/dsp_tools/utils/excel_to_json/properties.py
```diff
@@ -0,0 +1,457 @@
...
+        UserError: if any of the checks are failed
+    """
+    # Every row in these columns must have a value
+    required_values = ["name", "super", "object", "gui_element"]
+    # If there are no problems, it returns an empty dict
+    missing_dict = utl.check_required_values(check_df=to_check_df, required_values_columns=required_values)
+    # This checks if the label columns have at least one value per row
+    missing_labels = utl.find_one_full_cell_in_cols(check_df=to_check_df, required_columns=language_label_col)
+    # If everything is ok, we get None, otherwise we update the dict
+    if missing_labels is not None:
```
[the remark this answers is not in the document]
jnussbaum: First: My fault, I wasn't aware of this. For most Python object, the shorthand `if missing_labels` works, that's why we often do it like this in DSP-TOOLS. But for pandas Series, it really doesn't work, so your code `if missing_labels is not None` is the correct way to do it.

Second: Sure, that's a matter of taste. For shorter expressions, I like the walrus operator, but for such long lines, I wouldn't use it.

src/dsp_tools/utils/excel_to_json/utils.py
Nora-Olivia-Ammann: Done

src/dsp_tools/utils/excel_to_json/utils.py
Nora-Olivia-Ammann: Personally I like it there because we don't have to explain where it happens. This way it is clear...

src/dsp_tools/utils/excel_to_json/utils.py
Nora-Olivia-Ammann: I like to put it in the read function, because I always want to clean the excel file... So it makes it easier. Do you think it would make sense instead to write ead_clean_excel_file? Or should we make a function that only reads it and a function that calls those two functions?

src/dsp_tools/utils/excel_to_json/utils.py
Nora-Olivia-Ammann: I have planned to look at all of the excel_to_json files, as it makes no sense to have a new util and not use them consistently. How much we want to go into detail we can discuss at a later point if that is ok with you.

src/dsp_tools/utils/excel_to_json/utils.py
jnussbaum: And actually, it's too specific to go to the docstring. So according to my own advice of yesterday, it's better to leave it in the inline-comments 😃

src/dsp_tools/utils/excel_to_json/utils.py
Nora-Olivia-Ammann: In the case of columns I understand your point. But with the df you can get bugs that are annoying to find. Because if you do a change in memory in a df then it changes it everywhere... 
Screenshot 2023-08-30 at 10 22 55

But maybe it would be worth a discussion to remove all inplace where there is no performance improvement (https://sourcery.ai/blog/pandas-inplace/)?

src/dsp_tools/utils/excel_to_json/utils.py
jnussbaum: I think renaming this function to `read_and_clean_excel_file()` would be fine. Then the caller knows what he will get, and it's not any more a blackbox that does unexpected things.

src/dsp_tools/utils/excel_to_json/utils.py
jnussbaum: Are you sure this problem is tied to variable names? I think the problem is the inplace, not how the variables are named.

src/dsp_tools/utils/excel_to_json/utils.py
Nora-Olivia-Ammann: There is not built-in feature in this function that let's us ignore nan. To solve this it requires some extra steps. But in this case there should not be any nan values in the column anyway (that is checked elsewhere) which is why I decided not to take that into account. I mentioned it in the function so that for other application we are aware of it. In that case we could still create another function that takes it into account.

src/dsp_tools/utils/excel_to_json/utils.py
Nora-Olivia-Ammann: True, you are correct. Probably I resolved it by removing the inplace in the past (it was a while ago). But generally I also tend to add information to the variables as in the past I often worked with several different dfs at the same time and it makes keyword arguments (which I prefer) clearer to me...

src/dsp_tools/utils/excel_to_json/properties.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I will definitely keep walrus operators in mind in the future.

src/dsp_tools/utils/excel_to_json/properties.py
Nora-Olivia-Ammann: Discussed below.

src/dsp_tools/utils/excel_to_json/properties.py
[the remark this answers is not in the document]
jnussbaum: "Clean code reads like well-written prose." (Robert C. Martin)

If we look at it from the caller side: When we use a function, we want to know what it does. Some examples from my suggestions in other comments:

| your name                            | my suggestion                               |
| ------------------------------------ | ------------------------------------------- |
| check_duplicate_raise_error()        | check_column_for_duplicates()               |
| _search_validation_error()           | _prepare_err_msg_from_json_validation_error |
| check_required_columns_raise_error() | check_that_df_contains_required_columns()   |

The reason I prefer my suggestions is that it's almost like telling the computer: "do that for me". So the reason I struggle with your name `_check_missing_values_in_row_raise_error()` is that I cannot translate it into a spoken command like "Hey computer, please check the missing values in the row and raise an error".

See also: https://dev.to/pksilen/27-tips-for-making-your-code-read-like-beautifully-written-prose-4aab

src/dsp_tools/utils/excel_to_json/utils.py
jnussbaum: Me too, I prefer keyword arguments 😄. 

I think the reason I dislike `check_df` is that in natural language, it sounds like a command ("Hey computer, please check this df"), which would be a good function name, but a bad variable name. That's why I prefer `df_to_be_checked`, because in natural language, it's "a dataframe that needs to be checked".

But actually, I prefer Balduin's input over my initial input: The variable only lives during 4 lines of code. And every function is a separate scope, so if so set a breakpoint here, and execution halts here, you only see the variables of this scope (=this function). So I don't see a potential to confuse it with other variables, if it's just named `df`

src/dsp_tools/utils/excel_to_json/properties.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Tip #6: Use a preposition in the function name only when necessary: You don’t need to add a preposition to a function name if the preposition can be assumed (i.e., the preposition is implicit).
?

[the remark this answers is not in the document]
jnussbaum: What is your question?

src/dsp_tools/utils/excel_to_json/utils.py
jnussbaum: 1) 
   a) I think if we prefer shortness over understandability, we don't do a favor to other developers and to our future selves :-) 
   b) For me, verbs are very much tied to function names, because functions **do** something. So, a dataframe should not have a verb in its variable name.

2) I think it really depends on the context. In a short function with only 1 dataframe, I prefer `df`. But in a complex function with multiple dataframes I prefer longer names. I don't think this is inconsistent. Because it has to make sense in the context

But now we have discussed a lot about a single variable name 😂  . I like such discussions, because we can learn something, but at some point, we should just take a decision. I leave this decision to you 😄

src/dsp_tools/utils/excel_to_json/properties.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Would this not violate this rule?

src/dsp_tools/utils/excel_to_json/utils.py
Nora-Olivia-Ammann: It is good to have the conversation now i think

src/dsp_tools/utils/excel_to_json/properties.py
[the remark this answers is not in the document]
jnussbaum: Do you mean that Tip no. 6 from dev.to violates Robert Martin's quote? These rules are not natural laws, rather they point you in a certain direction. So in the present case (discussing about `_check_missing_values_in_row_raise_error`), it's too far away from natural language. Especially, I struggle with the `_raise_error` part. If you delete that, it would already be better.

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,296 @@
...
+    # with the column name as key and a boolean series as value that contain true for every pd.NA
+    res_dict = {col: df[col].isnull() for col in required_values_columns if df[col].isnull().any()}
+    # If all the columns are filled, then it returns an empty dictionary.
+    return res_dict
+
+
+def turn_bool_array_into_index_numbers(series: pd.Series[bool], true_remains: bool = True) -> list[int]:
+    """
+    This function takes a pd.Series which only contains boolean values, by default, the index numbers of the True values
+    are extracted.
```
jnussbaum, line 132: I don't understand the first sentence of the docstring. I think it's grammatically not correct.

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,296 @@
...
+    If the parameter "true_remains" is True, then it creates a list with the index numbers of the True values.
+    If the parameter is False, then it inverses the pd.Series, and returns a list with the index numbers of
```
jnussbaum, line 134: The info "then it inverses the pd.Series" overcomplicates the docstring, because the user/caller of this function only wants to know what he will get in the end. He doesn't care if the list is inversed or not. 

If you want to inform the user that the input parameter will be silently modified, then I think that the silent modification is a problem. Then I would make a copy of the parameter before modifying it.

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,296 @@
...
+    df: pd.DataFrame,
+    substring_list: list[str],
+    substring_colname: str,
+    check_empty_colname: str,
+    must_have_value: bool,
+) -> pd.Series | None:
+    """
+    It is presumed that the column "substring_colname" has no empty cells.
+    Based on the string content of the individual rows, which is specified in the "substring_list",
+    the cell is the column "check_empty_colname" is checked whether it is empty or not.
```
jnussbaum, line 265:
```suggestion
    the cell in the column "check_empty_colname" is checked whether it is empty or not.
```

src/dsp_tools/utils/excel_to_json/utils.py
```diff
@@ -0,0 +1,296 @@
...
+    The "substring_list" contains the different possibilities regarding the content of the cell.
+    They are joined in a RegEx "|" which denotes "or".
+    If it does not match any of the sub-strings, then the RegEx returns
+    False which means that the value in the column "check_empty_colname" is not relevant.
```
jnussbaum, line 269: Docstrings should explain to users what can be expected from this function, not how the function works internally. And we always have to assume that users have only access to the docstring, but not to the code. (DaSCH is open source, but imagine a closed source project: There, the user has only the docstring, but not the code.)

For the user of this function, it can be frustrating to read about return values of regexes, because s/he is not in direct contact with these regexes. S/he is only interested in:

1. What can I feed into the function? (input)
2. What will the function give me? (output)

Everything in between is implementation detail, that should be hidden from the user!

test/unittests/test_excel_to_json/test_properties.py
```diff
@@ -0,0 +1,592 @@
+"""unit tests for excel to properties"""
+
+# pylint: disable=missing-class-docstring,missing-function-docstring,disable=protected-access,
+# disable=wrong-import-order mypy: allow_untyped_calls
```
jnussbaum, line 4:
```suggestion
# pylint: disable=missing-class-docstring,missing-function-docstring,protected-access,wrong-import-order
# mypy: allow_untyped_calls
```

Does `mypy: allow_untyped_calls` have an effect? If I remove it and run mypy, nothing happens...

test/unittests/test_excel_to_json/test_properties.py
```diff
@@ -0,0 +1,592 @@
...
+    def setUpClass(cls) -> None:
+        """Is executed once before the methods of this class are run"""
+        os.makedirs("testdata/tmp", exist_ok=True)
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        """Is executed after the methods of this class have all run through"""
+        for file in os.listdir("testdata/tmp"):
+            os.remove("testdata/tmp/" + file)
+        os.rmdir("testdata/tmp")
```
jnussbaum, line 34: These 3 lines can be simplified to `shutil.rmtree("testdata/tmp")`

jnussbaum (approved):

test/unittests/test_excel_to_json/test_properties.py
Nora-Olivia-Ammann: I'll try it out

Nora-Olivia-Ammann: Works for me too

PR #488, opened 2023-08-24
chore!: refactor id2iri: remove --outfile and --verbose flags (DEV-2576)

jnussbaum:


docs/cli-commands.md
```diff
@@ -264,10 +264,7 @@ by IRIs provided in a mapping file.
 dsp-tools id2iri xmlfile.xml mapping.json
 ```
 
-The following options are available:
-
-- `--outfile` (optional, default: `id2iri_replaced_[timestamp].xml`): path to the output file
-- `-v` | `--verbose` (optional): print more information about the progress to the console
+The output file is written to `[original name]_replaced_[timestamp].xml`.
```
jnussbaum, line 267: both flags are not necessary:
 - outfile ist not necessary to be configurable. Just write the file with a suffix and a timestamp, that's enough
 - there were no substantial things printed with the verbose flag

src/dsp_tools/utils/id_to_iri.py
```diff
@@ -1,90 +1,156 @@
-"""
-This module handles the replacement of internal IDs with their corresponding IRIs from DSP.
-"""
```
jnussbaum, line 3: This module-level docstrings is superfluous. Important are the docstrings of the functions

PR #491, opened 2023-08-28
feat(id2iri): add flag to remove created resources from XML (DEV-2571)

jnussbaum:


src/dsp_tools/utils/id_to_iri.py
```diff
@@ -207,5 +239,10 @@ def id_to_iri(
         tree=tree,
```
gNahcab, line 239: Is it on purpose that success can be overwritten by the method '_remove_resources_if_id_in_mapping' which follows?

src/dsp_tools/utils/id_to_iri.py
```diff
@@ -207,5 +239,10 @@ def id_to_iri(
...
         mapping=mapping,
     )
+    if remove_resource_if_id_in_mapping:
+        tree, success = _remove_resources_if_id_in_mapping(
+            tree=tree,
+            mapping=mapping,
+        )
     _write_output_file(orig_xml_file=xml_file_as_path, tree=tree)
     return success
```
gNahcab, line 248: Is it on purpose that success can be overwritten by the method '_remove_resources_if_id_in_mapping' which follows?

gNahcab, line 248:
```
  Returns:
        True if everything went well, False otherwise
```
description of methode 'id_to_iri' is misleading in my opinion. There is no possibility that success returns 'False', why is it in the description of the method?

gNahcab (changes requested):

src/dsp_tools/utils/id_to_iri.py
```diff
@@ -160,6 +160,36 @@ def _replace_ids_by_iris(
...
+) -> tuple[etree._Element, bool]:
+    """
+    Remove all resources from the XML file if their ID is in the mapping.
+
+    Args:
+        tree: parsed XML file
+        mapping: mapping of internal IDs to IRIs
+
+    Returns:
+        modified XML tree
```
gNahcab, line 175: returns modified XML tree and success

src/dsp_tools/utils/id_to_iri.py
jnussbaum: Good point. This is part of a overarching design pattern that I often use in DSP-TOOLS: every top-level-function (like xmlupload, create, id2iri, ...) always returns True if successful, and False if not. The file `src/dsp_tools/cli.py` is designed in this way, so I really need to return a boolean. 

Now the question is: Does it make sense to define `success = True` at the beginning of the function, and then at the end of the function return the unmodified `success` variable? Or shall I just `return True` at the end of the function?

Pros and cons of the current code:

- Con: For the current use case, there's some superfluous code that "inflates" the code. (e.g. the functions have 2 return values, namely the actual one and the success variable)
- Pro: if in the future, one of the lower-level functions introduces a "unhappy path" (where success=False), everything is already in place, and I don't have to modify the signatures of the functions (e.g. introducing a second return value)

@BalduinLandolt Have you got an opinion on this?

src/dsp_tools/utils/id_to_iri.py
jnussbaum: Done

docs/cli-commands.md
```diff
@@ -265,8 +265,19 @@ by IRIs provided in a mapping file.
 dsp-tools id2iri xmlfile.xml mapping.json
 ```
 
+The following options are available:
+
+- `-r` | `--remove-resources` (optional): remove resources if their ID is in the mapping 
+  (this prevents doubled resources on the DSP server,
+  because the resources occurring in the mapping already exist on the DSP server)
+
```
Nora-Olivia-Ammann, line 273: Maybe add here that there can't be further statements with the resource in subject position. Because to me that would not be obvious.

docs/cli-commands.md
```diff
@@ -265,8 +265,19 @@ by IRIs provided in a mapping file.
...
 The output file is written to `[original name]_replaced_[timestamp].xml`.
 
+If the flag `--remove-resources` is set,
+all resources of which the ID is in the mapping are removed from the XML file.
+This prevents doubled resources on the DSP server,
+because normally, the resources occurring in the mapping already exist on the DSP server.
+
```
Nora-Olivia-Ammann, line 280: Didn't you state this just above?

docs/incremental-xmlupload.md
```diff
@@ -85,3 +86,28 @@ dsp-tools xmlupload --incremental additional_data_replaced_[timestamp].xml
 | <center>Important</center>                                                                                                                      |
 |-------------------------------------------------------------------------------------------------------------------------------------------------|
 | Internal IDs and IRIs cannot be mixed within the same file. An XML file uploaded with the incremental option must not contain any internal IDs. |
+
+
+
+## 4. Continue an interruped xmlupload
+
+If an xmlupload didn't finish successfully, 
+some resources have already been created, but others not.
```
Nora-Olivia-Ammann, line 95: but others have not.

docs/incremental-xmlupload.md
```diff
@@ -85,3 +86,28 @@ dsp-tools xmlupload --incremental additional_data_replaced_[timestamp].xml
...
+If the remaining resources have references to created ones,
+these references must be made with IRIs,
+and the created resources must be removed from the XML file
+(otherwise they would be created a second time).
+
```
Nora-Olivia-Ammann, line 100: If the resources not uploaded to the DSP-API and reference those, which were successfully created, the original reference IDs must be replaced with the new IRIs.

docs/incremental-xmlupload.md
```diff
@@ -85,3 +86,28 @@ dsp-tools xmlupload --incremental additional_data_replaced_[timestamp].xml
...
+dsp-tools xmlupload data.xml
+# crash: some resources have been uploaded, and a id2iri_mapping_[timestamp].json file has been written
+# fix the reason for the crash
+
+# replace the IDs and remove the created resources with:
+dsp-tools id2iri data.xml --remove-resources id2iri_mapping_[timestamp].json
+
+# upload the outputted XML file with
+dsp-tools xmlupload data_replaced_[timestamp].xml
+```
```
Nora-Olivia-Ammann, line 113: I would remove the # comments and make those normal text. Because this is not stuff you type in.

src/dsp_tools/utils/id_to_iri.py
```diff
@@ -160,6 +160,36 @@ def _replace_ids_by_iris(
...
+        tree: parsed XML file
+        mapping: mapping of internal IDs to IRIs
+
+    Returns:
+        a tuple of the modified XML tree and the success status
+    """
+    success = True
+    resources = tree.xpath("|".join([f"/knora/{x}" for x in ["resource", "annotation", "link", "region"]]))
+    resources_to_remove = [x for x in resources if x.attrib.get("id") in mapping]
+    for resource in resources_to_remove:
```
Nora-Olivia-Ammann, line 180: This could also be done in a comprehension (they are faster)

src/dsp_tools/utils/id_to_iri.py
```diff
@@ -160,6 +160,36 @@ def _replace_ids_by_iris(
...
+    Remove all resources from the XML file if their ID is in the mapping.
+
+    Args:
+        tree: parsed XML file
+        mapping: mapping of internal IDs to IRIs
+
+    Returns:
+        a tuple of the modified XML tree and the success status
+    """
+    success = True
```
Nora-Olivia-Ammann, line 177: I don't understand why there is a success status if it remains the same always?

src/dsp_tools/utils/id_to_iri.py
Nora-Olivia-Ammann: To me too it is unclear. If I have a variable with a bool which you return then I expect there to be a possibility for it to change. In this case I would just return True. If we need to implement a change, creating a new variable is not too difficult.

Nora-Olivia-Ammann (changes requested):

src/dsp_tools/utils/id_to_iri.py
BalduinLandolt: This is an issue with exceptions in general: They behave like an additional, implicit and untyped return channel - something we despise in the world of functional programming 🙂   

The clean way to solve this is to either return something useful (some aggregation of error messages?) that is consumed by `cli.py` instead of throwing exceptions; or more pragmatically, remove all boolean return values that can not be false, because they are meaningless, and remove the `success` variable from `cli.py` because it can only be true as well and is therefor just as meaningless.  
Then the distinction between success and failure would be determined exclusively by wether or not an exception is thrown.

In any case, this is probably out of scope for this PR and should be addressed separately

gNahcab: Or maybe to avoid a possible bug in the future: just remove success from the methods that get called from within id_to_iri? since the latter could - if changes were made later in the first method that gets called - overwrite the success message of the first one 
something like this:
```
success = True
 xml_file_as_path, json_file_as_path = _check_input_parameters(xml_file=xml_file, json_file=json_file)
    mapping = _parse_json_file(json_file_as_path)
    tree = parse_xml_file(xml_file_as_path)
    tree = _replace_ids_by_iris(
        tree=tree,
        mapping=mapping,
    )
    if remove_resource_if_id_in_mapping:
        tree = _remove_resources_if_id_in_mapping(
            tree=tree,
            mapping=mapping,
        )
    _write_output_file(orig_xml_file=xml_file_as_path, tree=tree)
    return success
```

docs/cli-commands.md
jnussbaum: I think we speak of 2 different things: You talk about triples (subject - predicate - object) in the database, and I talk about resources that can be uploaded (=`<resource>` tags in the XML file --> a resource like "Iliad Prooem" that can be linked at with a link like https://ark.dasch.swiss/ark:/72163/1/082E/40kW9f9=SzOnQiyvhBNSqw=.20220414T072754555597Z).

In the context of an xmlupload, we don't care about what kind of triples can be in the database. We only care about `<resource>`s that are uploaded. And if "Iliad Prooem" has been uploaded already, I don't want to upload it a second time.

docs/cli-commands.md
jnussbaum: I removed it above

docs/incremental-xmlupload.md
jnussbaum: thanks. fixed

src/dsp_tools/utils/id_to_iri.py
jnussbaum: A comprehension would indeed work. I would then have 1 line that looks like this: 

```
[resource.getparent().remove(resource) for resource in resources_to_remove]
```

This is an expression that is assigned to nothing, so it invokes this pylint warning: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/expression-not-assigned.html.

Now, I could assign it to `_`:

```
_ = [resource.getparent().remove(resource) for resource in resources_to_remove]
```

That would work for both python and pylint, but I dislike it, because I think it's harder to understand.

docs/incremental-xmlupload.md
jnussbaum: Done

src/dsp_tools/utils/id_to_iri.py
jnussbaum: To clarify: `cli.py` is designed in a way that every function must return a success status. I did that (back then) because for the `create` and `xmlupload` commands, there are 3 possibilites:

- return with success=True and no errors raised: everything perfect
- return with success=False and no errors raised: some minor abnormalities occurred, but the process could finish nevertheless
- error raised: a big abnormality occurred, the process could not finish

At least for the `xmlupload` command, this distinction is important.

This means for id2iri: Currently, it has to return a success status to `cli.py`. But of course, internally, the many `success` variables that are never modified are questionable. That's why I have removed them, and only at the very end, I `return True`.

jnussbaum: And thank you @gNahcab, you noticed that in my previous code, the 2nd function overwrote the success status from the 1st function, which was a bug, of course.

gNahcab (approved):

Nora-Olivia-Ammann (approved):

PR #503, opened 2023-09-01
chore: documenation cleaning (DEV-2602)

Nora-Olivia-Ammann:


README.md
```diff
@@ -11,22 +11,24 @@
...
 
-DSP-TOOLS is a command line tool that helps you to interact with the DaSCH Service Platform API. 
+DSP-TOOLS is a command line tool that helps you to interact with the DaSCH Service Platform (DSP) API. 
 This document is intended for developers who want to work with the code of DSP-TOOLS. 
 
 | <center>Hint</center>                                                                                                                                                                |
 |:-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| This technical document was written as a guide for developers. For the end user documentation, please consult [https://docs.dasch.swiss](https://docs.dasch.swiss/latest/DSP-TOOLS). |
+| This technical document was written as a guide for developers. For the end user documentation, 
+please consult [https://docs.dasch.swiss](https://docs.dasch.swiss/latest/DSP-TOOLS). |
```
jnussbaum, line 20: This additional line break breaks the markdown table. Or was it your intention that from "please" on, it is in a new row?

If you want it like this, then I suggest writing the entire 2nd sentence in the new row.

README.md
```diff
@@ -11,22 +11,24 @@
...
+[developers documentation](https://docs.dasch.swiss/latest/DSP-TOOLS/developers/).
 
 
 
-## Quick start
+## Quick Start
 
-To get started quickly, without reading the details, just execute these commands:
+To get started quickly, without reading the details, just execute these commands (a prior installation of 
+[homebrew](https://brew.sh/) is required):
```
jnussbaum, line 31: 1. I would put the line break before the opening parentheses (semantic line breaks)
2. I would capitalize Homebrew

README.md
```diff
@@ -65,14 +67,16 @@ There are two files defining the dependencies:
...
+install it with `poetry add package`.
 
-- installs the package (incl. sub-dependencies) in your virtual environment
-- adds the package to the section `[tool.poetry.dependencies]` of `pyproject.toml`
-- adds the pinned versions of the package and all sub-dependencies to `poetry.lock`
+This
+
+- Installs the package (incl. sub-dependencies) in your virtual environment,
+- Adds the package to the section `[tool.poetry.dependencies]` of `pyproject.toml`,
+- Adds the pinned versions of the package and all sub-dependencies to `poetry.lock`.
```
jnussbaum, line 79: I just wonder: is it correct to capitalize the start of bullet points, if they are part of a sentence? It seems counter-intuitive to me to write "This Installs ...", I would write "installs". 

I just wonder because I'd like to do it correct in the future :-)

README.md
Nora-Olivia-Ammann: No i had trouble with the line length but i did not intend for that break to be displayed.

jnussbaum (changes requested):
Good PR with many valuable changes. 
See my comments for some points where I think it should be improved.

In addition, a general point to consider:

Concerning the changes in the titles in these files: 
- docs/file-formats/excel2json.md
- docs/file-formats/json-project/overview.md
- docs/file-formats/json-project/ontologies.md
- docs/file-formats/xml-data-file.md

The orthographical changes are okay, but we should closely look at the rephrasing/rewording of some of the titles.
- In some instances, you removed "in detail": Have you considered that there is a system behind these "in detail"s? (If yes, feel free to change them. If no, consider it first, and then you might change it).
- In some instances, you added "... object". See my comments that this is technically not correct in most of the cases. 

We should also consider that there are some user guides, and perhaps bookmarks on people's machine, linking to certain titles. So we shouldn't change them too often.
One manual I know of is this here: "How to archive your data at DaSCH.pdf" on https://www.dasch.swiss/our-manuals.

Every time we change titles in the docs, we must also create a new version of the manual.

jnussbaum (approved):

PR #504, opened 2023-09-01
feat: get project without credentials (DEV-841)

jnussbaum:


docs/cli-commands.md
```diff
@@ -118,6 +118,9 @@ it is necessary to specify the following options:
 dsp-tools get -s https://api.dasch.swiss -u 'your@email.com' -p 'password' -P my_project project_definition.json
 ```
 
+It is possible to get a project from a DSP server without giving credentials.
+But in this case, the users won't be retrieved, due to privacy reasons.
+
```
Nora-Olivia-Ammann, line 123: What do you mean by the users won't be retrieved? The error message is clearer to me

src/dsp_tools/utils/project_get.py
```diff
@@ -41,7 +42,10 @@ def get_project(
     """
     con = Connection(server=server, dump=dump)
     if user and password:
-        con.login(user, password)
+        try:
+            con.login(user, password)
+        except BaseError:
+            warnings.warn("WARNING: Missing or wrong credentials. You won't get data about the users of this project.")
```
Nora-Olivia-Ammann, line 48: I think this info should be mentioned in the docs.

Nora-Olivia-Ammann (changes requested):

docs/cli-commands.md
jnussbaum: Better now?

Nora-Olivia-Ammann (approved):

PR #511, opened 2023-09-08
refactor: pandas deprecates inplace parameter (DEV-2618)

Nora-Olivia-Ammann:


jnussbaum (approved):
Can you add yourself as Assignee? Apart from that, the PR is good, although I cannot explain why `test__rename_deprecated_lang_cols()` succeeded before.

PR #514, opened 2023-09-12
chore: fix gui_attributes notation (DEV-2651)

Nora-Olivia-Ammann:


docs/file-formats/json-project/ontologies.md
```diff
@@ -479,7 +479,7 @@ See the [xmlupload documentation](../xml-data-file.md#list-prop) for more inform
 
 - `List`: A GUI element for *ListValue*. A dropdown to select a list node.
     - *gui_attributes*:
-        - `hlist=<list-name>` (required): 
+        - `hlist:<list-name>` (required):
```
jnussbaum, line 482: Here we have the `<>` to indicate that "list-name" is a placeholder, and not the literal string "list-name". Would it perhaps help to do this everywhere consistently?

Like this:

```
 - `maxlength: <integer>` (optional): maximal length (number of characters accepted)
```

jnussbaum, line 482: And I would then also add a whitespace everywhere:

```suggestion
        - `hlist: <list-name>` (required): 
```

jnussbaum (approved):

PR #516, opened 2023-09-13
feat: generate UMLs with pyreverse (DEV-2661)

jnussbaum:
**After merging this PR, don't forget to merge the respective PR in the dsp-docs repo: https://github.com/dasch-swiss/dsp-docs/pull/95**

### DIFF
--- a/.github/workflows/tests-on-push.yml
+++ b/.github/workflows/tests-on-push.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
 
       # Checkout source, install Python, install dependencies
-      ######################################################
+      #######################################################
       - name: Checkout source
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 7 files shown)


Nora-Olivia-Ammann (approved):

BalduinLandolt (commented):
I'm absolutely missing context for this PR, I didn't even know we had diagrams in the DSP-TOOLS repo.  
So maybe this is not a useful comment. But I found that Graphviz for diagramming introduces lots of overhead (i.e. everything you need to do in this PR), adds an extra dependency, and generally makes things unnecessarily complicated (you can't run the docs directly anymore, you're forced to use the make commands). And for those reasons I stopped using .dot files and instead use mermaid diagrams. The mermaid code can be written directly in markdown and will be rendered by mkdocs directly without any explicit conversion (maybe a plugin is needed); plus in VSCode there are plugins that give you live preview and syntax highlighting.

PR #517, opened 2023-09-14
fix: no-op test that doesn't test anything (DEV-2655)

Nora-Olivia-Ammann:


src/dsp_tools/utils/excel2json/properties.py
```diff
@@ -293,7 +293,7 @@ def _check_missing_values_in_row_raise_error(df: pd.DataFrame, excelfile: str) -
     if missing_dict:
         # Get the row numbers from the boolean series
         missing_dict = utl.get_wrong_row_numbers(wrong_row_dict=missing_dict, true_remains=True)
-        error_str = "\n".join([f" - Column Name: {k} Row Number: {v}" for k, v in missing_dict.items()])
+        error_str = "\n".join([f"- Column Name: {k} Row Number: {v}" for k, v in missing_dict.items()])
```
jnussbaum, line 296:
```suggestion
        error_str = "\n".join([f"- Column '{k}', Row Number(s) {v}" for k, v in missing_dict.items()])
```

This would produce a much nicer output:
```
The file 'Test' is missing values in the following rows:
- Column 'name', Row Number(s) [9]
- Column 'super', Row Number(s) [2, 4, 9]
```

If I look at you output below, I find it hard to read

jnussbaum (changes requested):
Looks good 👍 

One question: If we observe this no-op problem from a meta-perspective, I think we made the mistake that we wrote a test, and because it succeeded, we thought it would test what we want. 

To avoid this, there is the general pattern Red - Green - Refactor (see here: https://www.youtube.com/watch?v=B1j6k2j2eJg). Our mistake was to skip the "Red" step.

So I wanted to ask: When you created this PR, did you also do the "Red" step? Have you tried out if these tests fail, e.g. if you add some nonsense to the regex?

Nora-Olivia-Ammann:
> Looks good 👍
> 
> One question: If we observe this no-op problem from a meta-perspective, I think we made the mistake that we wrote a test, and because it succeeded, we thought it would test what we want.
> 
> To avoid this, there is the general pattern Red - Green - Refactor (see here: https://www.youtube.com/watch?v=B1j6k2j2eJg). Our mistake was to skip the "Red" step.
> 
> So I wanted to ask: When you created this PR, did you also do the "Red" step? Have you tried out if these tests fail, e.g. if you add some nonsense to the regex?

No that is what i think went wrong.

jnussbaum (approved):

PR #518, opened 2023-09-14
fix(excel2json): don't crash if optional columns are deleted in "properties" Excel file (DEV-2652)

Nora-Olivia-Ammann:


src/dsp_tools/utils/excel2json/utils.py
```diff
@@ -284,3 +284,26 @@ def col_must_or_not_empty_based_on_other_col(
         return pd.Series(combined_array)
     else:
         return None
+
+
+def add_optional_columns(df: pd.DataFrame, optional_col_set: set[str]) -> pd.DataFrame:
+    """
+    This function takes a df and a set of columns which may not be in the df, but whose absence could cause errors in
+    the code following.
+    As a quick fix, they are added to the df.
```
jnussbaum, line 293: "Quick fix" is a term used in work organization, but it seems inappropriate in a function docstring. Because the docstring should describe what the function does, no matter if it is a quick fix or not.

src/dsp_tools/utils/excel2json/utils.py
```diff
@@ -284,3 +284,26 @@ def col_must_or_not_empty_based_on_other_col(
...
+
+    Args:
+        df: Original df
+        optional_col_set: set of columns that may not be in the df.
+        If they are not, they will be added.
```
jnussbaum, line 298: these should fit on one line. 
If it is really, really impossible to shorten it to 1 line, then indent it

src/dsp_tools/utils/excel2json/utils.py
```diff
@@ -284,3 +284,26 @@ def col_must_or_not_empty_based_on_other_col(
         return pd.Series(combined_array)
     else:
         return None
+
+
+def add_optional_columns(df: pd.DataFrame, optional_col_set: set[str]) -> pd.DataFrame:
+    """
+    This function takes a df and a set of columns which may not be in the df, but whose absence could cause errors in
+    the code following.
```
jnussbaum, line 292: semantic line breaks?

src/dsp_tools/utils/excel2json/utils.py
jnussbaum, line 293: I think it would be worth mentioning that the added columns will be empty

src/dsp_tools/utils/excel2json/utils.py
```diff
@@ -284,3 +284,26 @@ def col_must_or_not_empty_based_on_other_col(
...
+        df: Original df
+        optional_col_set: set of columns that may not be in the df.
+        If they are not, they will be added.
+
+    Returns:
+        The df with the added columns.
+        If all are already there, the df is returned unchanged.
+    """
+    in_df_cols = set(df.columns)
+    if not optional_col_set.issubset(in_df_cols):
```
jnussbaum, line 305: Cool, I just learned that dicts have this handy built-in method. Thank you 😄

test/unittests/test_excel2json/test_utils.py
```diff
@@ -179,6 +179,33 @@ def test_get_comments(self) -> None:
...
+            }
+        )
+        returned_df = utl.add_optional_columns(df=original_df, optional_col_set=optional_cols)
+        # as the columns are extracted via a set, they are not sorted and may appear in any order,
+        # this would cause the validation to fail
+        returned_df = returned_df.sort_index(axis=1)
+        assert_frame_equal(expected_df, returned_df)
+        # here all the columns exist
+        returned_df = utl.add_optional_columns(df=expected_df, optional_col_set=optional_cols)
+        assert_frame_equal(expected_df, returned_df)
```
jnussbaum, line 207:
```suggestion

        # if all columns exist, the df should be returned unchanged
        unchanged_df = utl.add_optional_columns(df=expected_df, optional_col_set=optional_cols)
        assert_frame_equal(expected_df, unchanged_df)
```

... just to make it more clear

src/dsp_tools/utils/excel2json/utils.py
Nora-Olivia-Ammann: This is not a dictionary but a set?

jnussbaum:
@Nora-Olivia-Ammann This PR is okay now, but could you improve the title + text of the linear ticket? As it is now, I only understand it because I know the context. But imagine we discuss it in the DSP-TOOLS meeting: no one would understand. 

BTW: The same applies for some other tickets you created. If you have time, could you reformulate them?

jnussbaum:
@Nora-Olivia-Ammann : IMHO, the title is not expressive enough. What do you think about "fix(excel2json): don't crash if optional columns are deleted in "properties" Excel file"

--> and the linear ticket accordingly

jnussbaum (approved):

PR #520, opened 2023-09-14
refactor(excel2json): replace excel reader in excel2resources (DEV-2668)

Nora-Olivia-Ammann:


src/dsp_tools/utils/excel2json/resources.py
```diff
@@ -119,24 +120,9 @@ def _row2resource(
...
-        p = mock.patch("openpyxl.styles.fonts.Font.family.max", new=100)
-        p.start()
-        try:
-            details_df = pd.read_excel(excelfile, sheet_name=name)
-        except ValueError as err:
-            raise BaseError(str(err)) from None
-        p.stop()
+        details_df = utl.read_and_clean_excel_file(excelfile=excelfile, sheetname=name)
+    except ValueError as err:
+        raise BaseError(str(err)) from None
```
jnussbaum, line 125: I think this isn't necessary any more. All other usages of `read_and_clean_excel_file()` use it without try-catch.

I don't know why I had added this additional try-catch inside the first try-catch: If I look at it now, it doesn't make sense.

src/dsp_tools/utils/excel2json/utils.py
```diff
@@ -12,28 +12,29 @@
...
-def read_and_clean_excel_file(excelfile: str) -> pd.DataFrame:
+def read_and_clean_excel_file(excelfile: str, sheetname: str | int = 0) -> pd.DataFrame:
     """
     This function reads an Excel file, if there is a ValueError then it patches the openpyxl part that creates the
     error and opens it with that patch.
     It cleans and then returns the pd.DataFrame.
 
     Args:
         excelfile: The name of the Excel file
+        sheetname: The name of the Excel sheet, by default it reads the first
```
jnussbaum, line 23:
```suggestion
        sheetname: The name or index (zero-based) of the Excel sheet, by default it reads the first
```

jnussbaum:
@Nora-Olivia-Ammann : The PR is good, but the ticket should have a more telling title + text

src/dsp_tools/utils/excel2json/resources.py
jnussbaum: This conversation is still unresolved. Do you agree, or would you like to leave the code as it is?

jnussbaum:
@Nora-Olivia-Ammann IMHO, the title is still not expressive enough. What do you think about "refactor(excel2json): replace excel reader in excel2resources" ?

And the linear ticket accordingly.

Nora-Olivia-Ammann:
Done

src/dsp_tools/utils/excel2json/resources.py
jnussbaum: Okay, that's fine. Thanks for the explanation, I didn't know that.

jnussbaum (approved):

PR #523, opened 2023-09-15
chore(excel2json): replace BaseError with UserError (DEV-2671)

Nora-Olivia-Ammann:


src/dsp_tools/utils/excel2json/lists.py
```diff
@@ -272,13 +272,13 @@ def validate_lists_section_with_schema(
...
     Raises:
-        BaseError: if the validation fails
+        UserError: if the validation fails
 
     Returns:
         True if the "lists" section passed validation
     """
     if bool(path_to_json_project_file) == bool(lists_section):
-        raise BaseError("Validation of the 'lists' section works only if exactly one of the two arguments is given.")
+        raise UserError("Validation of the 'lists' section works only if exactly one of the two arguments is given.")
```
jnussbaum, line 281: This should be a base error. Because it can only occur if a developer has used the function in a wrong way. It is not possible that a user will ever get this error

src/dsp_tools/utils/excel2json/lists.py
```diff
@@ -350,7 +350,7 @@ def excel2lists(
         verbose: verbose switch
 
     Raises:
-        BaseError if something went wrong
+        UserError if something went wrong
```
jnussbaum, line 353: Is it impossible to get a BaseError? If not, then we should list them both

src/dsp_tools/utils/excel2json/lists.py
jnussbaum: It doesn't need to be handled, because if the code of DSP-TOOLS was written correctly, it will never occur. It can only occur if we ourselves make a programming mistake and call this function with both arguments.

src/dsp_tools/utils/excel2json/lists.py
```diff
@@ -272,6 +272,7 @@ def validate_lists_section_with_schema(
         lists_section: the "lists" section as Python object
 
     Raises:
+        UserError: if the validation fails
         BaseError: if the validation fails
```
jnussbaum, line 276:
```suggestion
        BaseError: ifthis function is called with invalid parameters
```

src/dsp_tools/utils/excel2json/resources.py
```diff
@@ -135,7 +135,7 @@ def _row2resource(
         try:
             details_df = pd.read_excel(excelfile, sheet_name=name)
         except ValueError as err:
-            raise BaseError(str(err)) from None
+            raise UserError(str(err)) from None
```
jnussbaum, line 138: Here, it depends on how our error system should look like. My implicit assumption is:
 - we use UserError if the user can do something / fix the input data, and if the message is easy enough to be helpful for the user. No stack trace is printed, only the message.
 - we use BaseError if it is handled by the caller
 - we use BaseError if something weird happens which is not handled --> stack trace will be printed

Here we have the 3rd point. If this ever happens, the message will be too complicated for the user - and honestly even for us, because I have never found out what the problem actually is. So I was not able to fix the Excel file.

Now the question is: What should we do in the 3rd case, when something weird happens. My philosophy is: raise a BaseError, so the entire stack trace is printed. The user must get in touch with us anyway, and then we directly have the stack trace (we don't have to search in the logging file first).

What do you think about that?

jnussbaum: P.S. As I have commented here, I would remove this try-except anyways. Have you seen that comment? What are your thoughts about that?

Nora-Olivia-Ammann: In my mind, the verification did not pass because of some bad user imput, meaning the user caused the error. But if we treat the UserError as you described above then a BaseError makes sense...

jnussbaum (approved):

PR #528, opened 2023-09-25
chore(xmlupload): performance measurement (DEV-2703)

jnussbaum:


docs/developers/trace-xmlupload.md
```diff
@@ -0,0 +1,31 @@
+# Trace xmlupload
+
+## Preparation
+
+- update the dependencies in the virtual environment: `poetry install`
+- activate the virtual environment: `poetry shell`
```
Nora-Olivia-Ammann, line 6: maybe we should also mention that a stack has to be running, it's obvious if you know dsp-tools well but but since this is a step-by-step guide, it feels like it should be here too.

Nora-Olivia-Ammann:
You mentioned that we should not deploy this branch. But I disagree, we also have testing on the branch and while this is not testing, it is important to analyse code quality and may become relevant in the future. However, the Readme would have to be altered slightly and I would not include results here.

PR #533, opened 2023-09-28
fix(xmlupload): fix performance slowdown during stash applying (DEV-2717)

Nora-Olivia-Ammann:


pyproject.toml
```diff
@@ -114,6 +114,7 @@ exclude = [
     "src/dsp_tools/models/user.py",               # TODO: activate this
     "src/dsp_tools/models/value.py",              # TODO: activate this
     "src/dsp_tools/models/xmlresource.py",        # TODO: activate this
+    "test/unittests/test_xml_upload_stash.py",
```
Nora-Olivia-Ammann, line 117: I don't understand why mypy was complaining about this file, while all the other test files which are not typed either were fine. As a quick fix, I added it to the ignored files.

test/unittests/test_xml_upload_stash.py
BalduinLandolt, line 1: Test coverage is something important. But if you look at the two functions tested here, they only do a call to an external library, and one should never test third party code on ones own codebase. So all those tests do, is check that the function call is forwarded correctly. Maybe 90 lines of test code is a bit much overhead for that.

Also, I'm not a fan of "parametrising" tests, like it is commonly done in this repo. But that is a discussion for a different place than a PR review.

src/dsp_tools/utils/xml_upload_stash.py
```diff
@@ -79,26 +98,26 @@ def _get_text_hash_value(old_xmltext: str) -> str:
...
     """
-    new_xmltext = hash_to_value[pure_text]
-    # replace the outdated internal ids by their IRI
-    for _id, _iri in id2iri_mapping.items():
-        new_xmltext.regex_replace(f'href="IRI:{_id}:IRI"', f'href="{_iri}"')
-    return new_xmltext
+    for internal_id in id_set:
+        xml_with_id.regex_replace(
+            pattern=r'href="IRI:' + internal_id + r':IRI"',
+            repl='href="' + id2iri_mapping[internal_id] + '"',
```
BalduinLandolt, line 118: I would use an f-string here

src/dsp_tools/utils/xml_upload_stash.py
```diff
@@ -179,13 +198,18 @@ def upload_single_link_xml_property(
...
+        xml_from_stash = hash_to_value[text_hash_value]
+    except KeyError as err:
+        _log_iri_does_not_exist_error(received_error=err, stashed_resource=stashed_resource, all_link_props=link_prop)
         # no action necessary: this property will remain in nonapplied_xml_texts,
         # which will be handled by the caller
         return nonapplied_xml_texts
 
-    new_xmltext = _replace_internal_ids_with_iris(
-        pure_text=text_hash_value, id2iri_mapping=id2iri_mapping, hash_to_value=hash_to_value
+    id_set = xml_from_stash.find_all_substring_in_xmlstr(pattern='href="IRI:(.*?):IRI"')
```
BalduinLandolt, line 209: this is a pattern that always remains the same and is part of the "domain logic" of the application, so it might be worth extracting it into a top level constant

BalduinLandolt (approved):
Overall looks good. I added some minor remarks

test/unittests/test_xml_upload_stash.py
```diff
@@ -0,0 +1,89 @@
+# pylint: disable=missing-class-docstring,missing-function-docstring,protected-access
+
+from unittest import TestCase
+
+import pytest
+
+import dsp_tools.utils.xml_upload_stash as upld
```
jnussbaum, line 7: I'm not a fan of importing modules under a different name, because it confuses me when I read "upld" further down in the code. I normally import functions in this way:

```
from path.to.module import function1, function2, ...
```

src/dsp_tools/utils/xml_upload_stash.py
```diff
@@ -65,12 +62,34 @@ def _log_unable_to_upload_xml_resource(
     logger.warning(err_msg, exc_info=True)
 
 
+def _log_iri_does_not_exist_error(
```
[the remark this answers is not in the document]
jnussbaum: `log_unable_to_retrieve_resource` should be private as well (with _ prefixed), just to be consistent.

src/dsp_tools/utils/xml_upload_stash.py
jnussbaum, line 1: This file contains 2 times a parameter `resource_in_triplestore` typed as `Any`, and once typed as `dict[str, Any]`. Type annotations should be as exact as possible. `Any` should be avoided. `dict[str, Any]` is okay in this case, because I guess it's not possible to make it more exactly.

src/dsp_tools/utils/xml_upload_stash.py
```diff
@@ -179,13 +198,18 @@ def upload_single_link_xml_property(
...
         # no action necessary: this property will remain in nonapplied_xml_texts,
         # which will be handled by the caller
         return nonapplied_xml_texts
 
-    new_xmltext = _replace_internal_ids_with_iris(
-        pure_text=text_hash_value, id2iri_mapping=id2iri_mapping, hash_to_value=hash_to_value
+    id_set = xml_from_stash.find_all_substring_in_xmlstr(pattern='href="IRI:(.*?):IRI"')
+
+    xml_from_stash = _replace_internal_ids_with_iris(
+        id2iri_mapping=id2iri_mapping, xml_with_id=xml_from_stash, id_set=id_set
```
jnussbaum, line 212: Here again, I dislike having multiple parameters on 1 line. But in the caller, it's not as problematic as in the definition.

jnussbaum (changes requested):

src/dsp_tools/utils/xml_upload_stash.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: This function is not private because it is called up by the xmlupload.py script. I put it in this script because this is a "helper" script for that function during the code that uploads the stashed links.

poetry.lock
jnussbaum, line 1: The file poetry.lock should only be modified in one of these cases:
 a) in a version bumping PR like this one: https://github.com/dasch-swiss/dsp-tools/pull/535
 b) if you add/remove a dependency in this PR

Since this PR is neither of these 2 cases, you should revert this file.

src/dsp_tools/models/value.py
```diff
@@ -26,11 +26,15 @@ def __str__(self) -> str:
     def get_all_iris(self) -> Optional[list[str]]:
         return self.__iriregexp.findall(self.__xmlstr)
 
-    def replace(self, fromStr: str, toStr: str) -> None:
-        self.__xmlstr = self.__xmlstr.replace(fromStr, toStr)
-
-    def regex_replace(self, pattern: str, repl: str) -> None:
-        self.__xmlstr = regex.sub(pattern=repr(pattern)[1:-1], repl=repl, string=self.__xmlstr)
+    def find_all_iri_in_xmlstr(self) -> set[str]:
```
jnussbaum, line 29: This method name is misleading: It doesn't return IRIs, but IDs. I suggest something like: `find_ids_referenced_in_salsah_links()`

jnussbaum:
@Nora-Olivia-Ammann I just realized that there was already a Linear ticket for this: [DEV-2667](https://linear.app/dasch/issue/DEV-2667/). Once your PR is merged, could you close that ticket with a comment that it was resolved by this PR?

And could you also close the ticket [DEV-1981](https://linear.app/dasch/issue/DEV-1981/) with a comment that it was probably resolved by this PR? (But only after this PR is merged)

src/dsp_tools/models/value.py
```diff
@@ -23,14 +23,21 @@ def __init__(self, xmlstr: str) -> None:
     def __str__(self) -> str:
         return self.__xmlstr
 
+    def replace(self, fromStr: str, toStr: str) -> None:
```
jnussbaum, line 26: For a cleaner diff, it would be good if you move this method down to the place where it was before

src/dsp_tools/models/value.py
```diff
@@ -23,14 +23,21 @@ def __init__(self, xmlstr: str) -> None:
...
         return self.__iriregexp.findall(self.__xmlstr)
 
-    def replace(self, fromStr: str, toStr: str) -> None:
-        self.__xmlstr = self.__xmlstr.replace(fromStr, toStr)
+    def find_all_iri_in_xmlstr(self) -> set[str]:
+        return set(regex.findall(pattern='href="IRI:(.*?):IRI"', string=self.__xmlstr))
 
-    def regex_replace(self, pattern: str, repl: str) -> None:
-        self.__xmlstr = regex.sub(pattern=repr(pattern)[1:-1], repl=repl, string=self.__xmlstr)
+    def replace_one_internal_id_with_iri_in_string(self, internal_id: str, iri: str) -> None:
```
jnussbaum, line 35: The method name doesn't mention that the replacement happens only in salsah-links. I think `replace_id_with_iri_in_salsah_links()` would be clearer.

src/dsp_tools/utils/xml_upload_stash.py
```diff
@@ -79,31 +98,33 @@ def _get_text_hash_value(old_xmltext: str) -> str:
...
 def _replace_internal_ids_with_iris(
-    pure_text: str,
     id2iri_mapping: dict[str, str],
-    hash_to_value: dict[str, KnoraStandoffXml],
+    xml_with_id: KnoraStandoffXml,
+    id_set: set[str],
 ) -> KnoraStandoffXml:
     """
-    This function replaces the internal ids with the new IRIs from the triplestore.
+    This function takes an XML string and a set with internal ids that are in that string
```
jnussbaum, line 106:
```suggestion
    This function takes an XML string and a set with internal ids that are referenced in salsah-links in that string
```

src/dsp_tools/utils/xml_upload_stash.py
```diff
@@ -79,26 +98,26 @@ def _get_text_hash_value(old_xmltext: str) -> str:
...
+    It replaces all internal ids of that set with the corresponding iri according to the mapping dictionary
 
     Args:
-        pure_text: the text with the ids
-        id2iri_mapping: the dictionary that contains the mapping information
-        hash_to_value: the dictionary that contains the hash of the string and the xml value
+        id2iri_mapping: dictionary with id and iri mapping
+        xml_with_id: KnoraStandoffXml with the string that should have replacements
```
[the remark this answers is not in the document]
jnussbaum: Don't you like this suggestion?

I just realised that the variable name should have ids (plural)

src/dsp_tools/utils/xml_upload_stash.py
```diff
@@ -65,12 +62,34 @@ def _log_unable_to_upload_xml_resource(
...
+    if a linked resource does not have an IRI in the triplestore.
+
+    Args:
+        received_error: Error received
+        stashed_resource: resource that is stashed
+        all_link_props: all the link properties from that resource
+    """
+    err_msg = (
+        f"Unable to upload the xml text of '{all_link_props.name}' of resource '{stashed_resource.id}'"
+        f"because the resource with the internal id '{received_error.args[0]}' does not exist in the triplestore."
```
jnussbaum, line 81: I find the pattern try - except - log error more complicated and more verbose than the previously used pattern:

```
try:
    my_val = my_dict["foo"]
except KeyError as kerr:
    _log_key_error(kerr)

def _log_key_error(kerr: KeyError):
    logger.warning(f"The error kerr.args[0] occurred!")
```

```
if my_val not in my_dict:
    _log_that_resource_doesnt_exist("The resource doesn't exist")
my_val = my_dict["foo"]

def _log_that_resource_doesnt_exist(msg: str):
    logger.warning(msg)
```

In addition, passing a string to a function is more compliant to the Law of Demeter (=Principle of Least Knowledge) than passing a KeyError to it. Because in fact you don't need the entire KeyError, you only need the ID that was not found in the dictionary. It would be much easier to pass the failing ID as string, than passing a KeyError and then extracting the string with received_error.args[0].

Because

src/dsp_tools/utils/xml_upload_stash.py
```diff
@@ -137,11 +158,11 @@ def _create_XMLResource_json_object_to_update(
     return json.dumps(jsonobj, indent=4, separators=(",", ": "), cls=KnoraStandoffXmlEncoder)
 
 
-def upload_single_link_xml_property(
+def _upload_single_link_xml_property(
     link_prop_in_triplestore: Any,
```
jnussbaum, line 162: This type hint should be improved.

src/dsp_tools/utils/xml_upload_stash.py
```diff
@@ -79,31 +98,33 @@ def _get_text_hash_value(old_xmltext: str) -> str:
...
-    pure_text: str,
     id2iri_mapping: dict[str, str],
-    hash_to_value: dict[str, KnoraStandoffXml],
+    xml_with_id: KnoraStandoffXml,
+    id_set: set[str],
 ) -> KnoraStandoffXml:
     """
-    This function replaces the internal ids with the new IRIs from the triplestore.
+    This function takes an XML string and a set with internal ids that are referenced in salsah-links in that string
+    It replaces all internal ids of that set with the corresponding iri according to the mapping dictionary
```
jnussbaum, line 107:
```suggestion
    This function takes an XML string and a set with internal ids that are referenced in salsah-links in that string.
    It replaces all internal ids of that set with the corresponding iri according to the mapping dictionary.
```

src/dsp_tools/utils/xml_upload_stash.py
Nora-Olivia-Ammann: 1. Regarding: try - except - log error
I wrote it that way because a try, except is faster than an if clause if it occurs less than 50% of the time. Which I prioritized here since performance time is our focus here.

2. Regarding the KeyError Logging
This I can understand and can fix easily.

test/unittests/test_xml_upload_stash.py
```diff
@@ -0,0 +1,86 @@
...
+                '<resource label="r1_label" restype="r1_restype" id="r1_id" permissions="res-default">'
+                '<text-prop name=":hasRichtext"><text permissions="res-default" encoding="xml">'
+                '<a class="salsah-link" href="IRI:r2_id:IRI">r2_id</a>'
+                "</text></text-prop></resource>"
+            )
+        )
+        returned_set = one_link_KnoraStandoffXml.find_ids_referenced_in_salsah_links()
+        self.assertEqual({"r2_id"}, returned_set)
+
+    def test_find_ids_referenced_in_salsah_links_three_links(self) -> None:
```
jnussbaum, line 24: Wow, these small methods are much more readable than the parametrized ones that I'm used to. This will be a lesson for myself, because now is the first time I realize the benefits!

src/dsp_tools/models/value.py
```diff
@@ -26,11 +26,18 @@ def __str__(self) -> str:
...
 
+    def find_ids_referenced_in_salsah_links(self) -> set[str]:
+        return set(regex.findall(pattern='href="IRI:(.*?):IRI"', string=self.__xmlstr))
+
     def replace(self, fromStr: str, toStr: str) -> None:
         self.__xmlstr = self.__xmlstr.replace(fromStr, toStr)
 
-    def regex_replace(self, pattern: str, repl: str) -> None:
-        self.__xmlstr = regex.sub(pattern=repr(pattern)[1:-1], repl=repl, string=self.__xmlstr)
+    def replace_one_id_with_iri_in_salsah_link(self, internal_id: str, iri: str) -> None:
```
jnussbaum, line 35: Sorry to complain again, but it should be salsah_links (plural). 
(my fault that I see it only now)

jnussbaum (approved):

PR #541, opened 2023-10-04
refactor(xmlupload): modularise individual functions in file read_validate_xml_file.py (DEV-2767)

Nora-Olivia-Ammann:


src/dsp_tools/utils/xml_utils.py
```diff
@@ -0,0 +1,95 @@
...
+            elem.attrib["restype"] = "Region"
+            elem.tag = "resource"
+    return input_tree
+
+
+def _remove_comments_from_element_tree(
+    input_tree: etree._ElementTree[etree._Element],
+) -> etree._ElementTree[etree._Element]:
+    """
+    This function comments and processing instructions
```
BalduinLandolt, line 67:
```suggestion
    This function removes comments and processing instructions.
```
typo?

src/dsp_tools/utils/xml_utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+    Returns:
+        the root element of the parsed XML file
+    """
+
+    # remove comments and processing instructions (commented out properties break the XMLProperty constructor)
+    if isinstance(input_file, (str, Path)):
+        tree = _parse_xml_file(input_file)
+    else:
+        tree = _remove_comments_from_element_tree(input_file)
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I didn't know that. Since here it is relatively small, I will look into that.

src/dsp_tools/utils/xml_utils.py
jnussbaum, line 1: Why is this file not in the "xmlupload" subfolder? It seems to belong there

Nora-Olivia-Ammann: Well it is also used by the file id3iri, which really is not related. I consider this new file a general util as all it does is either read a xml or get an etree and then clean it. No other validation etc. is done here, so I would expect it to be universal.

src/dsp_tools/utils/xml_utils.py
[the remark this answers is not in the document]
jnussbaum: AFAIU this PR is just splitting up big functions into smaller functions. So I would argue that we don't spend time on code beautification. 

But thanks for mentioning this, it's always insightful to read your comments. Would you generally recommend to not use if-else anymore, and always prefer pattern matching?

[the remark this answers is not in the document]
Nora-Olivia-Ammann: The functions in the file will look the same regardless in which directory they are. The original function was also used by the id2iri script. I changed nothing there. The only difference is that I thought it made more sense to put this file in the general directory.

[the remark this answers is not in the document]
BalduinLandolt: It's up to you @Nora-Olivia-Ammann but I'm with @jnussbaum that we should try to move forward as efficiently as possible. And by that I mean: I will try to focus my reviews on the relevant things at hand; and if I forget myself and comment on beautification (like here) I encourage you not to get sidetracked too much 😉 

> Would you generally recommend to not use if-else anymore, and always prefer pattern matching?
I would not go that far: If you check a boolean condition, `if ... else ` is definitely the right tool for the job. But if you want to check the structure of something (e.g. here: is this object of type A, B or C?), then pattern matching is nicer for sure.

src/dsp_tools/utils/xml_utils.py
```diff
@@ -0,0 +1,95 @@
...
+            elem.tag = "resource"
+    return input_tree
+
+
+def _remove_comments_from_element_tree(
+    input_tree: etree._ElementTree[etree._Element],
+) -> etree._ElementTree[etree._Element]:
+    """
+    This function removes comments and processing instructions.
+    Commented out properties break the XMLProperty constructor
```
jnussbaum, line 70:
```suggestion
    Commented out properties break the XMLProperty constructor.
```

src/dsp_tools/utils/xml_utils.py
```diff
@@ -0,0 +1,112 @@
...
+    Args:
+        input_file: path to the XML file, or parsed ElementTree
+
+    Returns:
+        the root element of the parsed XML file
+    """
+
+    # remove comments and processing instructions (commented out properties break the XMLProperty constructor)
+
+    match input_file:
```
Nora-Olivia-Ammann, line 32: @BalduinLandolt 
The pattern matching didn't work so far. When I run it isolated it is fine. But the e2e fail.
Do you see a quick fix? Otherwise I would suggest we table this and go back to the if else. Then we can close this PR from my point of view.

jnussbaum: I was unsure about the way you implemented the SPM, so I researched it, because I was curious if it's allowed to do it the way you do it. But it seems to be correct.

So no, I don't see why it fails. I suggest going back to the if-else, so that we can move forward.

src/dsp_tools/utils/xmlupload/read_validate_xml_file.py
```diff
@@ -9,50 +8,39 @@
...
     """
-    Parse an XML file with DSP-conform data,
-    remove namespace URI from the elements' names,
-    and transform the special tags <annotation>, <region>, and <link>
-    to their technically correct form
-    <resource restype="Annotation">, <resource restype="Region">, and <resource restype="LinkObj">.
+    This function takes an element tree or a path to an XML file
+    It validates the file against the XML schema
+    It checks if all the mentioned bitstream files are in the specified location
+    It retrieves the shortcode and default ontology from the XML file
```
jnussbaum, line 26: It's really necessary to have a correct punctuation, because reStructuredText behaves like markdown: The line breaks are not rendered. So you will see a mess if you hover over this function:
Screenshot 2023-10-04 at 11 26 20

src/dsp_tools/utils/xml_utils.py
BalduinLandolt: I think it will work, if you just remove the parenthesis after the type (you want to say "in case it is a string", etc. so you don't want to instantiate a ew string; `str` is the type string, `str()` is a new object of type string), it will work

src/dsp_tools/utils/xmlupload/read_validate_xml_file.py
```diff
@@ -9,50 +8,39 @@
...
 from dsp_tools.models.xmlresource import XMLResource
 from dsp_tools.utils.create_logger import get_logger
+from dsp_tools.utils.shared import validate_xml_against_schema
+from dsp_tools.utils.xml_utils import parse_and_clean_xml_file
 
 logger = get_logger(__name__)
 
 
-def parse_xml_file(input_file: Union[str, Path, etree._ElementTree[Any]]) -> etree._Element:
+def validate_and_parse_xml_file(
```
jnussbaum, line 17: I'm a bit unhappy with the order: The correct order would be: First parse, then validate. In the code, down on lines 36-37, the order is also wrong.

When looking into it, I realized that you cannot simply inverse lines 36-37, because you inherited a mess from me: there are a bunch of functions, but the way they are organized is not intuitive.

I see a way how I could improve it, but I would need to do it myself (or in pair programming with you). It's too complicated to explain :-) 

In fact I would like to improve it now, because there is no worth splitting long functions into smaller ones, if the "Zuständigkeiten" of them are not clean. What do you think?

src/dsp_tools/utils/xml_utils.py
BalduinLandolt: or even better:
```python
match input_file:
    case str(s):
        # here `s` is strictly typed as str, rather than as a union
        do_something(s)
```

src/dsp_tools/utils/xmlupload/read_validate_xml_file.py
Nora-Olivia-Ammann: I agree with the order, it is not intuitive. But the only thing I did was use the extract method from pycharm refactor (smart copy-paste) and moved the functions around. I would prefer to close this down quickly. So it would be fine for me if you did it later.

src/dsp_tools/utils/xml_utils.py
Nora-Olivia-Ammann: Unfortunately leaving the braces away is causing a syntax error. 
Screenshot 2023-10-04 at 13 07 07

Placing the s in the variable does not fix it unfortunately, I have a feeling it has something to do that we use a type from a package that is private. So I'll reverse it so that we can finish this quickly.

PR #542, opened 2023-10-04
refactor(xmlupload): file upload_stashed_resptr_props.py (DEV-2774)

Nora-Olivia-Ammann:


src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
```diff
@@ -97,6 +74,125 @@ def upload_stashed_resptr_props(
...
+            logger.warning(err_msg, exc_info=True)
+            continue
+        if verbose:
+            print(f'  Successfully uploaded resptr-prop of "{link_prop.name}". Value: {resptr}')
+            logger.info(f'Successfully uploaded resptr-prop of "{link_prop.name}". Value: {resptr}')
+        nonapplied_resptr_props[stashed_resource][link_prop].remove(resptr)
+    return nonapplied_resptr_props
+
+
+def _unable_to_retrieve_resource(
```
jnussbaum, line 133: As already discussed, I'm not yet convinced of this approach. We have dozens of print and logging statements, every time specific for each case. Should we every time make a new function out of them? Perhaps @BalduinLandolt can give us some advice how this should best be solved. 

If you want to keep it, then I would rename the function. Functions should describe what they do (with a verb), and this one logs/prints if the resource cannot be retrieved, so I suggest `_log_if_unable_to_retrieve_resource()` or something similar.

jnussbaum, line 133: Concerning the inline comment: They explain why after this function call, the loop is continued with the `continue` keyword. So it doesn't make sense to have the inline comment in the function. I would move it up, to where the function is called.

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
```diff
@@ -97,6 +74,125 @@ def upload_stashed_resptr_props(
...
+    stashed_resource: XMLResource,
+    resptrs: list[str],
+    id2iri_mapping: dict[str, str],
+    con: Connection,
+    nonapplied_resptr_props: dict[XMLResource, dict[XMLProperty, list[str]]],
+    verbose: bool,
+) -> dict[XMLResource, dict[XMLProperty, list[str]]]:
+    """
+    This function takes one resource stashed resource and resptr-props that are specific to a property.
+    It uploads them to the triplestore and removes them from the nonapplied_resptr_props dictionary.
```
jnussbaum, line 90: We don't have access to the database (=triplestore), we always send a request to DSP-API, which in turn writes into the database.

```suggestion
    This function takes one stashed resource and the resptr-props that are specific to a property.
    It uploads them to DSP-API and removes them from the nonapplied_resptr_props dictionary.
```

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
```diff
@@ -97,6 +74,125 @@ def upload_stashed_resptr_props(
...
+
+    Args:
+        resource_in_triplestore: The resource retried from the triplestore
```
jnussbaum, line 93:
```suggestion
        resource_in_triplestore: The resource retrieved from the triplestore
```

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
```diff
@@ -97,6 +74,125 @@ def upload_stashed_resptr_props(
...
+def _upload_all_resptr_props_of_single_resource(
+    resource_in_triplestore: dict[str, Any],
+    link_prop: XMLProperty,
+    res_iri: str,
+    stashed_resource: XMLResource,
+    resptrs: list[str],
+    id2iri_mapping: dict[str, str],
+    con: Connection,
+    nonapplied_resptr_props: dict[XMLResource, dict[XMLProperty, list[str]]],
+    verbose: bool,
```
jnussbaum, line 86: I suggest reordering the parameters to make it better understandable:
```suggestion
    resource_in_triplestore: dict[str, Any],
    stashed_resource: XMLResource,
    res_iri: str,
    link_prop: XMLProperty,
    resptrs: list[str],
    id2iri_mapping: dict[str, str],
    con: Connection,
    nonapplied_resptr_props: dict[XMLResource, dict[XMLProperty, list[str]]],
    verbose: bool,
```

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
```diff
@@ -97,6 +74,125 @@ def upload_stashed_resptr_props(
...
+) -> dict[XMLResource, dict[XMLProperty, list[str]]]:
+    """
+    This function takes one resource stashed resource and resptr-props that are specific to a property.
+    It uploads them to the triplestore and removes them from the nonapplied_resptr_props dictionary.
+
+    Args:
+        resource_in_triplestore: The resource retried from the triplestore
+        link_prop: the link property object
```
jnussbaum, line 94:
```suggestion
        link_prop: the property to which the stashed resptrs belong to
```

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
Nora-Olivia-Ammann: I agree with the name change.

I know it is a bit of a weird approach. The reason, why I extract them in some cases is, because when I want to revise the code, unless I am particularly interested in the logging and so on, I really don't care to see these. And in larger functions where a lot of logging occurs it inflates the function and makes them, in my opinion, very messy and hard to read.

Nora-Olivia-Ammann: Done.

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
Nora-Olivia-Ammann: Done

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
Nora-Olivia-Ammann: Done

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
BalduinLandolt: I see both sides, here. And I don't have a solution for it either. I might go as far as to rename it to `__warn()` and I definitely would change the call
```python
            _unable_to_retrieve_resource(
                error=err,
                resource=resource,
            )
```
to
```python
            _unable_to_retrieve_resource(err, resource)
```
so it doesn't get in the way as much.

Also, I would remove the docstrings and inline comments from this method. They don't add value, and they add 11 lines of code for a method that consists of 7 lines.

I think, the underlying issue is twofold here: Our error model is not precise enough, so that it is necessary to create informative error messages manually, from the error messages of the exceptions. And secondly, the concepts of logging and user feedback in the console are still not separated properly. Why print errors to the console? One could have an error counter in the background, display a progress bar, in the end print "XYZ finished with 5 errors. See the logs for more information."  
But this is an issue for another day, and should not concern us here in this PR.

BalduinLandolt (approved):
I think it's good to merge

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
Nora-Olivia-Ammann: I made the changes.

However, I am unable to leave out the docstring because of the linters.

BalduinLandolt: Then we will have to change the linting settings. It can not be that private methods *require* docstrings. That's not linting, that's tyranny! 😉

Nora-Olivia-Ammann: I looked it up and it ignores private methods by default. I didn't know that, so in this case I was able to delete the docstring.

jnussbaum (approved):
Sorry that I blocked you, here is now my approval. In the future I will proceed as Balduin suggested in the chat.

PR #543, opened 2023-10-04
refactor(xmlupload): file xmlupload.py (DEV-2775)

Nora-Olivia-Ammann:


src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -97,38 +97,29 @@ def xmlupload(
     preparation_start = datetime.now()
 
     # establish connection to DSP server
-    con = login(server=server, user=user, password=password, dump=dump)
+    con = login(
+        server=server,
+        user=user,
+        password=password,
+        dump=dump,
+    )
```
BalduinLandolt, line 105: I liked the previous better

src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -209,6 +203,106 @@ def xmlupload(
...
+def _get_project_permissions_and_classes_from_server(
+    server_connection: Connection,
+    permissions: dict[str, XmlPermission],
+    shortcode: str,
+) -> tuple[dict[str, Permissions], dict[str, type]]:
+    """
+    This function tries to connect to the server and retrieve the project information
+    If the project is not on the server, it raises a UserError
+    From the information from the server, it creates a dictionary with the permission information
+    And a dictionary with the information about the classes
```
BalduinLandolt, line 215:
```suggestion
    This function tries to connect to the server and retrieve the project information.
    If the project is not on the server, it raises a UserError.
    From the information from the server, it creates a dictionary with the permission information
    and a dictionary with the information about the classes.
```

Nora-Olivia-Ammann:
The file is again rather large. I would suggest, to move out all the sipi stuff. This way it will also be easier to change if we move to ingest. What do you think?

src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -209,6 +203,106 @@ def xmlupload(
...
+def _extract_resources_and_permissions_from_xml(
+    root: etree._Element,
+    proj_context: ProjectContext,
+    default_ontology: str,
+) -> tuple[dict[str, XmlPermission], list[XMLResource]]:
+    """
+    This function takes the root of the tree the project context on the server and the name of the default ontology
+    From the root it separates the permissions from the resources
+    It creates corresponding Python objects of the elements
+    It returns a collection with the Python objects
```
BalduinLandolt, line 283:
```suggestion
    This function takes the root of the tree, the project context on the server and the name of the default ontology.
    From the root it separates the resource permissions.
    It returns a collection of corresponding Python objects.
```

src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -242,63 +336,39 @@ def _upload_resources(
...
+        upload_bitstream_successful, resource_bitstream = _check_for_bitstream_and_upload_if_there(
+            resource=resource,
+            sipi_server=sipi_server,
+            imgdir=imgdir,
+            filesize=filesize,
+            bitstream_size_uploaded_mb=bitstream_size_uploaded_mb,
+            permissions_lookup=permissions_lookup,
+            metrics=metrics,
+            bitstream_size_total_mb=bitstream_size_total_mb,
+            preprocessing_done=preprocessing_done,
```
BalduinLandolt, line 365: boolean parameters to a function are generally a red flag: They indicate that the function will have a top level if-else (which is the case here), so the function does two fundamentally different things. Ergo, it should be two separate functions.

src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -345,6 +417,153 @@ def _upload_resources(
...
+        preprocessing_done: If True, then no upload is necessary
+
+    Returns:
+        The information from sipi which is needed to establish a link from the resource
+    """
+    if preprocessing_done:
+        resource_bitstream = resource.get_bitstream(
+            internal_file_name_bitstream=resource.bitstream.value,  # type: ignore[union-attr]
+            permissions_lookup=permissions_lookup,
+        )
```
BalduinLandolt, line 450: I would factor that out to a function with a nice name (`_get_bitstream_information_from_sipi()`)

src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -345,6 +417,153 @@ def _upload_resources(
...
+            filepath=str(Path(imgdir) / Path(pth)),
+        )
+        bitstream_duration = datetime.now() - bitstream_start
+        bitstream_duration_ms = bitstream_duration.seconds * 1000 + int(bitstream_duration.microseconds / 1000)
+        mb_per_sec = round((filesize / bitstream_duration_ms) * 1000, 1)
+        metrics.append(
+            MetricRecord(resource.id, filetype, filesize, "bitstream upload", bitstream_duration_ms, mb_per_sec)
+        )
+
+        internal_file_name_bitstream = img["uploadedFiles"][0]["internalFilename"]  # type: ignore[index]
```
BalduinLandolt, line 466: this method with the name "upload multimedia to sipi" should only contain this code, and should only be called if preprocessing is not done

src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -345,6 +417,153 @@ def _upload_resources(
...
+        resource_bitstream = resource.get_bitstream(
+            internal_file_name_bitstream=internal_file_name_bitstream,
+            permissions_lookup=permissions_lookup,
+        )
+    return resource_bitstream
+
+
+def _check_for_bitstream_and_upload_if_there(
```
BalduinLandolt, line 474: this function feels like a detour, also it is somewhat misleading because it says that it does an upload, which is not necessarily true (when preprocessing is done), but it can still produce an error (when getting the bitstream information), which would then be handled as an "upload error"

src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -138,7 +129,10 @@ def xmlupload(
     )
 
     # temporarily remove circular references
-    resources, stashed_xml_texts, stashed_resptr_props = remove_circular_references(resources, verbose)
+    resources, stashed_xml_texts, stashed_resptr_props = remove_circular_references(
+        resources=resources,
+        verbose=verbose,
+    )
```
BalduinLandolt, line 135: again, not a fan of this

src/dsp_tools/utils/xmlupload/xmlupload.py
BalduinLandolt, line 1: I added some comments regarding bitstreams, so here I'll try to "gather my thoughts".  
In `_upload_resource()` I would do something along the lines of the following (pseudo code):

```python
def _upload_resources():
    for resource in resources:
        # ...
        if resource.bitstream:
            if not perprocessing_done:
                check_if_bitstream_there()
                # if not, hanlde
               upload_bitstream()
               # if failed, hanlde
            get_bitstream_info()
            # if none found, hanlde
            # ...
```

Like this you don't need to pass boolean flags and you can do away with all top level if-else.

BalduinLandolt (approved):
Over all it looks good. I would propose refining the bitstream handling a bit more (see my comments), and there are some marginal things I remarked. But otherwise it looks good.

BalduinLandolt:
> The file is again rather large. I would suggest, to move out all the sipi stuff. This way it will also be easier to change if we move to ingest. What do you think?

Feel free, it will certainly be good. But I would say we should eventually look at package structure in a broader way, which might make some changes of that nature obsolete.

src/dsp_tools/utils/xmlupload/xmlupload.py
Nora-Olivia-Ammann: Thanks for your suggestion. This is something I am always discussing with myself.

PR #545, opened 2023-10-05
fix(fast-xmlupload): CLI-command in documentation has outdated information (DEV-2779)

Nora-Olivia-Ammann:


jnussbaum (approved):
Thanks for spotting this 😄

PR #547, opened 2023-10-05
test(xmlupload): artificial test data with complex circular references (DEV-2780)

Nora-Olivia-Ammann:


testdata/generate_test_data/test_circles_onto.json
```diff
@@ -0,0 +1,88 @@
+{
+    "$schema": "../../src/dsp_tools/resources/schema/project.json",
+    "project": {
+        "shortcode": "4123",
+        "shortname": "systematic-tp",
```
jnussbaum, line 5: to prevent name clashes with our current e2e tests, I suggest taking another shortcode / shortname that is not used yet in any JSON file in the DSP-TOOLS repo

Nora-Olivia-Ammann: is there a shortcode you recommend?

testdata/generate_test_data/create_complex_circular_references.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+    This function creates a graph with circular references.
+    By default, it saves it in the current directory; this can be changed with the parameter save_location.
+
+    Args:
+        replication_counter: number of times the sub-graphs should be created in one root-graph.
+        save_location: path to the folder, where the file should be saved
+    """
+    root = create_circular_references_test_graph(replication_counter=replication_counter)
+    excel2xml.write_xml(root, os.path.join(save_location, f"test_circular_references_{replication_counter}.xml"))
```
jnussbaum, line 19:
```suggestion
    excel2xml.write_xml(root, Path(save_location / f"test_circular_references_{replication_counter}.xml"))
```

It is recommended to use pathlib instead of os.path. Just add `from pathlib import Path`

testdata/generate_test_data/test_circles_onto.json
jnussbaum: No, just take a hexadecimal number with four digits, and quickly search in the repo that it is not used yet.

testdata/generate_test_data/create_complex_circular_references.py
Nora-Olivia-Ammann: Now it crashes:
TypeError: unsupported operand type(s) for /: 'str' and 'str'
The IDE also tells me: Expected type 'str', got 'Path' instead

jnussbaum: Sorry, my fault: `Path(save_location) / f"test_circular_references_{replication_counter}.xml"` will work

Nora-Olivia-Ammann: Not with the current linter. Because the save function only allows str. I changed that, now it should be fine.

jnussbaum: Perfect. Thanks for adapting the type signature of the write_xml function.

testdata/generate_test_data/__init__.py
BalduinLandolt, line 1: just wondering: if we have code in the `testdata/` folder, will that be bundled with the release or is this excluded? if it gets bundled, you could consider moving this into the `test/` folder where it certainly will be excluded.

testdata/generate_test_data/test_circles_onto.json
```diff
@@ -0,0 +1,86 @@
...
+            "test",
+            "testing"
+        ],
+        "ontologies": [
+            {
+                "name": "simcir",
+                "label": "Test ontology",
+                "comment": "To test the circular references.",
+                "properties": [
+                    {Å
```
BalduinLandolt, line 20:
```suggestion
                    {
```
typo?

testdata/generate_test_data/test_circles_onto.json
```diff
@@ -0,0 +1,86 @@
...
+                                "cardinality": "0-n",
+                                "gui_order": 2
+                            }
+                        ]
+                    }
+                ]
+            }
+        ]
+    }
+}
```
BalduinLandolt, line 86: missing final newline (which is considered good practice)

I have my IDE configured so that it adds that automatically, which I would recommend you do too

testdata/generate_test_data/create_complex_circular_references.py
```diff
@@ -0,0 +1,203 @@
...
+    This function creates a graph with circular references.
+    It is capable of reproducing one graph with all the references a specified number of times.
+    Each circle type has its own sub-graph number in order to differentiate them by their ID.
+    The resulting etree is suitable for an upload to the DSP-API.
+
+    Args:
+        replication_counter: number of times the sub-graphs should be created in one root-graph.
+
+    Returns:
+        A etree which is suitable for an upload into the DSP-API
```
BalduinLandolt, line 33:
```suggestion
        An etree which is suitable for an upload into the DSP-API
```

testdata/generate_test_data/create_complex_circular_references.py
```diff
@@ -0,0 +1,203 @@
...
+    return [
+        excel2xml.make_resource(
+            restype=":TestThing", label=f"res_{let}_{replication_counter}", id=f"res_{let}_{replication_counter}"
+        )
+        for let in letters
+    ]
+
+
+def _make_salsah_link(target_res: etree._Element) -> str:
+    return f'<a class="salsah-link" href="IRI:{target_res.attrib["id"]}:IRI">{target_res.attrib["id"]}</a>'
```
BalduinLandolt, line 57: One rule I like to follow is to try to provide a function with only as much information as it really needs. Therefor I would rather pass a string `target_id` than an entire XML element, of which you only use a single attribute.

testdata/generate_test_data/create_complex_circular_references.py
```diff
@@ -0,0 +1,203 @@
...
+            link = excel2xml.make_resptr_prop(name=":hasResource", value=target_res.attrib["id"])
+        # one resource with many targets
+        case list():
+            link = excel2xml.make_resptr_prop(name=":hasResource", value=[x.attrib["id"] for x in target_res])
+    return link
+
+
+def _make_simple_text(resource: etree._Element) -> etree._Element:
+    resource.append(excel2xml.make_text_prop(name=":hasSimpleText", value="This is text."))
+    return resource
```
BalduinLandolt, line 82: if you do an in-place modification, there is really no need to return the object, because every caller already has a reference to the modified object

BalduinLandolt (approved):
Looks good to me. How would you feel about also checking in the produced data?

testdata/generate_test_data/__init__.py
jnussbaum: No, the only distributed folder is `src/dsp_tools`

testdata/generate_test_data/create_complex_circular_references.py
jnussbaum: Not only Balduin likes to follow it, but it even has a name: Law of Demeter, a.k.a. principle of least knowledge 😄

Nora-Olivia-Ammann: The reason I did this was because accessing the ID needs a bit of unpacking which I only have to do once instead of having to type this every time. But I do also see your point that this limits the usability.

testdata/generate_test_data/create_complex_circular_references.py
Nora-Olivia-Ammann: My problem here is, that if I do this in a list comprehension, I get a list back from the function which contains only None values, because the function does not return anything.

jnussbaum: I always considered it good practice to return modified objects, so that for the caller, it is easier to know that the function modified that object. Imagine this case:

```
my_object = create_object()
function_with_bad_name()  # silently modifies my_object
# my_object is now modified, but I don't know!
```

I already dislike that many functions modify the objects they receive, but then I try to always return them, to make things clearer. 

What do you think?

Nora-Olivia-Ammann:
> Looks good to me. How would you feel about also checking in the produced data?

What do you mean by check? I did a test upload, but would you like an automatic test?

BalduinLandolt:
> > Looks good to me. How would you feel about also checking in the produced data?
> 
> What do you mean by check? I did a test upload, but would you like an automatic test?

As discussed: "check in" as in "add to the the repo" :)

jnussbaum:
> Looks good to me. How would you feel about also checking in the produced data?

I'm not too enthusiastic with adding too much stuff to the repo. It feels like a duplication if we have the code + the produced files. I think the code alone is enough

PR #548, opened 2023-10-05
refactor: use sets instead of lists for stashing circular references (DEV-2771)

BalduinLandolt:


src/dsp_tools/models/xmlresource.py
```diff
@@ -109,6 +110,14 @@ def get_resptrs(self) -> list[str]:
                         resptrs.extend(value.resrefs)
         return resptrs
 
+    def get_internal_resptrs(self) -> set[str]:
+        """
+        Get a set of all resource ID that are referenced by this resource by means of a document internal ID.
```
jnussbaum, line 115:
```suggestion
        Get a set of all resource IDs that are referenced by this resource by means of an internal ID.
```

jnussbaum (approved):
I didn't look closely at the algorithm, I just trust you :-)

BalduinLandolt:
> I didn't look closely at the algorithm, I just trust you :-)

sounds risky! xD

Nora-Olivia-Ammann (approved):
Looks good to me.
What is the advantage to switch to sets in this case?

BalduinLandolt:
> Looks good to me. What is the advantage to switch to sets in this case?

In this case an insignificant performance increase (`if x in y: ...` is considerably faster with sets, and that happens a lot here; but this is not the slow part of the code, so it doesn't matter.)  
But mostly I can do `isSubsetOf()` which does away with having to check for each link, if it is in the ok set. This could be done with `all(...)` too, but less nicely. And this allowed me to get rid of the `ok` boolean which made the whole loop ugly and caused an extra step of if-nesting.

But this was mostly about getting a PR merged to warm up with the repo; and only marginally to make it easier to change stuff in a later step.

PR #550, opened 2023-10-05
chore: create composite setup action and allow more PR titles

BalduinLandolt:


.github/actions/preparation/action.yml
```diff
@@ -0,0 +1,25 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout source
+      uses: actions/checkout@v3
+      with:
+        submodules: recursive
+    - name: Install poetry
+      uses: abatilo/actions-poetry@v2
+    - name: Setup a local virtual environment (if no poetry.toml file)
```
jnussbaum, line 13:
```suggestion
    - name: Setup a local virtual environment (if no poetry.lock file)
```

Here I'm sceptical. There is always a poetry.lock file present. Is this step necessary? What does it do?

Nora-Olivia-Ammann: I don't know the set-up well enough to judge this.

.github/actions/preparation/action.yml
```diff
@@ -0,0 +1,25 @@
...
+description: "Preparation steps to run tests"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout source
+      uses: actions/checkout@v3
+      with:
+        submodules: recursive
+    - name: Install poetry
```
jnussbaum, line 11: I once ran into problems when I installed poetry before python. Because poetry will then be installed using a built-in (probably old) python version, and will keep that version "forever". And if you then install the dependencies, it will fail, because it doesn't take the python version that you want.

.github/actions/preparation/action.yml
BalduinLandolt: I copy-pasted this from the action's docs, seems you could provide a `poetry.toml` file to configure these things (which we don't do) that would override this. But this specifies that the virtual environment is created whichin the repo, so that it is accessible to the caching action.

Again, I don't really understand too much of this, I just copied it from the action's docs. But to be able to test it, we need to merge it

jnussbaum: If you want we can make YOLO and just merge it, and then see if there are problems.

But what is the purpose of this PR? I suppose you want to factor out these steps to economize time. But I argue that these steps are very fast anyways, because GitHub CI provides us with pre-existing ubuntu and python. So we could also try to repeat these steps every time.

BalduinLandolt: It is more a code-organisational question that I want to factor this out: If I make everything a separate job, I would have to repeat this too many times for my taste. So I thought I'd follow the pattern we have in the API repo instead.

But I will remove the premade poetry action and go back to what there is currently... that would be less YOLO 😉

.github/actions/preparation/action.yml
BalduinLandolt: Currently the setup is:
```
      - name: Checkout source
        uses: actions/checkout@v3
        with:
          submodules: recursive


      - name: Install poetry
        run: | 
          curl -sSL https://install.python-poetry.org | python3 -
          poetry self add poetry-exec-plugin


      - name: Set up Python 3.11
        uses: actions/setup-python@v4
        with:
          python-version: 3.11
          # attempt to restore dependencies from cache (if successful, the following `poetry install` will be a no-op)
          cache: poetry


      - name: Install Python dependencies
        run: poetry install
```

I use this now - with that I should be ok, right?

jnussbaum: yes, that should be fine.
I don't remember why I thought that python must be installed before poetry. In one repo, it really caused me problems, but apparently, it works here. So never mind my comment from before.

jnussbaum (approved):

.github/actions/preparation/action.yml
BalduinLandolt: Fingers crossed...

PR #553, opened 2023-10-05
chore(CI): split up test action into several jobs

BalduinLandolt:


.github/workflows/tests-on-push.yml
```diff
@@ -10,70 +10,145 @@ on:
...
+      # run tests
+      - name: start stack
+        run: poetry run dsp-tools start-stack --no-prune
+      - name: e2e tests
+        run: poetry run pytest test/e2e/
+      - name: stop stack  # see if this command can run (it isn't tested anywhere else)
+        run: poetry run dsp-tools stop-stack
+
+  others:
+    name: Other Tests
```
jnussbaum, line 132: Look out: these are no tests, that's the setup that is required for one of the e2e tests (for the local image processing (fast xmlupload)). 
That's why the e2 tests failed.

BalduinLandolt: thanks for the hint, clearly I wasn't paying attention

BalduinLandolt: > On line 127, we could tell pytest to execute all except `test/e2e/test_fast_xmlupload.py`. And then, you can add a job on line 171 that runs only that one.

I simply put this in front of all of them. Good enough?

jnussbaum: sure

jnussbaum (approved):

PR #557, opened 2023-10-06
refactor: create specific objects for stash information (DEV-2787)

BalduinLandolt:


src/dsp_tools/utils/xmlupload/stash/stash_models.py
```diff
@@ -49,3 +46,60 @@ def make(tups: list[tuple[XMLResource, StandoffStashItem]]) -> StandoffStash | N
...
+
+    res_id: str
+    res_type: str
+    prop_name: str
+    target_id: str
+
+
+@dataclass(frozen=True)
+class LinkValueStash:
+    """Holds information about a number of stashed link values, organized by resource instance."""
```
jnussbaum, line 63: I'm not sure if the terminology is clear enough. I often refer to this as `resptr-prop`. Could you perhaps add in the docstring that LinkValue = resptr?

BalduinLandolt: I'm reluctant to perpetuate this horrible terminology we have inherited 😉   
No, seriously, of course I can do that, but it is definitely my goal to replace as much of the terminological idiosyncrasies of dsp-tools by the terms we use in the rest of our system. So on the long run, I will try to phase out "resptr"

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
```diff
@@ -33,179 +31,100 @@ def upload_stashed_resptr_props(
         nonapplied_resptr_props: the resptr props that could not be uploaded
     """
 
+    # XXX: how to handle not found resources. currently nothing happens apart from a warning. # pylint: disable=fixme
```
jnussbaum, line 34: What's the plan with this?

BalduinLandolt: we should discuss it in the next meeting and then see how to move on; but I don't want to have the PR waiting that long (but of course I can delete the comment, if you prefer - I just thought it would make it easier to find the place in question again)

src/dsp_tools/utils/xmlupload/upload_stashed_xml_texts.py
```diff
@@ -117,6 +116,7 @@ def upload_stashed_xml_texts(
         nonapplied_xml_texts: the xml texts that could not be uploaded
     """
 
+    # XXX: how to handle not found resources. currently nothing happens apart from a warning. # pylint: disable=fixme
```
jnussbaum, line 119: what's the plan with this?

jnussbaum (approved):
Really difficult to review. If there was a bug in this PR, I wouldn't find it 🙈 . Is there a way to write tests?

BalduinLandolt:
> Really difficult to review. If there was a bug in this PR, I wouldn't find it 🙈 . Is there a way to write tests?

I know, this turned out more complicated than I hoped. But I'm actually rather confident, there were some bugs in it, and all of them were caught by the e2e tests, so I think it should be ok.  
I'm working towards making things more testable, but I'm not yet there; and I don't want to write bad tests just for the sake of it.

jnussbaum:
> > Really difficult to review. If there was a bug in this PR, I wouldn't find it 🙈 . Is there a way to write tests?
> 
> I know, this turned out more complicated than I hoped. But I'm actually rather confident, there were some bugs in it, and all of them were caught by the e2e tests, so I think it should be ok. I'm working towards making things more testable, but I'm not yet there; and I don't want to write bad tests just for the sake of it.

Good to hear, that sounds good.

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
BalduinLandolt: I'll remove it - should be easy enough to find again :)

src/dsp_tools/utils/xmlupload/stash/stash_models.py
```diff
@@ -49,3 +46,60 @@ def make(tups: list[tuple[XMLResource, StandoffStashItem]]) -> StandoffStash | N
...
+
+
+@dataclass(frozen=True)
+class LinkValueStashItem:
+    """Holds information about a single stashed link value."""
+
+    res_id: str
+    res_type: str
+    prop_name: str
+    target_id: str
```
Nora-Olivia-Ammann, line 58: The permissions are still missing right?

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
```diff
@@ -35,177 +33,97 @@ def upload_stashed_resptr_props(
...
-    )
-    return nonapplied_resptr_props
-
-
-def _upload_all_resptr_props_of_single_resource(
-    resource_in_triplestore: dict[str, Any],
-    stashed_resource: XMLResource,
-    link_prop: XMLProperty,
+        if verbose:
+            print(f'  Upload resptrs of resource "{res_id}"...')
```
Nora-Olivia-Ammann, line 50: Do you want to take this out the print statement in a future PR or could you do it now? At the moment, we do not get prints for Standoff and the console output is different for the two stashes.

Nora-Olivia-Ammann (approved):
Looks good, but would it be possible to add the permissions?

src/dsp_tools/utils/xmlupload/stash/stash_models.py
BalduinLandolt: yes, I would like to keep that in a separate task, as it would require quite a bit of work to understand how the permissions work/should work

src/dsp_tools/utils/xmlupload/upload_stashed_resptr_props.py
BalduinLandolt: I tried to harmonize it, I think in both cases now, we get prints, if verbose=True... but I may be mistaken

Nora-Olivia-Ammann: I haven't tried it out, but it'll be a quick fix in case it isn't.

PR #562, opened 2023-10-10
test(xmlupload): create artificial test data with multiple links between the same resource (DEV-2799)

Nora-Olivia-Ammann:


BalduinLandolt (approved):
This is starting to get pretty complex... but from what I can tell, looks good

testdata/generate_test_data/create_complex_circular_references.py
```diff
@@ -67,13 +71,15 @@ def _make_xml_text_prop(target_res: etree._Element | list[etree._Element]) -> et
     return excel2xml.make_text_prop(name=":hasRichtext", value=excel2xml.PropertyElement(salsah_link, encoding="xml"))
 
 
-def _make_resptr_prop(target_res: list[etree._Element] | etree._Element) -> etree._Element:
+def _make_resptr_prop(
+    target_res: list[etree._Element] | etree._Element, property_name: str = ":hasResource1"
```
jnussbaum, line 75: my personal preference is to put multiple parameters on multiple lines

testdata/generate_test_data/create_complex_circular_references.py
```diff
@@ -181,12 +187,14 @@ def _make_inverted_complex_dependencies(replication_counter: str) -> list[etree.
...
     all_resources = _make_list_of_resources(number_of_resources=5, replication_counter=replication_counter)
-    all_resources = _make_inverted_complex_dependencies_resource_ABC(all_resources)
+    all_resources = _make_single_link_inverted_complex_dependencies_resource_ABC(all_resources)
     all_resources = _make_inverted_complex_dependencies_resource_D(all_resources)
     return _make_inverted_complex_dependencies_resource_E(all_resources)
 
 
-def _make_inverted_complex_dependencies_resource_ABC(resource_list: list[etree._Element]) -> list[etree._Element]:
+def _make_single_link_inverted_complex_dependencies_resource_ABC(
+    resource_list: list[etree._Element],
```
jnussbaum, line 196: For 1 single parameter, I would also be okay to put it on 1 line (if the line is long enough)

testdata/generate_test_data/create_complex_circular_references.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I can do that

testdata/generate_test_data/create_complex_circular_references.py
Nora-Olivia-Ammann: In this case the line was too long. Black added the trailing comma

testdata/generate_test_data/create_complex_circular_references.py
Nora-Olivia-Ammann: I do like that too, if there are many or the type hints are extremely long and complex. But in this case the types are quite repetitive and the functions very short. So I do not want to end up with a function that has a longer parameter and output declaration than code. So I decided against that.

PR #563, opened 2023-10-10
test(xmlupload): create analysis for xml data (DEV-2800)

Nora-Olivia-Ammann:


src/dsp_tools/analyse_xml_data/xml_to_graph_models.py
```diff
@@ -0,0 +1,62 @@
...
+        subject_id: str,
+        object_ids: set[str],
+        edge_weight: int | float = 1,
+        reified_ede_weight: int | float = 999999999,
+    ):
+        self.subject_id = subject_id
+        self.object_ids = object_ids
+        self.reified_object_id = uuid.uuid1()
+        self.edge_weight = edge_weight
+        self.reified_ede_weight = reified_ede_weight
```
BalduinLandolt, line 57: the beauty of dataclasses is that you never write a constructor again :)  
as for the the ID, which you don't want to include, that you can define as `reified_object_id: UUID = field(init=False, default_factory=uuid.uuid4)` - the first parameter means that it's not included in the constructor and the second defines the factory function that creates the value for each new instance. (I would normally use UUID4 not 1, as it is fully random, but it doesn't matter too much)

src/dsp_tools/analyse_xml_data/xml_to_graph_models.py
```diff
@@ -0,0 +1,62 @@
...
+
+@dataclass
+class ResptrLink:
+    """
+    This class represents a link between two resources.
+    """
+
+    subject_id: str
+    object_id: str
+    edge_weight: float | int = 1
```
BalduinLandolt, line 27: could you use float everywhere to avoid union types? it will still work to pass integers, I think.

src/dsp_tools/analyse_xml_data/xml_to_graph_models.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@dataclass
+class TripleGraph:
+    """
+    Contains all classes that express links between resources.
+    """
+
+    resptr_links: Optional[list[ResptrLink]]
+    xml_links: Optional[list[XMLLink]]
```
BalduinLandolt, line 16: `Optional[A]` can be expressed as `A | None` since some versions of Python. But in many cases, lists don't need to be optional anyways, as an empty list can have the same meaning as None.

src/dsp_tools/analyse_xml_data/xml_to_graph_models.py
```diff
@@ -0,0 +1,62 @@
...
+class ResptrLink:
+    """
+    This class represents a link between two resources.
+    """
+
+    subject_id: str
+    object_id: str
+    edge_weight: float | int = 1
+
+    def to_networkX_format(self):
```
BalduinLandolt, line 29: typehints! :)

src/dsp_tools/analyse_xml_data/xml_to_graph_models.py
Nora-Olivia-Ammann: Great thanks!

src/dsp_tools/analyse_xml_data/xml_to_graph_models.py
Nora-Olivia-Ammann: I chose optional because I would not assign a value at all if there are no xml links...

jnussbaum: In that case, the parameter can have the value `[]`, which works perfect, because `bool([]) == False`. It would make type hints easier

jnussbaum: No, that will work fine: `my_list: list[str] = []` is valid.

Nora-Olivia-Ammann: Great thanks!

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -0,0 +1,169 @@
...
+        cycle = rx.digraph_find_cycle(g)  # type: ignore[attr-defined]  # pylint: disable=no-member
+        print("-" * 10)
+        print(f"cycle: {cycle}")
+        node = _find_cheapest_node(g, cycle)
+        source, targets = node
+        removed_nodes.append(UploadResource(g[source][0], targets))
+        g.remove_node(source)
+        print(f"removed link: {node}")
+        leaf_nodes = _remove_leaf_nodes(g)
+        removed_nodes.extend(leaf_nodes)
```
Nora-Olivia-Ammann, line 143: @BalduinLandolt  Is it relevant that the leaf nodes are documented? Because they do not influence the size of the stash that has to be uploaded and is therefore not relevant to the end performance.

BalduinLandolt: as discussed: this is to keep track of when in the process which resource can be uploaded because it no longer has unresolvable dependencies

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -0,0 +1,185 @@
...
+        if xml:
+            xml_instances.extend(xml)
+        if all_links:
+            all_link_ids.extend(all_links)
+    return resptr_instances, xml_instances, set(all_link_ids)
+
+
+def _create_classes_single_resource(
+    resource: etree._Element,
+) -> tuple[list[ResptrLink] | None, list[XMLLink] | None, list[str] | None]:
```
jnussbaum, line 30:
```suggestion
) -> tuple[list[ResptrLink], list[XMLLink], list[str]]:
```

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+def analyse_circles_in_data(xml_filepath: str, tracer_output: str) -> None:
+    """
+    This function takes an XML filepath
+    It analyzes how many and which links have to be removed
+    so that all circular references are broken up.
+
+    Args:
+        xml_filepath: path to the file
```
jnussbaum, line 158: The argument list is incomplete. (Since you mention this argument in the body, it's probably not necessary to repeat it here).

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -0,0 +1,185 @@
...
+        source, targets = node
+        removed_nodes.append(UploadResource(g[source][0], targets))
+        g.remove_node(source)
+        print(f"removed link: {node}")
+        leaf_nodes = _remove_leaf_nodes(g)
+        removed_nodes.extend(leaf_nodes)
+    return removed_nodes
+
+
+def analyse_circles_in_data(xml_filepath: str, tracer_output: str) -> None:
```
jnussbaum, line 151:
```suggestion
def analyse_circles_in_data(xml_filepath: str, tracer_output_file: str) -> None:
```

and even better: create a pathlib.Path from the output file right in the beginning, and then pass it around in the program as Path.

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -0,0 +1,185 @@
...
+        if resptr:
+            resptr_instances.extend(resptr)
+        if xml:
+            xml_instances.extend(xml)
+        if all_links:
+            all_link_ids.extend(all_links)
+    return resptr_instances, xml_instances, set(all_link_ids)
+
+
+def _create_classes_single_resource(
```
jnussbaum, line 28:
```suggestion
def _create_classes_from_single_resource(
```

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -0,0 +1,185 @@
...
+    if resptr_links:
+        all_used_ids.extend(resptr_links)
+        resptr_link_objects = [ResptrLink(subject_id, object_id) for object_id in resptr_links]
+    if xml_links:
+        all_used_ids.extend(chain.from_iterable(xml_links))
+        xml_link_objects = [XMLLink(subject_id, x) for x in xml_links]
+    return resptr_link_objects, xml_link_objects, all_used_ids
+
+
+def _get_all_links_one_resource(resource: etree._Element) -> tuple[list[str], list[set[str]]]:
```
jnussbaum, line 47:
```suggestion
def _get_all_links_from_one_resource(resource: etree._Element) -> tuple[list[str], list[set[str]]]:
```

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -0,0 +1,185 @@
...
+                xml_links.extend(_extract_id_one_text_prop(prop))
+    return resptr_links, xml_links
+
+
+def _extract_id_one_resptr_prop(resptr_prop: etree._Element) -> list[str]:
+    return [x.text for x in resptr_prop.getchildren() if x.text]
+
+
+def _extract_id_one_text_prop(text_prop: etree._Element) -> list[set[str]]:
+    # the same ID is in several separate <text> in one <text-prop> are considered separate links
```
jnussbaum, line 64:
```suggestion
    # if the same ID is in several separate <text> values of one <text-prop>, they are considered separate links
```

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -0,0 +1,185 @@
...
+            case "{https://dasch.swiss/schema}text-prop":
+                xml_links.extend(_extract_id_one_text_prop(prop))
+    return resptr_links, xml_links
+
+
+def _extract_id_one_resptr_prop(resptr_prop: etree._Element) -> list[str]:
+    return [x.text for x in resptr_prop.getchildren() if x.text]
+
+
+def _extract_id_one_text_prop(text_prop: etree._Element) -> list[set[str]]:
```
jnussbaum, line 63:
```suggestion
def _extract_ids_from_text_prop(text_prop: etree._Element) -> list[set[str]]:
```

BalduinLandolt (approved):

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -0,0 +1,186 @@
...
+    # the same ID is in several separate <text> in one <text-prop> are considered separate links
+    xml_props = []
+    for text in text_prop.getchildren():
+        links = _extract_id_one_text(text)
+        if links:
+            xml_props.append(links)
+    return xml_props
+
+
+def _extract_id_one_text(text: etree._Element) -> set[str]:
```
jnussbaum, line 73:
```suggestion
def _extract_ids_from_one_text_value(text: etree._Element) -> set[str]:
```

src/dsp_tools/analyse_xml_data/models_xml_to_graph.py
```diff
@@ -0,0 +1,37 @@
...
+class ResptrLink:
+    """This class represents a link between two resources."""
+
+    subject_id: str
+    object_id: str
+
+
+@dataclass
+class XMLLink:
+    """This class represents a link between a resource and an XMl text which contains links to other resources."""
```
jnussbaum, line 16: I don't fully understand this. The variable names `subject_id` and `object_link_ids` suggest that this class represents 1 resource that contains 1 xml-text with several links to other resources.

If my assumption is correct, I would rephrase the docstring: `This class represents a resource with an XML text which contains links to other resources.`

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
```diff
@@ -0,0 +1,142 @@
...
+        '<resource xmlns="https://dasch.swiss/schema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" '
+        'label="res_A_19" restype=":TestThing" id="res_A_19" permissions="res-default"><resptr-prop '
+        'name=":hasResource1"><resptr permissions="prop-default">res_B_19</resptr><resptr '
+        'permissions="prop-default">res_C_19</resptr></resptr-prop><text-prop name=":hasRichtext"><text '
+        'permissions="prop-default" encoding="xml"><a class="salsah-link" href="IRI:res_B_19:IRI">res_B_19</a><a '
+        'class="salsah-link" href="IRI:res_C_19:IRI">res_C_19</a></text></text-prop></resource>'
+    )
+    res_resptr_links, res_xml_links, res_all_used_ids = _create_classes_from_single_resource(test_ele)
+    #########
+    res_B_19 = [obj.object_id for obj in res_resptr_links]  # type: ignore[union-attr]
```
jnussbaum, line 27: You could remove all 3 ignores of this function, if `_create_classes_from_single_resource()` doesn't return `tuple[None, None, None]` (as I suggested above).

jnussbaum: Plus, you can then make this test function more concise: `assert "res_B_19" in [obj.object_id for obj in res_resptr_links]`

Like this, you have less lines, and you can remove the ugly `#####` lines

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
```diff
@@ -0,0 +1,142 @@
...
+    res_C_19 = [obj.object_id for obj in res_resptr_links]  # type: ignore[union-attr]
+    assert "res_C_19" in res_C_19
+    #########
+    expected_ids = ["res_B_19", "res_C_19", "res_B_19", "res_C_19", "res_A_19"]
+    assert expected_ids == unordered(res_all_used_ids)
+    ########
+    expected_xml = {"subject_id": "res_A_19", "object_link_ids": {"res_B_19", "res_C_19"}}
+    res_xml_dict = res_xml_links[0].__dict__  # type: ignore[index]
+    for k, v in expected_xml.items():
+        assert v == res_xml_dict[k]
```
jnussbaum, line 38: I believe that you can achieve what you want in an easier way:

```suggestion
    assert res_xml_links[0].subject_id == "res_A_19"
    assert res_xml_links[0].object_link_ids == {"res_B_19", "res_C_19"}
```

The entire thing with the dictionary, and then iterating over `k, v` of it is too complicated.

Plus, using a dunder method like `__dict__` should be avoided.

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
```diff
@@ -0,0 +1,142 @@
...
+        '<resource xmlns="https://dasch.swiss/schema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" '
+        'label="res_A_11" restype=":TestThing" id="res_A_11" permissions="res-default"><text-prop '
+        'name=":hasRichtext"><text permissions="prop-default" encoding="xml"><a class="salsah-link" '
+        'href="IRI:res_B_11:IRI">res_B_11</a></text></text-prop><resptr-prop name=":hasResource1"><resptr '
+        'permissions="prop-default">res_B_11</resptr></resptr-prop></resource>'
+    )
+    res_resptr, res_xml = _get_all_links_from_one_resource(test_ele)
+    expected_resptr, expected_xml = ["res_B_11"], [{"res_B_11"}]
+    assert expected_resptr == res_resptr
+    assert expected_xml == unordered(res_xml)
```
jnussbaum, line 52: This can be done easier:

```suggestion
    assert res_resptr == ["res_B_11"]
    assert unordered(res_xml) == [{"res_B_11"}]
```

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def test_get_all_links_from_one_resource_no_links() -> None:
+    test_ele = etree.fromstring(
+        '<resource label="res_B_18" restype=":TestThing" id="res_B_18" permissions="res-default"/>'
+    )
+    res = _get_all_links_from_one_resource(test_ele)
+    assert ([], []) == res
```
jnussbaum, line 60: I normally assert the other way round: first what I actually have, then what I expect

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
```diff
@@ -0,0 +1,142 @@
...
+def test_text_only_get_all_links_from_one_resource() -> None:
+    test_ele = etree.fromstring(
+        '<resource xmlns="https://dasch.swiss/schema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" '
+        'label="res_C_18" restype=":TestThing" id="res_C_18" permissions="res-default"><text-prop '
+        'name=":hasRichtext"><text permissions="prop-default" encoding="xml"><a class="salsah-link" '
+        'href="IRI:res_A_18:IRI">res_A_18</a></text><text permissions="prop-default" encoding="xml"><a '
+        'class="salsah-link" href="IRI:res_B_18:IRI">res_B_18</a></text></text-prop></resource>'
+    )
+    res_resptr, res_xml = _get_all_links_from_one_resource(test_ele)
+    expected_xml = [{"res_A_18"}, {"res_B_18"}]
```
jnussbaum, line 72: Is it worth having a variable for something as short as this?

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
```diff
@@ -0,0 +1,142 @@
...
+    assert expected_xml == unordered(res_xml)
+
+
+def test_extract_id_one_text_with_one_id() -> None:
+    test_ele = etree.fromstring(
+        '<text permissions="prop-default" encoding="xml"><a class="salsah-link" '
+        'href="IRI:res_A_11:IRI">res_A_11</a></text>'
+    )
+    res = _extract_ids_from_one_text_value(test_ele)
+    expected = {"res_A_11"}
```
jnussbaum, line 83: Is it worth having a variable for something as short as this?

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def test_extract_id_one_text_with_several_id() -> None:
+    test_ele = etree.fromstring(
+        '<text permissions="prop-default" encoding="xml"><a class="salsah-link" '
+        'href="IRI:res_A_11:IRI">res_A_11</a><a class="salsah-link" href="IRI:res_B_11:IRI">res_A_11</a><a '
+        'class="salsah-link" href="IRI:res_B_11:IRI">res_A_11</a></text>'
+    )
+    res = _extract_ids_from_one_text_value(test_ele)
+    expected = {"res_A_11", "res_B_11"}
```
jnussbaum, line 103: Is it worth having a variable for something as short as this?

(several more occurrences are further down, I won't comment them any more)

jnussbaum (commented):
See my comments

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
Nora-Olivia-Ammann: how would you test it? The result is a class instance.

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
Nora-Olivia-Ammann: i do it consistently expected, result. But it seems that pytest prefers it the other way around

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -0,0 +1,228 @@
...
+    print("=" * 80)
+    print("Start time:", start)
+    print("End time:", datetime.now())
+    return resource_upload_order
+
+
+if __name__ == "__main__":
+    analyse_circles_in_data(
+        xml_filepath=Path("testdata/xml-data/circular-references/test_circular_references_1.xml"),
+        tracer_output_file="circular_references_tracer.json",
```
jnussbaum, line 227: Could/should be a path, too

src/dsp_tools/analyse_xml_data/sgv.json
jnussbaum, line 1: What's the purpose of this file?

t.json
jnussbaum, line 1: What's the purpose of this file?

PR #566, opened 2023-10-12
chore: integrate DSP-INGEST in start-stack

BalduinLandolt:
@jnussbaum and/or @Nora-Olivia-Ammann could you check if the workflow with fast-xmlupload and ingest works with that.  
I can say for sure that both images have access to the images; what I'm not so sure about is if all the plumbing is quite correct. But I wasn't sure how to trigger this workflow, so I thought I'd leave this up to you :)


Nora-Olivia-Ammann:
I do not know the compose file at all. Maybe @jnussbaum is a better judge.

jnussbaum (approved):

PR #568, opened 2023-10-13
refactor: modularize xmlupload (DEV-2813)

BalduinLandolt:


Nora-Olivia-Ammann (approved):
I haven't looked at it in extreme detail but it looks good to me. Please tell me if you would like a closer inspection.

BalduinLandolt:
> I haven't looked at it in extreme detail but it looks good to me. Please tell me if you would like a closer inspection.

I'm pretty confident that I haven't broken anything (says he, right after breaking something and having to fix it...). The most important thing to me would be to know if the steps that are now being done in the big `xmlupload` function, are logical and reasonably well divided into sub-steps

Nora-Olivia-Ammann:
> > I haven't looked at it in extreme detail but it looks good to me. Please tell me if you would like a closer inspection.
> 
> I'm pretty confident that I haven't broken anything (says he, right after breaking something and having to fix it...). The most important thing to me would be to know if the steps that are now being done in the big `xmlupload` function, are logical and reasonably well divided into sub-steps

I will play it trough in the code this afternoon.

src/dsp_tools/utils/xmlupload/upload_config.py
```diff
@@ -0,0 +1,64 @@
...
+class UploadConfig:
+    """Configuration for the upload process."""
+
+    verbose: bool = False
+    dump: bool = False
+    save_metrics: bool = False
+    preprocessing_done: bool = False
+    server_as_foldername: str = field(default="unknown")
+    save_location: Path = field(default_factory=lambda: Path.home() / Path(".dsp-tools") / "xmluploads")
+    timestamp_str: str = field(default_factory=lambda: datetime.now().strftime("%Y-%m-%d_%H%M%S"))
```
jnussbaum, line 47: Do you really need a factory for these two, or would a normal `default` value work, too?

src/dsp_tools/utils/xmlupload/write_diagnostic_info.py
```diff
@@ -3,117 +3,54 @@
...
+from typing import Any
 
 import pandas as pd
-import regex
 from lxml import etree
 
 from dsp_tools.utils.create_logger import get_logger
+from dsp_tools.utils.xmlupload.upload_config import UploadConfig
 
 MetricRecord = namedtuple("MetricRecord", ["res_id", "filetype", "filesize_mb", "event", "duration_ms", "mb_per_sec"])
```
jnussbaum, line 15: @Nora-Olivia-Ammann @BalduinLandolt Not the topic of this PR, but just FYI: We once added some metrics to the xmlupload, which I guess are now redundant, because you are going to implement better measurements. The metrics we implemented back then were made "by hand", i.e. there are a lot of lines all over the code doing time measurements. If you want, you can remove them.

src/dsp_tools/utils/xmlupload/write_diagnostic_info.py
```diff
@@ -3,117 +3,54 @@
...
-        success = False
-    if metrics:
-        os.makedirs("metrics", exist_ok=True)
-        df = pd.DataFrame(metrics)
-        df.to_csv(f"metrics/{metrics_filename}")
-        print(f"Total time of xmlupload: {sum(int(record.duration_ms) for record in metrics) / 1000:.1f} seconds")
 
-    return success
+
+def write_metrics(
```
jnussbaum, line 38: @Nora-Olivia-Ammann @BalduinLandolt Not the topic of this PR, but just FYI: We once added some metrics to the xmlupload, which I guess are now redundant, because you are going to implement better measurements. The metrics we implemented back then were made "by hand", i.e. there are a lot of lines all over the code doing time measurements. If you want, you can remove them.

BalduinLandolt: good to know, probably makes sense to do that, but I will keep that for a separate task, I think

src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -81,53 +72,87 @@ def xmlupload(
...
-
-    permissions = _extract_permissions_from_xml(root, proj_context)
-    resources = _extract_resources_from_xml(root, default_ontology)
-
-    permissions_lookup, resclass_name_2_type = _get_project_permissions_and_classes_from_server(
-        server_connection=con,
-        permissions=permissions,
-        shortcode=shortcode,
+    resources, permissions_lookup, resclass_name_2_type, stash = _prepare_upload(
+        root, con, default_ontology, shortcode, config.verbose, metrics
```
jnussbaum, line 92: I always prefer keyword arguments, because it makes the program more robust against involuntary changes, e.g. if you move the order of the parameters in the function signature

src/dsp_tools/utils/xmlupload/upload_config.py
BalduinLandolt: I think you're right. for the date, the semantics of it will change ever so slightly, as it will capture the date and time of loading this class to memory, rather than the date and time of instantiating the class. But that doesn't matter here, right? It's just about having the same timestamp everywhere

jnussbaum: I wasn't aware of this subtle (but important) difference between `default` and `default_factory`. But you're right, in our case that doesn't matter.

BalduinLandolt: the may reason for default-factory is e.g. with lists: if you use `default=[]` then all instances of the dataclass will share the same list, which is practically never what you want; on the other hand, `default_factory=list` will be called once for each instance, and hence each instance gets its own list

jnussbaum (approved):

BalduinLandolt:
> > > I haven't looked at it in extreme detail but it looks good to me. Please tell me if you would like a closer inspection.
> > 
> > 
> > I'm pretty confident that I haven't broken anything (says he, right after breaking something and having to fix it...). The most important thing to me would be to know if the steps that are now being done in the big `xmlupload` function, are logical and reasonably well divided into sub-steps
> 
> I will play it trough in the code this afternoon.

I'll merge it as it is for now, but we can still improve on it, if there is something you don't like. Just let me know!

PR #569, opened 2023-10-14
continue network examination

BalduinLandolt:


src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -23,85 +25,145 @@ def create_classes_from_root(root: etree._Element) -> list[ResptrLink] and list[
 
 def _create_classes_single_resource(
     resource: etree._Element,
-) -> list[ResptrLink] | None and list[XMLLink] | None and list[str] | None:
+) -> tuple[list[ResptrLink] | None, list[XMLLink] | None, list[str] | None]:
     subject_id = resource.attrib.get("id")
+    if not subject_id:
```
Nora-Olivia-Ammann, line 30: this feels redundant, because we assume that a valid XML according to our xsd is uploaded. There the attribute "id" is mandatory. Instead of checking it everytime, I would  build a validation according to our xsd into the code.

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -23,85 +25,145 @@ def create_classes_from_root(root: etree._Element) -> list[ResptrLink] and list[
...
     return resptr_links, xml_links
 
 
 def _extract_id_one_resptr_prop(resptr_prop: etree._Element) -> list[str]:
-    return [x.text for x in resptr_prop.getchildren()]
+    return [x.text for x in resptr_prop.getchildren() if x.text]
 
 
-def _extract_id_one_text_prop(text_prop: etree._Element) -> list[set[str]] | None:
+def _extract_id_one_text_prop(text_prop: etree._Element) -> list[set[str]]:
```
Nora-Olivia-Ammann, line 61: if there are no text prop there would be no sets in the lists. So should this not be a problem with the typing?

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -23,85 +25,145 @@ def create_classes_from_root(root: etree._Element) -> list[ResptrLink] and list[
...
+            if searched:
+                all_links.add(searched.group(1))
+    return all_links
+
+
+def _make_graph(
+    resptr_instances: list[ResptrLink], xml_instances: list[XMLLink], all_link_ids: set[str]
+) -> rx.PyDiGraph:  # type: ignore[type-arg] # pylint: disable=no-member
+    g: rx.PyDiGraph = rx.PyDiGraph()  # type: ignore[type-arg] # pylint: disable=no-member
+    nodes = [(id_, None, None) for id_ in all_link_ids]
```
Nora-Olivia-Ammann, line 86: believe in python it is convention to write _id

src/dsp_tools/analyse_xml_data/models_xml_to_graph.py
```diff
@@ -1,46 +1,37 @@
...
 
     @property
-    def cost_links(self):
+    def cost_links(self) -> float:
+        """The cost of this outgoing link (1 / number of links in the XML text)"""
         return 1 / len(self.object_link_ids)
+
+
+@dataclass
+class UploadResource:
```
Nora-Olivia-Ammann, line 28: The name is a bit unclear to me when I read it in the code as at this point we are far away from actually uploading it.

Nora-Olivia-Ammann (approved):

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
jnussbaum: Type hints for containers (here: the list) tell you what kind of item will be in the container. There can also be 0 of those items in the container, without any problems

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
jnussbaum: I don't know the exact context here, but normally, we don't only **assume** that the XML is valid - we **know** that it has been validated against the XSD.

PR #570, opened 2023-10-14
refactor: introduce abstraction layer for connection (DEV-2820)

BalduinLandolt:


Nora-Olivia-Ammann (approved):

src/dsp_tools/connection/__init__.py
jnussbaum, line 1: Should `connection` really be a top-level package? I would rather make it a subpackage of `models`. Because connection is not one of the big functionalities that we offer (like `excel2xml`, which rightly is a top-level package).

src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -85,7 +85,7 @@ def xmlupload(
     metrics: list[MetricRecord] = []
 
     # establish connection to DSP server
-    con = login(server=server, user=user, password=password, dump=config.dump)
+    con: Connection = login(server=server, user=user, password=password, dump=config.dump)
```
jnussbaum, line 88: unnecessary type hint: function returns Connection

test/e2e/test_group.py
```diff
@@ -18,7 +19,7 @@ def setUp(self) -> None:
         Creates a connection to DSP-API.
         For each test method, a new TestCase instance is created, so setUp() is executed before each test method.
         """
-        self.con = Connection(server="http://0.0.0.0:3333")
+        self.con: Connection = ConnectionLive(server="http://0.0.0.0:3333")
```
jnussbaum, line 22: just out of curiosity: why do you type that as Connection? In the other tests you don't.

jnussbaum (approved):
Some remarks from my side, but it looks good.

test/e2e/test_listnode.py
```diff
@@ -3,7 +3,8 @@
 
 import pytest
 
-from dsp_tools.models.connection import Connection
+from dsp_tools.connection.connection import Connection
```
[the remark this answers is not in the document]
BalduinLandolt: That's the general idea behind using interfaces: you tell your application that it has an interface, not an implementation. Admittedly, for test code this doesn't matter too much - but I would prefer to type stuff as the interface wherever possible, including all test code.

test/e2e/test_group.py
BalduinLandolt: just an experiment I forgot to remove :)

src/dsp_tools/connection/__init__.py
BalduinLandolt: Models is completely overloaded at the moment so I don't want to add anything there for now. But we should soon talk about the over-arching package structure :)

PR #573, opened 2023-10-17
refactor(xmlupload): optimize stash links (DEV-2847)

Nora-Olivia-Ammann:


src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -6,62 +6,65 @@
...
+                resptr_links.extend(_create_class_instance_resptr_link(subject_id, prop))
             case "{https://dasch.swiss/schema}text-prop":
-                xml_links.extend(_extract_ids_from_text_prop(prop))
+                xml_links.extend(_create_class_instance_text_prop(subject_id, prop))
     return resptr_links, xml_links
 
 
-def _extract_ids_from_one_resptr_prop(resptr_prop: etree._Element) -> list[str]:
-    return [x.text for x in resptr_prop.getchildren() if x.text]
+def _create_class_instance_resptr_link(subject_id: str, resptr_prop: etree._Element) -> list[ResptrLink]:
```
BalduinLandolt, line 49: not a fan of this function name. maybe `_create_resptr_link` or `_create_resptr_link_object`?

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -6,62 +6,65 @@
...
+    for resptr in resptr_prop.getchildren():
+        if r_text := resptr.text:
+            instance = ResptrLink(subject_id, r_text)
+            resptr.attrib["stashUUID"] = instance.link_uuid
+            resptr_links.append(instance)
+    return resptr_links
 
 
-def _extract_ids_from_text_prop(text_prop: etree._Element) -> list[set[str]]:
+def _create_class_instance_text_prop(subject_id: str, text_prop: etree._Element) -> list[XMLLink]:
```
BalduinLandolt, line 60: as above

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -93,94 +101,128 @@ def _make_graph(
...
-    removed_target_ids: list[str] = [node_index_lookup[x[1]] for x in edges_out]
-    return cheapest_node, removed_target_ids
+        costs.append((source, target, node_value, edges_out))
+    cheapest_nodes = sorted(costs, key=lambda x: x[2])[0]
+    cheapest_links = [x for x in edge_list if x[0] == cheapest_nodes[0] and x[1] == cheapest_nodes[1]]
+    return cheapest_links
 
 
-def _generate_upload_order(
+def _remove_edges_get_removed_class_instances(
```
BalduinLandolt, line 177: how about `_stash_links`? or `_remove_edges_to_stash`?

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -93,94 +101,128 @@ def _make_graph(
...
+    phantom_links = []
+    for instance in links_to_stash:
+        if isinstance(instance, XMLLink):
+            phantom_links.extend(_find_remove_phantom_xml_edges(source, target, edge_list, instance, remaining_nodes))
+    to_remove_list.extend(phantom_links)
+    g.remove_edges_from(to_remove_list)
+    return ResourceStashInfo(node_index_lookup[source], links_to_stash)
+
+
+def _find_remove_phantom_xml_edges(
```
BalduinLandolt, line 212: I would say "find" or "remove" but not both

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -197,32 +240,32 @@ def analyse_circles_in_data(xml_filepath: Path, tracer_output_file: str) -> list
     print("=" * 80)
     tracer = VizTracer(
         minimize_memory=True,
-        ignore_c_function=True,
-        ignore_frozen=True,
-        include_files=["extract_links_from_XML.py", "models.py"],
+        max_stack_depth=3,
     )
```
BalduinLandolt, line 322: I think for now we have VizTracer only in the dev dependencies, so this might break when we ship it with only the prod dependencies. I think either we leave it in and add it to the dependencies, or we remove it entirely.

src/dsp_tools/utils/xmlupload/create_upload_order_stash_circles.py
BalduinLandolt, line 1: can be removed?

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
```diff
@@ -102,28 +103,54 @@ def test_extract_ids_from_text_prop_with_several_text_links() -> None:
...
+        b'<?xml version="1.0" encoding="UTF-8"?><knora xmlns="https://dasch.swiss/schema" '
+        b'xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://dasch.swiss/schema '
+        b'https://raw.githubusercontent.com/dasch-swiss/dsp-tools/main/src/dsp_tools/resources/schema/data.xsd" '
+        b'shortcode="0700" default-ontology="simcir"><resource label="res_A_11" restype=":TestThing" id="res_A_11" '
+        b'permissions="res-default"><resptr-prop name=":hasResource1"><resptr '
+        b'permissions="prop-default">res_B_11</resptr></resptr-prop></resource><resource label="res_B_11" '
+        b'restype=":TestThing" id="res_B_11" permissions="res-default"><text-prop name=":hasRichtext"><text '
+        b'permissions="prop-default" encoding="xml">Start text<a class="salsah-link" '
+        b'href="IRI:res_C_11:IRI">res_C_11</a>end text.</text></text-prop></resource><resource label="res_C_11" '
+        b'restype=":TestThing" id="res_C_11" permissions="res-default"></resource></knora>'
```
BalduinLandolt, line 144: any reason you don't use multiline (triple-quote) strings?

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
```diff
@@ -102,28 +103,54 @@ def test_extract_ids_from_text_prop_with_several_text_links() -> None:
...
         '<resptr-prop xmlns="https://dasch.swiss/schema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" '
         'name=":hasResource1"><resptr permissions="prop-default">res_A_13</resptr><resptr '
         'permissions="prop-default">res_B_13</resptr><resptr permissions="prop-default">res_C_13</resptr></resptr-prop>'
     )
-    res = _extract_ids_from_one_resptr_prop(test_ele)
-    assert unordered(res) == ["res_A_13", "res_B_13", "res_C_13"]
+    res = _create_class_instance_resptr_link("res_D_13", test_ele)
+    assert res[0].object_id == "res_A_13"
+    assert res[1].object_id == "res_B_13"
+    assert res[2].object_id == "res_C_13"
```
BalduinLandolt, line 130:
```suggestion
    assert res == ["res_A_13", "res_B_13",  "res_C_13"]
```

BalduinLandolt (commented):
Functionally looks good to me, thanks!  
I added some "stylistic" remarks, feel free to apply or ignore them as you see fit.

But I'd like to re-review, once you're done with the modifications to the models.

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -6,62 +6,67 @@
...
-            resptr_instances.extend(resptr)
-        if xml:
-            xml_instances.extend(xml)
-        if subject_id:
-            all_resource_ids.add(subject_id)
-    return resptr_instances, xml_instances, set(all_resource_ids)
+        all_resource_ids.append(subject_id)
+        resptr_instances.extend(resptr)
+        xml_instances.extend(xml)
+    return root, resptr_instances, xml_instances, all_resource_ids
```
BalduinLandolt, line 24: why do you return the root here? In my logic, you should never have to return the thing you passed in

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -93,136 +103,239 @@ def _make_graph(
...
 
     Args:
         g: graph
         node_index_lookup: reference between graph indices and original id
+        edge_list: list of edges in the graph as tuple (source_node, target_node, Class Instance)
+        node_indices: index numbers of the nodes still in the graph
 
     Returns:
         List of instances that contain the information of the resource id and its links.
+        The number of links in the stash.
```
BalduinLandolt, line 275: One of the three return values is not described

BalduinLandolt (approved):
Some more details, but otherwise it looks good to me

src/dsp_tools/analyse_xml_data/models.py
```diff
@@ -1,17 +1,24 @@
...
 
 
-@dataclass
+@dataclass(frozen=True)
 class ResptrLink:
     """This class represents a link between two resources."""
 
     subject_id: str
     object_id: str
+    link_uuid: str = field(default_factory=lambda: str(uuid.uuid4()))
```
jnussbaum, line 13: can you add in the docstring what this field means?

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -6,62 +6,67 @@
...
 
-from dsp_tools.analyse_xml_data.models import ResptrLink, UploadResource, XMLLink
+from dsp_tools.analyse_xml_data.models import ResptrLink, XMLLink
 
 
-def _create_info_from_xml_for_graph(root: etree._Element) -> tuple[list[ResptrLink], list[XMLLink], set[str]]:
+def _create_info_from_xml_for_graph(
+    root: etree._Element,
+) -> tuple[etree._Element, list[ResptrLink], list[XMLLink], list[str]]:
     """Create instances of the classes ResptrLink and XMLLink from the root of the XML file."""
```
jnussbaum, line 15: You return 4 things, but only 2 of them are described here. 
 1. Can you reduce the number of things you return? 
 2. Can you describe everything that you return?

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
Nora-Olivia-Ammann: In triplequote string the new lines are actually encoded newlines. In this case I would prefer not to encode any tabs or linebreaks, so I prefer a single-line string

src/dsp_tools/analyse_xml_data/models.py
```diff
@@ -1,17 +1,24 @@
...
+    def cost_links(self) -> float:
+        """The cost of this outgoing is consistently 1"""
+        return 1
 
 
-@dataclass
+@dataclass(frozen=True)
 class XMLLink:
     """
     This class represents a link between a resource and an XML text
```
jnussbaum, line 24: This docstring is a bit cryptic:
 - actually, the link is not between the resource and the XML text
 - The 2nd line is grammatically out of place

src/dsp_tools/analyse_xml_data/models.py
```diff
@@ -20,21 +27,9 @@ class XMLLink:
 
     subject_id: str
     object_link_ids: set[str]
+    link_uuid: str = field(default_factory=lambda: str(uuid.uuid4()))
```
jnussbaum, line 30: could you document this field?

Nora-Olivia-Ammann:
> But I'd like to re-review, once you're done with the modifications to the models.

Thanks!
The models though will not be redone. I removed the UploadResource model as the information about the stashes and upload order are now represented by a list and a dictionary respectively. In my opinion this separates the information semantically better and the results are easier to test.

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -93,136 +103,239 @@ def _make_graph(
...
-        ignore_c_function=True,
-        ignore_frozen=True,
-        include_files=["extract_links_from_XML.py", "models.py"],
+        max_stack_depth=3,
     )
     tracer.start()
     tree = etree.parse(xml_filepath)
     root = tree.getroot()
-    resptr_instances, xml_instances, all_resource_ids = _create_info_from_xml_for_graph(root)
+    root, resptr_instances, xml_instances, all_resource_ids = _create_info_from_xml_for_graph(root)
```
jnussbaum, line 319: Perhaps, `resptr_links` and `xml_links` would be better variable names

--> and in `_make_graph()`, too

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
BalduinLandolt: ah, no, my bad, overlooked that... in that case it's ok. The most thorough thing to do would be to test for the entire class objects, but if you think that's overkill, then leave it as it is now

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -93,136 +103,239 @@ def _make_graph(
     """
```
jnussbaum, line 103: The docstring is a bit misleading:

```
xml_instances: Instances of links to texts containing links to other resources
```

The links are not TO texts, but IN texts

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -93,136 +103,239 @@ def _make_graph(
...
     for xml in xml_instances:
-        xml_edges.extend(
-            [(node_id_lookup[xml.subject_id], node_id_lookup[x], xml.cost_links) for x in xml.object_link_ids]
-        )
-    g.add_edges_from(xml_edges)
-    print(f"number of xml edges: {len(xml_edges)}")
-    return g, node_index_lookup
+        edges.extend([(node_id_lookup[xml.subject_id], node_id_lookup[x], xml) for x in xml.object_link_ids])
+    g.add_edges_from(edges)
+    return g, node_index_lookup, edges, set(node_indices)
```
jnussbaum, line 116: 4 return values are quite a lot. Could you reduce the number?

If not, could you add the missing ones to the docstring?

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
```diff
@@ -6,62 +6,65 @@
...
-    if xml_links:
-        xml_link_objects = [XMLLink(subject_id, x) for x in xml_links]
-    return resptr_link_objects, xml_link_objects, subject_id
-
-
-def _get_all_links_from_one_resource(resource: etree._Element) -> tuple[list[str], list[set[str]]]:
-    resptr_links: list[str] = []
-    xml_links: list[set[str]] = []
+    resptr_links, xml_links = _get_all_links_from_one_resource(subject_id, resource)
+    return resptr_links, xml_links, subject_id
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I did this to make the function more concise, but I can integrate the two if both of you prefer it.

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
BalduinLandolt: That seems weird... if you do in-place modification, it should also apply to the caller side. Sorry I insist on that... but I take some offence by this pattern, because from the method signature it suggests that the passed object will not be modified (but a modified copy gets returned - by convention that's the reason to return something like that), so to me this is misleading behaviour.

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
BalduinLandolt: there is a simple way of returning fewer things: group them into an object. :)  
I know to some people this feels like adding overhead - but for me as a functional programmer, this adds type safety, which is a good thing.

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
Nora-Olivia-Ammann: I am happy with either. @jnussbaum what do you think?

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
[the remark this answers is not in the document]
BalduinLandolt: (no idea why I wrote this comment in German?! ^^)  
I don't feel strongly about it, but it just seems like adding 3 lines of method declaration and 1 line of return, for just 2 lines of actual functionality (which could even be reduced to 1 line if you inline `resource.attrib["id"]`, though I wouldn't vote for that) seems a bit much.

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
Nora-Olivia-Ammann: True.

test/unittests/test_analyse_xml_data/test_extract_links_from_XML.py
Nora-Olivia-Ammann: I did it like this because I wanted to test only a certain component.

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
jnussbaum: I thought this function won't be shipped anyway? But generally, I don't think that VizTracer should be in production code.

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
Nora-Olivia-Ammann: Weirdest thing. I tested it again with root and here it works. If I check for the UUID attribute in a element for example in test_create_info_from_xml_for_graph_from_one_resource. Then the UUID attribute is not there. But in root (I did not check separately) it works. Do you understand why?

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
Nora-Olivia-Ammann: The root is now removed, so that reduces it to 3. For me in this case it feels like overkill. I create an object at this point to return it. And after it is returned I immediately unpack it again...

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
Nora-Olivia-Ammann: Well if we look at it how it will be in the triplestore, I disagree. I found it very awkward to describe these things as the representation in the graph is not the same as in the triplestore. But still we have to understand how the XMLLinks differ from ResptrLinks when we work with the rx graph. At this point would you focus on how it is in the rx represenation or what it represents in the triplestore?

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
BalduinLandolt: I have no idea... but if I had to guess, maybe if you are on the iterator, modifications do not get applied there right away?

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
BalduinLandolt: Technically there is no link relationship between the starting resource and its text values; the text values are related to their containing resource by a property, but for text values this is not a link property but a regular value property pointing to a TextValue.  
And although this may seem like hair-splitting: DSP-TOOLS does not know anything about the existence of a triplestore, it only knows of an arbitrary API which we expose, and which may or may not directly correspond to the model and technology we use to persist the data behind this API. And, in fact, this is a striking place of where this is not the case: The API exposes XML, but in the triplestore we persist the markup as standoff.  
And on top of that, even "resptr" is a concept that is exclusive to the DSP-TOOLS upload XML and is used nowhere else in our system.  
I think it's important to keep these distinct layers separated in our minds.

As for this docstring: This function is where the XML model is turned in the Graph model, so the goal should probably be to make clear how the XML maps to the graph model.

Here is a suggestion:

```
    Args:
        resptr_links: A list of objects representing a direct link (resptr) between a starting resource and a target resource
        xml_links: A list of objects representing one or more links from a single starting resource to a set of target resources, where all targed resources are linked to from a single text value on the starting resource.
        all_resource_ids: IDs of all the resources in the graph
```

I'm sure this has room for improvement (and, admittedly, it's a mouth full!), but I think it captures the inherent logic and stresses the differences between the two kinds of link-objects at hand.

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
BalduinLandolt: How about moving this one function over to the testdata package, unlike the rest of this file, it's really only for us and doesn't need to be shipped. But that doesn't mean we have to change the function or even remove it; that just means it may want to live with the rest of the non-production code of the repo

Nora-Olivia-Ammann: I also renamed the file, because it does not only extract the links but also analyses them.

src/dsp_tools/analyse_xml_data/extract_links_from_XML.py
Nora-Olivia-Ammann: Thanks, I am happy for your input, as I said I was unsure myself how describe it.

PR #574, opened 2023-10-17
refactor: modularize xmlupload method (DEV-2836)

BalduinLandolt:


src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -352,93 +332,81 @@ def _upload_resources(
...
-            print(f"WARNING: Unable to create resource '{resource.label}' ({resource.id}): {err_msg}")
-            log_msg = (
-                f"Unable to create resource '{resource.label}' ({resource.id})\n"
-                f"Resource details:\n{vars(resource)}\n"
-                f"Property details:\n" + "\n".join([str(vars(prop)) for prop in resource.properties])
-            )
-            logger.warning(log_msg, exc_info=True)
+            if not bitstream_information:
+                # Note: previously, if the bitstream could not be uploaded, the resource was skipped without adding to
+                # the failed_uploads list, which I'm not sure was as intended
```
BalduinLandolt, line 352: @jnussbaum what do you think?

BalduinLandolt:
This turned out in a way that looks complicated in the diff, but really, it's just moving stuff into functions and deleting the metrics

Nora-Olivia-Ammann (approved):
Looks good to me.

src/dsp_tools/utils/xmlupload/xmlupload.py
jnussbaum: No, that was not intended. Thanks that you caught this!

BalduinLandolt: Ok, in that case I'll leave it as it is now and just remove the comment. Thanks for the clarification

jnussbaum: But now I realize that the old lines 390-392 (https://github.com/dasch-swiss/dsp-tools/blob/03d7fca22b7d30f232bae5f09e5a93439420f675/src/dsp_tools/utils/xmlupload/xmlupload.py#L390) had a wrong indentation! Did you notice that, and correct it in your PR?

jnussbaum: If I see correctly, the resource is silently skipped if the file cannot be uploaded. I don't think that's ideal...

PR #580, opened 2023-10-18
DON'T MERGE: measure stash applying with viztracer

jnussbaum:
Here is the used XML file and the result: https://drive.google.com/drive/folders/1XEgukeRSkDKzSFha-H7tuiwRht5CSABp?usp=sharing

### DIFF
--- a/src/dsp_tools/utils/xmlupload/generate-test-data-for-viztracer.py
+++ b/src/dsp_tools/utils/xmlupload/generate-test-data-for-viztracer.py
@@ -1,0 +1,49 @@
+from enum import Enum
+from pathlib import Path
+
+from lxml import etree
+
+from dsp_tools.excel2xml import (
+    PropertyElement,
+    make_boolean_prop,
+    make_resource,
+    make_resptr_prop,
+    make_root,
+    make_text_prop,
+    write_xml,
+)
+
+
+class __Letter(Enum):
+    A = 0
+    B = 1
+    C = 2
+
+
+def __make_circle(resource_counter: int) -> list[etree._Element]:
+    return [__make_one_resource(resource_counter=resource_counter, letter=x) for x in __Letter]
+
+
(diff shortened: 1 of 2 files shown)


jnussbaum:
@Nora-Olivia-Ammann @BalduinLandolt Unfortunately I cannot run this with `circles-10000.xml` (30000 very small resources, without images, just a lot of links). Because every time I try it, after a while the `db` container behaves strangely. This time, it was exactly after the 30000th resource was uploaded, when DSP-API began to respond every request with `org.knora.webapi.store.triplestore.errors.TriplestoreResponseException: Triplestore responded with HTTP code 500`, later with `org.knora.webapi.store.triplestore.errors.TriplestoreConnectionException: Failed to connect to triplestore`.

The stats and logs of the `db` container are not helpful: `db` uses 411.3MB / 23.45GB RAM, and at the time where the error first occurred, it logged

```
2023-10-19 12:39:47 org.apache.jena.fuseki.servlets.ActionErrorException: prepareCommit
2023-10-19 12:39:47     at org.apache.jena.fuseki.servlets.ServletOps.errorOccurred(ServletOps.java:275) ~[fuseki-server.jar:4.9.0]
...
```

The resources I allotted to Docker are quite generous: 6 CPUs, 24 GB RAM, 1GB Swap, 64 GB Virtual Disk Limit.

But what I actually wanted to ask: Do we still need this PR/branch, or are we now certain enough that there is no problem with the stash applying mechanism?

BalduinLandolt:
@jnussbaum thanks for the info! That's somewhat concerning to hear, but as long as we never observe that in a real-world scenario, we shouldn't worry too much about it.  
However, this perfectly illustrates why Ivan and I thought to run these kinds of tests with a mock API instead of in an integrated setting. You just eliminate a lot more sources of potential error.

Realistically, we probably don't need this anymore; but if you feel like a challenge, you could after all try to run such a test with a mocked connection.

PR #581, opened 2023-10-18
refactor: create JSON-LD context without API request (DEV-2845)

BalduinLandolt:


jnussbaum (approved):
Very good!

Nora-Olivia-Ammann (approved):
I am also happy, because it will unlikely to be a problem in the future. But we should not forget about the limitations for the future.

PR #582, opened 2023-10-20
test(xmlupload): logic of link removal in stash links (DEV-2854)

Nora-Olivia-Ammann:


test/unittests/test_analyse_xml_data/test_construct_and_analyze_graph.py
```diff
@@ -1,15 +1,26 @@
-# pylint: disable=missing-class-docstring,missing-function-docstring,protected-access
+# pylint: disable=missing-class-docstring,missing-function-docstring,protected-access,no-member
+# mypy: disable-error-code="var-annotated,assignment,arg-type"
```
BalduinLandolt, line 2: how many of those can you get rid of, if you don't use `patch`?

BalduinLandolt (approved):
Apart from the thing we discussed, looks good! If you can get rid of the `patch`, all the better

test/unittests/test_analyse_xml_data/test_construct_and_analyze_graph.py
Nora-Olivia-Ammann: None, the last pylint is everytime we use rx instances. The other ones I don't remember exactly but they are not about the patch.

PR #591, opened 2023-10-25
refactor: use class Edge instead of complicated type annotation

jnussbaum:
The complicated type annotation `list[tuple[int, int, XMLLink | ResptrLink]]` is replaced by a new model: `class Edge` in `models.py`

### DIFF
--- a/src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
+++ b/src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
@@ -1,13 +1,12 @@
 from __future__ import annotations
 
-from collections import namedtuple
 from typing import Any
 
 import regex
 import rustworkx as rx
 from lxml import etree
 
-from dsp_tools.analyse_xml_data.models import Edge, ResptrLink, XMLLink
+from dsp_tools.analyse_xml_data.models import Cost, Edge, ResptrLink, XMLLink
 
 
 def create_info_from_xml_for_graph(
@@ -167,7 +166,6 @@ def _find_cheapest_outgoing_links(
     Returns:
         The edges (i.e. links) that should be stashed (containing all the edges connecting the two nodes)
     """
-    Cost = namedtuple("Cost", ["source", "target", "node_value", "edges_out"])
     costs: list[Cost] = []
     for source, target in cycle:
         edges_in = graph.in_edges(source)
@@ -175,7 +173,7 @@ def _find_cheapest_outgoing_links(
         edges_out = graph.out_edges(source)
         node_cost = sum(x[2].cost_links for x in edges_out)
         node_value = node_cost / node_gain
-        costs.append(Cost(source, target, node_value, edges_out))
+        costs.append(Cost(source, target, node_value))
     cheapest_cost = sorted(costs, key=lambda x: x.node_value)[0]
     cheapest_links = [
         x for x in edges if x.source_rx_index == cheapest_cost.source and x.target_rx_index == cheapest_cost.target
--- a/src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
+++ b/src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
+from collections import namedtuple
 from typing import Any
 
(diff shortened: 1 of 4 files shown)


src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
```diff
@@ -112,12 +113,12 @@ def make_graph(
     node_indices = list(graph.add_nodes_from(nodes))
     id_to_rustworkx_index = dict(zip(all_resource_ids, node_indices))
     rustworkx_index_to_id = dict(zip(node_indices, all_resource_ids))
```
BalduinLandolt, line 115: again, please remove "rustworkx" from variables.

Also, I would strongly encourage you to use shorter variable names, if the variables only live for 10 lines of code. There is no need for them to be so long, and it will make the code far more readable.  
I would suggest `index_lookup` and `id_lookup`

src/dsp_tools/analyse_xml_data/models.py
```diff
@@ -45,3 +45,23 @@ class XMLLink:
...
+
+@dataclass(frozen=True)
+class Edge:
+    """
+    This class represents an edge in the rustworkx graph.
+
+    Attributes:
+        source_rx_index: rustworkx index of the resource from which the link originates
+        target_rx_index: rustworkx index of the resource where the link points to
+        link_object: the link that connects the source with the target
```
BalduinLandolt, line 58: this seems like over-documentation

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
```diff
@@ -152,38 +153,40 @@ def _remove_leaf_nodes(
...
-        edge_list: list of all the edges that were in the original graph
+        edges: edges in the graph (contains info about source node, target node, and link info)
 
     Returns:
-        - A list with the links that should be stashed.
-        - It contains all the edges connecting the two nodes.
+        The edges (i.e. links) that should be stashed (containing all the edges connecting the two nodes)
     """
-    costs = []
+    Cost = namedtuple("Cost", ["source", "target", "node_value", "edges_out"])
```
BalduinLandolt, line 170: I'm not a fan of mixing named tuples and dataclasses. In this case, does this really add value, here?

BalduinLandolt (approved):
I generally like the changes. How confident are you that this does not have a negative performance impact for large datasets?

jnussbaum:
> I generally like the changes. How confident are you that this does not have a negative performance impact for large datasets?

How could it have a negative impact? I didn't think about it, but I cannot think of a way how it impacts performance negatively.

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
jnussbaum: The added value comes some lines further down, where I can refer to the names of the `Cost` object, instead of indexing it:

Now:

```
... if x.source_rx_index == cheapest_cost.source and x.target_rx_index == cheapest_cost.target
```

Without namedtuple:

```
... if x.source_rx_index == cheapest_cost[0] and x.target_rx_index == cheapest_cost[1]
```

A dataclass would introduce too much overhead, but I wanted to have a one-liner

BalduinLandolt: I actually liked the old one better, there. I don't think we should go so far as to completely avoid index access - and with the named arguments, it actually doesn't fit on one line anymore.

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
jnussbaum: done

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
jnussbaum: It's now a dataclass (defined outside of the function, which improves performance)

BalduinLandolt: Interesting... did not expect that to make it faster. But all the better

PR #592, opened 2023-10-25
refactor: refactor _remove_edges_to_stash() and _find_phantom_xml_edges()

jnussbaum:


src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
```diff
@@ -299,13 +301,13 @@ def generate_upload_order(
...
         stash_counter += len(links_to_remove)
-        links_to_stash = _remove_edges_to_stash(
+        _remove_edges_to_stash(
             graph=graph,
             edges_to_remove=links_to_remove,
             all_edges=edges,
             remaining_nodes=remaining_node_indices,
         )
-        stash_lookup = _add_stash_to_lookup_dict(stash_lookup, links_to_stash)
+        stash_lookup = _add_stash_to_lookup_dict(stash_lookup, [x.link_object for x in links_to_remove])
```
BalduinLandolt, line 310: I preferred the previous version: both more readable and most likely faster

test/unittests/test_analyse_xml_data/test_construct_and_analyze_graph.py
```diff
@@ -384,8 +384,7 @@ def test_remove_edges_to_stash_phantom_xml() -> None:
         Edge(2, 5, c_bdf_xml),
     ]
     graph.add_edges_from([e.as_tuple() for e in edges])
-    res_links_to_stash = _remove_edges_to_stash(graph, edges_to_remove, edges, remaining_nodes)
-    assert res_links_to_stash == [c_bdf_xml]
+    _remove_edges_to_stash(graph, edges_to_remove, edges, remaining_nodes)
```
BalduinLandolt, line 387: also here, the previous implementation was more explicit and probably easier to debug

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
```diff
@@ -188,73 +188,75 @@ def _remove_edges_to_stash(
...
-        the links that should be stashed
     """
-    source, target = edges_to_remove[0].source_rx_index, edges_to_remove[0].target_rx_index
-    links_to_stash = [x.link_object for x in edges_to_remove]
-    # if only one (source, target) is entered, it removes only one edge, not all
-    # therefore we need as many entries in the list as there are edges between the source and target to break the cycle
-    to_remove_list = [(x.source_rx_index, x.target_rx_index) for x in edges_to_remove]
-    phantom_links = []
-    for link_to_stash in links_to_stash:
+    # find the normal edges to remove
```
BalduinLandolt, line 201: this comment is redundant

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
```diff
@@ -188,73 +188,75 @@ def _remove_edges_to_stash(
...
-    graph.remove_edges_from(to_remove_list)
-    return links_to_stash
+            new_phantoms = _find_phantom_xml_edges(
+                edges_to_remove[0].source_rx_index,
+                edges_to_remove[0].target_rx_index,
+                all_edges,
+                link_to_stash,
+                remaining_nodes,
+            )
+            phantom_edges_to_remove.extend(new_phantoms)
```
BalduinLandolt, line 217: I liked this as a one-liner

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
```diff
@@ -188,73 +188,75 @@ def _remove_edges_to_stash(
...
 
 def _find_phantom_xml_edges(
-    source: int,
-    target: int,
-    edges: list[Edge],
-    xml_instance: XMLLink,
+    source_rx_index: int,
+    target_rx_index: int,
+    all_edges: list[Edge],
+    xml_link_to_stash: XMLLink,
```
BalduinLandolt, line 228: to me, the old variables were clearer and more concise. Only `xml_instance` I would rename to `xml_link`

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
```diff
@@ -188,73 +188,75 @@ def _remove_edges_to_stash(
...
-    the link may contain further links to other resources.
-    If we stash the XMLLink then in the real data all the links are stashed.
-    This is not automatically the case in the rx graph.
-    We identify all the edges that need to be removed so that the rx graph represents the links that remain
-    in the real data.
+    If an edge that will be removed represents an XML link,
+    the text value may contain further links to other resources.
+    If we stash the XMLLink, then in the real data all links of that text value are stashed.
+    So, the rx graph must be purged from these "phantom" links that are not in the real data any more.
+    This function identifies the edges that must be removed from the rx graph.
```
BalduinLandolt, line 236:
```suggestion
    This function identifies these "phantom" links, so they can be removed from the graph.
```
I don't like the phrasing "purge", also "rx" is not relevant here.

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
```diff
@@ -188,73 +188,75 @@ def _remove_edges_to_stash(
...
         return all(
             (
-                x.source_rx_index == source,
-                x.target_rx_index != target,
-                x.link_object == xml_instance,
-                x.target_rx_index in remaining_nodes,
+                x.source_rx_index == source_rx_index,
+                x.target_rx_index != target_rx_index,
+                x.link_object == xml_link_to_stash,
+                x.target_rx_index in remaining_nodes,  # could be removed because of leaf node -> NoEdgeBetweenNodes err
```
BalduinLandolt, line 255: not clear what this comment intends to tell me. If it's a call to action, it should be a linear ticket; if not, it should be clearly informative

BalduinLandolt (approved):

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
jnussbaum: The reason I took it out of `_remove_edges_to_stash()` is: This function removes the edges, as its name says. But then, it also extracted the link objects from a parameter it received: `[x.link_object for x in links_to_remove]` was done inside the function, and the result was returned. 

That's completely unnecessary, this can be done on the caller side. The old version was misleading: The function did not only what its name said, but it also extracted something from one of its parameters, and returned it.

I don't think that this would be faster or more readable. But perhaps it's a misunderstanding?

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
jnussbaum: Thanks for pointing this out, I must find a clearer formulation.

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
jnussbaum: You're right, I reverted it to 1 line, and it's more readable :-)

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
BalduinLandolt: Let me clarify:
- I think it's more readable to have a named list than a comprehension
- speed: internally, this list that previously has been returned is still created inside the function, previously that was just returned, now you re-create it on the caller side. My assumption - without testing it - is that creating this list twice is slower than passing it back from the function.

jnussbaum: - I agree that a named list can be more readable than a comprehension.
- But I disagree that the list is still created inside the function: The list `[x.link_object for x in links_to_remove]` is completely unrelated to what happens in `_remove_edges_to_stash()`. Inside there, the `ResptrLink | XMLLink`s are completely ignored. So I found it misleading to create a list with `ResptrLink | XMLLink`s inside that function, and then return it. 

Or am I completely confusing things?

BalduinLandolt: maybe I'm confused... but from what I see:

previously:

```python
links_to_stash = [x.link_object for x in edges_to_remove]
...
for link_to_stash in links_to_stash:
    ...
...
return links_to_stash
```

now:

```python
...
for link_to_stash in [x.link_object for x in edges_to_remove]:
    ....
...
```
so I think you just in-lined the list creation into the for loop, discard the list because you don't hold on to it; and then on the caller side, you create the same list again.

jnussbaum: Oh I see, you are right. So from performance side, my code will be slower, because the same list is built 2 times.

I didn't realize it, because if you analyze it from a "flow of information" view, then it doesn't make sense that a function called `_remove_edges_to_stash()` returns an information that can easily be extracted from a parameter that you pass to that function.

So, I still think that my version is "nicer", but the previous is faster. I'll text you in chat concerning performance measurements.

BalduinLandolt: sounds good, thanks!

PR #593, opened 2023-10-25
DO NOT MERGE: Feature/dev 2859 get rid of resource instance factory

BalduinLandolt:


jnussbaum:
@BalduinLandolt Can we close this?

BalduinLandolt:
> @BalduinLandolt Can we close this?

Not yet, if that's ok with you: This will never get merged as such, but some of the changes (especially in terms of package structure) I really like, and I would like to have them quickly at hand for when I need them - and for that, the PR is still useful.

PR #595, opened 2023-10-26
chore: tidy up graph analyzing

jnussbaum:


testdata/xml-data/circular-references/analyse_circles_in_xml.py
```diff
@@ -30,19 +30,18 @@ def analyse_circles_in_data(xml_filepath: Path, tracer_output_file: str, save_tr
         max_stack_depth=3,
     )
     tracer.start()
-    tree = etree.parse(xml_filepath)
-    root = tree.getroot()
+    root = parse_and_clean_xml_file(xml_filepath)
```
jnussbaum, line 33: This function transforms the special tags <annotation>, <region>, and <link> to their technically correct form <resource restype="Annotation">, <resource restype="Region">, and <resource restype="LinkObj">. This is necessary, otherwise these special tags aren't considered when analyzing the links

src/dsp_tools/analyse_xml_data/construct_and_analyze_graph.py
jnussbaum, line 1: Here I just improved the flow of information: It's not necessary for `_create_info_from_xml_for_graph_from_one_resource()` to return `subject_id`. This can be extracted on the caller side.

test/unittests/test_analyse_xml_data/test_construct_and_analyze_graph.py
jnussbaum, line 1: Now, `_create_info_from_xml_for_graph_from_one_resource()` doesn't return the subject_id any more. So, the tests become easier, because they don't have to test this unnecessary thing any more.

PR #599, opened 2023-10-26
feat(xmlupload): reduce stash size of circular references (DEV-2848)

jnussbaum:


src/dsp_tools/utils/xmlupload/stash_circular_references.py
```diff
@@ -18,122 +24,92 @@
...
-                            res_id=res.id,
-                            res_type=res.restype,
-                            uuid=uuid,
-                            prop_name=link_prop.name,
-                            value=standoff_xml,
-                        )
-                        stashed_standoff_values.append(standoff_stash_item)
-                        value.value = KnoraStandoffXml(uuid)
-                        value.resrefs = [_id for _id in value.resrefs if _id in ok_res_ids]
+                    if not value.link_uuid in stash_lookup[res.id]:
```
BalduinLandolt, line 51: I think it's recommended to write `x not in y` rather than `not x in y` but I may be mistaken

src/dsp_tools/utils/xmlupload/stash_circular_references.py
```diff
@@ -18,122 +24,92 @@
...
+
+    Returns:
+        stash: an object that contains the stashed references
     """
     stashed_standoff_values: list[StandoffStashItem] = []
     stashed_link_values: list[LinkValueStashItem] = []
-    ok_resources: list[XMLResource] = []
 
-    for res in nok_resources.copy():
+    for res in resources.copy():
```
BalduinLandolt, line 45: is it still necessary to copy those?

src/dsp_tools/utils/xmlupload/stash_circular_references.py
```diff
@@ -18,122 +24,92 @@
...
+        if not res.id in stash_lookup:
+            continue
         for link_prop in res.get_props_with_links():
             if link_prop.valtype == "text":
```
BalduinLandolt, line 49: how about factoring out everything in the `if x == "text"` and `if x == resptr` into a helper function, so that this function gets a bit more readable?

src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -118,15 +118,16 @@ def _prepare_upload(
...
     resources, permissions_lookup, resclass_name_2_type = _get_data_from_xml(
         con=con,
         root=root,
         default_ontology=default_ontology,
         shortcode=shortcode,
         verbose=verbose,
     )
-    # temporarily remove circular references
-    resources, stash = remove_circular_references(resources, verbose=verbose)
+    resources = sorted(resources, key=lambda res: upload_order.index(res.id))
```
BalduinLandolt, line 129: I'm a bit concerned about the performance of this: the way I understand sorting in python, the lambda is called for each element in the list; and looking up the index of a list is O(n). So in the worst case we end up with O(n^2) here.  
If I understand the code correctly, we could get this down to O(n), if resources was a dict {res.id: res}, and then we could say `sorted_resources = [resources[res_id] for res_id in upload_order]`.

I'm happy with either of the following two options:
1. you merge this as is, and then check if it's a performance issue with large datasets.
2. you try to address the potential issue in this PR.

But I wouldn't like to take our chances and just hoping that this doesn't slow us down.

BalduinLandolt (approved):
All in all very nice, thanks! I noticed one potential performance issue, that may require some investigation/work

src/dsp_tools/utils/xmlupload/xmlupload.py
jnussbaum: Wow, it was indeed a huge performance problem: With SGV, my sorting approach took 4 minutes, yours takes 0.1 seconds... Thanks for pointing this out!

BalduinLandolt: cool, I didn't even expect that much difference

PR #603, opened 2023-10-27
fix: ignore IRIs in stashing algorithm (DEV-2885)

jnussbaum:


test/unittests/test_analyse_xml_data/test_construct_and_analyze_graph.py
```diff
@@ -117,8 +117,8 @@ def test_extract_id_one_text_with_several_id() -> None:
     test_ele = etree.fromstring(
         """<text permissions="prop-default" encoding="xml">
             <a class="salsah-link" href="IRI:res_A_11:IRI">res_A_11</a>
-            <a class="salsah-link" href="IRI:res_B_11:IRI">res_A_11</a>
-            <a class="salsah-link" href="IRI:res_B_11:IRI">res_A_11</a>
+            <a class="salsah-link" href="IRI:res_B_11:IRI">res_B_11</a>
+            <a class="salsah-link" href="IRI:res_B_11:IRI">res_B_11</a>
         </text>"""
```
BalduinLandolt, line 122: I'm not sure I understand 100%... is this a typo you're fixing? or is it a different scenario? if the latter is the case, I would propose not changing the existing test, but to make this an additional test case. In the fist case, this makes sense, of course

BalduinLandolt (approved):
The code you changed looks good, thanks!  
I didn't take the time, though, to remind myself of the exact order of how things happen: If this happens after the weight calculation for the XML links (which I don't think is the case, but just to be sure), would it also require to make adjustments to the weight calculation? Or is this still correct?

test/unittests/test_analyse_xml_data/test_construct_and_analyze_graph.py
jnussbaum: just a typo

jnussbaum:
> The code you changed looks good, thanks! I didn't take the time, though, to remind myself of the exact order of how things happen: If this happens after the weight calculation for the XML links (which I don't think is the case, but just to be sure), would it also require to make adjustments to the weight calculation? Or is this still correct?

It's before the weighting.

PR #606, opened 2023-10-27
fix: IRI regex used in multiple places is incorrect (DEV-2888)

jnussbaum:


BalduinLandolt (approved):
would it make sense to store this regex somewhere in a central place, instead of repeating it in multiple places in the code?

PR #609, opened 2023-10-27
fix: don't create logging handler multiple times (DEV-2891)

jnussbaum:


src/dsp_tools/utils/create_logger.py
```diff
@@ -41,5 +37,35 @@ def get_logger(
...
+        the logger instance
+    """
+    global rotating_file_handler
+    if not rotating_file_handler:
+        _make_handler(
+            Path.home() / Path(".dsp-tools"),
+            filesize_mb=20,
+            backupcount=10,
+        )
+        rotating_file_handler = cast(logging.handlers.RotatingFileHandler, rotating_file_handler)
```
BalduinLandolt, line 67: if you let the function return the handler instead of writing it to the global variable, and here you do `rotating_file_handler = _make_handler(...)`, then no cast is needed

src/dsp_tools/utils/create_logger.py
```diff
@@ -1,38 +1,34 @@
 import logging
 import logging.handlers
 from pathlib import Path
+from typing import cast
 
+rotating_file_handler: logging.handlers.RotatingFileHandler | None = None
```
BalduinLandolt, line 6: should probably be a private variable, so `_rotating_file_handler `

BalduinLandolt (approved):
Two minor comments, otherwise looking good. Did you notice any difference with only one handler created?

PR #612, opened 2023-10-30
chore: log dsp-tools version

jnussbaum:


src/dsp_tools/cli.py
```diff
@@ -289,6 +290,21 @@ def _parse_arguments(
...
+    # _detail_version has one of the following formats:
+    # - 'dsp-tools==5.0.3\n'
+    # - 'dsp-tools @ git+https://github.com/dasch-swiss/dsp-tools.git@1f95f8d1b79bd5170a652c0d04e7ada417d76734\n'
+    # - '-e git+ssh://git@github.com/dasch-swiss/dsp-tools.git@af9a35692b542676f2aa0a802ca7fc3b35f5713d#egg=dsp_tools\n'
+    # - ''
+    if version_number := regex.search(r"\d+\.\d+\.\d+", _detail_version):
+        return version_number.group(0)
+    if regex.search(r"github.com", _detail_version):
+        return _detail_version.replace("\n", "")
+    return version("dsp-tools")
```
BalduinLandolt, line 305: could this be implemented in terms of `importlib.metadata.version()`? https://docs.python.org/3.11/library/importlib.metadata.html#module-importlib.metadata

BalduinLandolt (commented):
I love the intent, but I think this could be simplified, see my comment below

src/dsp_tools/cli.py
jnussbaum: Line 305 (`return version("dsp-tools")`) is already `importlib.metadata.version()` ;-)
But I only use it as fallback. I prefer the more elaborated approach via `pip freeze`, because `importlib.metadata` doesn't know anything about editable installations, nor about installations from a git source.

BalduinLandolt: oops... I overlooked that line.  
To me this feels like a lot of overhead and logic to maintain, but ok, it doesn't matter too much I think.

BalduinLandolt (approved):

PR #613, opened 2023-10-30
refactor: add specific model for bitstream information (DEV-2893)

BalduinLandolt:
Another little PR to pave the way for my upcoming PR

### DIFF
--- a/src/dsp_tools/models/xmlresource.py
+++ b/src/dsp_tools/models/xmlresource.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass
 from typing import Optional, Union
 
 import regex
@@ -9,6 +10,18 @@ from dsp_tools.models.permission import Permissions
 from dsp_tools.models.value import KnoraStandoffXml
 from dsp_tools.models.xmlbitstream import XMLBitstream
 from dsp_tools.models.xmlproperty import XMLProperty
+
+
+@dataclass(frozen=True)
+class BitstreamInfo:
+    """
+    Represents a bitstream object, consisting of it's file name on the local file system,
+    the internal file name assigned by SIPi and optionally its permissions.
+    """
+
+    local_file: str
+    internal_file_name: str
+    permissions: Permissions | None = None
 
 
 class XMLResource:  # pylint: disable=too-many-instance-attributes
@@ -175,7 +188,7 @@ class XMLResource:  # pylint: disable=too-many-instance-attributes
 
     def get_bitstream_information(
         self, internal_file_name_bitstream: str, permissions_lookup: dict[str, Permissions]
-    ) -> Optional[dict[str, Union[str, Permissions]]]:
+    ) -> BitstreamInfo | None:
         """
         Get the bitstream object belonging to the resource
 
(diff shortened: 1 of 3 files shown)


src/dsp_tools/models/xmlresource.py
```diff
@@ -11,6 +12,18 @@
 from dsp_tools.models.xmlproperty import XMLProperty
 
 
+@dataclass(frozen=True)
+class BitstreamInfo:
+    """
+    Represents a bitstream object, consisting of it's file name on the local file system,
```
jnussbaum, line 18:
```suggestion
    Represents a bitstream object, consisting of its file name on the local file system,
```

src/dsp_tools/models/xmlresource.py
```diff
@@ -11,6 +12,18 @@
...
+    the internal file name assigned by SIPi and optionally its permissions.
```
jnussbaum, line 19: Additionally, I always promote semantic line breaks :-)

jnussbaum, line 19:
```suggestion
    the internal file name assigned by SIPI and optionally its permissions.
```

src/dsp_tools/models/xmlresource.py
```diff
@@ -184,14 +197,13 @@ def get_bitstream_information(
             permissions_lookup: Is used to resolve the permission id's to permission sets
```
jnussbaum, line 197:
```suggestion
            permissions_lookup: Is used to resolve the permission IDs to permission sets
```

src/dsp_tools/models/xmlresource.py
```diff
@@ -175,7 +188,7 @@ def get_propvals(
 
     def get_bitstream_information(
         self, internal_file_name_bitstream: str, permissions_lookup: dict[str, Permissions]
-    ) -> Optional[dict[str, Union[str, Permissions]]]:
+    ) -> BitstreamInfo | None:
         """
         Get the bitstream object belonging to the resource
```
jnussbaum, line 193: I know this was not written by you, but if you understand what happens here: Could you write a description that is helpful?  Because I don't get it...

src/dsp_tools/utils/xmlupload/xmlupload.py
```diff
@@ -400,20 +400,29 @@ def _create_resource(
...
     try:
+        if bitstream_information:
+            bitstream_info: dict[str, str | Permissions] | None = {
+                "value": bitstream_information.local_file,
+                "internal_file_name": bitstream_information.internal_file_name,
+            }
+            if bitstream_information.permissions and bitstream_info:
+                bitstream_info["permissions"] = bitstream_information.permissions
+        else:
+            bitstream_info = None
```
jnussbaum, line 418: Would it be too difficult to change the class `ResourceInstance`? It would be more natural to teach `ResourceInstance` how to cope with `BitstreamInfo`, if possible.

jnussbaum (approved):
Some comments, but otherwise LGTM

src/dsp_tools/utils/xmlupload/xmlupload.py
BalduinLandolt: You're right that this would be the clean implementation. But as this is just a preparatory step to get rid of ResourceInstance alsong with the ResourceInstanceFactory, I considered it the "way of least resistance" not to touch the internals of ResourceInstance. This conversion will of course be of a short life span, and the replacement of ResourceInstance will be able to handle BitstreamInfo directly.

src/dsp_tools/models/xmlresource.py
BalduinLandolt: Done. In this case I'm with you; and in documentation markdown I'm all for semantic line breaks. In docstrings I'm sometimes torn, because I also like docstrings to take up as little screen real estate as possible - and semantic line breaks can easily double or tripple the number of lines the docstring uses up.

PR #614, opened 2023-10-31
chore: improve logging for upload-files

jnussbaum:


src/dsp_tools/fast_xmlupload/upload_files.py
```diff
@@ -324,6 +324,8 @@ def _upload_files_in_parallel(
     result: list[tuple[Path, bool]] = []
     for uploaded in as_completed(upload_jobs):
         result.append(uploaded.result())
+        if len(result) % 1000 == 0:
+            print(f"{datetime.now()}: Uploaded {len(result)} files so far, continue...")
     return result
```
BalduinLandolt, line 329: this really looks like a case for tqdm

src/dsp_tools/fast_xmlupload/upload_files.py
```diff
@@ -385,6 +387,9 @@ def upload_files(
     """
     dir_with_processed_files_path = _check_processed_dir(dir_with_processed_files)
     pkl_file_paths = get_pkl_files()
+    logger.info("Found the following pickle files:")
+    for pkl in pkl_file_paths:
+        logger.info(f" - {str(pkl)}")
 
     # read paths from pkl file
     internal_filenames_of_processed_files = _get_paths_from_pkl_files(pkl_files=pkl_file_paths)
```
BalduinLandolt, line 395: I'm not really familiar with this code - I'm assuming the paths will be logged too, somewhere?

BalduinLandolt (approved):

src/dsp_tools/fast_xmlupload/upload_files.py
jnussbaum: Don't think this would work. `as_completed(upload_jobs)` yields results from multiprocessing, i.e. waiting until more results come.

src/dsp_tools/fast_xmlupload/upload_files.py
jnussbaum: Yes, every time a file is uploaded successfully, its path is logged.

src/dsp_tools/fast_xmlupload/upload_files.py
BalduinLandolt: I'm sure this could be made to work, potentially one would have to provide it the length of all expected results, if that is not a fixed-length iterable but an iterator. But admittedly, this is a different story, so no need to speculate about it now :)

PR #624, opened 2023-11-07
test: add tests for newly added functionality (DEV-2915)

BalduinLandolt:


test/e2e/xmlupload/test_list_client_live.py
```diff
@@ -0,0 +1,16 @@
...
+
+def test_ist_clienbt_live() -> None:
+    con = ConnectionLive("http://localhost:3333")
+    con.login("root@example.com", "test")
+    project_client = ProjectClientLive(con, "0001")
+    project_iri = project_client.get_project_iri()
+    assert project_iri == "http://rdfh.ch/projects/0001"
+    list_client = ListClientLive(con, project_iri)
+    list_node_id_to_iri_lookup = list_client.get_list_node_id_to_iri_lookup()
+    assert len(list_node_id_to_iri_lookup) >= 28
```
BalduinLandolt, line 16: Note: this test is bad and brittle, but right now it's very hard to test this better. I would suggest leaving it in until it hurts us too much, or until we can re-write it to a more reasonable form.

test/unittests/util/test_date_util.py
```diff
@@ -0,0 +1,269 @@
...
+    def test_parse_date_fewer_digits_for_month_and_day(self) -> None:
+        date_string = "0860-1-1"
+        result = _parse_single_date(date_string, Era.CE)
+        expected = SingleDate(era=Era.CE, year=860, month=1, day=1)
+        assert result == expected
+
+    def test_do_not_allow_fewer_digits_in_year(self) -> None:
+        date_string = "860-1-1"
+        with pytest.raises(BaseError):
+            _parse_single_date(date_string, Era.CE)
```
BalduinLandolt, line 244: not sure how I feel about that - maybe the regex should be changed to allow this

test/unittests/xmlupload/test_resource_create_client.py
BalduinLandolt, line 1: One could add way more tests here, but it always requires XML to write these tests, which I find inappropriate (the format of the very input to be able to test the very output). I would suggest writing more tests once the Resource Create Client no longer depends on XMLResource

pyproject.toml
```diff
@@ -93,10 +97,11 @@ build-backend = "poetry.core.masonry.api"
 [tool.pytest.ini_options]
 addopts = ["--import-mode=importlib"]
 # see https://docs.pytest.org/en/latest/explanation/goodpractices.html#tests-outside-application-code
+pythonpath = [".", "src", "test"]
```
BalduinLandolt, line 100: the only real change in this file, which was needed for the tests to run (because I'm importing stuff from within the test package - i.e. `tests/unittests/helpersconnection_mock.py`), all other changes are auto-formatting... sorry about that

BalduinLandolt:
I discovered some minor issues during writing of the tests - but mostly it's just tests

test/unittests/xmlupload/test_list_client_live.py
```diff
@@ -0,0 +1,244 @@
...
+                "listinfo": {
+                    "id": "http://www.example.org/lists#a",
+                    "labels": [{"value": "some label"}],
+                },
+                "children": [],
+            }
+        }
+        con = ConnectionMock([list_response])
+        result = _get_list_from_server(con, "")
+        expected_node = ListNode(node_iri="http://www.example.org/lists#a", node_name="some label")
```
jnussbaum, line 136: I'm confused that "some label" is the **name** of the node. According to line 129, I expect it to be the **label**. Perhaps you could improve naming in line 129? (on that line, I also don't get what "value" is)

test/unittests/xmlupload/test_project_client_live.py
```diff
@@ -0,0 +1,71 @@
...
+            "@graph": [
+                {"@id": "http://www.example.org/ontologies/a.1/v2"},
+                {"@id": "http://www.example.org/ontologies/a.2/v2"},
+            ]
+        }
+        con = ConnectionMock([project_iri_response, project_ontologies_response])
+        project_client = ProjectClientLive(con, "")
+        ontology_name_dict = project_client.get_ontology_name_dict()
+        expected = {
+            "a.1": "http://www.example.org/ontologies/a.1/v2",
```
jnussbaum, line 51: Does that mean that "a.2" is automatically derived from "http://www.example.org/ontologies/a.1/v2"?

jnussbaum (approved):
LGTM, just some remarks

pyproject.toml
```diff
@@ -4,7 +4,9 @@
 name = "dsp-tools"
 version = "5.1.0"
 description = "DSP-TOOLS is a Python package with a command line interface that helps you interact with a DaSCH service platform (DSP) server."
-authors = ["DaSCH - Swiss National Data and Service Center for the Humanities <info@dasch.swiss>"]
+authors = [
+    "DaSCH - Swiss National Data and Service Center for the Humanities <info@dasch.swiss>",
+]
```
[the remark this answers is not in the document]
BalduinLandolt: done

pyproject.toml
```diff
@@ -31,7 +33,9 @@ requests = "^2.31.0"
 jsonschema = "^4.19.1"
 openpyxl = "^3.1.2"
 networkx = "^3.1"
-pandas = {version = "^2.0.3", extras = ["excel"]}  # extra package that contains xlrd that is necessary for reading old .xls Excel files
+pandas = { version = "^2.0.3", extras = [
+    "excel",
+] } # extra package that contains xlrd that is necessary for reading old .xls Excel files
```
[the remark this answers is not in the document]
BalduinLandolt: done. (but we should make sure that our auto-formatting results in the same, to avoid these things in the future)

src/dsp_tools/utils/xmlupload/list_client.py
```diff
@@ -98,7 +98,8 @@ def _get_list_from_server(con: Connection, list_iri: str) -> List:
     list_info = list_object["listinfo"]
     children: list[dict[str, Any]] = list_object["children"]
     root_iri = list_info["id"]
-    list_name = list_info["name"]
+    # if the root node does not have a name, use the label instead
```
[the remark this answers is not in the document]
BalduinLandolt: yes, according to the API docs, that is the case. And I actually only realized that when tests failed for me, because in the anything test data we have precisely this case.

test/e2e/xmlupload/test_list_client_live.py
```diff
@@ -0,0 +1,16 @@
...
+
+def test_list_client_live() -> None:
+    con = ConnectionLive("http://localhost:3333")
+    con.login("root@example.com", "test")
+    project_client = ProjectClientLive(con, "0001")
+    project_iri = project_client.get_project_iri()
+    assert project_iri == "http://rdfh.ch/projects/0001"
+    list_client = ListClientLive(con, project_iri)
+    list_node_id_to_iri_lookup = list_client.get_list_node_id_to_iri_lookup()
+    assert len(list_node_id_to_iri_lookup) >= 28
```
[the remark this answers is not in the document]
BalduinLandolt: the second test doesn't make too much sense to me, the fist one I'm happy to apply.  
However, in a good testing scenario we should be able to know exactly what is the expected result, which is here not always given, because other tests may add more lists to the project (which is indeed what is happening)

test/unittests/util/test_date_util.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+class TestParseDateStringWithDateOnly:
+    """Tests parsing a date string with only dates, no era or calendar information."""
+
+    def test_start_date(self) -> None:
+        date_string = "2020-01-01"
+        result = parse_date_string(date_string)
+        expected_start_and_end = SingleDate(era=Era.CE, year=2020, month=1, day=1)
+        expected = Date(Calendar.GREGORIAN, expected_start_and_end, expected_start_and_end)
```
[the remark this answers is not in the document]
BalduinLandolt: yes, I suppose the test and the docs should not contradict each other

test/unittests/xmlupload/test_list_client_live.py
BalduinLandolt: no, I can't change that, the JSON you're seeing is of the same shape as the JSON returned by the list endpoint of the API (just with lots of non-important information removed). And because the list at hand here, does not have a `name` specified, the label will be used as a name instead. (Which is not ideal, but the problem is not that in and by itself, but that we can even have list nodes without a unique name.)

test/unittests/xmlupload/test_project_client_live.py
BalduinLandolt: yes

PR #625, opened 2023-11-07
chore(excel2xml): suppress irrelevant warnings, turn non-fatal errors into warnings (DEV-2917)

jnussbaum:
https://linear.app/dasch/issue/DEV-2917/

### DIFF
--- a/src/dsp_tools/excel2xml.py
+++ b/src/dsp_tools/excel2xml.py
@@ -2009,7 +2009,7 @@ def _append_bitstream_to_resource(
                 f"'{row['file']}' (Resource ID '{row['id']}', Excel row {row_number}). "
                 f"An attempt to deduce them from the resource permissions failed."
             )
-    with warnings.catch_warnings():  # ignore warnings about not existing files
+    with warnings.catch_warnings():  # ignore warnings about non-existing files
         resource.append(
             make_bitstream_prop(
                 path=str(row["file"]),
--- a/src/dsp_tools/excel2xml.py
+++ b/src/dsp_tools/excel2xml.py
@@ -1912,7 +1912,6 @@ def _validate_and_prepare_cli_input_file(dataframe: pd.DataFrame) -> pd.DataFram
 def _convert_rows_to_xml(
     dataframe: pd.DataFrame,
     max_num_of_props: int,
-    print_errors: bool,
 ) -> list[etree._Element]:
     """
     Iterate through the rows of the CSV/Excel input file,
@@ -1952,7 +1951,7 @@ def _convert_rows_to_xml(
             # in all cases (except for the very first iteration), a previous resource exists
             if resource is not None:
                 resources.append(resource)
-            resource = _convert_resource_row_to_xml(row_number=row_number, row=row, print_errors=print_errors)
+            resource = _convert_resource_row_to_xml(row_number=row_number, row=row)
 
         # this is a property-row
         else:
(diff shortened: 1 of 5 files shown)


src/dsp_tools/excel2xml.py
```diff
@@ -2003,18 +2003,20 @@ def _append_bitstream_to_resource(
...
                 f"Missing file permissions for file "
                 f"'{row['file']}' (Resource ID '{row['id']}', Excel row {row_number}). "
                 f"An attempt to deduce them from the resource permissions failed."
             )
-    resource.append(
-        make_bitstream_prop(
-            path=str(row["file"]),
-            permissions=str(file_permissions),
-            calling_resource=row["id"],
+    with warnings.catch_warnings():  # ignore warnings about not existing files
```
Nora-Olivia-Ammann, line 2012: non-existing

src/dsp_tools/excel2xml.py
```diff
@@ -2041,17 +2043,20 @@ def _convert_resource_row_to_xml(
     resource_id = row["id"]
     resource_label = row.get("label")
     if pd.isna([resource_label]):
-        raise BaseError(f"Missing label for resource '{resource_id}' (Excel row {row_number})")
-    if not check_notna(resource_label):
+        resource_label = ""
+        warnings.warn(f"Missing label for resource '{resource_id}' (Excel row {row_number})")
```
Nora-Olivia-Ammann, line 2047: Do you really want to make the label optional? That seems not like a good idea since the label is used in the APP

src/dsp_tools/excel2xml.py
```diff
@@ -2041,17 +2043,20 @@ def _convert_resource_row_to_xml(
...
+    elif not check_notna(resource_label):
         warnings.warn(
             f"The label of resource '{resource_id}' looks suspicious: '{resource_label}' (Excel row {row_number})"
         )
     resource_restype = row.get("restype")
     if not check_notna(resource_restype):
-        raise BaseError(f"Missing restype for resource '{resource_id}' (Excel row {row_number})")
+        resource_restype = ""
+        warnings.warn(f"Missing restype for resource '{resource_id}' (Excel row {row_number})")
```
Nora-Olivia-Ammann, line 2055: Am I understanding this wrong or are you making the type of resource optional?

Nora-Olivia-Ammann (commented):
From what I see here is that with this structure we create XMLs based on excels that are missing mandatory values. Why would this be desirable?

jnussbaum:
> From what I see here is that with this structure we create XMLs based on excels that are missing mandatory values. Why would this be desirable?

This was not the intention. Rather, I want that the entire program runs through and gives me a list of warnings/errors. Previously, the first error interrupted the program, so the user never knew the entire list of problems. The idea behind this PR is that creating an invalid XML is acceptable, because the user has then a list of problems, so that they can fix them directly in the XML, or they can fix it in the Excel and then run the command again.

This PR was born out of a real-world-scenario: I exported dokubib from salsah, converted it to XML, and sent the corrupt XML with the problem list to Rita, who takes care about fixing the problems.

Nora-Olivia-Ammann:
> > From what I see here is that with this structure we create XMLs based on excels that are missing mandatory values. Why would this be desirable?
> 
> This was not the intention. Rather, I want that the entire program runs through and gives me a list of warnings/errors. Previously, the first error interrupted the program, so the user never knew the entire list of problems. The idea behind this PR is that creating an invalid XML is acceptable, because the user has then a list of problems, so that they can fix them directly in the XML, or they can fix it in the Excel and then run the command again.
> 
> This PR was born out of a real-world-scenario: I exported dokubib from salsah, converted it to XML, and sent the corrupt XML with the problem list to Rita, who takes care about fixing the problems.

Could we not set a flag so that we can stop the program in the end if any error occurred?

Nora-Olivia-Ammann (approved):

PR #630, opened 2023-11-10
feat(xmlupload): check consistency with ontology on server (DEV-2921)

Nora-Olivia-Ammann:


test/unittests/commands/xmlupload/test_ontology_client.py
```diff
@@ -0,0 +1,285 @@
...
+                    "owl:onProperty": {"@id": "testonto:sequenceOfAudio"},
+                    "owl:cardinality": 1,
+                },
+            ],
+            "@type": "owl:Class",
+            "@id": "testonto:AudioSequence",
+        },
+    ]
+    res_cls = _get_all_classes_from_graph(test_graph)
+    assert res_cls == ["testonto:AudioSequence"]
```
BalduinLandolt, line 46: I'm not 100% sure about this myself, but I feel it should not be necessary to provide "fully realistic" json in cases like this - I would personally leave out the properties that do not matter at all because they will just be ignored.

BalduinLandolt: Doing so, would drastically shorten this test file, which would be nice :)

BalduinLandolt (commented):
This ended up being quite a number of comments... sorry about that. Of course not all of them are equally significant... And let me know if you have any questions

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,74 @@
...
+
+def _iterate_over_xml_find_problems(root: etree._Element, onto_tool: OntoCheckTool) -> None:
+    unknown_eles = UnknownOntologyElements(save_path=onto_tool.save_location)
+    for resource in root.iterchildren():
+        if not _diagnose_classes(resource.tag, onto_tool):
+            unknown_eles.classes.append((resource.attrib["id"], resource.attrib["restype"]))
+        if problems := _diagnose_all_properties(resource, onto_tool):
+            unknown_eles.properties.extend(problems)
+    if unknown_eles.not_empty():
+        unknown_eles.execute_problem_protocol()
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Regarding the first part.
I did a version of this in the beginning and it became overly complicated. But if I were to extract it in the tuple format then maybe I could cleanly separate this. Thanks!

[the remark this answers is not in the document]
Nora-Olivia-Ammann: > Also, I'm not a fan of creating an object and then modifying its members. How about creating lists first and then creating the object with the finished lists? Then it can also be a frozen object.

Is there a specific reason?

src/dsp_tools/commands/xmlupload/ontology_client.py
```diff
@@ -0,0 +1,127 @@
...
+        self.ontology_names = onto_names
+
+    def _get_ontology_from_server(self, ontology_name: str) -> list[dict[str, Any]]:
+        try:
+            url = f"/ontology/{self.shortcode}/{ontology_name}/v2"
+            res: dict[str, Any] = try_network_action(self.con.get, route=url)
+        except BaseError:
+            raise BaseError(
+                f"Ontologies for project {self.shortcode} could not be retrieved from the DSP server"
+            ) from None
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I see your point. Which method would you think is better?

src/dsp_tools/commands/xmlupload/ontology_diagnose_models.py
```diff
@@ -0,0 +1,93 @@
...
+        }
+        prop_di = {
+            "type": "property",
+            "resource id": [x[0] for x in self.properties],
+            "problematic type": [x[1] for x in self.properties],
+        }
+        df = pd.DataFrame.from_records([cls_di, prop_di])
+        df.to_excel(
+            excel_writer=Path(self.save_path, "ProblematicOntologyElements_in_XML.xlsx"), sheet_name=" ", index=False
+        )
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I have thought about that. If you prefer it I'll change it.

test/unittests/commands/xmlupload/test_ontology_client.py
Nora-Olivia-Ammann: I thought I would leave all the top level keys in the file. But I can delete the rdfs:subClassOf elements if you think it makes it too long.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def _diagnose_classes(class_str: str, onto_tool: OntoCheckTool) -> bool:
+    prefix, cls = _identify_ontology(class_str, onto_tool)
```
jnussbaum, line 49: `cls` is a keyword in the python language, so I wouldn't recommend naming a variable `cls`

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,74 @@
...
+    return prop in onto_tool.onto_lookup[prefix].properties
+
+
+def _identify_ontology(prop_cls: str, onto_tool: OntoCheckTool) -> tuple[str, ...]:
+    if onto_tool.default_ontology_colon.match(prop_cls):
+        return onto_tool.default_ontology_prefix, prop_cls.lstrip(":")
+    elif onto_tool.knora_undeclared.match(prop_cls):
+        return "knora-api", prop_cls
+    elif onto_tool.generic_prefixed_ontology.match(prop_cls):
+        return tuple(prop_cls.split(":"))
```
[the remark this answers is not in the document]
jnussbaum: Do you consider module level constants a good practice? I think that the current approach of passing an object as parameter that carries this info is much cleaner

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,74 @@
...
+    prefix, prop = _identify_ontology(prop_str, onto_tool)
+    return prop in onto_tool.onto_lookup[prefix].properties
+
+
+def _identify_ontology(prop_cls: str, onto_tool: OntoCheckTool) -> tuple[str, ...]:
+    if onto_tool.default_ontology_colon.match(prop_cls):
+        return onto_tool.default_ontology_prefix, prop_cls.lstrip(":")
+    elif onto_tool.knora_undeclared.match(prop_cls):
+        return "knora-api", prop_cls
+    elif onto_tool.generic_prefixed_ontology.match(prop_cls):
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: There are three different version I found in the XML

onto -> as written in the XML:
knora -> knora-api:someThing
knora -> someThing
defaultOnto -> defaultOnto:someThing
defaultOnto -> :someThing
otherOnto -> otherOnto:someThing

I called it generic prefixed because it splits any prefix whether it is default, knora etc. doesn't matter. In the lookup we save knora as "knora-api", the default under its prefix and the others also with their prefix. If the default or knora is prefixed it still is correctly identified, so there does not need to be a specification. But the default ontology only with the : or knora without any declaration need to be identified separately. What we could do is instead of raising a BaseError if it is structurally incorrect, just add it to the incorrect results. What do you think?

src/dsp_tools/commands/xmlupload/ontology_client.py
```diff
@@ -0,0 +1,127 @@
...
+    shortcode: str
+    default_ontology: str
+    ontology_names: list[str] = field(default_factory=list)
+
+    def get_all_ontologies_from_server(self) -> dict[str, list[dict[str, Any]]]:
+        """
+        This function retrieves all ontologies from a project plus the knora-api ontology from a server.
+
+        Returns:
+            Dictionary with the ontology names and an instance of Ontology.
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Old docstring. Thanks

jnussbaum:
General remark on error handling: 

Currently, _iterate_over_xml_find_problems() raises single BaseErrors when a prefixed identifier doesn't follow a known ontology pattern. These BaseErrors are raised in _identify_ontology() and then escalate further up, without being handled. This is subpar. We would rather like to accumulate all errors, and then give a list of all errors to the user.
(Besides, do_xml_consistency_check() doesn't document these BaseErrors in its docstring.)

Within the models, you often raise BaseErrors. There I have the same concern: are these handled by their callers? Or how would they escalate?

Nora-Olivia-Ammann:
> General remark on error handling:
> 
> Currently, _iterate_over_xml_find_problems() raises single BaseErrors when a prefixed identifier doesn't follow a known ontology pattern. These BaseErrors are raised in _identify_ontology() and then escalate further up, without being handled. This is subpar. We would rather like to accumulate all errors, and then give a list of all errors to the user. (Besides, do_xml_consistency_check() doesn't document these BaseErrors in its docstring.)
> 
> Within the models, you often raise BaseErrors. There I have the same concern: are these handled by their callers? Or how would they escalate?

Thinking again about this, I actually would prefer to do it differently here. Since we are searching for problems, and the other types of problems do not cause an immediate crash, I think it would be more useful and consistent if we document it in the way we do with the ones where the prefix is known. We would need to change the way the error message is constructed somewhat but I think it would be ok.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    onto_tool = OntoCheckTool(
+        default_ontology_prefix=onto_client.default_ontology,
+        onto_lookup=_get_project_and_knora_ontology_from_server(onto_client),
+        save_location=onto_client.save_location,
+    )
+    _iterate_over_xml_find_problems(root, onto_tool)
+
+
+def _get_project_and_knora_ontology_from_server(onto_client: OntologyClient) -> dict[str, Ontology]:
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Generally I agree, but here what we get from the API is an onology called knora-api. We have to deal with this in the strings and json. So in this case I think it would be more confusing

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,74 @@
...
+        if not _diagnose_properties(prop.attrib["name"], onto_tool)
+    ]
+
+
+def _diagnose_properties(prop_str: str, onto_tool: OntoCheckTool) -> bool:
+    prefix, prop = _identify_ontology(prop_str, onto_tool)
+    return prop in onto_tool.onto_lookup[prefix].properties
+
+
+def _identify_ontology(prop_cls: str, onto_tool: OntoCheckTool) -> tuple[str, ...]:
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Well that is actually not the case, as it does not split it in every case. But I can rename it to get_prefix and prop_cls_name

src/dsp_tools/commands/xmlupload/ontology_client.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: If we follow Balduin's logic with the project_client.py which is not a information storage but something that does things, it should remain in this folder. Since the project_client is in this folder too.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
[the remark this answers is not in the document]
BalduinLandolt: Generally preferring immutability is also my main reason. But there is also a cosmetic aspect to it, namely the accessing of the variable is not nested. just to illustrate (notice the comments):

```python
@dataclass
class Something:
    a: list[int]
    b: list[int]

def mutable() -> Something:
    something = Something()
    for i in range(99):
        something.a.append(i+1)  # compare this line
    for i in range(99):
        something.b.append(i * i)
    return something

def immutable() -> Something:
    a = []
    for i in range(99):
        a.append(i+1)  # with this line
    b = []
    for i in range(99):
        b.append(i * i)
    return Something(a, b)
```

of course this is a stupid example because this would normally be solved with list comprehension... but you get my point

src/dsp_tools/commands/xmlupload/ontology_client.py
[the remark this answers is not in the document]
BalduinLandolt: tiny detail: `from e` and omiting it is not strictly the same - I figured out recently that there is a difference: Omiting it combines the stack traces with "During handling of the above exception, another exception occurred" (or along those lines), and `from e` combines them with "Which is directly caused by the above exception".

I generally prefer more detail - but as I said, we should discuss exception handling anyway, so it's your pick for now, I would say.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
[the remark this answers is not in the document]
BalduinLandolt: I think module level constants are a good thing for certain use cases - and this might be one. But if you prefer to organize it, I would at least have it be a dedicated class and use it as static class variables:

```python
class OntologyRegex:
    default_ontology_colon = r"^:[A-Za-z]+$"
    knora_undeclared = r"^[A-Za-z]+$"
    generic_prefixed_ontology = r"^[A-Za-z]+-?[A-Za-z]+:[A-Za-z]+$"

def foo() -> None:
    x = regex.match(OntologyRegex.default_ontology_colon, "...")
```

Then you just require the class, and not an instance of it.  
These are constants and having different versions of them at any given time would simply be a bug - so they should also be treated as singletons. And there is no point in having instances of singletons - that's literally just duplication 😉

test/unittests/commands/xmlupload/test_ontology_client.py
BalduinLandolt: again, I'm not 100% sure what the best idea is - both realistic data and brevity are good things. But in my last PR I opted for brevity and it felt good. But it's absolutely up to you

Nora-Olivia-Ammann: Thanks for the comment. I will leave the top level keys, because I do like to be able to somewhat see what we actually get in real life, but delete the longer content.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,123 @@
...
+        class_info.append("Resource type does not follow a known ontology pattern")
+        return class_info
+    try:
+        if cls_ not in onto_check_info.onto_lookup[prefix].classes:
+            class_info.append("Invalid Class Type")
+            return class_info
+        return None
+    except KeyError:
+        class_info.append("Unknown ontology prefix")
+        return class_info
```
BalduinLandolt, line 88: I personally don't like the over-use of exceptions, and all of this could be done by if-checking (and the performance difference should be negligible, even in large projects)

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
BalduinLandolt, line 1: After reading this file all the way through, I feel it would become way more easily understandable, if you added 3 datatypes, instead of using complicated list and tuple structures:

```python
@dataclass(frozen=True)
class ClassInfo:
    class_id: str
    res_type: str

@dataclass(frozen=True)
class PropertyInfo:
    prop_id: str
    prop_name: str

@dataclass(frozen=True)
class OntologyProblem:
    entity: ClassInfo | PropertyInfo
    message: str
```

Then most of your function would accept or return an instance or a list of one or more of those classes, and things might become a lot clearer to understand.

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -0,0 +1,102 @@
...
+        if cls_msg:
+            msg += cls_msg + extra_separator
+        prop_msg = self._print_problem_string_props()
+        if prop_msg:
+            msg += prop_msg
+        if len(self.classes) + len(self.properties) > 50:
+            ex_name = "InvalidOntologyElements_in_XML.xlsx"
+            self._get_and_save_problems_as_excel(ex_name)
+            msg += extra_separator + f"\nAn excel: '{ex_name}' was saved at '{self.save_path}' listing the problems."
+        raise UserError(msg)
```
BalduinLandolt, line 57: how about letting this method return an optional string, and letting the raising of the error be done on the caller side. That would make this a more pure model, because it would then not rely on producing a sie effect

test/unittests/commands/xmlupload/test_check_consistency_with_ontology.py
```diff
@@ -0,0 +1,314 @@
...
+        onto_lookup={
+            "test": Ontology(classes=["classA", "classB"], properties=["propA", "propB"]),
+            "knora-api": Ontology(classes=["knoraClassA"], properties=["knoraPropA"]),
+        },
+    )
+    classes = [["idA", "knoraClassA"]]
+    properties = [["idA", "knora-api:knoraPropA"]]
+    assert not _find_problems_in_classes_and_properties(
+        classes, properties, onto_check_info
+    )  # type: ignore[func-returns-value]
```
BalduinLandolt, line 94: does it really need this ignore here? seems strange

test/unittests/commands/xmlupload/test_check_consistency_with_ontology.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+class TestDiagnoseClass:
+    @staticmethod
```
BalduinLandolt, line 98: this is purely cosmetic - but I think to my eye it would look less cluttered to add `self` to each of those test functions (and not use it), than to have the `@staticmethod` decorator everywhere

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,123 @@
...
+def _get_all_classes_and_properties(root: etree._Element) -> tuple[list[list[str]], list[list[str]]]:
+    classes = _get_all_class_types_and_ids(root)
+    properties = []
+    for resource in root.iterchildren(tag="resource"):
+        properties.extend(_get_all_property_names_and_resource_ids(resource))
+    return classes, properties
+
+
+def _get_all_class_types_and_ids(root: etree._Element) -> list[list[str]]:
+    return [[resource.attrib["id"], resource.attrib["restype"]] for resource in root.iterchildren(tag="resource")]
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Originally it was a tuple, but the reason it changed is because i append the problem identification at the end. For example, unknown prefix etc.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def _get_all_property_names_and_resource_ids(resource: etree._Element) -> list[list[str]]:
+    return [[resource.attrib["id"], prop.attrib["name"]] for prop in resource.iterchildren() if prop.tag != "bitstream"]
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Same as above

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,123 @@
...
+    except KeyError:
+        class_info.append("Unknown ontology prefix")
+        return class_info
+
+
+def _diagnose_all_properties(properties: list[list[str]], onto_check_info: OntoCheckInformation) -> list[list[str]]:
+    problem_list = []
+    for prop_info in properties:
+        if problem := _diagnose_properties(prop_info, onto_check_info):
+            problem_list.append(problem)
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Not if I continue to use the walrus operator. If I do then I get a usage before assignment crash. I would have to call the function at both places in the comprehension for it to work, which I don't like. I also tried to switch the location of the walrus to the end of the comprehension `if not problem:= etc.` which did not even pass the basic syntax check.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
Nora-Olivia-Ammann: I understand your point. I tend to do them not only for performance but also if I consider it to be rare. I will have a think about it.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
Nora-Olivia-Ammann: I have thought about that. But I am unsure, when using classes become a layer too far... I will revisit this, when I think about the general error structure.

test/unittests/commands/xmlupload/test_check_consistency_with_ontology.py
Nora-Olivia-Ammann: Yes, and I don't understand why it appears in the first place. I could not find the inconsistencies at all.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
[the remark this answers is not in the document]
BalduinLandolt: right... what about

```python
[x for x in (_diagnose_properties(prop_info, onto_check_info) for prop_info in properties) if x]
```

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,123 @@
...
+
+def _get_all_classes_and_properties(root: etree._Element) -> tuple[list[list[str]], list[list[str]]]:
+    classes = _get_all_class_types_and_ids(root)
+    properties = []
+    for resource in root.iterchildren(tag="resource"):
+        properties.extend(_get_all_property_names_and_resource_ids(resource))
+    return classes, properties
+
+
+def _get_all_class_types_and_ids(root: etree._Element) -> list[list[str]]:
```
jnussbaum, line 46: The order of the return values is actually the other way round: your function should be named `_get_all_class_ids_and_types`

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,123 @@
...
+from lxml import etree
+from regex import Pattern
+
+from dsp_tools.commands.xmlupload.models.ontology_diagnose_models import InvalidOntologyElements, OntoCheckInformation
+from dsp_tools.commands.xmlupload.ontology_client import OntologyClientLive
+from dsp_tools.models.exceptions import BaseError
+
+defaultOntologyColon: Pattern[str] = regex.compile(r"^:\w+$")
+knoraUndeclared: Pattern[str] = regex.compile(r"^\w+$")
+genericPrefixedOntology: Pattern[str] = regex.compile(r"^[A-Za-z]+-?[A-Za-z]+:\w+$")
```
jnussbaum, line 12: Are you sure ontos can contain only 1 single dash? I'd feel more comfortable with:

```suggestion
genericPrefixedOntology: Pattern[str] = regex.compile(r"^[\w\-]+:\w+$")
```

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def do_xml_consistency_check(onto_client: OntologyClientLive, root: etree._Element) -> None:
```
jnussbaum, line 15: I thought the entire point of having a protocol is that the protocol should be used here? 

```suggestion
def do_xml_consistency_check(onto_client: OntologyClient, root: etree._Element) -> None:
```

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,123 @@
...
+    for resource in root.iterchildren(tag="resource"):
+        properties.extend(_get_all_property_names_and_resource_ids(resource))
+    return classes, properties
+
+
+def _get_all_class_types_and_ids(root: etree._Element) -> list[list[str]]:
+    return [[resource.attrib["id"], resource.attrib["restype"]] for resource in root.iterchildren(tag="resource")]
+
+
+def _get_all_property_names_and_resource_ids(resource: etree._Element) -> list[list[str]]:
```
jnussbaum, line 50: same here: the return values are the other way round: `_get_all_resource_ids_and_property_names`

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
jnussbaum: For me, only readability counts in this matter. I dislike too many try-exepts, no matter how often they are executed. For me the question is rather: How often do I have to read it? The question "How often is it executed?" doesn't matter for me.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,123 @@
...
+
+def _diagnose_all_properties(properties: list[list[str]], onto_check_info: OntoCheckInformation) -> list[list[str]]:
+    problem_list = []
+    for prop_info in properties:
+        if problem := _diagnose_properties(prop_info, onto_check_info):
+            problem_list.append(problem)
+    return problem_list
+
+
+def _diagnose_properties(prop_info: list[str], onto_check_info: OntoCheckInformation) -> list[str] | None:
```
jnussbaum, line 99: It's a single property, so this function should be named `_diagnose_property`

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -0,0 +1,123 @@
...
+        if prop not in onto_check_info.onto_lookup[prefix].properties:
+            prop_info.append("Invalid Property")
+            return prop_info
+        return None
+    except KeyError:
+        prop_info.append("Unknown ontology prefix")
+        return prop_info
+
+
+def _get_prefix_and_prop_cls_identifier(prop_cls: str, default_ontology_prefix: str) -> tuple[str, ...]:
```
jnussbaum, line 115: what about renaming `prop_cls` to `prop_or_cls`? (Especially in the parameter, perhaps also in the function name)

jnussbaum, line 115: The return value is of fixed length, so `tuple[str, str]` would be the better return type

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -0,0 +1,102 @@
...
+class Ontology:
+    """This class saves the properties and the classes from an ontology."""
+
+    classes: list[str] = field(default_factory=list)
+    properties: list[str] = field(default_factory=list)
+
+
+@dataclass
+class OntoCheckInformation:
+    """This class saves information that is needed to check if consistency with the ontology."""
```
jnussbaum, line 21:
```suggestion
    """This class saves information that is needed to check consistency with the ontology."""
```

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -0,0 +1,102 @@
...
+
+    default_ontology_prefix: str
+    onto_lookup: dict[str, Ontology]
+    save_location: Path
```
jnussbaum, line 25: This field seems counter-intuitive to me: What does a save location have to do with an ontology check information? Could you explain it in the docstring?

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -0,0 +1,102 @@
...
+from dataclasses import dataclass, field
+from pathlib import Path
+
+import pandas as pd
+
+from dsp_tools.models.exceptions import UserError
+
+
+@dataclass(frozen=True)
+class Ontology:
```
jnussbaum, line 12: we already have a class with this name in `src/dsp_tools/commands/project/models/ontology.py`, so we now have 2 classes with the same name in the codebase. However, the name `OntoInfo` is not taken yet, so perhaps you should rename it?

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -0,0 +1,102 @@
...
+                    "resource id": x[0],
+                    "problematic type": x[1],
+                    "problem": x[2],
+                }
+                for x in self.properties
+            ]
+        )
+        return pd.DataFrame.from_records(problems)
+
+    def _print_problem_string_cls(self) -> str:
```
jnussbaum, line 84: This method name is misleading, because it doesn't print anything. What about `compose_problem_string_cls`?

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -0,0 +1,102 @@
...
+        if self.classes:
+            separator = "\n----------------------------\n"
+            problems = [
+                f"\tResource ID: '{x[0]}'\n\tResource Type: '{x[1]}'\n\tProblem: '{x[2]}'" for x in self.classes
+            ]
+            return "The following resource(s) have an invalid resource type:\n\n" + separator.join(problems)
+        else:
+            return ""
+
+    def _print_problem_string_props(self) -> str:
```
jnussbaum, line 94: This method name is misleading, because it doesn't print anything. What about `compose_problem_string_props`?

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -0,0 +1,102 @@
...
+    """This class saves and prints out the information regarding ontology classes and properties
+    that are in the XML but not the ontology."""
+
+    save_path: Path
+    classes: list[list[str]] = field(default_factory=list)
+    properties: list[list[str]] = field(default_factory=list)
+
+    def execute_problem_protocol(self) -> None:
+        """
+        This function is executed if there are any elements in properties or classes.
```
jnussbaum, line 39: That's not true. You can execute it also when there are no elements :-) To be more precise: 

```suggestion
        If there are any elements in properties or classes, this method composes an error message.
```

I agree with Balduin that this method should only compose a string, but not raise an error.

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -0,0 +1,102 @@
...
+        extra_separator = "\n\n----------------------------\n\n"
+        msg = "Some property and or class type(s) used in the XML are unknown:" + extra_separator
+        cls_msg = self._print_problem_string_cls()
+        if cls_msg:
+            msg += cls_msg + extra_separator
+        prop_msg = self._print_problem_string_props()
+        if prop_msg:
+            msg += prop_msg
+        if len(self.classes) + len(self.properties) > 50:
+            ex_name = "InvalidOntologyElements_in_XML.xlsx"
```
jnussbaum, line 54: Not so user friendly :-) What about 

```suggestion
            ex_name = f"{orig_filename}_syntax_errors.xlsx"
```

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
BalduinLandolt: you're right, I missed that

test/unittests/commands/xmlupload/test_ontology_client.py
```diff
@@ -0,0 +1,245 @@
...
+
+from dsp_tools.commands.xmlupload.ontology_client import (
+    OntologyClientLive,
+    _get_all_classes_from_graph,
+    _get_all_properties_from_graph,
+    _remove_prefixes,
+    deserialize_ontology,
+)
+
+# pylint: disable=missing-class-docstring,missing-function-docstring,unused-argument,redefined-outer-name,protected-access
```
jnussbaum, line 16: redefined-outer-name isn't an error that occurs in this module

```suggestion
# pylint: disable=missing-class-docstring,missing-function-docstring,unused-argument,protected-access
```

test/unittests/commands/xmlupload/test_ontology_client.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+@dataclass
+class ConnectionMock(ConnectionMockBase):
+    get_responses: dict[Any, Any]
```
jnussbaum, line 21: it is only one single response, so it should be singular

test/unittests/commands/xmlupload/test_ontology_client.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def get(self, route: str, headers: dict[str, str] | None = None) -> dict[Any, Any]:
+        return self.get_responses
+
+
+class TestGetAllClassesFromGraph:
+    @staticmethod
+    def test_single_class() -> None:
```
jnussbaum, line 29: For my taste, this is too fine-grained. As I commented in `_get_all_classes_from_graph`, I would merge the 2 functions, and then test the higher-level one.

src/dsp_tools/commands/xmlupload/ontology_client.py
```diff
@@ -0,0 +1,133 @@
...
+from dsp_tools.utils.create_logger import get_logger
+from dsp_tools.utils.shared import try_network_action
+
+# pylint: disable=too-few-public-methods
+
+logger = get_logger(__name__)
+
+
+@dataclass
+class OntologyClient(Protocol):
```
jnussbaum, line 17: This protocol is never used to create a `class OntologyClientMock`. There are only instances of `OntologyClientLive`. So I wonder if it is worth having this protocol?

test/unittests/commands/xmlupload/test_ontology_client.py
```diff
@@ -0,0 +1,245 @@
...
+                "http://www.knora.org/ontology/0801/beol",
+            ],
+            "selfjoin": False,
+            "shortcode": "0801",
+            "shortname": "beol",
+            "status": True,
+        }
+    }
+    con = ConnectionMock(response)
+    onto_cli = OntologyClientLive(con, "0801", "beol", Path(""))
```
jnussbaum, line 193: the abbreviation "cli" stands for Command Line Interface. So I wouldn't abbreviate at all, just write `onto_client`

test/unittests/commands/xmlupload/test_check_consistency_with_ontology.py
```diff
@@ -0,0 +1,314 @@
...
+    _get_prefix_and_prop_cls_identifier,
+)
+from dsp_tools.commands.xmlupload.models.ontology_diagnose_models import OntoCheckInformation, Ontology
+from dsp_tools.models.exceptions import BaseError
+
+# pylint: disable=missing-function-docstring,missing-class-docstring
+
+
+def test_get_all_classes_and_properties() -> None:
+    test_ele = etree.fromstring(
```
jnussbaum, line 25: General remark: For me, it feels more natural to abbreviate element with elem. But feel free to leave it as it is :-)

test/unittests/commands/xmlupload/test_read_validate_xml_file.py
```diff
@@ -5,6 +5,8 @@
     _check_if_salsah_targets_exist,
 )
 
+# pylint: disable=missing-function-docstring
```
jnussbaum, line 8: All functions in this module have docstrings. And pylint passes on main branch, so this change should be reverted

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -0,0 +1,102 @@
...
+    def execute_problem_protocol(self) -> None:
+        """
+        This function is executed if there are any elements in properties or classes.
+        If there are more than 100 entries combined, then the result is also saved as an excel.
+
+        Raises:
+            UserError: If properties or classes have any entries.
+        """
+        extra_separator = "\n\n----------------------------\n\n"
+        msg = "Some property and or class type(s) used in the XML are unknown:" + extra_separator
```
jnussbaum, line 46:
```suggestion
        msg = "Some property and/or class type(s) used in the XML are unknown:" + extra_separator
```

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
Nora-Olivia-Ammann: Right, and in that case, I as opposed to you often times I prefer Errors...

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -35,24 +35,40 @@ def do_xml_consistency_check(onto_client: OntologyClientLive, root: etree._Eleme
...
+    for resource in root.iterchildren(tag="resource"):
+        restype = resource.attrib["restype"]
+        if restype in cls_dict:
+            cls_dict[restype].append(resource.attrib["id"])
+        else:
+            cls_dict[restype] = [resource.attrib["id"]]
+    return cls_dict
+
+
+def _get_all_property_names_and_resource_ids_one_resouce(
```
jnussbaum, line 78: typo: `resouce`

src/dsp_tools/commands/xmlupload/ontology_client.py
```diff
@@ -0,0 +1,133 @@
...
+        return {onto_name: deserialize_ontology(onto_graph) for onto_name, onto_graph in ontologies.items()}
+
+    def _get_all_ontology_jsons_from_server(self) -> dict[str, list[dict[str, Any]]]:
+        self._get_ontology_names_from_server()
+        project_ontos = {onto: self._get_ontology_from_server(onto) for onto in self.ontology_names}
+        project_ontos["knora-api"] = self._get_knora_api_from_server()
+        return project_ontos
+
+    def _get_ontology_names_from_server(self) -> None:
+        try:
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I don't understand your comment.

src/dsp_tools/commands/xmlupload/ontology_client.py
```diff
@@ -0,0 +1,133 @@
...
+    return Ontology(classes, properties)
+
+
+def _get_all_cleaned_classes_from_graph(onto_graph: list[dict[str, Any]]) -> list[str]:
+    classes = _get_all_classes_from_graph(onto_graph)
+    return _remove_prefixes(classes)
+
+
+def _get_all_classes_from_graph(onto_graph: list[dict[str, Any]]) -> list[str]:
+    return [elem["@id"] for elem in onto_graph if elem.get("knora-api:isResourceClass")]
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I did it so I could test the removal of prefixes and the extraction of the classes/ properties separately...

src/dsp_tools/commands/xmlupload/ontology_client.py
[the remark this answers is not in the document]
jnussbaum: if `self.ontology_names` is already populated, you don't have to do an API call to get the info a 2nd time. But I don't know if this case can ever happen.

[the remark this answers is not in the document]
Nora-Olivia-Ammann: Well, if more code were to use that model then perhaps. But not the way it is written now. This feels superfluous to me until we change code...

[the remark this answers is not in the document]
jnussbaum: That's fine for me, you can leave it away then

PR #637, opened 2023-11-14
fix(xmlupload): print reason of failure if resource cannot be created (DEV-2925)

jnussbaum:


src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -40,17 +40,10 @@ def create_resource(
...
         resource_json_ld = json.dumps(resource_dict, ensure_ascii=False)
-        try:
-            res = try_network_action(self.con.post, route="/v2/resources", jsondata=resource_json_ld)
-            iri = res["@id"]
-            label = res["rdfs:label"]
-            return iri, label
-        except BaseError as e:
-            msg = f"Could not create resource {resource.id} (label: {resource.label}, iri: {resource.iri})"
-            raise BaseError(msg) from e
-        except KeyError as e:
```
jnussbaum, line 51: I cannot find a place where a KeyError could be triggered. And even if it did, I think it's better not to catch it here.

BalduinLandolt: key error can always happen if you do a dict lookup. But you are indeed right, there is no reason why this should happen - and I don't recall why I added this.

BalduinLandolt (approved):

PR #639, opened 2023-11-14
feat(xmlupload): add check if links point to valid resource ID (DEV-2902)

jnussbaum:


src/dsp_tools/commands/xmlupload/read_validate_xml_file.py
```diff
@@ -34,15 +36,62 @@ def validate_and_parse_xml_file(
...
+        root: parsed XML file
+
+    Raises:
+        UserError: if a link target does not exist in the XML file
+    """
+    resptr_errors = _check_if_resptr_targets_exist(root)
+    salsah_errors = _check_if_salsah_targets_exist(root)
+    errors = resptr_errors + salsah_errors
+    if errors:
+        sep = "\n - "
```
Nora-Olivia-Ammann, line 63: To have this as an extra line seems a bit like overkill to me.

Nora-Olivia-Ammann (approved):

PR #647, opened 2023-11-16
chore: more sourcery refactorings

jnussbaum:


src/dsp_tools/utils/shared.py
```diff
@@ -163,7 +163,7 @@ def try_network_action(
         try:
             if args and not kwargs:
                 return action(*args)
-            elif kwargs and not args:
+            elif not args and kwargs:
                 return action(**kwargs)
             elif args and kwargs:
                 return action(*args, **kwargs)
```
BalduinLandolt, line 169: I would propose re-writing this entire section to a `match` roundabout like this:

```python
try:
    match (args, kwargs):
        case ([], {}):
            return action()
        case  (_, {}):
            return action(*args)
        case ([], _):
            return action(**kwargs)
        case _:
            return action(*args, **kwargs)
```

and this could even be extracted to a function `_execute_action()`, so your function only does one thing, namely retry.

src/dsp_tools/utils/shared.py
```diff
@@ -271,17 +268,16 @@ def _validate_xml_tags_in_text_properties(doc: Union[etree._ElementTree[etree._E
...
-                or len(list(text.iterchildren())) > 0
-            ):
-                sourceline = f" line {text.sourceline}: " if text.sourceline else " "
-                propname = text.getparent().attrib["name"]  # type: ignore[union-attr]
-                resname = text.getparent().getparent().attrib["id"]  # type: ignore[union-attr]
-                resources_with_illegal_xml_tags.append(f" -{sourceline}resource '{resname}', property '{propname}'")
+        regex_finds_tags = bool(regex.search(r'<([a-zA-Z/"]+|[^\s0-9].*[^\s0-9])>', str(text.text)))
+        etree_finds_tags = bool(list(text.iterchildren()))
+        has_tags = regex_finds_tags or etree_finds_tags
+        if text.attrib["encoding"] == "utf8" and has_tags:
```
BalduinLandolt, line 276: I would extract all this checking into a function

BalduinLandolt (approved):
I added some suggestions - feel free to apply them in this PR, later or not at all, as you see fit

src/dsp_tools/utils/shared.py
jnussbaum: Nora an me, we tried to do the same thing, but in a match case, `[]` catches empty list **and** non-empty ones. AFAIU, there is no way in a match statement to differentiate between an empty container and a non-empty one - except if we would say `case (args, kwargs) if args and not kwargs`, but then there's no point any more in translating the if-else into a match case!

BalduinLandolt: no, `case []` does not match a non-empty list, I'm sure of that

BalduinLandolt: image

PR #649, opened 2023-11-16
chore: refactor import-scripts

jnussbaum:


Nora-Olivia-Ammann (approved):

test/e2e/commands/test_rosetta.py
```diff
@@ -1,8 +1,8 @@
 # pylint: disable=protected-access,missing-function-docstring,redefined-outer-name
 
 import tempfile
+from collections.abc import Iterator
 from pathlib import Path
-from typing import Iterator
```
BalduinLandolt, line 5: what's the advantage of this change? I would normally prefer a type interface over an abstract implementation (=composition over inheritance)

BalduinLandolt (commented):
I added a question, apart from that it looks good

test/e2e/commands/test_rosetta.py
jnussbaum: https://docs.python.org/3.11/library/typing.html#typing.Iterator is a deprecated alias to collections.abc.Iterator, so I thought it would be better to use collections.abc?

BalduinLandolt: Interesting, I wasn't aware. In that case the change definitely makes sense

PR #653, opened 2023-11-23
chore: replace pylint, isort, black with ruff (DEV-2972)

jnussbaum:


src/dsp_tools/commands/fast_xmlupload/process_files.py
```diff
@@ -370,7 +370,7 @@ def _create_orig_file(
         shutil.copyfile(in_file, orig_file_full_path)
         logger.info(f"Created .orig file {orig_file_full_path}")
         return True
-    except Exception:  # pylint: disable=broad-exception-caught
+    except Exception:
```
jnussbaum, line 373: strange: ruff BLE001 should be triggered here ?!?

pyproject.toml
```diff
@@ -123,51 +121,65 @@ exclude = [
...
-    "too-many-return-statements",    # TODO: activate this
-    "too-many-statements",           # TODO: activate this
+[tool.ruff.lint.per-file-ignores]
+"__init__.py" = [
+    "F401",     # unused-import
+    "F403",     # undefined-local-with-import-star
+    "PLC0414",  # useless-import-alias
+]
+"src/dsp_tools/import_scripts/import_script.py" = [
+    "PLW2901",  # pylint: redefined-loop-name
```
Nora-Olivia-Ammann, line 172: It is correct that here we need a pylint ignore?

docs/developers/code-quality-tools/python-linting.md
Nora-Olivia-Ammann, line 1: Do you want to leave pylint here? (Line 7)

Nora-Olivia-Ammann (approved):

pyproject.toml
jnussbaum: Ruff reimplements rules from many linters, so the historical origin of a rule is often mentioned. Here, "PLW2901" means that this rule belongs to the group of reimplemented Pylint Warnings (hence PLW). 

Concerning my comment `# pylint: redefined-loop-name`: I need a human-friendly name, so I decided to make comments of the pattern `# original-linter: rule-name`.

Nora-Olivia-Ammann: So ruff understands pylint syntax? or this is solely for us? If it is solely for us, I may stay away from specifying with pylint. Because if they change, then we may be tempted to go to pylint and not read the ruff docs. Maybe you can comment it in your own language?

docs/developers/code-quality-tools/python-linting.md
jnussbaum: These files are an overview of existing tools. I created them when I dived into this subject, and took notes for myself. My idea was that I quickly present all existing tools, with their pros and cons, to justify the choice that I made for DSP-TOOLS. 

So if we keep these files, then we should also keep the reference to pylint. But we could also decide that the dev-docs of DSP-TOOLS is not the right place to describe the python ecosystem, and then we should remove all files except `docs/developers/code-quality-tools/code-quality-tools.md`

pyproject.toml
jnussbaum: Yes, it is solely for us. The reason I commented every rule with the name of its ruleset is the section `[tool.ruff.lint] > select`: Here, I need a way to describe each ruleset, so I wrote 

```
"PL",       # pylint`
```

But arguably, when ignoring single rules, the ruleset name could be omitted. So instead of 

```
ignore = [
    "PLR0913",  # pylint: too-many-arguments
    "PLR2004",  # pylint: magic-value-comparison
    "PT009",    # flake8-pytest-style: pytest-unittest-assertion 
    "PT027",    # flake8-pytest-style: pytest-unittest-raises-assertion
    "ISC001",   # flake8-implicit-str-concat: single-line-implicit-string-concatenation
]
```

I could also do this: 

```
ignore = [
    "PLR0913",  # too-many-arguments
    "PLR2004",  # magic-value-comparison
    "PT009",    # pytest-unittest-assertion 
    "PT027",    # pytest-unittest-raises-assertion
    "ISC001",   # single-line-implicit-string-concatenation
]
```

Would you prefer the second over the first?

Nora-Olivia-Ammann: If ruff is implementing flake8 then I would be fine with the flake annotation.
If they are implementing their own rule, which is alike another, then I prefer the second.
Thanks!

src/dsp_tools/cli.py
```diff
@@ -34,7 +34,7 @@
 logger = get_logger(__name__)
 
 
-def _make_parser(
+def _make_parser(  # noqa: PLR0915 (too-many-statements)
```
BalduinLandolt, line 37: could this be resolved rather than ignored?

src/dsp_tools/cli.py
```diff
@@ -421,7 +421,7 @@ def _derive_sipi_url(
     return parsed_arguments
 
 
-def _call_requested_action(args: argparse.Namespace) -> bool:
+def _call_requested_action(args: argparse.Namespace) -> bool:  # noqa: PLR0912, PLR0915 (too-many-branches/statements)
```
BalduinLandolt, line 424: could this be resolved?

src/dsp_tools/cli.py
```diff
@@ -449,7 +449,7 @@ def _call_requested_action(args: argparse.Namespace) -> bool:
                     password=args.password,
                     dump=args.dump,
                 )
-        else:
+        else:  # noqa: PLR5501 (collapsible-else-if)
```
BalduinLandolt, line 452: I'm with ruff on this one

src/dsp_tools/commands/excel2json/lists.py
```diff
@@ -261,8 +261,8 @@ def _make_json_lists_from_excel(
         The finished "lists" section
     """
     # reset the global variables
-    global list_of_previous_node_names
-    global list_of_lists_of_previous_cell_values
+    global list_of_previous_node_names  # noqa: PLW0603 (global-statement)
+    global list_of_lists_of_previous_cell_values  # noqa: PLW0603 (global-statement)
```
BalduinLandolt, line 265: seems like a bad pattern indeed. could this be resolved?

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -154,7 +154,7 @@ def _row2resource(
 
     cards = []
     for j, detail_row in details_df.iterrows():
-        j = int(str(j))  # j is a label/index/hashable, but we need an int
+        j = int(str(j))  # noqa: PLW2901 (redefined-loop-name) (j is a label/index/hashable, but we need an int)
```
BalduinLandolt, line 157: could this be changed rather than ignored?

src/dsp_tools/commands/project/models/group.py
```diff
@@ -107,7 +107,7 @@ def iri(self) -> Optional[str]:
         return self._iri
 
     @iri.setter
-    def iri(self, value: str) -> None:
+    def iri(self, value: str) -> None:  # noqa: ARG002 (unused-method-argument)
         raise BaseError("Group iri cannot be modified!")
```
BalduinLandolt, line 111: this setter should be removed entirely

BalduinLandolt (commented):
I stopped reviewing about 1/3 through. I'm not convinced by the approach of adding a powerful tool and silencing it all over the place. To me this is even worse than not having the tool in use at all (because as a dev I can't even get those warnings locally, now).  
I would say, either we fix *all* the issues right away (which is probably not realistic), we *only* use ignore statements in the pyproject.toml file, so they are easily removeable again, or we remove ruff from the CI and just see it as a tool that devs use locally to see warnings while working.  
Like this, the changes to me seem like being for the worse; but if the two of you like it, I'm ok with being outvoted.

Nora-Olivia-Ammann:
> I stopped reviewing about 1/3 through. I'm not convinced by the approach of adding a powerful tool and silencing it all over the place. To me this is even worse than not having the tool in use at all (because as a dev I can't even get those warnings locally, now). I would say, either we fix _all_ the issues right away (which is probably not realistic), we _only_ use ignore statements in the pyproject.toml file, so they are easily removeable again, or we remove ruff from the CI and just see it as a tool that devs use locally to see warnings while working. Like this, the changes to me seem like being for the worse; but if the two of you like it, I'm ok with being outvoted.

I see your point. I am fine with both because: We currently silence the warnings already, this is nothing new. The advantage here is it is probably faster and it is only one tool instead of three. We could keep this perhaps as a background project, like the error messages, where we clean up slowly?

BalduinLandolt:
> > I stopped reviewing about 1/3 through. I'm not convinced by the approach of adding a powerful tool and silencing it all over the place. To me this is even worse than not having the tool in use at all (because as a dev I can't even get those warnings locally, now). I would say, either we fix _all_ the issues right away (which is probably not realistic), we _only_ use ignore statements in the pyproject.toml file, so they are easily removeable again, or we remove ruff from the CI and just see it as a tool that devs use locally to see warnings while working. Like this, the changes to me seem like being for the worse; but if the two of you like it, I'm ok with being outvoted.
> 
> I see your point. I am fine with both because: We currently silence the warnings already, this is nothing new. The advantage here is it is probably faster and it is only one tool instead of three. We could keep this perhaps as a background project, like the error messages, where we clean up slowly?

It's true that we do that already, but far less than with this PR.

By "keep this as a background task" you mean introducing Ruff (i.e. instead of this PR) or getting rid of the ignores (i.e. we merge this and go from there)?  
In the end I can live with either, but I think the latter will be a bit annoying for the person doing it because there is no tooling to help you get rid of ignores, you have to search for them manually; and for the reviewers it will create lots of noise in the diffs.

Nora-Olivia-Ammann:
> > > I stopped reviewing about 1/3 through. I'm not convinced by the approach of adding a powerful tool and silencing it all over the place. To me this is even worse than not having the tool in use at all (because as a dev I can't even get those warnings locally, now). I would say, either we fix _all_ the issues right away (which is probably not realistic), we _only_ use ignore statements in the pyproject.toml file, so they are easily removeable again, or we remove ruff from the CI and just see it as a tool that devs use locally to see warnings while working. Like this, the changes to me seem like being for the worse; but if the two of you like it, I'm ok with being outvoted.
> > 
> > 
> > I see your point. I am fine with both because: We currently silence the warnings already, this is nothing new. The advantage here is it is probably faster and it is only one tool instead of three. We could keep this perhaps as a background project, like the error messages, where we clean up slowly?
> 
> It's true that we do that already, but far less than with this PR.
> 
> By "keep this as a background task" you mean introducing Ruff (i.e. instead of this PR) or getting rid of the ignores (i.e. we merge this and go from there)? In the end I can live with either, but I think the latter will be a bit annoying for the person doing it because there is no tooling to help you get rid of ignores, you have to search for them manually; and for the reviewers it will create lots of noise in the diffs.

By background task I mean that we introduce ruff and then slowly clean it up. But yes it is not a sexy job and not one person alone should do it... I can also live with either option. I see its advantages and disadvantages...

jnussbaum:
@Nora-Olivia-Ammann @BalduinLandolt I have now created a concept for how we use ruff: https://www.notion.so/dasch-swiss/Ruff-concept-9fc9658302ad4b35a79be1b83d219d06
Then I reworked this PR, so that it aligns with the concept. What are your thoughts?

src/dsp_tools/commands/excel2json/resources.py
Nora-Olivia-Ammann: maybe we could do this directly when we hand over the variable and don't assign it, that should work no?

jnussbaum: This is outdated, I found a better solution for it.

src/dsp_tools/cli.py
BalduinLandolt: I'm happy to agree to disagree on this one :)  
but generally I think it's not a wise idea to employ an opinionated tool and not agree with the opinions it has. So I would vote for not making that the rule but the exception

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -345,10 +345,10 @@ def append_permissions(root_element: etree._Element) -> etree._Element:
     return root_element
 
 
-def make_resource(
+def make_resource(  # noqa: D417 (undocumented-param)
     label: str,
     restype: str,
-    id: str,  # pylint: disable=redefined-builtin
+    id: str,  # noqa: A002 (builtin-argument-shadowing)
```
BalduinLandolt, line 351: those should be fixed, not ignored in my opinion, and they would be very easy to fix

src/dsp_tools/commands/start_stack.py
```diff
@@ -180,7 +180,7 @@ def _wait_for_fuseki(self) -> None:
                 )
                 if response.ok:
                     break
-            except Exception:  # pylint: disable=broad-exception-caught
+            except Exception:  # noqa: BLE001 (blind-except)
```
BalduinLandolt, line 183: should those indeed be ignored? I'm not convinced

BalduinLandolt (approved):
I added some last remarks, but generally I can get behind this

src/dsp_tools/cli.py
```diff
@@ -616,7 +616,7 @@ def run(args: list[str]) -> None:
         print("\nThe process was terminated because of an Error:")
         print(err.message)
         sys.exit(1)
-    except Exception as err:
+    except Exception as err:  # noqa: BLE001 (blind-except)
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I believe in this single case I think this makes sense. Because this is the outer most function of the CLI, we want to catch all exceptions that were not anticipated by us. Log them and tell the user to contact us.

Nora-Olivia-Ammann (approved):

src/dsp_tools/cli.py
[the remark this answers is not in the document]
jnussbaum: Yes exactly. I believe BLE001 is a rule that we generally want to be activated, but in this specific case, it is okay. We will never have to fix it. Due to the ignore comment, it doesn't appear as warning in the IDE.

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
jnussbaum: There are 2 reasons against:
 - If we change it, it is a breaking change.
 - The arguments of this function correspond to the attributes of the <resource> element:
   ```
   >>> str(make_resource(label="label", restype="restype", id="id", permissions="permissions"))
   <resource label="label" restype="restype" id="id" permissions="permissions"/>
   ```

I'm not against fixing it, I just don't have a good idea what the new name of the parameter could be?

src/dsp_tools/commands/start_stack.py
jnussbaum: What would your suggestion then be how to refactor the code? I'm afraid of catching a more specific exception class here...

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
Nora-Olivia-Ammann: underscore it? _id

src/dsp_tools/commands/start_stack.py
BalduinLandolt: Looking at the code now: why do we even retry here, if we call `http_call_with_retry` anyways?

jnussbaum: `http_call_with_retry` catches only timeout errors. I'm not sure, but perhaps we get an error as long as fuseki is not up yet. As soon as it's up, we `break` from the loop.

BalduinLandolt: just seems counter-intuitive to have a retry mechanism around a retry mechanism... but let's not get caught up on this one. It is a detail after all!

PR #654, opened 2023-11-27
feat(excel2json): new user error (InputError) implementation in properties section (DEV-3037)

Nora-Olivia-Ammann:


src/dsp_tools/models/exceptions.py
```diff
@@ -64,7 +65,56 @@ def __str__(self) -> str:
...
+                "Please contact the dsp-tools development team with the following information:\n"
+                "\t- Which command was used.\n"
+                "\t- If applicable, any files that were used in conjunction with the command.\n"
+                "\t- A file with the terminal output copied into.\n"
+                f"\t- The log files called 'logging.log', if there are several, include all.\n"
+                f"\t  They can be found at: {Path.home() / Path('.dsp-tools')}\n"
+            )
+            if custom_msg:
+                default_msg = "\n\n" + custom_msg + "\n--------------------------" + default_msg
+            super().__init__(default_msg)
```
BalduinLandolt, line 86: I think this could be streamlined:

```suggestion
        default_msg = ...
        match keep_default_ms, custom_msg:
            case False, str():
                super().__init__(custom_msg)
            case _, None:
                super().__init__(default_msg)
            case _:
                 super().__init__(f"{custom_msg}\n--------------------------\n{default_msg}")
```

... or around those lines - it's not exactly identical now. But I personally find it easier to grasp the logic like this.  
(also, I think the concatenated message in your version may have the line breaks in the wrong spot)

src/dsp_tools/models/exceptions.py
```diff
@@ -64,7 +65,56 @@ def __str__(self) -> str:
...
+                "please contact the dsp-tools development team with the following information:\n"
+                "\t- Which command was used.\n"
+                "\t- If applicable, any files that were used in conjunction with the command.\n"
+                "\t- A file with the terminal output copied into.\n"
+                f"\t- The log files called 'logging.log', if there are several, include all.\n"
+                f"\t  They can be found at: {Path.home() / Path('.dsp-tools')}\n"
+            )
+            if custom_msg:
+                default_msg = "\n\n" + custom_msg + "\n--------------------------" + default_msg
+            super().__init__(default_msg)
```
BalduinLandolt, line 112: as above

src/dsp_tools/models/exceptions.py
```diff
@@ -78,7 +128,9 @@ class UserError(BaseError):
 
 
 class XmlError(Exception):
```
BalduinLandolt, line 130: maybe `XmlUploadError` would be clearer? Like this, it made me think it refers to an error in the XML file.

src/dsp_tools/models/input_error.py
```diff
@@ -0,0 +1,131 @@
...
+    """This class contains information if the docker program is not started."""
+
+    user_msg: str
+
+
+@dataclass(frozen=True)
+class DirectoryProblem:
+    """This class contains information if a path does not exist."""
+
+    user_msg: str
```
BalduinLandolt, line 34: IMO this class should have a field `directory` which stores the non-existing directory. That might even make the message obsolete in this case?

src/dsp_tools/models/input_error.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@dataclass(frozen=True)
+class FileNotFoundProblem:
+    """This class contains information if a file is not found at the expected location."""
+
+    user_msg: str
```
BalduinLandolt, line 41: I'm sure this is intentional - but these problem classes are missing the execute protocol method

src/dsp_tools/models/input_error.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@dataclass(frozen=True)
+class ExcelStructureProblem:
+    """This class contains information if the structure (column names, etc.) is incorrect."""
+
+    user_msg: str
```
BalduinLandolt, line 48: I wonder if in some cases, the user message could actually be stored on the problem class, rather than being passed and stored on the instance, if the message is always the same. What do you think?

src/dsp_tools/models/input_error.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def _generate_error_msg(self) -> str:
+        msg = [self.user_msg]
+        if self.column:
+            msg.append("Column: " + self.column)
+        if self.rows:
+            rws = [str(x) for x in self.rows]
+            msg.append("Row(s):" + list_separator + list_separator.join(rws))
+        if self.values:
+            msg.append("Value(s):" + list_separator + list_separator.join(self.values))
```
BalduinLandolt, line 95: I'm having a hard time imagining how these will look, with the two list_separators... won't there be an "empty bulletpoint"?

src/dsp_tools/models/input_error.py
```diff
@@ -0,0 +1,131 @@
...
+class JsonValidationProblem:
+    """This class contains information about a JSON that fails its validation against the schema."""
+
+    user_msg: str
+    original_msg: str | None = None
+    message_path: str | None = None
+    property: str | None = None
+    resource_class: str | None = None
+    excel_column: str | None = None
+    excel_row: int | None = None
```
BalduinLandolt, line 109: these signatures suggest to me that there is room for improvement here 😉   
are some of those mutually exclusive? in that case, it might be nicer to store them as unions in the same field.

src/dsp_tools/models/input_error.py
```diff
@@ -0,0 +1,131 @@
+from dataclasses import dataclass
+from typing import Protocol
+
+# pylint: disable=too-few-public-methods
+
+
+separator = "\n\t"
+list_separator = "\n\t- "
```
BalduinLandolt, line 8: I tend to avoid tab stops where ever possible. How would you feel about replacing them with 2 or 4 spaces?

test/unittests/commands/excel2json/test_properties.py
```diff
@@ -432,20 +389,46 @@ def test__do_property_excel_compliance(self) -> None:
...
+                "\t- 9\n"
+                "There are missing values in a column that must not be empty:\n"
+                "\tColumn: label\n"
+                "\tRow(s):\n"
+                "\t- 3\n"
+                "\t- 8\n"
+                "There are missing values in a column that must not be empty:\n"
+                "\tColumn: gui_attributes\n"
+                "\tRow(s):\n"
+                "\t- 7"
```
BalduinLandolt, line 424: does this really have to be checked with a regex? looks like you know the literal string, so a simple equality check should be possible, right? And then, a multi-line string would probably be more readable

test/unittests/commands/excel2json/test_properties.py
BalduinLandolt, line 1: over all, I haven't researched/tried how else this would be done best. But this pattern of regex-matching on the message of an exception, when you know the exact string, seems overly complicated to me.

BalduinLandolt (approved):
I think this is a huge improvement. I didn't do the most thorough review, but I added some comments with thoughts on further improvements (and cosmetics). But I definitely feel we're on the right track here, with this error handling. And further refinements can always be done later on.

src/dsp_tools/models/input_error.py
Nora-Olivia-Ammann: No, if you join a list, that has only one value, then the separator is invisible. That is the advantage of it and why I need to put a separator in before the .join as the first value does not have it

test/unittests/commands/excel2json/test_properties.py
Nora-Olivia-Ammann: I will check with the equality, but regarding the multi-line strings, I have always had problems in the past with matching them, as the implicit linebreaks etc. are sometimes not recognised as equal to the explicit ones...

src/dsp_tools/models/input_error.py
Nora-Olivia-Ammann: When I started this PR I though I would do the whole command, so those classes would be necessary. But as I only did the properties the problems would not arise here and they are therefore not implemented. Should I delete them in this PR?

src/dsp_tools/models/input_error.py
Nora-Olivia-Ammann: Same

src/dsp_tools/models/input_error.py
Nora-Olivia-Ammann: The reason why I did not do that is because I wanted to keep this class fairly generic. The required structure of the excel is very different when we look at the lists and resources, so the same message won't work. Do you think we should create specific classes for each?

src/dsp_tools/models/input_error.py
Nora-Olivia-Ammann: I researched the best practice. And I am convinced 
https://stackoverflow.blog/2017/06/15/developers-use-spaces-make-money-use-tabs/
(I think it is about the tabs or spaces in code which is a non-issue here, but still funny)

src/dsp_tools/models/exceptions.py
BalduinLandolt: I see now - I think I missed the fact that default_msg starts with linebreaks

src/dsp_tools/models/input_error.py
BalduinLandolt: I don't mind having them here as placeholders

src/dsp_tools/models/input_error.py
BalduinLandolt: To me, this sounds like the more precise approach, but I'm honestly not sure it's worth the effort. What do you prefer? We could also consider this change as a second step

src/dsp_tools/models/input_error.py
BalduinLandolt: right, makes sense

test/unittests/commands/excel2json/test_properties.py
BalduinLandolt: now that I think about it, that makes sense - if you indent multiline-strings, you get extra whitespace. (in Scala we can strip this whitespace, which is very handy...)

test/unittests/commands/excel2json/test_properties.py
BalduinLandolt: Fair enough. Maybe the point here is that this is not actually a unit test, but an integration test (in the sense that it reads a file [which already should disqualify it as a unit test] and then measures a side effect [which is the very definition of integration tests], rather than checking the output of a unit of code for a given input), which I didn't realize but subconsciously made me suspicious.  
Without wanting to open huge cans of worms - how feasible would it be to change the tests of the important bits of functionality to "real" unit tests, and reduce the amount of integrated tests to the minimum needed to make sure that the integration works correctly? (meaning also: if we have unit tests for the error protocol that ensure that for a given problem, we get the right message, we no longer have to check that the raised exception has the right message [that's already ensured then], so it's enough to check that the right exception is raised)

Nora-Olivia-Ammann: I would favour to improve the test structure, especially if we do something where we have to do big changes anyway. I just opened a new PR where I want to restructure some tests in excel2json because there are many functions that test several things, this was the case here too and super annoying when trying to find the problem. So while I am at it, I am happy to change the locations. Would you do a new directory for the integration tests, so that we would have: e2e, unittest, integration?

src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -316,9 +337,17 @@ def _do_property_excel_compliance(df: pd.DataFrame, excelfile: str) -> None:
...
+    problems = [
+        utl.check_contains_required_columns_else_raise_error(df=df, required_columns=required_columns),
+        utl.check_column_for_duplicate(df=df, to_check_column="name"),
+    ]
+    if missing_vals_check := _check_missing_values_in_row(df=df):
+        problems.extend(missing_vals_check)
+    if any(problems):
+        extra = [problem.execute_error_protocol() for problem in problems if problem]
+
+        msg = [f"The excel file '{excelfile}' has some problems:", *extra]
```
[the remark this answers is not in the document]
BalduinLandolt: My thinking was along these lines:
- we have a `RequiredColumnMissingProblem` which can be checked and generated by one function
- we have a `DuplicateInColumnProblem`  which also gets created by a function
- we have a list of  `MissingRowValuesProblem` which gets created by a function
- we have an over-arching `ExcelPropertyProblem` (or whatever a good name might be) wrapper which contains the three above
- we have a factory function that takes the three optional inner problems and returns optionally a wrapper problem; the semantics would be: if none of the inner problems, then the wrapper would be none; if any problem, then the wrapper would be created containing at least one problem.
- the wrapper would provide you with an `execute_error_protocol()` that delegates this call to all of its members

Does that make some kind of sense?

test/unittests/commands/excel2json/test_properties.py
BalduinLandolt: Great, thanks!  
Hard to say where best to put those - then we would in principal have to go through all the tests and decide if they are integration tests or not... but probably it would make sense. Maybe e2e shoud even be a subfolder of integration?

src/dsp_tools/commands/excel2json/properties.py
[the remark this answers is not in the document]
BalduinLandolt: no, I didn't mean in terms of subclasses, but in terms of a dataclass representing the aggregated problem, that has three members, which each are dataclasses representing the sub-problems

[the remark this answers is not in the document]
Nora-Olivia-Ammann: I decided to leave out an aggregator class. The only further information it would have is the excel name. As the object is not currently handed around there seemed to me too little benefit to create an additional class, so I decided to add the name directly in the error msg string.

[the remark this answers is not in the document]
BalduinLandolt: for now, fair enough. In the long run we may want to re-think the fact that this method raises the error rather than just returning the problem directly

src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -316,9 +329,17 @@ def _do_property_excel_compliance(df: pd.DataFrame, excelfile: str) -> None:
...
-    _check_missing_values_in_row_raise_error(df=df, excelfile=excelfile)
+    problems: list[Any] = [
+        utl.check_contains_required_columns_else_raise_error(df=df, required_columns=required_columns),
+        utl.check_column_for_duplicate(df=df, to_check_column="name"),
+    ]
+    if missing_vals_check := _check_missing_values_in_row(df=df):
+        problems.extend(missing_vals_check)
+    if any(problems):
+        extra = [problem.execute_error_protocol() for problem in problems if problem]
+
```
jnussbaum, line 340: superfluous line break

src/dsp_tools/models/exceptions.py
```diff
@@ -64,7 +65,59 @@ def __str__(self) -> str:
...
+        default_msg = (
+            "\n\nAn internal error occurred.\n"
+            "Please contact the dsp-tools development team with the following information:\n"
+            "    - Which command was used.\n"
+            "    - If applicable, any files that were used in conjunction with the command.\n"
+            "    - A file with the terminal output copied into.\n"
+            "    - The log files called 'logging.log', if there are several, include all.\n"
+            f"      They can be found at: {Path.home() / Path('.dsp-tools')}\n"
+        )
+
```
jnussbaum, line 81: superfluous line break

src/dsp_tools/models/exceptions.py
```diff
@@ -64,7 +65,59 @@ def __str__(self) -> str:
...
+        default_msg = (
+            "\n\nAn internal error occurred.\n"
+            "Please contact the dsp-tools development team with the following information:\n"
+            "    - Which command was used.\n"
+            "    - If applicable, any files that were used in conjunction with the command.\n"
+            "    - A file with the terminal output copied into.\n"
+            "    - The log files called 'logging.log', if there are several, include all.\n"
+            f"     They can be found at: {Path.home() / Path('.dsp-tools')}\n"
+        )
+
```
jnussbaum, line 107: superfluous line break

test/unittests/commands/excel2json/test_properties.py
```diff
@@ -15,27 +13,15 @@
 from pandas.testing import assert_frame_equal
 
 from dsp_tools.commands.excel2json import properties as e2j
-from dsp_tools.models.exceptions import UserError
+from dsp_tools.models.exceptions import InputError
+from dsp_tools.models.input_error import InvalidExcelContentProblem
 
+excelfile = "testdata/excel2json/excel2json_files/test-name (test_label)/properties.xlsx"
+output_from_method, _ = e2j.excel2properties(excelfile, None)
```
jnussbaum, line 20: Perhaps not that important, but a while ago I learnt that there is a difference between function and method:
- function = standalone (without a class)
- method = inside a class

I did this naming wrong in many places in this codebase, because I didn't know the difference. But now that I learnt it, I thought I will spread the word :-)

jnussbaum (approved):

src/dsp_tools/models/input_error.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Currently they are, but I thought this may change in the future...?

src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -198,19 +207,24 @@ def _row2prop(df_row: pd.Series, row_num: int, excelfile: str) -> dict[str, Any]
...
-    # These functions may return None, this is checked before the update
-    _property = utl.update_dict_if_not_value_none(additional_dict=non_mandatory, to_update_dict=_property)
+
+    gui_attrib = _get_gui_attribute(df_row=df_row, row_num=row_num)
+    match gui_attrib:
+        case dict():
+            _property["gui_attributes"] = gui_attrib
+        case InvalidExcelContentProblem():
+            msg = f"There is a problem with the excel file: '{excelfile}'\n" + gui_attrib.execute_error_protocol()
+            raise InputError(msg) from None
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: These new classes are not actually exceptions. That is why they are called problem. This way it is easier to collect them than exceptions. It is not possible to raise a class that does not inherit from the Exceptions Base class To change this only in this case would in my point of view create a big inconsistency in the design.

I did not raise the exceptions on the gui_attrib class level because then the Problem classes have to know the excel name for a good error. I tried to keep the information in the classes as minimal as possible so that we can have "dumber" functions with less variables.

Additionally in my opinion the code why the try except is not much simpler.

But I agree the types in the return variables are quite horrible. We could discuss to do the pd.isnull check further outside. But I think that is something for another PR.

src/dsp_tools/models/input_error.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+separator = "\n    "
+list_separator = "\n    - "
+
+
+class Problem(Protocol):
+    """Information about input errors."""
+
+    def execute_error_protocol(self) -> str:
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Currently that is all the method does. But in the future it may also create and save a df etc. (Since we adapt the type of information needed here). Still the method have to be called the same so that the protocol works. 

At first I create a private function called create error_msg which was called by this function. But since we do not (at least at this point) create dfs or other files, it seemed superfluous. I only wanted to create this option when needed and as this is a ontology file, there won't be that so many errors that it won't fit the console.

src/dsp_tools/commands/excel2json/properties.py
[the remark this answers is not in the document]
BalduinLandolt: additionally, we could either alias `dict[str, int | str | float]` or - imo even better - create a model for gui attribute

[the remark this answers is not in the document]
Nora-Olivia-Ammann: I think ultimately, a model for a resource, list and property would be a good idea. But I think implementing the errors before is still a good idea as most of them come before the construction process is started. The reason why this is not checked before is because then the construction would have to be done twice, I did not want to do that.

PR #656, opened 2023-11-28
chore(process-files): diverse improvements

jnussbaum:


BalduinLandolt (approved):
there are things in this one that could be improved, but I would say we can just merge it as is and eventually improve stuff more systematic when we touch this file again

PR #659, opened 2023-11-30
refactor(excel2json): modularise unittests (DEV-3040)

Nora-Olivia-Ammann:


Nora-Olivia-Ammann:
As discussed with Balduin (https://github.com/dasch-swiss/dsp-tools/pull/654#discussion_r1410405743)
We plan to change some of these test to a integration test module. The modularisation for the file test_properties is in the PR from above, as tests needed to be changed there anyway. I would like to do the change to integration in a separate PR after both of the PRs are merged, so that we can start cleanly.

Nora-Olivia-Ammann:
> This PR produced a pretty gnarly diff, so I did more of a skimming than a proper review - but I didn't see any show stoppers, and I like the general notion of the PR, so I would just trust you on this one

Thanks. The only conceptual difference I made is that I got rid of this temp file writing and reading. So if that makes sense to you, the rest is just more pytest and breaking up big functions, but no content change.

Nora-Olivia-Ammann:
> > > This PR produced a pretty gnarly diff, so I did more of a skimming than a proper review - but I didn't see any show stoppers, and I like the general notion of the PR, so I would just trust you on this one
> > 
> > 
> > Thanks. The only conceptual difference I made is that I got rid of this temp file writing and reading. So if that makes sense to you, the rest is just more pytest and breaking up big functions, but no content change.
> 
> It absolutely does. You could consider turning `test_project.py` into pytest too, if you like.

That is a mish-mash I am not fond of myself. But The reason why it isn't, is because I could not find a simple and easy pytest equivalence for dictEqual at all, so I would have to compare keys and then iterate etc. So I kept the unittest. This is the same in the other files.

PR #664, opened 2023-12-01
feat(excel2json): new error implementation for json validation (DEV-3047)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/input_error.py
```diff
@@ -126,3 +126,39 @@ def execute_error_protocol(self) -> str:
...
+
+
+@dataclass(frozen=True)
+class JsonValidationResourceProblem:
+    """This class contains information about a JSON resource section that fails its validation against the schema."""
+
+    problematic_resource: str | None = None
+    excel_sheet: str | None = None
+    excel_column: str | None = None
+    excel_row: int | None = None
```
BalduinLandolt, line 138: I would assume that these are always defined or undefined at the same time (you can't have a problem in a row without a column or without a sheet), so it would make sense to me to define a class `ExcelPosition` or `TableField` or something like that, that can bundle and serialize these three datapoints.

src/dsp_tools/commands/excel2json/input_error.py
```diff
@@ -126,3 +126,39 @@ def execute_error_protocol(self) -> str:
...
+            f"{separator}Section of the problem: 'Resources'",
+        ]
+        if self.problematic_resource:
+            msg.append(f"Problematic Resource '{self.problematic_resource}'")
+        if self.excel_sheet:
+            msg.append(f"The problem is caused by the value in the Excel sheet '{self.excel_sheet}'")
+        if self.excel_row:
+            msg.append(f"The problem is caused by the value in the Excel row {self.excel_row}")
+        if self.excel_column:
+            msg.append(f"The problem is caused by the value in the Excel column '{self.excel_column}'")
```
BalduinLandolt, line 159: if my assumption is correct (see above) that these are all defined in the same cases, then they could be bundled:

```suggestion
        if self.excel_sheet and self.excel_row and self.excel_column:
            msg.append(f"The problem is caused by the value in the Excel sheet '{self.excel_sheet}' at row {self.excel_row}, column {self.excel_column}.")
```

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -40,52 +38,58 @@ def _validate_resources(
...
-
-    # check if resource names are unique
-    all_names = [r["name"] for r in resources_list]
-    if duplicates := {
-        index + 2: resdef["name"] for index, resdef in enumerate(resources_list) if all_names.count(resdef["name"]) > 1
-    }:
-        err_msg = (
-            f"Resource names must be unique inside every ontology, "
-            f"but your Excel file '{excelfile}' contains duplicates:\n"
+        err_msg = _search_json_validation_error_get_err_msg_str(
```
BalduinLandolt, line 41: I would simplify this name - how about `_find_validation_problem()`? And if you leave out the parameter names in the call, it would even fit on one line. (and since the params are of different type, there is no danger of inverting them by accident, the type checker would not allow this.)

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -40,52 +38,58 @@ def _validate_resources(
...
+    validation_error: jsonschema.ValidationError, resources_list: list[dict[str, Any]]
+) -> JsonValidationResourceProblem:
+    if json_path_to_resource := regex.search(r"^\$\[(\d+)\]", validation_error.json_path):
+        # fmt: off
+        wrong_res_name = (
+            jsonpath_ng.ext.parse(json_path_to_resource.group(0))
+            .find(resources_list)[0]
+            .value["name"]
+        )
+        # fmt: on
```
BalduinLandolt, line 59: why does this require `fmt: off`? would it work better if it were split up into multiple statements?

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -40,52 +38,58 @@ def _validate_resources(
...
-        for row_no, resname in duplicates.items():
-            err_msg += f" - Row {row_no}: {resname}\n"
-        raise UserError(err_msg)
-
-    return True
+        msg = f"\nThe Excel file '{excelfile}' did not pass validation." + err_msg.execute_error_protocol()
+        raise InputError(msg) from None
+
+
+def _search_json_validation_error_get_err_msg_str(
```
BalduinLandolt, line 49: I think even this function could be split up into smaller functions

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -118,7 +122,7 @@ def _row2resource(
     # load the cardinalities of this resource
     # if the excel sheet does not exist, pandas raises a ValueError
     try:
-        details_df = utl.read_and_clean_excel_file(excelfile=excelfile, sheetname=name)
+        details_df = read_and_clean_excel_file(excelfile=excelfile, sheetname=name)
     except ValueError as err:
         raise UserError(str(err)) from None
```
BalduinLandolt, line 127: I know this is not in scope of your PR - but this just screams for refactoring... :)

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -216,8 +221,22 @@ def excel2resources(
...
+    all_names = [r["name"] for r in resources]
+    if duplicates := {
+        index + 2: resdef["name"] for index, resdef in enumerate(resources) if all_names.count(resdef["name"]) > 1
+    }:
+        err_msg = (
+            f"Resource names must be unique inside every ontology, "
+            f"but your Excel file '{excelfile}' contains duplicates:\n"
+        )
+        for row_no, resname in duplicates.items():
+            err_msg += f" - Row {row_no}: {resname}\n"
```
BalduinLandolt, line 234: how about moving this logic into a util function? this seems too low level for such a high level function

src/dsp_tools/commands/excel2json/utils.py
```diff
@@ -78,7 +78,7 @@ def check_contains_required_columns_else_raise_error(
         required_columns: set of column names
 
     Returns:
-        An object if
+        An object if there is a problem else None.
```
BalduinLandolt, line 81: if you read the prose docstring, the args and returns section get 100% redundant. Could this be removed? Or otherwise the prose section reduced to " This function checks that a dataframe contains all required columns. Additional columns are allowed."?

BalduinLandolt (approved):
Nice, thanks. All my comments are purely cosmetic, feel free to ignore them

src/dsp_tools/commands/excel2json/resources.py
jnussbaum: The`fmt: off` allows putting it on several lines. If we remove the `fmt: off`, ruff and/or black will put it on 1 line. I think readability is better if it is split on several lines.

src/dsp_tools/commands/excel2json/utils.py
jnussbaum: In my opinion, the prose section shouldn't rephrase the args and returns sections.

jnussbaum (approved):
LGTM

src/dsp_tools/commands/excel2json/resources.py
BalduinLandolt: makes sense, thanks for the explanation. I think it would be worth trying to see how it looks with multiple statements

src/dsp_tools/commands/excel2json/resources.py
Nora-Olivia-Ammann: My intention for this file is to add fail fast behaviour and implement the new error protocoll there. I started to do it in one PR but the diff got too big (https://github.com/dasch-swiss/dsp-tools/pull/663/files) and I will let this one die later.

For my next PR I am planning to change all this excel validation behaviour, this is why I did not touch any of the other errors but the validation.

src/dsp_tools/commands/excel2json/resources.py
Nora-Olivia-Ammann: It is absolutely only parked there. This was in the json validation code, which is not actually what it is doing. When I do the fast fail (as mentioned above) this will get checked before anything is being constructed. So in the next PR it will be gone.

PR #668, opened 2023-12-05
fix: date validation forbids legal values (DEV-3059)

jnussbaum:


test/unittests/commands/excel2xml/test_excel2xml_lib.py
```diff
@@ -396,8 +396,9 @@ def test_make_date_prop(self) -> None:
             "2022",
             "GREGORIAN:CE:0476-09-04:CE:0476-09-04",
             "GREGORIAN:CE:2014-01-31",
+            "JULIAN:BC:1:AD:200",
         ]
-        invalid_values = ["GREGORIAN:CE:0476-09-04:CE:09-04", "GREGORIAN:CE:0476-09-010:CE:0476-09-04"]
+        invalid_values = ["GREGORIAN:CE:0476-09-010:CE:0476-09-04"]
```
jnussbaum, line 401: Unfortunately, if we accept years of 1-3 digits, this no longer produces an error. Because 09 could be the year, 04 the month, and the day could be omitted... So we just have to live with the fact that if we allow 1-3 digit years, we have to make validation less strict.

jnussbaum:
@Nora-Olivia-Ammann @BalduinLandolt Docs building fails, because of an outdated link. I opened a separate PR for that.

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -18,6 +18,7 @@
 from dsp_tools.commands.excel2xml.propertyelement import PropertyElement
 from dsp_tools.models.exceptions import BaseError
 from dsp_tools.models.helpers import DateTimeStamp
+from dsp_tools.utils.date_util import _full_date_pattern
```
BalduinLandolt, line 21: it should never be necessary to import a private member like this. Either we make the pattern public, or - which I would prefer - we expose a function that does the check for you. Then the caller also doesn't have to know that they need the verbose flag for the regex

BalduinLandolt (approved):
I suggested one improvement, apart from that it looks good to me

test/unittests/commands/excel2xml/test_excel2xml_lib.py
Nora-Olivia-Ammann: no we can specify in a regex that we do not allow numbers to start with a 0 through [1-9]

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
Nora-Olivia-Ammann: I agree.

Nora-Olivia-Ammann (approved):

test/unittests/commands/excel2xml/test_excel2xml_lib.py
jnussbaum: But this won't resolve the problem at hand, or would it?

src/dsp_tools/utils/date_util.py
```diff
@@ -8,16 +8,21 @@
 
 from dsp_tools.models.exceptions import BaseError
 
-_full_date_pattern = r"""
+_calendar = r"GREGORIAN|JULIAN|ISLAMIC"
+_era = r"CE|BCE|BC|AD"
+_year = r"\d{1,4}"
```
Nora-Olivia-Ammann, line 13: here we would need to specify r"[1-9]\d{0,3}"

jnussbaum: There might well be users who pad a date with zeroes. And even our test data often makes it, see here: https://github.com/dasch-swiss/dsp-tools/blob/4afc63ff47f117f6632de5b4316b628942c9e396/testdata/xml-data/test-data-systematic.xml#L117. 
So I don't think it would be wise to do that.

And besides, I don't think it would solve the validation problem: If we allow 1-3 digit years, it is impossible (or at least very tricky) to forbid the test case I have removed.

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
jnussbaum: Thanks for the suggestion, I adapted it.

src/dsp_tools/utils/date_util.py
Nora-Olivia-Ammann: Ok. I understand your case.

PR #670, opened 2023-12-05
feat(ingest-upload): create new ingest xmlupload cli command (DEV-3019)

Nora-Olivia-Ammann:


src/dsp_tools/commands/ingest_xmlupload/apply_ingest.py
```diff
@@ -0,0 +1,56 @@
...
+    This functions returns the information to replace the original filepaths with the identifier from dsp-ingest.
+
+    Returns:
+        dictionary with original: identifier from dsp-ingest
+
+    Raises:
+        InputError: if no file was found
+    """
+    if filepath := glob.glob("mapping-*.csv"):
+        df = pd.read_csv(filepath[0])
```
BalduinLandolt, line 24: this requires there to be only one mapping file, otherwise it will find the wrong one, which would lead the following steps go wrong. I'm not sure this is a problem (as it is unlikely that users have multiple mapping files in the directory. But alternatively, the mapping file could be passed as a CLI argument (like we do with the project json or the upload xml files), which would be more verbose but also a bit more stable.

src/dsp_tools/commands/ingest_xmlupload/apply_ingest.py
```diff
@@ -0,0 +1,56 @@
...
+    else:
+        raise InputError("No mapping csv file was found in the directory.")
+
+
+def replace_filepath_with_sipi_uuid(
+    xml_tree: etree._ElementTree[etree._Element],
+    orig_path_2_uuid_filename: dict[str, str],
+) -> tuple[etree._ElementTree[etree._Element], IngestInformation]:
+    """
+    Replace the original filepaths in the <bitstream> gags by the uuid filenames of the uploaded files.
```
BalduinLandolt, line 36:
```suggestion
    Replace the original filepaths in the <bitstream> tags by the uuid filenames of the uploaded files.
```

src/dsp_tools/commands/ingest_xmlupload/apply_ingest.py
```diff
@@ -0,0 +1,56 @@
...
+        print(f"The file '{filepath[0]}' is used to map the internal SIPI image IDs to the original filepaths.")
+        return dict(zip(df["original"].tolist(), df["derivative"].tolist()))
+    else:
+        raise InputError("No mapping csv file was found in the directory.")
+
+
+def replace_filepath_with_sipi_uuid(
+    xml_tree: etree._ElementTree[etree._Element],
+    orig_path_2_uuid_filename: dict[str, str],
+) -> tuple[etree._ElementTree[etree._Element], IngestInformation]:
```
BalduinLandolt, line 34: an alternative to changing the XML (which I always feel a bit uneasy when we do that) would be to loop over the created objects and modify them according to the mapping. For now, with the `XMLResource` object being tightly coupled to XML which it requires to create an instance, this would probably not be super convenient - but in the long run, once we have decoupled these things, I would actually prefer that approach.

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
```diff
@@ -0,0 +1,67 @@
...
+    xml_file: str,
+    user: str,
+    password: str,
+    dsp_url: str,
+    sipi_url: str,
+) -> None:
+    """
+    This function reads an XML file
+    and imports the data described in it onto the DSP server,
+    using the fast XML upload method.
```
BalduinLandolt, line 24: "fast XML upload" has quasi become a terminus for the workflow with the local processing, which is not what we do here. So maybe this should be re-phrased?

test/unittests/commands/ingest_xmlupload/test_user_information.py
```diff
@@ -0,0 +1,115 @@
...
+                "unused_media1",
+                "unused_media2",
+                "unused_media3",
+            ],
+            [
+                "no_up0",
+                "no_up1",
+                "no_up2",
+                "no_up3",
+            ],
```
BalduinLandolt, line 64: wouldn't we have the same test confidence with 1-2 elements in such test objects? It's not much of a difference, but I think the test file might look a bit more concise

testdata/dsp-ingest_data/dsp-ingest.json
```diff
@@ -0,0 +1,92 @@
...
+            "labels": {"en": "Attribution"},
+            "gui_element": "SimpleText"
+          },
+          {
+            "name": "hasSourceLink",
+            "super": ["hasValue"],
+            "object": "UriValue",
+            "labels": {"en": "External link"},
+            "gui_element": "SimpleText"
+          }
```
BalduinLandolt, line 51: to keep the test to the point, I don't think you need any property except `hasImage`, right?

testdata/dsp-ingest_data/dsp-ingest.xml
```diff
@@ -0,0 +1,108 @@
...
+        <allow group="KnownUser">V</allow>
+        <allow group="ProjectMember">D</allow>
+        <allow group="ProjectAdmin">CR</allow>
+        <allow group="Creator">CR</allow>
+    </permissions>
+    <permissions id="prop-restricted">
+        <allow group="ProjectMember">M</allow>
+        <allow group="ProjectAdmin">CR</allow>
+        <allow group="Creator">CR</allow>
+    </permissions>
```
BalduinLandolt, line 26: we probably don't need so many (if any?) permissions for a test like this, right?

testdata/dsp-ingest_data/dsp-ingest.xml
BalduinLandolt, line 1: I love to see so many dogs, here! :D  
But for brevity reasons, I think their number could be reduced to 1-2 to have the same test confidence; and although dogs need names and it's good to attribute images, I would recommend reducing test data to the bare minimum, meaning that all this non-required information would best be removed.

BalduinLandolt (approved):
Looks good, just some minor comments and one typo

src/dsp_tools/commands/ingest_xmlupload/apply_ingest.py
Nora-Olivia-Ammann: That sounds like a good suggestion.

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
Nora-Olivia-Ammann: My bad, I copy pasted this function and did not catch that, thanks.

testdata/dsp-ingest_data/dsp-ingest.json
Nora-Olivia-Ammann: The reason for the test data as it is, is that I want to have an image in a parent folder, one in a subfolder and one in each that has the same name. This way I think we covered the everything. Regarding the other properties, although the images are in the public domain wikimedia asks to add a attribution. Since this will be publicly available, I wanted to do it correctly. But I can delete the hasName.

Nora-Olivia-Ammann: I understand about the simplicity. I would however like to have distinct pictures that I can differentiate from each other. I took care to download the lowest quality. It was easier than taking my own pictures and reducing the quality...

src/dsp_tools/commands/ingest_xmlupload/apply_ingest.py
Nora-Olivia-Ammann: What I do not want to loose though is the check if all the media in the mapping file was used or not, and which media  was not found. At the end of the substitution rather than stopping as soon as one was unsuccessful. And I am not sure how to implement that without having to hand around a lot of information or doing the same work (checking for bitstream) twice, which I am also not a fan of...

src/dsp_tools/commands/ingest_xmlupload/user_information.py
```diff
@@ -0,0 +1,137 @@
...
+class UserInformation(Protocol):
+    """
+    Information about user messages.
+    """
+
+    def all_good_msg(self) -> str:
+        """
+        If everything went as it should this function return a success message for the user.
+        Returns:
+
```
jnussbaum, line 29:
```suggestion
```

test/unittests/commands/ingest_xmlupload/test_user_information.py
```diff
@@ -0,0 +1,103 @@
...
+        res_msg = IngestInformation([], [("no upload id", "media path")])._get_error_msg()
+        assert res_msg == expected
+
+    def test_all_problem_with_df_msg(self) -> None:
+        expected = (
+            "The upload cannot continue as there are problems with the media referenced in the XML.\n"
+            "    Media was uploaded to Sipi which was not referenced in the XML file.\n"
+            "    The file 'UnusedMediaUploadedInSipi.csv' was saved in '.' with the filenames.\n\n"
+            "    Media was referenced in the XML file but not previously uploaded to sipi:\n"
+            "    The file 'NotUploadedFilesToSipi.csv' was saved in '.' with the resource IDs and filenames."
```
jnussbaum, line 50: For me, this message is not clear. Perhaps we discuss it together in person how to reformulate?

test/unittests/commands/ingest_xmlupload/test_user_information.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def test_unused_media_to_df() -> None:
+    unused_media_list = [
+        "unused_media0",
+        "unused_media1",
+    ]
+    expected = pd.DataFrame({"Media Filenames": unused_media_list})
+    res_df = IngestInformation(unused_media_list, [])._unused_media_to_df()
+    assert_frame_equal(res_df, expected)
```
jnussbaum, line 72: I suggest importing this differently:
```
import pandas.testing
...
pandas.testing.assert_frame_equal(res_df, expected)
```

That's more verbose, but gives the reader more context. On first sight, it's not clear that `assert_frame_equal` is a pandas function.

test/unittests/commands/ingest_xmlupload/test_user_information.py
```diff
@@ -0,0 +1,103 @@
...
+        assert res_msg == expected
+
+
+def test_unused_media_to_df() -> None:
+    unused_media_list = [
+        "unused_media0",
+        "unused_media1",
+    ]
+    expected = pd.DataFrame({"Media Filenames": unused_media_list})
+    res_df = IngestInformation(unused_media_list, [])._unused_media_to_df()
```
jnussbaum, line 71: It took me a while before I understood that you monkey-patch the module `user_information` on lines 11-12. I think that's not a good practice, because it can be confusing (hard to understand). 

An alternative could be that in `user_information.py`, you transform the module-level variables into class attributes of the class `IngestInformation`, like this:

```
class IngestInformation:
    csv_filepath = field(default=Path.home(), init=False)
```

With init=False, you can exclude it from the `__init__()`, so that you don't have to pass it as parameter every time you create an instance.

For testing purposes, you can then create an instance of IngestInformation with a custom csv_filepath.

testdata/dsp-ingest_data/dsp-ingest.xml
jnussbaum, line 1: A repo-wide search for `dsp-ingest.json`, or `dsp-ingest.xml` show 0 occurrences. So these files are actually unused? And all the dog images are also unused? (I mean, you reference them in the unit tests, but for that purpose, no file has to exist...)

jnussbaum (changes requested):
Looks generally good. I have a lot of renaming suggestions, which you can adapt, or also ignore :-) But 2 things should be addressed:
 - the new image format jpx that you added
 - the unused test data

src/dsp_tools/commands/ingest_xmlupload/apply_ingest.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Replace the original filepaths in the <bitstream> tags by the uuid filenames of the uploaded files.
+
+    Args:
+        xml_tree: The parsed original XML tree
+        orig_path_2_uuid_filename: Mapping from original filenames to uuid filenames from the mapping.csv
+
+    Returns:
+        The XML tree with the replaced filepaths (modified in place)
```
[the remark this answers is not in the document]
BalduinLandolt: You may want to have a look at my comment above: In the long run, we should not modify the XML at all, but rather we should think of DSP-TOOLS as a data pipeline which applies a series of transformations to the data that flows through. Ideally, this would just be a transformation step by itself, which would then render this entire question obsolete. And for that reason I'm not sure we have to necessarily follow this kind of best practice here.  
But yes, in principle I think you're right.

src/dsp_tools/commands/ingest_xmlupload/apply_ingest.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    no_uuid_found = []
+    used_media_paths = []
+    for elem in xml_tree.iter():
+        if etree.QName(elem).localname.endswith("bitstream"):
+            if (img_path := elem.text) in orig_path_2_uuid_filename:
+                elem.text = orig_path_2_uuid_filename[img_path]
+                used_media_paths.append(img_path)
+            else:
+                no_uuid_found.append((cast("etree._Element", elem.getparent()).attrib.get("id"), elem.text))
```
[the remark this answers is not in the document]
BalduinLandolt: yes, please! `from __future__` should go away soon-ish (not even sure it's needed in python 3.12), and it's generally better for tooling than the quoted variant

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Args:
+        xml_file: path to XML file containing the resources
+        user: the user's e-mail for login into DSP
+        password: the user's password for login into DSP
+        dsp_url: URL to the DSP server
+        sipi_url: URL to the Sipi server
+
+    Raises:
+        InputError: if any media was not uploaded of uploaded media was not referenced.
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: should be or

src/dsp_tools/commands/ingest_xmlupload/apply_ingest.py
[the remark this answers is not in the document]
jnussbaum: To prevent misunderstandings: With "should go away soon-ish" you mean that it will soon become default behavior, so that it isn't necessary any more. You mean that quoting type annotations is not a good idea. So, you don't mean that `from __future__ import annotations` should be removed from this file here.

I'm not sure, but the last time I read something about type hints handling in python, they wrote that `from __future__ import annotations` won't disappear soon. I think we still need it now, and even in the middle term future.

test/unittests/commands/ingest_xmlupload/test_user_information.py
Nora-Olivia-Ammann: I understand your point, but does it make sense to create a class variable that we only want because of testing?

How would you solve the maximum print statements? If we keep the current number it would bloat the test data enormously...

testdata/dsp-ingest_data/dsp-ingest.xml
Nora-Olivia-Ammann: I was planning on an e2e test, which I did not yet write as the ingest does not actually work.

test/unittests/commands/ingest_xmlupload/test_user_information.py
jnussbaum: The only elegant solution I see is to turn the max print statements into a class attribute, too. 
And I don't think we should do that only for testing purposes. It's not considered good practice to have module-level variables at all. So anyway, it would be better to turn them into class attributes.

testdata/dsp-ingest_data/dsp-ingest.xml
jnussbaum: Ah okay, I see. Sorry, I was quite confused by those files :-)

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -155,7 +155,7 @@ def _make_bitstream_file_value(bitstream_info: BitstreamInfo) -> dict[str, Any]:
         case "mp4":
             prop = "knora-api:hasMovingImageFileValue"
             value_type = "MovingImageFileValue"
-        case "jpg" | "jpeg" | "jp2" | "png" | "tif" | "tiff":
+        case "jpg" | "jpeg" | "jp2" | "png" | "tif" | "tiff" | "jpx":
```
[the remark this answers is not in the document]
jnussbaum: Ah okay, I see. could you explain that in an inline comment?

src/dsp_tools/commands/ingest_xmlupload/apply_ingest.py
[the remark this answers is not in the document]
jnussbaum: Hmmm... You are speaking in riddles... What comment above? I don't find it.

test/unittests/commands/ingest_xmlupload/test_user_information.py
Nora-Olivia-Ammann: I will do that almost as said. Because if you give inti=False then you can never hand over another value, which we want (https://docs.python.org/3/library/dataclasses.html#default-factory-functions)

test/unittests/commands/ingest_xmlupload/test_apply_ingest.py
```diff
@@ -0,0 +1,61 @@
...
+        reference_dict = {"images/Fluffy.jpg": "fluffy_uuid"}
+        res_tree, res_msg = replace_filepath_with_sipi_uuid(root, reference_dict)  # type: ignore[arg-type]
+        assert not res_msg.media_no_uuid
+        assert not res_msg.unused_media_paths
+        expected = (
+            b'<knora><resource label="Fluffy Image" restype=":Image2D" id="Fluffy1" '
+            b'permissions="res-default"><bitstream permissions="prop-default">fluffy_uuid</bitstream>'
+            b"</resource></knora>"
+        )
+        assert etree.tostring(res_tree) == expected
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: That is really weird because for me it often does not work, sometimes it does for whatever reasons.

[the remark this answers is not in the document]
jnussbaum: It is indeed tricky to not getting knocked out by additional spaces and/or newlines. It took me a while to figure out the suggestion I gave you in my first comment :-)

But I think it's worth it, because unformatted XML-strings are horror :-)

src/dsp_tools/commands/ingest_xmlupload/user_information.py
```diff
@@ -0,0 +1,137 @@
...
+list_separator = "\n    - "
+
+maximum_prints = 20
+
+csv_filepath = Path.home()
+unused_media_filename = "UnusedMediaUploadedInSipi.csv"
+no_uuid_filename = "NotUploadedFilesToSipi.csv"
+
+
+class UserInformation(Protocol):
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I started a protocol because of the assumption that other errors will join that will also be of the same protocol. But until that is the case we can delete it.

src/dsp_tools/commands/ingest_xmlupload/user_information.py
```diff
@@ -0,0 +1,137 @@
...
+        If that is the case it returns the message,
+        if not, it ends without an effect.
+
+        Returns:
+            Message if all went well.
+        """
+        if not self.unused_media_paths and not self.media_no_uuid:
+            return (
+                "All media referenced in the XML file were uploaded to sipi.\n"
+                "No media was uploaded to sipi that was not referenced in the XML file."
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: But they are uploaded in sipi, just through ingest.

test/unittests/commands/ingest_xmlupload/test_user_information.py
Nora-Olivia-Ammann: I disagree that it it sufficiently beneficial to be more verbose. If you encounter new code / code you have not written it can be expected that you encounter things you don't yet know. We import a lot of things without declaring where it comes from in the code itself. I think it is not too much work and too much to expect that if a person does not know that, they can scroll to the top of the file or do a search and then it becomes obvious. This is something that will be done once and from then it is clear.

test/unittests/commands/ingest_xmlupload/test_apply_ingest.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: No what I actually mean is that tests that have not failed in the past, now suddenly start to fail...

src/dsp_tools/commands/ingest_xmlupload/apply_ingest.py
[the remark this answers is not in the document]
BalduinLandolt: sorry, yes, in file view and before it was marked as resolved, it was easier to see ^^  
here you go: https://github.com/dasch-swiss/dsp-tools/pull/670/files/f2ed9e9877d78287ad5c2984e319baf4176216b2#r1421403136

src/dsp_tools/commands/ingest_xmlupload/apply_ingest.py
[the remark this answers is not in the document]
BalduinLandolt: Correct, I mean that we should use `from __future__ import annotations` rather than quoted type annotations, because it will be easier to change, once the annotations are default behaviour.  
And you're right, this seems to be an ongoing thing - I just did a bit of reading on it... and things seem to move slowly - so for now we will just have to live with the future imports

test/unittests/commands/ingest_xmlupload/test_apply_ingest.py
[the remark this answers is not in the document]
BalduinLandolt: The main danger of failing tests when using multiline strings (at least for me) has always been the fact that there may be different whitespace (newlines that are easy to forget, indentation spaces, etc.), so the best thing I found was not to compare strings in tests (whitespace shouldn't matter!) but to compare actual information content wherever possible.  
In this case, you actually compare two XML strings, one of which you let lxml serialize and the other one is hard coded. Why not ask lxml for the thing you're actiually interested in (namely the text value inside the `<bitstream>` tag) which will either be `"images/Fluffy.jpg"` or `"fluffy_uuid"`, depending on if the function does what it should - so you could just assert that.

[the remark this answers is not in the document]
BalduinLandolt: (maybe not the most clear comment from my side... sorry! does it make sense?)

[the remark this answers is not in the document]
Nora-Olivia-Ammann: It does to me, thanks! The only thing that could be a problem with your method is that we don't check, that the rest of the string is unchanged...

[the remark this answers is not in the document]
BalduinLandolt: that's true - I'm not sure we have to test that, though, because unless there is good reason why we think it might (in which case that should be tested specifically), it's essentially just a guessing/hoping game, if the test string fulfils the condition where this might happen. But it's up to you if you deem that necessary

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
```diff
@@ -0,0 +1,70 @@
...
+
+    shortcode = xml_tree_orig.getroot().attrib["shortcode"]
+    orig_path_2_uuid_filename = get_mapping_dict_from_file(shortcode)
+    xml_tree_replaced, ingest_info = replace_filepath_with_sipi_uuid(
+        xml_tree=xml_tree_orig,
+        orig_path_2_uuid_filename=orig_path_2_uuid_filename,
+    )
+    if good := ingest_info.all_good_msg():
+        print(good)
+        logger.info(good)
```
BalduinLandolt, line 56: this reminds me a lot of some typical patterns in the Go programming language; and there the variable would conventionally be called `ok`, not `good`. But that's just a remark, not that it matters...

test/e2e/commands/test_ingest_xmlupload.py
```diff
@@ -0,0 +1,48 @@
...
+import pytest
+
+from dsp_tools.commands.ingest_xmlupload.upload_xml import ingest_xmlupload
+from dsp_tools.models.exceptions import InputError
+
+
+@pytest.fixture()
+def base_path() -> Path:
+    """Get the root directory of the repository."""
+    return Path(__file__).parents[3]
```
BalduinLandolt, line 12: this looks hacky... but as long as it works... :)

test/unittests/commands/ingest_xmlupload/test_apply_ingest_uuid.py
```diff
@@ -0,0 +1,72 @@
...
+        res_tree, res_msg = replace_filepath_with_sipi_uuid(root, reference_dict)
+        assert not res_msg.media_no_uuid
+        assert not res_msg.unused_media_paths
+        expected = (
+            b"<knora>"
+            b'<resource label="Fluffy Image" restype=":Image2D" id="Fluffy1" permissions="res-default">'
+            b'<bitstream permissions="prop-default">fluffy_uuid</bitstream>'
+            b"</resource></knora>"
+        )
+        assert etree.tostring(res_tree) == expected
```
BalduinLandolt, line 28: I feel this kind of thing is better tested directly on the xml tree, rather than comparing serialization:  
Your assertion doesn't really care if the strings are the same, you really only care thet the XPath `./knora/resource/bitstream/text()` no longer contains "images/Fluffy.jpg", but rather "fluffy_uuid".  
I don't recall the syntax from the top of my head, but there are at least 3 ways (object navigation, xpath, css selectors) to get to a certain place in an lxml tree; and the node should give you its text; and then you can just compare the string values.  
(you could even generate the xml tree in lxml rather than from a string, but I'm not sure how nice that would look/feel)

BalduinLandolt (approved):
looks goos! Just some minor remarks

test/e2e/commands/test_ingest_xmlupload.py
Nora-Olivia-Ammann: The problem I had with the solution that I discuses with Johannes (shutil.copy) was, that it actually moved the file to the root directory. Meaning that even after the test session the csv was no longer in the folder it actually needs to be. (It took me the longest time to figure out, why I kept getting the wrong error messages) I could not find how I can prevent that with shutil so I went with this.

Any other ideas?

jnussbaum: I really don't like this hack, but I'm optimistic that we can find a better way. 

You said that https://docs.python.org/3.10/library/shutil.html#shutil.copy moved the file instead of copying it? I cannot believe that... Perhaps you deleted the wrong file after the `yield`?

jnussbaum: Yup, and it works:

```
@pytest.fixture()
def _retrieve_mapping_file() -> None:
    """Put the mapping file into the cwd."""
    mapping_file = Path("testdata/dsp-ingest-data/mapping-00A2.csv")
    shutil.copy(mapping_file, ".")
    yield
    Path(mapping_file.name).unlink()


@pytest.mark.usefixtures("_retrieve_mapping_file")
def test_ingest_xmlupload() -> None:
    print("test function that does nothing")
```

test/unittests/commands/ingest_xmlupload/test_apply_ingest_id.py
```diff
@@ -0,0 +1,72 @@
...
+        res_tree, res_msg = replace_filepath_with_sipi_id(root, reference_dict)
+        assert not res_msg.media_no_id
+        assert not res_msg.unused_media_paths
+        expected = (
+            b"<knora>"
+            b'<resource label="Fluffy Image" restype=":Image2D" id="Fluffy1" permissions="res-default">'
+            b'<bitstream permissions="prop-default">fluffy_id</bitstream>'
+            b"</resource></knora>"
+        )
+        assert etree.tostring(res_tree) == expected
```
jnussbaum, line 28: Me too, I strongly suggest comparing the content of the XML, not the string serialization. I made bad experiences with comparing XML strings...
Here an example how you can do it, and it's even simpler and shorter:

```suggestion
        assert res_tree.getroot()[0][0].text == "fluffy_id"
```

test/e2e/commands/test_ingest_xmlupload.py
Nora-Olivia-Ammann: thanks

Nora-Olivia-Ammann: > copying

The problem was I think that I didn't delete a file at all. What happened is that it disappeared from the original folder and ended up in the root folder...

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
```diff
@@ -0,0 +1,70 @@
...
+    user: str,
+    password: str,
+    dsp_url: str,
+    sipi_url: str,
+) -> None:
+    """
+    This function reads an XML file
+    and imports the data described in it onto the DSP server,
+    using the ingest XML upload method.
+    Before using this method,
```
jnussbaum, line 31:
```suggestion
    Before using this function,
```

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
```diff
@@ -0,0 +1,70 @@
...
+    xml_file: Path,
+    user: str,
+    password: str,
+    dsp_url: str,
+    sipi_url: str,
+) -> None:
+    """
+    This function reads an XML file
+    and imports the data described in it onto the DSP server,
+    using the ingest XML upload method.
```
jnussbaum, line 30:
```suggestion
```

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Args:
+        xml_file: path to XML file containing the resources
+        user: the user's e-mail for login into DSP
+        password: the user's password for login into DSP
+        dsp_url: URL to the DSP server
+        sipi_url: URL to the Sipi server
+
+    Raises:
+        InputError: if any media was not uploaded or uploaded media was not referenced.
```
jnussbaum, line 43: I think we shouldn't prevent the xmlupload if there are files that were ingested but not in the XML. Imagine you ingested 100'000 files, and by mistake 1 was doubled. Now you want to make the xmlupload, but you are not allowed because there is 1 superfluous file on the server. That would be quite annoying...

Or is there a reason why we should disallow it?

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
```diff
@@ -0,0 +1,70 @@
...
+    xml_tree_replaced, ingest_info = replace_filepath_with_sipi_id(
+        xml_tree=xml_tree_orig,
+        orig_path_2_id_filename=orig_path_2_id_filename,
+    )
+    if ok := ingest_info.ok_msg():
+        print(ok)
+        logger.info(ok)
+    else:
+        err_msg = ingest_info.execute_error_protocol()
+        logger.exception(err_msg)
```
jnussbaum, line 59: This shouldn't be done, see here for the reason: https://docs.python.org/3/library/logging.html#logging.exception.

Correct is:

```suggestion
        logger.error(err_msg)
```

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -156,7 +156,8 @@ def _make_bitstream_file_value(bitstream_info: BitstreamInfo) -> dict[str, Any]:
         case "mp4":
             prop = "knora-api:hasMovingImageFileValue"
             value_type = "MovingImageFileValue"
-        case "jpg" | "jpeg" | "jp2" | "png" | "tif" | "tiff":
+            # jpx is the extension of the files returned by dsp-ingest
```
jnussbaum, line 159: I suggest moving this comment 2 lines down, because it doesn't belong to mp4, but the images.

src/dsp_tools/utils/connection_live.py
```diff
@@ -240,7 +240,7 @@ def post(
         if not route.startswith("/"):
             route = f"/{route}"
         url = self.server + route
-        headers = {}
+        headers = {"X-Asset-Ingested": "true"}
```
jnussbaum, line 243: If you change that here, every single POST request that DSP-TOOLS makes, will have this header. I doubt that this is what we want... We should only send this header if a resource is created via ingest-xmlupload. That means, we have to parametrize this. I hope it won't be too much work, but I fear we cannot just leave it as it is.

test/e2e/commands/test_ingest_xmlupload.py
```diff
@@ -0,0 +1,52 @@
...
+from pathlib import Path
+
+import pytest
+
+from dsp_tools.commands.ingest_xmlupload.upload_xml import ingest_xmlupload
+from dsp_tools.models.exceptions import InputError
+
+
+@pytest.fixture()
+def _retrieve_mapping_file() -> None:  # type: ignore[misc]
```
jnussbaum, line 11: if you import Iterator from typing, then you don't need the ignore comment:

```suggestion
def _retrieve_mapping_file() -> Iterator[None]:
```

testdata/dsp-ingest-data/dsp-ingest.xml
jnussbaum, line 1: We have a subfolder `testdata/invalid-testdata` where all invalid files are collected. From your test, I see that this is an invalid file. So, you could also add the problem in the filename, so that this file would become `testdata/invalid-testdata/dsp-ingest/data-with-extra-file-and-uningested-file.xml`

testdata/dsp-ingest-data/dsp-ingest.json
jnussbaum, line 1: This file seems to be unused

testdata/dsp-ingest-data/images/Fluffy.jpg
jnussbaum, line 1: this file seems to be unused (for the unittests, the file doesn't have to exist)

testdata/dsp-ingest-data/images/GoodGirl.jpg
jnussbaum, line 1: this file seems to be unused (for the unittests, the file doesn't have to exist)

testdata/dsp-ingest-data/images/subFolder/Fluffy.jpg
jnussbaum, line 1: this file seems to be unused (for the unittests, the file doesn't have to exist)

testdata/dsp-ingest-data/images/subFolder/ShibaInu.jpg
jnussbaum, line 1: this file seems to be unused (for the unittests, the file doesn't have to exist)

test/unittests/commands/ingest_xmlupload/test_apply_ingest_id.py
```diff
@@ -0,0 +1,61 @@
...
+        xml = (
+            "<knora>"
+            '<resource label="Fluffy Image" restype=":Image2D" id="Fluffy1" permissions="res-default">'
+            '<bitstream permissions="prop-default">images/Fluffy.jpg</bitstream>'
+            "</resource>"
+            "</knora>"
+        )
+        root = etree.ElementTree(etree.fromstring(xml))
+        reference_dict = {"images/Fluffy.jpg": "fluffy_id"}
+        res_tree, res_msg = replace_filepath_with_sipi_id(root, reference_dict)
```
jnussbaum, line 19: The variable name `res_msg` is subpar. This is not a message (=str), but an `IngestInformation` object, so what about `ingest_info`?

test/unittests/commands/ingest_xmlupload/test_apply_ingest_id.py
```diff
@@ -0,0 +1,61 @@
...
+
+class TestReplaceBitstreamPaths:
+    def test_all_good(self) -> None:
+        xml = (
+            "<knora>"
+            '<resource label="Fluffy Image" restype=":Image2D" id="Fluffy1" permissions="res-default">'
+            '<bitstream permissions="prop-default">images/Fluffy.jpg</bitstream>'
+            "</resource>"
+            "</knora>"
+        )
```
jnussbaum, line 16: Personally, I prefer this:

```suggestion
        xml = """
        <knora>
            <resource label="Fluffy Image" restype=":Image2D" id="Fluffy1" permissions="res-default">
                <bitstream permissions="prop-default">images/Fluffy.jpg</bitstream>
            </resource>
        </knora>
        """
```

jnussbaum (changes requested):
Overall, it looks very good. Many of my comments are cosmetical, but some should be addressed before merging. They are commented, but here an overview:

- type annotation of IngestInformation.media_no_id
- should we disallow an xmlupload if there are files that were ingested but not in the XML?
- wrong file given in  `_get_unused_path_msg()`
- We shouldn't add the header "X-Asset-Ingested" to every single POST request that DSP-TOOLS makes, but only if a resource is created via ingest-xmlupload.
- unused test data files (should be removed)

src/dsp_tools/commands/ingest_xmlupload/user_information.py
```diff
@@ -0,0 +1,121 @@
...
+                "The data XML file contains references to the following multimedia files "
+                "which were not previously uploaded through dsp-ingest:"
+                + list_separator
+                + list_separator.join([f"Resource ID: '{x[0]}' | Filepath: '{x[1]}'" for x in self.media_no_id])
+            )
+        elif len(self.media_no_id) > self.maximum_prints:
+            return (
+                "The data XML file contains references to multimedia files "
+                "which were not previously uploaded through dsp-ingest:\n"
+                f"    The file with the resource IDs and filenames was saved at "
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: https://owl.purdue.edu/owl/general_writing/grammar/using_articles.html

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
jnussbaum: When I suggested to delete this line, it was not clear to me what the "ingest XML upload method" is. But probably that's the term you use for sending a request with the header "X-Asset-Ingested". Then it makes sense.

src/dsp_tools/commands/fast_xmlupload/upload_xml.py
```diff
@@ -118,7 +118,7 @@ def fast_xmlupload(
         password=password,
         imgdir=".",
         sipi=sipi_url,
-        config=UploadConfig(preprocessing_done=True),
+        config=UploadConfig(media_previously_uploaded=True),
```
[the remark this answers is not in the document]
jnussbaum: I don't associate the term "media" with "multimedia files". I would rather think of Washington Post, New York Times, etc. :-) For me, `multimedia_files` would be clearer. But I think your term will also be understood, so if you prefer that term, it's okay.

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
jnussbaum: I see your point. But if a user gets this error, is there a way to clean it up? I imagine it to be difficult to access the server and then delete the unused file, and then delete it from the CSV.

src/dsp_tools/commands/ingest_xmlupload/user_information.py
```diff
@@ -0,0 +1,121 @@
...
+from pathlib import Path
+from typing import Any
+
+import pandas as pd
+
+# pylint: disable=too-few-public-methods
+
+
+separator = "\n    "
+list_separator = "\n    - "
```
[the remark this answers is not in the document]
jnussbaum: Now I see that these 2 lines are repeated in 3 files. Then it would be better to define them only once, and import in the other files.

src/dsp_tools/commands/ingest_xmlupload/user_information.py
```diff
@@ -0,0 +1,121 @@
...
+@dataclass(frozen=True)
+class IngestInformation:
+    """
+    This class stores the information about the mapping of ids provided by the dsp-ingest service
+    and the filepaths used in the XML file.
+    """
+
+    unused_media_paths: list[str]
+    media_no_id: list[tuple[str, str] | Any]
+    maximum_prints: int = field(default=20)
```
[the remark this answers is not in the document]
jnussbaum: AFAIK, the only thing one should not do is specifying a mutable default argument, like list. For this, one should write `mylist = field(default_factory=list)`. But if it's a non-mutable default, I don't see a reason to use `field(default=x)`. 

The 2nd code example of https://docs.python.org/3/library/dataclasses.html#dataclasses.field suggests to me that `z` and `t` work the same way (copy pasted from there):

```
    z: int = field(repr=False, default=10)
    t: int = 20
```

But of course, this is cosmetical :-)

src/dsp_tools/commands/ingest_xmlupload/user_information.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Would that not be a little overkill? I like it to be consistent, but I am hesitant to import something where it is not crucial that it is identical.

src/dsp_tools/commands/xmlupload/resource_create_client.py
jnussbaum: But then, the indentation is misleading. I would then de-indent it, so that it is aligned with the `case`

src/dsp_tools/commands/fast_xmlupload/upload_xml.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: let's compromise and write media_files, then the variable is not too long

src/dsp_tools/commands/ingest_xmlupload/user_information.py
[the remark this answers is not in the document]
jnussbaum: From a general point of view, I think that repeating stuff is less scalable. Once we have many files, and each repeats this, it can be inadvertently modified in 1 place. So I think defining it only once is the cleaner way. But I agree that the separators aren't that important, so I leave it up to you :-)

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
Nora-Olivia-Ammann: Do you mean a different error message?

src/dsp_tools/commands/ingest_xmlupload/user_information.py
```diff
@@ -0,0 +1,121 @@
...
+        return (
+            pd.DataFrame(
+                {"Resource ID": [x[0] for x in self.media_no_id], "Filepath": [x[1] for x in self.media_no_id]}
+            )
+            if len(self.media_no_id) > self.maximum_prints
+            else None
+        )
+
+
+def _save_as_csv(df: pd.DataFrame, filepath: Path, filename: str) -> None:
```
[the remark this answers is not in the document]
jnussbaum: okay

src/dsp_tools/commands/ingest_xmlupload/user_information.py
```diff
@@ -0,0 +1,120 @@
...
+                "which were not previously uploaded through dsp-ingest:"
+                + list_separator
+                + list_separator.join([f"Resource ID: '{x[0]}' | Filepath: '{x[1]}'" for x in self.mediafiles_no_id])
+            )
+        elif len(self.mediafiles_no_id) > self.maximum_prints:
+            return (
+                "The data XML file contains references to multimedia files "
+                "which were not previously uploaded through dsp-ingest:\n"
+                f"    The file with the resource IDs and problematic filenames was saved at "
+                f"'{self.csv_directory_path}/{self.mediafiles_no_id_csv}'."
```
jnussbaum, line 74: This way of concatenating a Path object with a "/" to a string is not recommended. "/" might be the correct separator on UNIX, but not on windows. You can let pathlib do the work for you, it will always choose the right separator: 

```suggestion
                f"'{self.csv_directory_path / self.mediafiles_no_id_csv}'."
```

jnussbaum: But thinking about your variable self.csv_directory_path: I think it's redundant. If you save a file, it will always be saved in the cwd. I believe you can completely remove the variable self.csv_directory_path from this class.

See this code example:
```python
>>> assert Path("foo").absolute() == Path.cwd() / "foo"
```

src/dsp_tools/commands/ingest_xmlupload/user_information.py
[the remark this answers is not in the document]
jnussbaum: My reasoning is: For us devs, the file is known, we refer to it as "the file". But the user hears it the first time, it was just saved. So for the user, it would be more natural to hear that we saved "a" file for them. But that's a detail, so if you prefer "the", that's fine for me :-)

src/dsp_tools/commands/ingest_xmlupload/user_information.py
Nora-Olivia-Ammann: It served two purposes for me. If the path changed it is easy to modify at only one place.
It is easier for testing. This way I don't have to make these over complicated path patches, but can just overwrite it.

jnussbaum: okay, I didn't know that, that's a good reason

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
Nora-Olivia-Ammann: I think coping with bad user behaviour is something we only have to address once it is used widely. Apart from us, no one should use this route at the moment. To make it public we would need to add a start ingest command anyway, so I think the details of the implementation can wait.
As to more tolerance I strongly disagree, this code is the gate-keeper to bad user behaviour.
Uploading files that are never referenced and therefore float in nirvana may be user friendly but neither sustainable from a memory point of view nor good practice from a data modelling view. We had the discussion with Balduin about the circular references if we could not just have a requested number of IRIs. He declined that because this way the api allows is vulnerable to get data that is incorrect. This in my point of view is the same, user friendliness does not trump data integrity.

src/dsp_tools/utils/connection_live.py
```diff
@@ -240,13 +239,12 @@ def post(
         if not route.startswith("/"):
             route = f"/{route}"
         url = self.server + route
-        headers = {}
         if jsondata:
-            headers["Content-Type"] = "application/json; charset=UTF-8"
+            self.headers["Content-Type"] = "application/json; charset=UTF-8"
```
jnussbaum, line 243: I wouldn't store `headers` as a class attribute. Because once you make a POST request with jsondata, the header for ALL SUBSEQUENT requests will be "Content-Type" = "application/json; charset=UTF-8". So also for subsequent GET requests, and POST requests without jsondata. 

A cleaner solution would be to pass the "X-Asset-Ingested" really only for this one POST request that creates the resource.

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
jnussbaum: Okay, that's a valid argument. So we know that we will have to work diligently and cleanly, if we ingest data.

testdata/dsp-ingest-data/dsp-ingest.xml
jnussbaum: So would you prefer to put the invalid files together with the valid ones? i.e. would you move `testdata/invalid-testdata/excel2json/` to `testdata/excel2json/invalid`? That's worth considering, we could do that.

testdata/dsp-ingest-data/images/Fluffy.jpg
jnussbaum: Since GitHub is primarily a code repo (for code text files, and not for multimedia files), and since we don't want to blow up the size of the repo, I would be very reluctant with adding multimedia files if they are not strictly necessary. We already have test files in `testdata/bitstreams`, so I would reuse these.

P.S. Regarding your comment that the test files should be specific to every test, and not collected in 1 folder and reused multiple times: That's a good thought, which I will store "im Hinterkopf" for future tests that I write. But I think for multimedia files, we should rather reuse the ones in `testdata/bitstreams`. Otherwise, the repo size will grow and grow.

testdata/dsp-ingest-data/dsp-ingest.xml
Nora-Olivia-Ammann: For me it is less about validity but rather where it is used. We have a test project systematic, with a json and xml, these to me belong in one folder. Because they are one project...

testdata/dsp-ingest-data/images/Fluffy.jpg
Nora-Olivia-Ammann: About the bitstream, that is true, however would it be possible to remove the readme.txt (in a separate PR)? Because it is not actually a bitstream that is used...

src/dsp_tools/utils/connection_live.py
Nora-Olivia-Ammann: In this case, either I would have to have a attribute solely about the ingest, which also does not seem like a good solution or write another post function...?

src/dsp_tools/commands/ingest_xmlupload/apply_ingest_id.py
```diff
@@ -0,0 +1,67 @@
...
+    Returns:
+        dictionary with original: identifier from dsp-ingest
+
+    Raises:
+        InputError: if no file was found
+    """
+    filepath = Path(f"mapping-{shortcode}.csv")
+    if filepath.exists():
+        df = pd.read_csv(filepath)
+        msg = f"The file '{filepath}' is used to map the internal SIPI image IDs to the original filepaths."
```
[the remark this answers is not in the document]
BalduinLandolt: strictly speaking, mappings work both ways, but I would also say that the key is mapped to the value, not the other way around

src/dsp_tools/utils/connection_live.py
Nora-Olivia-Ammann: @BalduinLandolt you worked on this a lot. Do you have a suggestion?

src/dsp_tools/commands/ingest_xmlupload/user_information.py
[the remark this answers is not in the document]
BalduinLandolt: just to throw it out there: maybe the fact that these separators are repeated in multiple files is actually an indication that something else is repeated too, namely the logic of creating bulletpoint list strings; and that may be something to put in a little util function?

src/dsp_tools/commands/ingest_xmlupload/upload_xml.py
BalduinLandolt: I think this is something that should not be decided by devs, that's a product decision. It could very well be a legit use case that projects ingest all their files at once, but create the resources in batches; but at the same time it is indeed a door-opener to bad user behaviour. Which one we pick is up to PM in the end.  
But surely this shouldn't keep this PR waiting, so I think for now, either way would be ok.

src/dsp_tools/commands/ingest_xmlupload/user_information.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: That is something we could certainly look at but I think it it outside of the scope of this PR...

src/dsp_tools/utils/connection_live.py
BalduinLandolt: I would suggest adding the param `headers: dict[str, str] | None = None` to *all* methods (Get, Put, Post, Delete) both on the protocol and the implementation, and add it to the default headers, if it is defined; with the default value, it shouldn't break any existing code, but it would allow for adding any header anywhere if needed, which gives us a lot of flexibility also for the future.  
But in principal it wouldn't break anything to do it the way it is now - superfluous headers are just ignored. However, I don't think that's the most sustainable solution in the long run: We may run into a situation where we want to be more specific on when we send which header.

BalduinLandolt (approved):
I didn't do a detailed review now, but I feel we should try to get this merged soon-ish and try it out in practice. And if things need improvement after all, we can still do that in a next PR.

src/dsp_tools/commands/ingest_xmlupload/user_information.py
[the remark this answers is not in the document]
BalduinLandolt: fair enough

testdata/dsp-ingest-data/dsp-ingest.xml
jnussbaum: Good idea. If you like, you can restructure the test data in a separate PR.

testdata/dsp-ingest-data/images/Fluffy.jpg
jnussbaum: You mean `testdata/bitstreams/README.txt`? Initially, I didn't have it, but Balduin thought it would be good practice to have it. And if not there, where else should it live? It only makes sense to have it directly next to the data that it describes.

src/dsp_tools/utils/connection_live.py
jnussbaum: Thanks for the clarification, Balduin, that's what I meant. I strongly vote for Balduin's solution.

testdata/dsp-ingest-data/images/Fluffy.jpg
BalduinLandolt: I don't remember that... ^^  
feel free to remove it

jnussbaum (approved):

src/dsp_tools/utils/connection_live.py
jnussbaum: Okay, that's fine. I just approved. You can then do the header in a separate PR.

Nora-Olivia-Ammann: Already made a ticket

PR #672, opened 2023-12-06
feat(cli): change user output when handling errors (DEV-3067)

Nora-Olivia-Ammann:


src/dsp_tools/cli.py
```diff
@@ -606,10 +612,20 @@ def run(args: list[str]) -> None:
...
+        raise err from None
+    except InternalError as err:
+        logger.error(err)
+        raise err from None
+    except RetryError as err:
+        logger.error(err)
+        raise err from None
+    except Exception as err:
+        logger.error(err)
+        raise InternalError from None
```
jnussbaum, line 628: From a technical point of view, you could merge the identical ones into tuples, like `except (InputError, InternalError)`. 

From a user point of view, I wonder if this would not be uglier to look at than before?

Nora-Olivia-Ammann: From a technical, True I didn't think of that.

From User. I don't see how that is uglier. The big difference is that they see the error we raised without a stack trace. The error that are unexpected they also don't see a stack trace (which to my understanding was one of the goals) but we log trace and tell them to sent it to us...

jnussbaum: With uglier, I meant the difference in the UserError: Before it was printed, and then `sys.exit(1)`, now it is `raise err from None`. So the user will see a stack trace. I quickly tried out this:

```
% dsp-tools create testdata/invalid-testdata/json-project/invalid-super-property.json
Using DSP server 'http://0.0.0.0:3333' and SIPI server 'http://0.0.0.0:1024'
  File "/Users/nussbaum/Library/Caches/pypoetry/virtualenvs/dsp-tools-vOmZUZ4F-py3.11/bin/dsp-tools", line 6, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/nussbaum/Desktop/dsp-tools/src/dsp_tools/cli.py", line 574, in main
    run(sys.argv[1:])
  File "/Users/nussbaum/Desktop/dsp-tools/src/dsp_tools/cli.py", line 619, in run
    raise InternalError from None
dsp_tools.models.exceptions.InternalError: 

An internal error occurred.
Please contact the dsp-tools development team with the following information:
    - Which command was used.
    - If applicable, any files that were used in conjunction with the command.
    - A file with the terminal output copied into.
    - The log files called 'logging.log', if there are several, include all.
      They can be found at: /Users/nussbaum/.dsp-tools
```

Nora-Olivia-Ammann: That is weird, otherwise if we do from None, the trace is not visible... can we leave out sys.exit?

Nora-Olivia-Ammann: Ah ok. That is an easy fix

BalduinLandolt (approved):

PR #674, opened 2023-12-07
feat(xmlupload): improve error message if ontology is unknown (DEV-3024)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -41,14 +46,18 @@ def execute_problem_protocol(self) -> tuple[str, pd.DataFrame | None]:
...
+            f"{list_separator.join(self.ontos_on_server)}{extra_separator}"
+        )
         cls_msg = self._compose_problem_string_cls()
         if cls_msg:
             msg += cls_msg + extra_separator
         prop_msg = self._compose_problem_string_props()
         if prop_msg:
             msg += prop_msg
-        if len(self.classes) + len(self.properties) > 50:
+        if len(self.classes) + len(self.properties) > 20:
```
jnussbaum, line 60: I think that's a too low number. It's easily possible to scroll through 50 lines in a terminal. I would be annoyed to have an unknown Excel file in my folder for such an unimportant reason.

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -81,29 +90,42 @@ def _get_problems_as_df(self) -> pd.DataFrame:
         unpacked: list[dict[str, str]] = list(itertools.chain(*problems))
         return pd.DataFrame.from_records(unpacked)
 
+    def _calculate_num_resources(self, to_count: list[tuple[str, list[str], str]]) -> int:
+        return sum((len(x[1]) for x in to_count))
+
     def _compose_problem_string_cls(self) -> str | None:
```
jnussbaum, line 96: My first thought was that this name means "compose a problem string for this class". It took me really long to understand that there is a corresponding method `_compose_problem_string_props`, and that one method does it for classes, the other for props. 

Now I see that in my mind, I have to add a "for" or a "of" before "cls".

I would rename this function to `_compose_problem_string_from_classes` or `_compose_problem_string_of_cls`. 

And then the same for `_compose_problem_string_props`

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -81,29 +90,42 @@ def _get_problems_as_df(self) -> pd.DataFrame:
...
         if self.classes:
-            separator = "\n----------------------------\n"
+            if self._calculate_num_resources(self.classes) > 20:
```
jnussbaum, line 98: It's error prone to have a magic value here. (For magic values, see here: https://docs.astral.sh/ruff/rules/magic-value-comparison/)

And in this case it's especially dangerous, because if you change the 20 on line 60, you might easily forget to change it here. So I would store this in a class variable.

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -81,29 +90,42 @@ def _get_problems_as_df(self) -> pd.DataFrame:
...
 
-            return "The following resource(s) have an invalid resource type:\n\n" + separator.join(problems)
+            return "The following resource(s) have an invalid resource type:\n\n" + grand_separator.join(problems)
         else:
             return None
 
     def _compose_problem_string_props(self) -> str | None:
         if self.properties:
-            separator = "\n----------------------------\n"
+            if self._calculate_num_resources(self.properties) > 20:
```
jnussbaum, line 117: number 20: same as above

jnussbaum (approved):
That's a good improvement. But I would definitively fix the "magic value" trap that I commented.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -44,9 +44,11 @@ def _find_problems_in_classes_and_properties(
     property_problems = _diagnose_all_properties(properties, onto_check_info)
     if not class_problems and not property_problems:
         return None
-    problems = InvalidOntologyElements(classes=class_problems, properties=property_problems)
+    problems = InvalidOntologyElements(
+        classes=class_problems, properties=property_problems, ontos_on_server=list(onto_check_info.onto_lookup.keys())
+    )
     msg, df = problems.execute_problem_protocol()
-    if df:
+    if df is not None:
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Yes pandas raises an error saying that the truth value of a df is ambiguous

PR #677, opened 2023-12-08
refactor(xmlupload): clean up handle media information (DEV-3086)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -15,40 +15,53 @@
...
     resource: XMLResource,
+    media_previously_uploaded: bool,
     sipi_server: Sipi,
     imgdir: str,
     permissions_lookup: dict[str, Permissions],
-) -> BitstreamInfo | None:
+) -> tuple[bool, None | BitstreamInfo]:
     """
-    This function uploads a specified bitstream file to SIPI and then returns the file information from SIPI.
+    This function checks if a resource has a bitstream
```
jnussbaum, line 26: That's incomplete info. This function does much more

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -15,40 +15,53 @@
...
 
     Args:
-        resource: resource with that has a bitstream
+        resource: resource holding the bitstream
+        media_previously_uploaded: True if the image is already in SIPI
         sipi_server: server to upload
-        imgdir: directory of the file
         permissions_lookup: dictionary that contains the permission name as string and the corresponding Python object
+        imgdir: directory of the file
```
jnussbaum, line 33: Why do you switch positions of imgdir? The previous position was correct

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -15,40 +15,53 @@
...
 
     Returns:
-        The information from sipi which is needed to establish a link from the resource
+        If the handling of the bitstream object was successful,
+        in case of a bitstream the object or None in case of unsuccessful try.
```
jnussbaum, line 37: At times, the grammar of your sentences is a bit difficult to understand :-) Can you reformulate that? Or we can also do it together ;-)

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -68,22 +80,49 @@ def handle_bitstream(
...
         return resource_bitstream
     except BaseError as err:
         err_msg = err.orig_err_msg_from_api or err.message
         msg = f"Unable to upload file '{bitstream.value}' of resource '{resource.label}' ({resource.id})"
         print(f"{datetime.now()}: WARNING: {msg}: {err_msg}")
         logger.warning(msg, exc_info=True)
         return None
+
+
+def _upload_bitstream(
```
jnussbaum, line 101: This function contains 10 LOC, and its call needs 6 LOC. So I wonder if it is worth factoring this out into an own function?

jnussbaum (approved):
LGTM. See my comments for some suggestions how to improve.

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -15,40 +15,53 @@
...
-    img: Optional[dict[Any, Any]] = try_network_action(
-        sipi_server.upload_bitstream,
-        filepath=str(Path(imgdir) / Path(pth)),
-    )
-    internal_file_name_bitstream = img["uploadedFiles"][0]["internalFilename"]  # type: ignore[index]
-    return resource.get_bitstream_information(
-        internal_file_name_bitstream=internal_file_name_bitstream,
-        permissions_lookup=permissions_lookup,
-    )
+    bitstream = cast(XMLBitstream, resource.bitstream)
```
BalduinLandolt, line 39: I'm sceptical about this cast: with this, you erase the type information that this may be None. Afterwards you correctly check for this case, so currently it's not broken. But if in the future we delete the line with this check, then the type system does not warn us that we may have None at hand when we assume an `XMLBitstream`. Is it really needed? And if so, why?

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -15,40 +15,53 @@
...
+    success = True
+    bitstream_information: None | BitstreamInfo = None
 
+    match bitstream, media_previously_uploaded:
+        case None, _:
+            return success, bitstream_information
```
BalduinLandolt, line 45: In this case, this could simply be a short-circuiting if-statement

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -68,22 +80,49 @@ def handle_bitstream(
...
+    Args:
+        resource: resource with that has a bitstream
+        sipi_server: server to upload
+        imgdir: directory of the file
+        permissions_lookup: dictionary that contains the permission name as string and the corresponding Python object
+
+    Returns:
+        The information from sipi which is needed to establish a link from the resource
+    """
+    pth = resource.bitstream.value  # type: ignore[union-attr]
```
BalduinLandolt, line 119: `_upload_bitstream()` is called where the none case for resource.bitstream was already done. If you would just pass `path` from there, then you wouldn't need the type ignore

src/dsp_tools/commands/xmlupload/resource_multimedia.py
Nora-Olivia-Ammann: I had to cast it for mypy. I could not find another solution apart from ignoring it.

BalduinLandolt: that seems strange to me, because it the type of `bitstream` is `XMLBitstream | None` then mypy should understand the following match-case to infer the correct type in each branch. Where do you get errors if you remove the cast?

Nora-Olivia-Ammann: Screenshot 2023-12-11 at 15 55 04

BalduinLandolt: would it work if you changed the last case to wildcard (`case _: ...`)?

src/dsp_tools/commands/xmlupload/resource_multimedia.py
Nora-Olivia-Ammann: I did it to make the distinction between the action of the upload and the error handling with the messaging. If we did not communicate at all with the user I could be convinced to combine them. But with the prints and logs as it is now it is too messy.

PR #679, opened 2023-12-11
fix(excel2json): find hidden worksheets in excel

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/utils.py
```diff
@@ -73,7 +74,7 @@ def read_and_clean_all_sheets(excelfile: str) -> dict[str, pd.DataFrame]:
     try:
         return {name.strip(""): clean_data_frame(df) for name, df in df_dict.items()}
     except AttributeError:
-        msg = InvalidSheetNameProblem(excelfile, list(df_dict.keys())).execute_error_protocol()
+        msg = InvalidSheetNameProblem(str(excelfile), list(df_dict.keys())).execute_error_protocol()
```
jnussbaum, line 77: even better: InvalidSheetNameProblem should store the excelfile as Path, and not as str.

src/dsp_tools/commands/excel2json/utils.py
```diff
@@ -44,7 +45,7 @@ def read_and_clean_excel_file(excelfile: str, sheetname: str | int = 0) -> pd.Da
     return read_df
 
 
-def read_and_clean_all_sheets(excelfile: str) -> dict[str, pd.DataFrame]:
+def read_and_clean_all_sheets(excelfile: str | Path) -> dict[str, pd.DataFrame]:
```
jnussbaum, line 48: even better: accept only Path (if the effort isn't too big)

src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -490,3 +492,11 @@ def excel2properties(
             print(f"properties section was created successfully and written to file '{path_to_output_file}'")
 
     return props, True
+
+
+def _read_check_property_df(excelfile: str) -> pd.DataFrame | None:
```
jnussbaum, line 497: I would rename this to `read_and_check_property_df`

src/dsp_tools/commands/excel2json/utils.py
jnussbaum, line 1: The function `read_and_clean_excel_file` is now unused and should be removed

src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -490,3 +492,11 @@ def excel2properties(
...
+    sheets_df_dict = read_and_clean_all_sheets(excelfile=excelfile)
+    if len(sheets_df_dict) != 1:
+        msg = str(MoreThanOneSheetProblem("properties.xlsx", list(sheets_df_dict.keys())))
+        raise InputError(msg)
+    return list(sheets_df_dict.values())[0]
```
jnussbaum, line 502: If a container is big and you want to return its first value, it's more performant to put it like this: `next(iter(sheets_df_dict.values()))`

But here, it doesn't matter, because it is always of length 1

jnussbaum (approved):
Some remarks, but apart from that, LGTM

src/dsp_tools/commands/excel2json/properties.py
Nora-Olivia-Ammann: why? this is only used in this file and nowhere else...

src/dsp_tools/commands/excel2json/utils.py
Nora-Olivia-Ammann: If it is used as a path I understand the argument but here it is only used in the message, so I don't see the benefit.

src/dsp_tools/commands/excel2json/properties.py
jnussbaum: Sorry, misunderstanding: I didn't mean to omit the _, I was rather pointing at the wording. So my suggestion is to rename it to `_read_and_check_property_df`.

PR #683, opened 2023-12-13
fix(excel2json): support uppercase classes sheet in resources.xlsx (DEV-3109)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -217,6 +212,23 @@ def excel2resources(
     return resources, True
 
 
+def _prepare_classes_df(resource_dfs: dict[str, pd.DataFrame]) -> tuple[pd.DataFrame, dict[str, pd.DataFrame]]:
+    if "classes" not in resource_dfs.keys():
```
Nora-Olivia-Ammann, line 216: There is one question for me. Is it sufficient to allow uppercase and lower case or should we support other spellings (that will be more difficult).

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -217,26 +216,35 @@ def excel2resources(
...
-) -> ResourcesSheetsNotAsExpected | None:
-    for i, row in classes_df.iterrows():
-        index = int(str(i))  # index is a label/index/hashable, but we need an int
-        if not check_notna(row["super"]):
-            raise UserError(
-                f"Sheet 'classes' of 'resources.xlsx' has a missing value in row {index + 2}, column 'super'"
-            )
+def _prepare_classes_df(resource_dfs: dict[str, pd.DataFrame]) -> tuple[pd.DataFrame, dict[str, pd.DataFrame]]:
+    resource_dfs = {k.strip(): v for k, v in resource_dfs.items()}
+    if "classes" in resource_dfs:
```
jnussbaum, line 221: It is tiresome to first check for "classes", then for "Classes", ... and then, what's with "CLASSES"? I suggest doing that the other way round: On line 219, transform all keys to lowercase:
`resource_dfs = {k.strip().lower(): v for k, v in resource_dfs.items()}`

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -217,6 +212,23 @@ def excel2resources(
...
 
+def _prepare_classes_df(resource_dfs: dict[str, pd.DataFrame]) -> tuple[pd.DataFrame, dict[str, pd.DataFrame]]:
+    if "classes" not in resource_dfs.keys():
+        if "Classes" not in resource_dfs.keys():
+            msg = ResourcesSheetsNotAsExpected(set(), names_sheets={"classes"}).execute_error_protocol()
+            raise InputError(msg)
+        else:
+            classes_df = resource_dfs.pop("Classes")
+    else:
+        classes_df = resource_dfs.pop("classes")
```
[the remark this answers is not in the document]
jnussbaum: I don't consider this ugly :-)

jnussbaum (approved):
What's the behavior if one of the resource sheets has wrong capitalization? In case this isn't handled yet, you could consider to first lowercase all resource sheets, similar as I suggested in my comment below for the "classes" sheet. 

Apart from that, I added some small suggestions how you could improve it.

But generally LGTM

src/dsp_tools/commands/excel2json/resources.py
Nora-Olivia-Ammann: I have thought about that, the problem with this is that we also change the all the other keys. When we compare the resource classes listed in the "classes" sheet to the sheets in the excel then we also need to change capitalisation for those. if there is a problem we will print out a message the names may not be identical to the ones used in the excel.

PR #684, opened 2023-12-14
chore: resolve Ruff warnings A003, B023, D103 and PLR0911

BalduinLandolt:


Nora-Olivia-Ammann (approved):

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -136,38 +137,43 @@ def find_date_in_string(string: str) -> Optional[str]:
...
+    "November": 11,
+    "Nov": 11,
+    "December": 12,
+    "Dec": 12,
+}
+
+
+def _find_date_in_string_throwing(string: str) -> str | None:
+    """
+    This method is the same as find_date_in_string(), but may raise a ValueError instead of returning None.
```
jnussbaum, line 175: AFAIK, this is a (standalone) function, not a (class) method (?)

```suggestion
    This function is the same as find_date_in_string(), but may raise a ValueError instead of returning None.
```

test/unittests/commands/excel2xml/test_excel2xml_lib.py
```diff
@@ -630,51 +630,125 @@ def test_make_uri_prop(self) -> None:
...
-                xml_returned = etree.tostring(xml_returned_as_element, encoding="unicode")
-                xml_returned = regex.sub(r" xmlns(:.+?)?=\".+?\"", "", xml_returned)
-                self.assertEqual(result, xml_returned)
-
-            with self.assertWarns(UserWarning):
-                method("label", "id", ark="ark", iri="iri")
-            with self.assertRaisesRegex(BaseError, "invalid creation date"):
-                method("label", "restype", "id", creation_date="2019-10-23T13:45:12")
+    def test_make_annotation(self) -> None:
+        """Tests a regular annotation."""
```
jnussbaum, line 634: Now that we don't require docstrings in test files any more, you could also remove some of the superfluous docstrings.

jnussbaum (approved):
Very good, appreciate this one, thank you :-)

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
BalduinLandolt: correct... let's blame copilot :)

test/unittests/commands/excel2xml/test_excel2xml_lib.py
jnussbaum: I meant the superfluous docstrings that you introduced in this PR :-)

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -136,38 +137,43 @@ def find_date_in_string(string: str) -> Optional[str]:
     # sanitize input, just in case that the method was called on an empty or N/A cell
     if not check_notna(string):
         return None
+    try:
+        return _find_date_in_string_throwing(string)
```
jnussbaum, line 141: I don't particularly like this pattern. In fact, no one of the called functions should raise an error, they should just re
turn None if they don't succeed. Then, we don't need a function with this strange name.

BalduinLandolt: I am with you on that, but I didn't want to make bigger changes in a simple linting PR. Feel free to adjust that in a separate PR and I will happily approve it :)

PR #685, opened 2023-12-14
fix(xmlupload): saving the csv file if there are more than 50 problems with the ontology names (DEV-3112)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -13,6 +15,8 @@
 knoraUndeclared: Pattern[str] = regex.compile(r"^\w+$")
 genericPrefixedOntology: Pattern[str] = regex.compile(r"^[\w\-]+:\w+$")
 
+cwd = os.getcwd()
```
jnussbaum, line 18: If it's not absolutely necessary, this shouldn't be global. (And here, I don't see a necessity)

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -49,11 +51,11 @@ def _find_problems_in_classes_and_properties(
     )
     msg, df = problems.execute_problem_protocol()
     if df is not None:
-        ex_name = "XML_syntax_errors.xlsx"
-        df.to_excel(excel_writer=Path(onto_check_info.save_location, ex_name), sheet_name=" ", index=False)
+        csv_file = f"XML_syntax_errors_{datetime.now().strftime('%Y-%m-%d_%H%M%S')}.csv"
+        df.to_csv(path_or_buf=Path(cwd, csv_file), index=False)
```
jnussbaum, line 55: More idiomatic would be to get rid of the `cwd` variable altogether, and then:
```
csv_file = Path.cwd() / f"XML_syntax_errors_{datetime.now().strftime('%Y-%m-%d_%H%M%S')}.csv"
...
msg = f"\nThe file: '{csv_file}' lists the problems."
```

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -49,11 +51,11 @@ def _find_problems_in_classes_and_properties(
...
         msg += (
             "\n\n---------------------------------------\n\n"
-            f"\nAn excel: '{ex_name}' was saved at '{onto_check_info.save_location}' listing the problems."
+            f"\nThe file: '{csv_file}' was saved at '{cwd}' listing the problems."
```
jnussbaum, line 58: The formulation "X was saved at Y listing the problems" is not so user-friendly.
Better: "full/path.ext lists the problems"

src/dsp_tools/commands/xmlupload/models/ontology_diagnose_models.py
```diff
@@ -58,7 +56,10 @@ def execute_problem_protocol(self) -> tuple[str, pd.DataFrame | None]:
         prop_msg = self._compose_problem_string_for_props()
         if prop_msg:
             msg += prop_msg
-        if len(self.classes) + len(self.properties) > maximum_prints:
+        if (
+            self._calculate_num_resources(self.classes) + self._calculate_num_resources(self.properties)
+            > maximum_prints
+        ):
```
jnussbaum, line 62: This could fit in 2 lines instead of 4:
```
problem_count = self._calculate_num_resources(self.classes) + self._calculate_num_resources(self.properties)
        if problem_count > maximum_prints:
```

jnussbaum (approved):
LGTM

PR #686, opened 2023-12-14
refactor: move retry logic into the connection implementation (DEV-3110)

BalduinLandolt:


Nora-Olivia-Ammann (commented):
LGTM, a lot cleaner. But as I said I don't know the original code very well.

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -268,7 +270,7 @@ def _get_project_context_from_server(connection: Connection) -> ProjectContext:
         UserError: If the project was not previously uploaded on the server
     """
     try:
-        proj_context: ProjectContext = try_network_action(lambda: ProjectContext(con=connection))
+        proj_context: ProjectContext = ProjectContext(con=connection)
```
jnussbaum, line 273: unnecessary type hint

test/e2e/commands/project/test_ontology.py
jnussbaum, line 1: I don't particularly like the tests in this file - but why are you confident to delete them?

src/dsp_tools/utils/connection_live.py
```diff
@@ -110,6 +179,7 @@ def _write_request_to_file(
         jsondata: Optional[str],
         params: Optional[dict[str, Any]],
         response: requests.Response,
+        filepath: Optional[str] = None,
```
jnussbaum, line 182: In a method named "_write_request_to_file", a parameter named "filepath" suggests that you write to that file. Better name suggestion: "uploaded_file"

src/dsp_tools/utils/connection_live.py
```diff
@@ -133,6 +204,7 @@ def _write_request_to_file(
             "headers": headers,
             "params": params,
             "body": json.loads(jsondata) if jsondata else None,
+            "filepath": filepath,
```
jnussbaum, line 207: same here: "filepath" doesn't tell me enough. "uploaded file" or so would be more telling.

src/dsp_tools/utils/connection_live.py
```diff
@@ -162,30 +236,34 @@ def post(
...
             # if data is not encoded as bytes, issues can occur with non-ASCII characters,
             # where the content-length of the request will turn out to be different from the actual length
-            data=jsondata.encode("utf-8") if jsondata else None,
-            timeout=timeout,
-        )
+            data = jsondata.encode("utf-8") if jsondata else None
+            request = partial(request, data=data)
+        elif files:
+            request = partial(request, files=files)
+            logger.warning(f"POST with file: {request}")
```
jnussbaum, line 257: Is a file upload so dangerous that we have to warn? ;-)

```suggestion
            logger.info(f"POST with file: {request}")
```

src/dsp_tools/utils/connection_live.py
```diff
@@ -162,30 +236,34 @@ def post(
         # otherwise the client can get a timeout error while the API is still processing the request
         # in that case, the client's retry will have undesired side effects (e.g. duplicated resources),
         # and the response of the original API call will be lost
-        timeout = 60
+        timeout = timeout or 60
```
jnussbaum, line 239: wouldn't it make more sense to set a default value for the parameter in the method signature? And then, you could move the long inline comment into the docstring

src/dsp_tools/utils/connection_live.py
```diff
@@ -162,30 +236,34 @@ def post(
...
         if jsondata:
-            headers["Content-Type"] = f"{content_type}; charset=UTF-8"
+            headers["Content-Type"] = "application/json; charset=UTF-8"
         if self.token:
             headers["Authorization"] = f"Bearer {self.token}"
 
-        response = requests.post(
-            url=url,
-            headers=headers,
+        request = partial(requests.post, url=url, headers=headers, timeout=timeout)
```
jnussbaum, line 249: good idea, but I think I have an even better one: collect these keywords in a dict, and pass that dict as `try_network_action(requests.post, **kwargs)`.

And then, the signature of `try_network_action` could be simplified: 
 - accept only `Literal["GET", "POST", "PUT", "DELETE"]` for the parameter `action`. 
 - remove parameter `*args` 
 - turn parameter `**kwargs` into `kwargs` (mandatory)

Then, the logic of `try_network_action` can be much simplified.

src/dsp_tools/utils/connection_live.py
```diff
@@ -1,12 +1,81 @@
...
+        kwargs: keyword arguments for the action
+
+    Raises:
+        BaseError: if the action fails permanently
+        unexpected exceptions: if the action fails with an unexpected exception
+
+    Returns:
+        the return value of action
+    """
+    action_as_str = f"{action=}, {args=}, {kwargs=}"
```
jnussbaum, line 43: not sure how much sense this makes. This is actually an unstructured log that corresponds 1:1 to the structured variant that is implemented in `_write_request_to_file()`. But I'm unsure how far we want to go in this PR, or if we want to improve this later.

src/dsp_tools/utils/connection_live.py
BalduinLandolt: actually, this line should be removed entirely, just used it for diagnosing some failing tests

src/dsp_tools/utils/connection_live.py
BalduinLandolt: good idea. I was also having some thoughts on how to simplify `try_network_action`, which were going a slightly different direction. How about we leave this as is for now (I think it works like that) and discuss our ideas for improvements, which we can then implement as follow-up tasks?  
My aim (which I managed only partially) was to keep this PR as "mechanical" as possible, simply inverting the hierarchy of those two concepts, and not tackle any bigger improvements just yet. (Agile... baby steps... short-lived PRs... yada-yada 😄 )

src/dsp_tools/utils/connection_live.py
BalduinLandolt: I agree that this doesn't make much sense as it is now, but I also think we should keep this for a future improvement. It doesn't make sense to log this each time we re-try either, as the information should stay the same with each attempt.

src/dsp_tools/utils/connection_live.py
jnussbaum: Sure, that's fine :-)

src/dsp_tools/utils/connection_live.py
BalduinLandolt: I would prefer to keep the default value hard-coded in as few places as possible (and the way python implements protocols, it's actually a bit awkward to have defaults in parameters...), but this ties in into some things I was going to suggest as future improvements anyways. I'd leave it for now.

src/dsp_tools/commands/project/create/project_create.py
```diff
@@ -171,9 +171,7 @@ def _create_groups(
     overall_success = True
     current_project_groups: dict[str, Group] = {}
     try:
-        remote_groups: list[Group] = try_network_action(
-            lambda: Group.getAllGroupsForProject(con=con, proj_iri=str(project.iri))
-        )
+        remote_groups: list[Group] = Group.getAllGroupsForProject(con=con, proj_iri=str(project.iri)) or []
```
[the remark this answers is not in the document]
BalduinLandolt: It's an easy enough fix to be justified, thanks!

src/dsp_tools/commands/xmlupload/models/sipi.py
```diff
@@ -21,8 +17,7 @@ class Sipi:
         dump_directory: directory where the HTTP requests are written
     """
 
-    sipi_server: str
-    token: str
+    con: Connection
     dump: bool = False
     dump_directory = Path("HTTP requests")
```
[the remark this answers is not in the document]
BalduinLandolt: right... I forgot. Thanks for spotting!

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -119,8 +118,7 @@ def _upload_bitstream(
     Returns:
         The information from sipi which is needed to establish a link from the resource
     """
-    img: Optional[dict[Any, Any]] = try_network_action(
-        sipi_server.upload_bitstream,
+    img: Optional[dict[Any, Any]] = sipi_server.upload_bitstream(
```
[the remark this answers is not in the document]
BalduinLandolt: Tried to apply your feedback. Could you quickly check if it still looks correct to you? (especially the returning None part if bitstream is None)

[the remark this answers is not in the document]
BalduinLandolt: (as soon as I have pushed the changes...)

src/dsp_tools/utils/connection_live.py
BalduinLandolt: you're right, fixed

test/e2e/commands/project/test_ontology.py
BalduinLandolt: fairly confident - but maybe I shouldn't have opened this can of worms, and just add this to the possible future tasks. Should I revert the change?

src/dsp_tools/commands/xmlupload/resource_multimedia.py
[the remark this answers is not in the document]
jnussbaum: perfect :-)

test/e2e/commands/project/test_ontology.py
jnussbaum: I'm very much in favor of removing crappy unused stuff. But then I'd like to double-check it. So if you give me a short explanation why this is safe to delete, I'm fine.

BalduinLandolt: sure! `Ontology.read()`/`.update()`/`.delete()` are only used in test code, so they are not production relevant (surprizingly to me, to be honest), so my reasoning was that it doesn't make sense to test test-functionality, and that we would want to remove this soon anyways, if it's only used in tests. Does that make sense?

jnussbaum: That makes sense, and I double-checked it. It's safe to delete :-)

PR #687, opened 2023-12-14
fix(xmlupload): increase timeout to prevent doubled resources (DEV-3114)

jnussbaum:


src/dsp_tools/utils/connection.py
BalduinLandolt, line 1: could we make timeout an optional int that defaults to none? then in the live implementation we can have a field defining the timeout that is used wherever no timeout is provided. that would be cleaner than hard-coding these values all over the place

BalduinLandolt (commented):
see my architectural suggestion.  
also, I would insist that users should be informed if the timeout happens (see linear) because in these situations we have reason to assume that something went wrong and there may be duplicate resources.

src/dsp_tools/utils/connection.py
jnussbaum: I think I was too fast with this PR. I will wait until you have tidied up the Connection class, and then we see how we can solve the problems of this PR.

BalduinLandolt: ok. sorry about the work done in vain!

PR #690, opened 2023-12-16
refactor: remove boilerplate from call_action

BalduinLandolt:


BalduinLandolt:
If you don't like it, I don't mind closing this PR. I just saw this as a proposed solution for PLR0912 (which didn't even complain here, funnily enough) and would be curious to hear what you think about it.

jnussbaum:
My impression is that this is a very sophisticated way to play around with python, very cool for pros, but not easily understandable. So while I like your PR from an academic/intellectual point of view, I dislike it from an business/enterprise point of view. My personal suggestion is to close this PR rather than merging it.

jnussbaum:
@Nora-Olivia-Ammann Do you have a feedback for this PR?

Nora-Olivia-Ammann (approved):

Nora-Olivia-Ammann:
Looks good, I keep getting a warning about that ignore, so that makes me happy!

BalduinLandolt:
I feel torn about this PR, on one side it is definitely more concise and a clean, functional solution; on the other side the old version was indeed more explicit and easier to understand. So I'm not sure if we should merge it or not...

Nora-Olivia-Ammann:
> I feel torn about this PR, on one side it is definitely more concise and a clean, functional solution; on the other side the old version was indeed more explicit and easier to understand. So I'm not sure if we should merge it or not...

I am happy with both. One was as you said simpler, the other seems to make all the AI happy. I don't know which is better...

BalduinLandolt:
> > I feel torn about this PR, on one side it is definitely more concise and a clean, functional solution; on the other side the old version was indeed more explicit and easier to understand. So I'm not sure if we should merge it or not...
> 
> I am happy with both. One was as you said simpler, the other seems to make all the AI happy. I don't know which is better...

Once the AI has enslaved humankind, this will be so much easier to answer... 😉

jnussbaum:
I have a better idea. What do you think about this one? https://github.com/dasch-swiss/dsp-tools/pull/701

PR #691, opened 2023-12-18
fix(excel2xml): make_text_prop: allow <, >, & in rich texts (DEV-3131)

jnussbaum:


test/unittests/commands/excel2xml/test_excel2xml_lib.py
```diff
@@ -124,7 +124,7 @@ def run_test(
             # a <text> has the additional attribute encoding="utf8" (the other encoding, xml, is tested in the caller)
             xml_expected = regex.sub(
                 r"<text (permissions=\".+?\")( comment=\".+?\")?",
-                '<text \\1\\2 encoding="utf8"',
+                '<text \\1 encoding="utf8"\\2',
```
jnussbaum, line 127: I excuse myself for the code & the tests that my past self wrote 🙈  They are so ugly that I had to adapt the order of the attributes here. If I have too much time some day, I will refactor all this :-)

test/unittests/commands/excel2xml/test_excel2xml_lib.py
```diff
@@ -554,6 +556,22 @@ def test_make_text_prop(self) -> None:
...
+                "text &amp; text",
+            ],
+            [
+                "text <unclosed> tag",
+                "text &lt;unclosed&gt; tag",
+            ],
+            [
+                'text <unclosed tag="tag"> text',
+                'text &lt;unclosed tag="tag"&gt; text',
+            ],
```
jnussbaum, line 574: The price that we have to pay for the current approach is, that it is possible to have invalid XML that just gets escaped. But I think the gain in usability for RDU is big enough, and the frequency of invalid XML tags low enough, to justify the approach of the current PR.

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -1333,6 +1334,55 @@ def make_text_prop(
...
+    ]
+    allowed_tags_regex = "|".join(allowed_tags)
+    lookahead = rf"(?!/?({allowed_tags_regex})>)"
+    lookbehind = rf"(?<!</?({allowed_tags_regex}))"
+    illegal_lt = rf"<{lookahead}"
+    illegal_gt = rf"{lookbehind}>"
+    illegal_amp = r"&(?![#a-zA-Z0-9]+;)"
+    text = regex.sub(illegal_lt, "&lt;", text)
+    text = regex.sub(illegal_gt, "&gt;", text)
+    text = regex.sub(illegal_amp, "&amp;", text)
```
BalduinLandolt, line 1382: to me, this feels extremely hard to read and understand. Maybe it would be worth going for a solution that has a bit more repetition, but instead is more readable?

test/unittests/commands/excel2xml/test_excel2xml_lib.py
```diff
@@ -501,7 +500,8 @@ def test_make_text_prop(self) -> None:
             lambda: excel2xml.make_text_prop(":test", excel2xml.PropertyElement(value="a", encoding="unicode")),
         )
 
-        # encoding="utf8"
+    def test_make_text_prop_utf8(self) -> None:
+        """test encoding=utf8"""
```
BalduinLandolt, line 504: is this docstring needed?

test/unittests/commands/excel2xml/test_excel2xml_lib.py
BalduinLandolt: I too think this is an improvement nonetheless

test/unittests/commands/excel2xml/test_excel2xml_lib.py
```diff
@@ -523,6 +523,10 @@ def test_make_text_prop(self) -> None:
                 'text <text text="text" > text',
                 'text &lt;text text="text" &gt; text',
             ],
+            [
+                "text &amp; text",
+                "text &amp;amp; text",
+            ],
```
BalduinLandolt, line 529: these could/should be tuples, as it's always a pair and doesn't need the variable size of the list data type. Also, more stylistically: I'd prefer having them on one line, so each test case takes up only one line, not four.

BalduinLandolt (approved):
LGTM, if you can find a more readable solution for the regex, I'd be happy; if not, that's ok too

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
jnussbaum: I have no idea how I could improve it...

test/unittests/commands/excel2xml/test_excel2xml_lib.py
jnussbaum: No, but I find it helpful

test/unittests/commands/excel2xml/test_excel2xml_lib.py
jnussbaum: I turned them into tuples, but I prefer them on 4 lines, so that input and output are vertically aligned.

test/unittests/commands/excel2xml/test_excel2xml_lib.py
BalduinLandolt: to me it seems like a duplication of the method name. But if you prefer to have it there, sure!

test/unittests/commands/excel2xml/test_excel2xml_lib.py
BalduinLandolt: Another thing you could consider: I've been pushing for less of those parametrized tests for a long time, and I was under the impression that lately, those have become a bit less (which to me is far more clear, both when reading tests, and when reading test reports), so you could re-write this to a non-parametrized variant of the same tests. But of course you don't have to

jnussbaum: It was indeed you good influence on me that led me away from writing new parametrized tests. If there is opportunity, I de-parametrize old tests. But the entire excel2xml code has very much room for improvement, and I could easily get lost if I start refactoring too much. So I won't do it in this PR.

BalduinLandolt: sure!

Nora-Olivia-Ammann (approved):

test/unittests/commands/excel2xml/test_excel2xml_lib.py
```diff
@@ -501,28 +500,33 @@ def test_make_text_prop(self) -> None:
...
+            ),
+            (
                 'text <text text="text" > text',
                 'text &lt;text text="text" &gt; text',
-            ],
+            ),
+            (
+                "text &amp; text",
+                "text &amp;amp; text",
+            ),
```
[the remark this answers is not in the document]
jnussbaum: The tests functions are now refactored, and look much nicer now :-)

PR #694, opened 2023-12-19
chore: get rid of http_call_with_retry

jnussbaum:


src/dsp_tools/commands/start_stack.py
```diff
@@ -173,11 +169,7 @@ def _wait_for_fuseki(self) -> None:
         """
         for _ in range(6 * 60):
```
BalduinLandolt, line 170: now that I think about it: does this really need to wait so long? can fuseki ever take more than a couple of seconds?

BalduinLandolt (approved):

src/dsp_tools/commands/start_stack.py
jnussbaum: Well,... I just mimic https://github.com/dasch-swiss/dsp-api/blob/main/webapi/scripts/wait-for-db.sh#L48.
If you feel confident, you can adapt the value there, and then I will adapt it also here. The only thing about DSP-API that I know, is that I never know :lol

I vaguely remember that we once had problems that these 6 minutes really timed out. But I think that was a special problem, that could then be fixed somehow.

BalduinLandolt: let's not touch it, then :)

PR #696, opened 2023-12-19
chore: move to Python 3.12

jnussbaum:


BalduinLandolt (approved):
I suppose you'll have to inform RDU about this change, so they don't wonder why things are failing

Nora-Olivia-Ammann (approved):

PR #699, opened 2023-12-20
chore: make dev-release to pypi.org on every commit to main (DEV-3130)

jnussbaum:


BalduinLandolt (approved):
see my comment in chat, but from what I can tell, it looks good

Nora-Olivia-Ammann (approved):

PR #701, opened 2023-12-26
chore: resolve PLR0911 in call_requested_action()

jnussbaum:


Nora-Olivia-Ammann (approved):

BalduinLandolt (approved):
Good to be rid of the warning for sure... To be honest, to me this feels somewhat artificial - not really like a code improvement, more like a change to silence a warning, which makes me think we're working for the tool, not the tool for us. But it doesn't hurt either, so I'm happy to see this merged and the other PR closed

PR #702, opened 2023-12-27
feat: option for custom headers in HTTP requests (DEV-3145)

Nora-Olivia-Ammann:


src/dsp_tools/utils/connection.py
```diff
@@ -15,18 +15,21 @@ class Connection(Protocol):
...
         Args:
             route: route that will be called on the server
+            headers: headers for the HTTP request
         """
 
     def put(
         self,
         route: str,
         jsondata: str | None = None,
+        headers: Optional[dict[str, str]] | None = None,
```
BalduinLandolt, line 32: No need for optional when you define it as union with None (Optional is a relict from a time befor python supported nice union syntax)
```suggestion
        headers: dict[str, str] | None = None,
```

src/dsp_tools/utils/connection_live.py
```diff
@@ -179,6 +178,7 @@ def _write_request_to_file(
         jsondata: Optional[str],
         params: Optional[dict[str, Any]],
         response: requests.Response,
+        headers: Optional[dict[str, str]] | None = None,
```
BalduinLandolt, line 181: see comment above

src/dsp_tools/utils/connection_live.py
```diff
@@ -327,15 +340,17 @@ def put(
         if not route.startswith("/"):
             route = f"/{route}"
         url = self.server + route
+        if not headers:
+            headers = {}
```
BalduinLandolt, line 344: this actually makes me think: should headers be optional and the default be none? or should it simply default to `{}`?

test/unittests/commands/xmlupload/connection_mock.py
```diff
@@ -8,25 +8,34 @@
...
 
     def get(self, route: str, headers: dict[str, str] | None = None) -> dict[str, Any]:
         raise AssertionError("GET not implemented in mock")
 
-    def put(self, route: str, jsondata: str | None = None, content_type: str = "application/json") -> dict[str, Any]:
+    def put(
+        self,
+        route: str,
+        jsondata: str | None = None,
+        headers: Optional[dict[str, str]] | None = None,
```
BalduinLandolt, line 21: as above

test/unittests/commands/xmlupload/stash/test_upload_stash_with_mock.py
BalduinLandolt, line 1: do we not use `ConnectionMockBase` in more places?

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -41,7 +42,10 @@ def create_resource(
         )
         resource_dict = self._make_resource_with_values(resource, bitstream_information)
         resource_json_ld = json.dumps(resource_dict, ensure_ascii=False)
-        res = self.con.post(route="/v2/resources", jsondata=resource_json_ld)
+        if self.media_previously_ingested:
+            res = self.con.post(route="/v2/resources", jsondata=resource_json_ld, headers={"X-Asset-Ingested": "true"})
+        else:
+            res = self.con.post(route="/v2/resources", jsondata=resource_json_ld)
```
BalduinLandolt, line 48:
```suggestion
        headers = {"X-Asset-Ingested": "true"} if self.media_previously_ingested else None # or {} respectively, see my other comments
        res = self.con.post(route="/v2/resources", jsondata=resource_json_ld, headers=headers)
```
I think this would look a bit nicer

BalduinLandolt (approved):
some minor comments, otherwise looks good

src/dsp_tools/utils/connection_live.py
Nora-Olivia-Ammann: If you mean give an empty dictionary as the default parameter that is unfortunately not a good idea. 

https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

test/unittests/commands/xmlupload/stash/test_upload_stash_with_mock.py
Nora-Olivia-Ammann: I found four usages, but in some of them there was already the headers parameter was already in the function.

src/dsp_tools/utils/connection_live.py
BalduinLandolt: then `None` it is :)  
I keep forgetting what a weird language Python is...

Nora-Olivia-Ammann: It is a pity, I was annoyed every time I wrote those two lines...

jnussbaum (approved):
LGTM

PR #715, opened 2024-01-03
chore(ci): move setup into reusable workflow

jnussbaum:
This PR reduces redundant parts of the GitHub Actions YAML file. If you agree to merge this PR, I will adapt the mandatory PR checks (right now the old mandatory ones are yellow, hanging). 

Please also check if you are happy with the way how it looks in the output: https://github.com/dasch-swiss/dsp-tools/actions/runs/7401472400/job/20137278424


BalduinLandolt (commented):
I'm not radically opposed to merging this, but I don't particularly like it, for two reasons: 1) the test report looks less nice now that all jobs have the name "all-jobs", and 2) this seems like mis-using a tool that is made for a different job instead of using the right tool for the job (which would be a set-up action), and I personally find that worse than a bit of code duplication.

Nora-Olivia-Ammann (commented):
I agree with Balduin. In order for me to see which test exactly failed, I now have to click several times until I get the information.

BalduinLandolt (approved):

PR #716, opened 2024-01-04
test: update initial stash size (DEV-3153)

Nora-Olivia-Ammann:


jnussbaum (approved):
LGTM.
(Even better than having 2 hard-coded values, would be 1 variable. Then you can't forget to modify one of them by accident.)

Nora-Olivia-Ammann:
> LGTM. (Even better than having 2 hard-coded values, would be 1 variable. Then you can't forget to modify one of them by accident.)

Good idea.

PR #722, opened 2024-01-05
chore: replace bandit action by ruff's bandit ruleset (DEV-3026)

jnussbaum:


pyproject.toml
```diff
@@ -179,6 +180,10 @@ select = [
 ]
 ignore = [
     "ISC001",   # flake8-implicit-str-concat: single-line-implicit-string-concatenation  # incompatible with the formatter
+    "S105",     # flake8-bandit: hardcoded password
+    "S106",     # flake8-bandit: hardcoded password
+    "S603",     # flake8-bandit: subprocess-without-shell-equals-true
```
jnussbaum, line 185: This rule is strange. It conflicts with S602:

- [S602](https://docs.astral.sh/ruff/rules/S602): `subprocess.run("foo", shell=True)` is discouraged
- [S603](https://docs.astral.sh/ruff/rules/S603): `subprocess.run("foo", shell=False)` is discouraged

After some research, I found that S602 is to be preferred, and S603 should best be ignored

pyproject.toml
```diff
@@ -179,6 +180,10 @@ select = [
...
+    "S320",     # flake8-bandit: xml parsing vulnerable to XML attacks
```
jnussbaum, line 186: In our codebase, we often parse XML input with lxml, like e.g. `etree.parse(file)`. lxml is considered unsafe by ruff. But I think in our case it's okay.

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -876,8 +876,16 @@ def make_geometry_prop(
...
+            if value_as_dict["type"] not in ["rectangle", "circle", "polygon"]:
+                raise BaseError(
+                    f"Failed validation in resource '{calling_resource}', property '{name}': "
+                    f"The 'type' of the JSON geometry object must be 'rectangle', 'circle', or 'polygon'."
+                )
+            if not isinstance(value_as_dict["points"], list):
+                raise BaseError(
+                    f"Failed validation in resource '{calling_resource}', property '{name}': "
+                    f"The 'points'of the JSON geometry object must be a list of points."
+                )
```
jnussbaum, line 888: As per https://docs.astral.sh/ruff/rules/assert/, the use of `assert` in production code is discouraged.

src/dsp_tools/commands/rosetta.py
```diff
@@ -21,7 +21,7 @@ def _update_possibly_existing_repo(rosetta_folder: Path) -> bool:
     is_rosetta_up_to_date = True
     if rosetta_folder.is_dir():
         print(f"Execute 'git pull' in {rosetta_folder}...")
-        completed_process = subprocess.run("git pull", shell=True, cwd=rosetta_folder, check=False)
+        completed_process = subprocess.run("git pull".split(), cwd=rosetta_folder, check=False)
```
jnussbaum, line 24: With the `split()` trick, I avoid `shell=True`, but I also trick https://docs.astral.sh/ruff/rules/start-process-with-partial-path/ away.

src/dsp_tools/commands/xmlupload/upload_config.py
```diff
@@ -27,7 +27,7 @@ def _transform_server_url_to_foldername(server: str) -> str:
         r"^api\.": "",
         r":\d{2,5}/?$": "",
         r"/$": "",
-        r"0.0.0.0": "localhost",
+        r"0.0.0.0": "localhost",  # noqa: S104 (hardcoded-bind-all-interfaces)
```
jnussbaum, line 30: Unfortunately, DSP-API and DSP-APP enforce 0.0.0.0 and don't accept localhost. As explained in https://docs.astral.sh/ruff/rules/hardcoded-bind-all-interfaces/, this is subpar, but there's nothing that dsp-tools can do against, so I just ignore it.

src/dsp_tools/commands/excel2xml/excel2xml_cli.py
```diff
@@ -146,7 +146,7 @@ def _convert_rows_to_xml(
 
         # this is a property-row
         else:
-            assert check_notna(row["prop name"])
+            assert check_notna(row["prop name"])  # noqa: S101 (use of assert in production code)
```
BalduinLandolt, line 149: any particular reason this is this ignored here? should be easy enough to resolve with a `if not check_notna(): raise AssertionError("...)` or something along these lines, right?

src/dsp_tools/commands/rosetta.py
```diff
@@ -48,8 +48,7 @@ def _clone_repo(
     """
     print(f"Clone into {rosetta_folder}...")
     completed_process = subprocess.run(
-        "git clone https://github.com/dasch-swiss/082E-rosetta-scripts.git",
-        shell=True,
+        "git clone https://github.com/dasch-swiss/082E-rosetta-scripts.git".split(),
         cwd=enclosing_folder,
         check=False,
     )
```
BalduinLandolt, line 54: IMO this would look nicer as two one-liners:
```python
    cmd =  "git clone https://github.com/dasch-swiss/082E-rosetta-scripts.git".split()
    completed_process = subprocess.run(cmd, cwd=enclosing_folder, check=False,)
```

BalduinLandolt: by the way, this could be done in a less low-level (but maybe not less complicated) way by using a library like `GitPython`

src/dsp_tools/commands/start_stack.py
```diff
@@ -152,8 +152,7 @@ def _start_up_fuseki(self) -> None:
             UserError: if the database cannot be started
         """
         completed_process = subprocess.run(
-            "docker compose up db -d",
-            shell=True,
+            "docker compose up db -d".split(),
             cwd=self.__docker_path_of_user,
             check=False,
         )
```
BalduinLandolt, line 158: as above

src/dsp_tools/commands/xmlupload/upload_config.py
BalduinLandolt: yes, we've been wanting to fix this for over a year, but it's a SIPI issue 🙄

test/e2e/test_cli.py
```diff
@@ -39,8 +39,8 @@ def tearDownClass(cls) -> None:
 
     def test_validate_lists_section_with_schema(self) -> None:
         """Test if the resource file 'src/dsp_tools/resources/schema/lists-only.json' can be accessed."""
-        cmd = f"dsp-tools create --lists-only --validate-only {self.test_project_systematic_file.absolute()}"
-        self._make_cli_call(cmd)
+        args = ["create", "--lists-only", "--validate-only", str(self.test_project_systematic_file.absolute())]
+        self._make_cli_call(args)
```
BalduinLandolt, line 43: personally, I find the `"...".split()` more readable than the list

BalduinLandolt: Actually, having now seen how it's used below, I would leave the string in here entirely, and only split it in `_make_cli_call()`

pyproject.toml
```diff
@@ -179,6 +180,10 @@ select = [
 ]
 ignore = [
     "ISC001",   # flake8-implicit-str-concat: single-line-implicit-string-concatenation  # incompatible with the formatter
+    "S105",     # flake8-bandit: hardcoded password
+    "S106",     # flake8-bandit: hardcoded password
```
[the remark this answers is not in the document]
BalduinLandolt: ideally they would be read from a config, I think, but that's a task for a different PR

pyproject.toml
BalduinLandolt: yea, S603 doesn't seem too great: https://github.com/PyCQA/bandit/issues/333

BalduinLandolt (approved):
LGTM, I added some cosmetic remarks, but nothing important

pyproject.toml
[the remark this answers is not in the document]
jnussbaum: I added this point in the already existing Notion page in the Technical Improvements.

test/e2e/test_cli.py
jnussbaum: Unfortunately, both of your suggestions are not possible, because if the path contains a whitespace, this will be splitted, too.

PR #725, opened 2024-01-09
test: add multimedia files with commas and semicola in filename (DEV-3159)

jnussbaum:


BalduinLandolt (approved):

Nora-Olivia-Ammann (approved):

Nora-Olivia-Ammann:
Should we specify somewhere that this does not work with sipi only ingest?

jnussbaum:
> Should we specify somewhere that this does not work with sipi only ingest?

That's a misunderstanding. We don't use ingest for this. The new test files are uploaded with `dsp-tools xmlupload`, so there is no ingest involved. Apparently, SIPI is able to handle filenames with semicola!

PR #731, opened 2024-01-11
chore: log all HTTP requests, get rid of --dump flag (DEV-3171)

jnussbaum:


BalduinLandolt (approved):
LGTM, one thought though, regarding the security alert: Could you actively suppress logging the `connection.login()` call. It's the only one I can think of, that contains sensitive information (apart from the Token, which you're already stripping out)

PR #732, opened 2024-01-11
chore: pass payload as dict

jnussbaum:


src/dsp_tools/utils/connection.py
```diff
@@ -28,7 +28,7 @@ def get(
     def put(
         self,
         route: str,
-        jsondata: str | None = None,
+        jsondata: dict[str, Any] | None = None,
```
BalduinLandolt, line 31: could we rename `jsondata`/`jsonobj`, etc. to `data` everywhere? (also on the caller side)

src/dsp_tools/utils/connection_live.py
```diff
@@ -232,7 +235,7 @@ def post(
         if jsondata:
             # if data is not encoded as bytes, issues can occur with non-ASCII characters,
             # where the content-length of the request will turn out to be different from the actual length
-            data = jsondata.encode("utf-8") if jsondata else None
+            data = json.dumps(jsondata, cls=SetEncoder).encode("utf-8")
             request = partial(request, data=data)
```
BalduinLandolt, line 239: with this PR you practically already did the work for https://www.notion.so/dasch-swiss/New-Improve-JSON-serialization-for-HTTP-requests-c344eea38112401d8ecbcc2a78998983?pvs=4  
if you want to keep that for a separate PR, sure. But I wouldn't let it wait, since the heavy lifting is already done, which makes this really low hanging fruit.

src/dsp_tools/utils/connection_live.py
```diff
@@ -330,7 +333,8 @@ def put(
                 headers=headers,
                 # if data is not encoded as bytes, issues can occur with non-ASCII characters,
                 # where the content-length of the request will turn out to be different from the actual length
-                data=jsondata.encode("utf-8") if jsondata else None,
+                data=json.dumps(jsondata, cls=SetEncoder).encode("utf-8") if jsondata else None,
+                # TODO: ensure_ascii=False?
```
BalduinLandolt, line 337: please do not check in TODO comments. If it's actually something to do, it should go on Notion or Linear

src/dsp_tools/utils/connection_live.py
```diff
@@ -330,7 +333,8 @@ def put(
                 headers=headers,
                 # if data is not encoded as bytes, issues can occur with non-ASCII characters,
                 # where the content-length of the request will turn out to be different from the actual length
-                data=jsondata.encode("utf-8") if jsondata else None,
+                data=json.dumps(jsondata, cls=SetEncoder).encode("utf-8") if jsondata else None,
```
BalduinLandolt, line 336: why do we need this `SetEncoder`?

BalduinLandolt (approved):
Nice!

src/dsp_tools/utils/connection_live.py
jnussbaum: Here I wanted to ask you if `ensure_ascii=False` would resolve the problem described on lines 334-35. You had used it in one single place (https://github.com/dasch-swiss/dsp-tools/blob/0d5627fb5de9f88ef5bab6ee05155a76fc76c791/src/dsp_tools/commands/xmlupload/resource_create_client.py#L44). Or was the problem even caused by `ensure_ascii=False`?

src/dsp_tools/utils/connection_live.py
jnussbaum: We need it to tell the json package how to serialize objects of type `set`, `Context`, and `OntoIri`. By the way, that's also the reason why we have to do the serialization ourselves, by hand. I don't see a possibility to pass a custom Encoder to the requests package.

src/dsp_tools/utils/connection.py
jnussbaum: `data` would be fine for me, but I even prefer `payload`. What do you think?

src/dsp_tools/utils/connection_live.py
Nora-Olivia-Ammann: Don't the linter prevent todo anymore? Previously I am not sure if it was darglint or pylint that made the checks fail. This could be remedied with this right? https://docs.astral.sh/ruff/rules/#flake8-fixme-fix

Because leaving todo is something that might happen for me accidentally.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -41,9 +41,8 @@ def create_resource(
             f"Attempting to create resource {resource.res_id} (label: {resource.label}, iri: {resource.iri})..."
         )
         resource_dict = self._make_resource_with_values(resource, bitstream_information)
-        resource_json_ld = json.dumps(resource_dict, ensure_ascii=False)
         headers = {"X-Asset-Ingested": "true"} if self.media_previously_ingested else None
-        res = self.con.post(route="/v2/resources", jsondata=resource_json_ld, headers=headers)
+        res = self.con.post(route="/v2/resources", jsondata=resource_dict, headers=headers)
```
Nora-Olivia-Ammann, line 45: Do we not dump the json here for consistency? There are no PWs...

Nora-Olivia-Ammann (approved):

src/dsp_tools/commands/xmlupload/resource_create_client.py
jnussbaum: I'm not sure what you mean. Before, every caller of the Connection class first dumped=serialized the dictionary to a JSON-formatted string. But this PR changes that: Now, we pass dictionaries to the Connection class.

So it was my intention to remove line 44 here. This is consistent, because everywhere else, we now do it the same way: passing dicts instead of strings. 

But perhaps I didn't get your point?

Nora-Olivia-Ammann: No I was just curious if consistency was the reason that is all

src/dsp_tools/utils/connection_live.py
jnussbaum: We could add "FIX" here: https://github.com/dasch-swiss/dsp-tools/blob/d3e238beb2a115f025ce99b7a0cc4861b26feaaf/pyproject.toml#L179-L180

Nora-Olivia-Ammann: works for me

BalduinLandolt: > Here I wanted to ask you if `ensure_ascii=False` would resolve the problem described on lines 334-35. You had used it in one single place (
> 
> https://github.com/dasch-swiss/dsp-tools/blob/0d5627fb5de9f88ef5bab6ee05155a76fc76c791/src/dsp_tools/commands/xmlupload/resource_create_client.py#L44
> ). Or was the problem even caused by `ensure_ascii=False`?

`ensure_ascii=True` escapes non-ascii characters with `\xy`, which then might end up in the data, which we don't want. So thinking about it now, we should definitely have it in there. (Or at least experiment with what the exact behaviour is, so we know for sure it does what we want. But I'm pretty sure we want our data to have all UTF-8 characters without escaping.)

the Issue I faced with the encoding seemed more like a byte-array vs. unicode (`b'...'` vs. `'...'`) problem, which messed up the content-length header that's auto-added by requests. I never cared to fully investigate it because that fixed it... but I'm pretty positive that 1) this is something else than the `ensure_ascii` and 2) that it could be solved with using the `json=dict` rather than `data=str` param on the requests (cf. my comment mentioning the notion page).

src/dsp_tools/utils/connection.py
BalduinLandolt: `data` would follow the naming used by requests. But I'm happy with either

src/dsp_tools/utils/connection_live.py
BalduinLandolt: we shouldn't pass sets to any json encoder, we should convert those to lists; and we absolutely should not pass custom types to a json encoder - but ok... then that's a slightly bigger refactoring, it seems.

src/dsp_tools/utils/connection_live.py
jnussbaum: Okay, so if I get you right: I will add `ensure_ascii=False`, and leave the byte encoding in place. Thanks for the clarification :-)

BalduinLandolt: > Okay, so if I get you right: I will add `ensure_ascii=False`, and leave the byte encoding in place. Thanks for the clarification :-)

Yes :)

PR #739, opened 2024-01-12
chore: improve performance with requests.Session object (DEV-3174)

jnussbaum:


src/dsp_tools/utils/connection_live.py
```diff
@@ -34,11 +34,15 @@ class ConnectionLive:
 
     server: str
     token: Optional[str] = None
+    session: requests.Session = field(init=False, default=requests.Session())
```
Nora-Olivia-Ammann, line 37: could we perhaps move this to an import? so that we write Session = .... 
it is less verbose and we can always check the imports if we want to know where it is from (we don't need that info on every line that object is called)

src/dsp_tools/utils/connection_live.py
```diff
@@ -195,15 +198,10 @@ def get(
         if not route.startswith("/"):
             route = f"/{route}"
         url = self.server + route
-        if not headers:
-            headers = {}
```
Nora-Olivia-Ammann, line 199: These are still in post and put, is there a reason why we don't have headers here?

src/dsp_tools/utils/connection_live.py
```diff
@@ -292,14 +286,9 @@ def delete(
...
```
Nora-Olivia-Ammann, line 296: same as above

src/dsp_tools/utils/connection_live.py
```diff
@@ -352,6 +341,8 @@ def _try_network_action(self, action: Callable[[], Response]) -> Response:
                 self._log_and_sleep(reason="Timeout Error", retry_counter=i)
                 continue
             except (ConnectionError, RequestException):
+                self.session.close()
+                self.session = requests.Session()
```
Nora-Olivia-Ammann, line 345: Superficial comment here. But if we want to keep with the new style of having the functions after they are used. _log_and_sleep should be moved down.

Nora-Olivia-Ammann (commented):
I don't know session enough at this point to feel confident in my approval. But apart from the comments it looks good to me.

BalduinLandolt (approved):

src/dsp_tools/utils/connection_live.py
jnussbaum: The only header we had here was the "User-Agent", which is now set in the `__post__init__()`. So there is no semantic change here, the User-Agent is still sent.

src/dsp_tools/utils/connection_live.py
jnussbaum: I will do that in the next PR which is already in the pipeline. (I don't want to create too many merge conflicts)

PR #740, opened 2024-01-15
fix: retry mechanism again catches return codes in the 500 range (DEV-3177)

jnussbaum:


src/dsp_tools/utils/connection_live.py
```diff
@@ -395,5 +310,57 @@ def delete(
...
+            except (ConnectionError, RequestException):
+                msg = f"Network Error: Try reconnecting to DSP server, next attempt in {2 ** i} seconds..."
+                print(f"{datetime.now()}: {msg}")
+                logger.error(f"{msg} (retry-counter {i=:})", exc_info=True)
+                time.sleep(2**i)
+                continue
+
+            in_500_range = 500 <= response.status_code < 600
+            try_again_later = "try again later" in response.text
+            if try_again_later or in_500_range:
```
BalduinLandolt, line 351: I would suggest factoring out this boolean logic into a `_should_retry(response: Response)` function

src/dsp_tools/utils/connection_live.py
```diff
@@ -395,5 +310,57 @@ def delete(
...
+            elif response.status_code != 200:
+                raise BaseError(
+                    message="Permanently unable to execute the network action. See logs for more details.",
+                    status_code=response.status_code,
+                    json_content_of_api_response=response.text,
+                    reason_from_api=response.reason,
+                    api_route=response.url,
+                )
+            else:
+                return response
```
BalduinLandolt, line 366:
```suggestion
            return response
```

no need to call else here, in any other code path, it either continues or raises.

src/dsp_tools/utils/connection_live.py
```diff
@@ -395,5 +310,57 @@ def delete(
             headers=headers,
             timeout=timeout,
         )
-        check_for_api_error(response)
         return cast(dict[str, Any], response.json())
+
+    def _try_network_action(self, action: Callable[..., Any]) -> Any:
```
BalduinLandolt, line 315: `self` is not called anywhere in this method, so this could just as well stay a top level function, I think

BalduinLandolt, line 315: Also, the return type could be `Response`, so the type annotations on the caller side would not be necessary

src/dsp_tools/utils/connection_live.py
```diff
@@ -395,5 +310,57 @@ def delete(
...
+            the return value of action
+        """
+        for i in range(7):
+            try:
+                response: requests.Response = action()
+            except (TimeoutError, ReadTimeout, ReadTimeoutError):
+                msg = f"Timeout Error: Try reconnecting to DSP server, next attempt in {2 ** i} seconds..."
+                print(f"{datetime.now()}: {msg}")
+                logger.error(f"{msg} (retry-counter {i=:})", exc_info=True)
+                time.sleep(2**i)
```
BalduinLandolt, line 340: these 4 lines look to be repeated 3 times, unless I'm missing something. I would factor that out into a little helper function.

BalduinLandolt (commented):
I added some suggestions. Also, I don't see any reason why `_try_network_action()` had to be turned into a method. Am I missing something? If not, I would actually prefer turning it back into a function.  
That would probably also make testing easier in the future.

src/dsp_tools/utils/connection_live.py
jnussbaum: That's true, but I find it better readable with the additional `else`

src/dsp_tools/utils/connection_live.py
jnussbaum: But then we have to restrict the `action`, and make sure that it really always is a `requests.xyz()` callable. Otherwise we cannot safely claim that the return value is a `requests.Response`. Or do you think we should still do it?

BalduinLandolt: ok, that makes sense

BalduinLandolt: in theory you could pass the session object to the function as an argument. But I'm absolutely ok with making it a method, I just didn't see the reason

BalduinLandolt: so you mean

```python
def _try_network_action(self, action: Callable[..., Response]) -> Response:
    ...
```

I don't see why not; and as it is a private method, we don't have to expect any other callers than the ones we know about in this file.

BalduinLandolt: actually, we don't need the `...` either, because the params have already been applied, so we know it's a callable that doesn't take any parameters, right?

BalduinLandolt (approved):

src/dsp_tools/utils/connection_live.py
jnussbaum: https://docs.python.org/3/library/typing.html#annotating-callable-objects

jnussbaum: Indeed, I can use `Callable[[], requests.Response]`, thanks for pointing that out :-)

PR #741, opened 2024-01-16
fix: properly log requests and their responses (DEV-3186)

jnussbaum:


src/dsp_tools/utils/connection_live.py
```diff
@@ -51,18 +51,15 @@ def login(self, email: str, password: str) -> None:
...
         response = self.post(
             route="/v2/authentication",
             data={"email": email, "password": password},
         )
         if not response.get("token"):
-            raise BaseError(
-                f"Error when trying to login with user '{email}' and password '{password} "
-                f"on server '{self.server}'",
-                json_content_of_api_response=json.dumps(response),
-                api_route="/v2/authentication",
```
jnussbaum, line 65: The info of these 2 lines has already been logged on line 328 (`self._log_response(response)`)

src/dsp_tools/utils/connection_live.py
```diff
@@ -336,13 +333,8 @@ def _try_network_action(self, action: Callable[[], Response]) -> Response:
                 time.sleep(2**i)
                 continue
             elif response.status_code != HTTP_OK:
-                raise BaseError(
-                    message="Permanently unable to execute the network action. See logs for more details.",
-                    status_code=response.status_code,
-                    json_content_of_api_response=response.text,
-                    reason_from_api=response.reason,
-                    api_route=response.url,
```
jnussbaum, line 344: The info of these 2 lines has already been logged on line 328 (`self._log_response(response)`)

src/dsp_tools/models/exceptions.py
```diff
@@ -1,63 +1,17 @@
...
-        if json_content_of_api_response:
-            self.json_content_of_api_response = json_content_of_api_response
-            with contextlib.suppress(json.JSONDecodeError):
-                parsed_json = json.loads(json_content_of_api_response)
-                if "knora-api:error" in parsed_json:
-                    knora_api_error = parsed_json["knora-api:error"]
-                    knora_api_error = regex.sub(r"^dsp\.errors\.[A-Za-z]+?: ?", "", knora_api_error)
-                    self.orig_err_msg_from_api = knora_api_error
-        self.reason_from_api = reason_from_api
-        self.api_route = api_route
```
jnussbaum, line 60: All this elaborate logic was actually useless. It was used nowhere in the entire codebase 🙈

src/dsp_tools/models/exceptions.py
```diff
@@ -128,6 +82,17 @@ class UserError(BaseError):
...
+    This error is raised when all attempts to reconnect to DSP have failed.
+
+    Attributes:
+        message: A message that describes the error
+    """
+
+    message: str
+
+
 class XmlUploadError(Exception):
```
Nora-Olivia-Ammann, line 96: I know this is not the scope, but while you're at it could you subclass this of a BaseError or is there are reason this is an Exception?

src/dsp_tools/utils/connection_live.py
```diff
@@ -51,18 +51,15 @@ def login(self, email: str, password: str) -> None:
...
             data={"email": email, "password": password},
         )
         if not response.get("token"):
-            raise BaseError(
-                f"Error when trying to login with user '{email}' and password '{password} "
-                f"on server '{self.server}'",
-                json_content_of_api_response=json.dumps(response),
-                api_route="/v2/authentication",
+            raise PermanentConnectionError(
+                f"Error when trying to login with user '{email}' and password '{password} on server '{self.server}'"
```
Nora-Olivia-Ammann, line 62: don't we log the errors in https://github.com/dasch-swiss/dsp-tools/blob/8296254cfeb54572da2e47788f389246176ecdba/src/dsp_tools/cli/entry_point.py#L68 so would it not log the password here even though we carefully extracted it previously?

jnussbaum: Good point. I will filter it out here, too.

src/dsp_tools/utils/connection_live.py
```diff
@@ -51,19 +51,18 @@ def login(self, email: str, password: str) -> None:
...
+        err_msg = f"Username and/or password are not valid on server '{self.server}'"
+        try:
+            response = self.post(
+                route="/v2/authentication",
+                data={"email": email, "password": password},
             )
+        except PermanentConnectionError:
+            raise UserError(err_msg)
+        if not response.get("token"):
+            raise UserError(err_msg)
```
jnussbaum, line 65: @Nora-Olivia-Ammann With this, the user gets a nice feedback, but no sensitive data is logged.

Nora-Olivia-Ammann (approved):

PR #744, opened 2024-01-17
fix: restore authorization header after ConnectionError (DEV-3190)

jnussbaum:


src/dsp_tools/utils/connection_live.py
```diff
@@ -342,11 +342,15 @@ def _try_network_action(self, action: Callable[[], Response]) -> Response:
...
                 self.session = Session()
+                self.session.headers["Authorization"] = f"Bearer {self.token}"
                 self._log_and_sleep(reason="Network Error", retry_counter=i)
                 continue
 
             if self._should_retry(response):
-                self._log_and_sleep(reason="Transient Error", retry_counter=i)
+                msg = f"Server unresponsive: Try reconnecting to DSP server, next attempt in {2 ** i} seconds..."
+                print(f"{datetime.now()}: {msg}")
+                logger.error(msg)
```
jnussbaum, line 352: these lines are actually not related to the bug, but they are problematic, too: the method `_log_and_sleep()` calls `logger.exception()`, which should only be done during error handling. Here, we should rather call `logger.error()`.

Nora-Olivia-Ammann: why do you not change the function _log_and_sleep then?

jnussbaum: Because in the other 2 cases where this method is used, it is from within an error handling. For that purpose, we need `logger.exception()`

PR #749, opened 2024-01-19
fix(ingest-xmlupload): apply mapping.csv even if extension has wrong casing (DEV-3197)

jnussbaum:


src/dsp_tools/commands/ingest_xmlupload/user_information.py
```diff
@@ -33,7 +33,7 @@ def ok_msg(self) -> str | None:
         if not self.unused_mediafiles and not self.mediafiles_no_id:
             return (
                 "All multimedia files referenced in the XML file were uploaded through dsp-ingest.\n"
-                "No multimedia files were uploaded through dsp-ingest that were not referenced in the XML file."
+                "All multimedia files uploaded through dsp-ingest were referenced in the XML file."
```
jnussbaum, line 36: The only reason why I changed this: When trying out SGV, it first scared me to see a negative formulation. The negation made me think that I made a mistake.

Nora-Olivia-Ammann (approved):

PR #750, opened 2024-01-19
test(connection): replace e2e tests by unit tests (DEV-3196)

jnussbaum:


test/unittests/utils/test_connection_live.py
```diff
@@ -22,3 +227,118 @@ def test_anonymize_different_lengths() -> None:
...
+    with patch("dsp_tools.utils.connection_live.time.sleep") as sleep_mock:
+        response = con._try_network_action(params)
+        assert [x.args[0] for x in sleep_mock.call_args_list] == [1, 2, 4]
+    assert con._renew_session.call_count == len(session_mock.responses) - 1
+    assert [x.args[0] for x in con._log_request.call_args_list] == [params] * len(session_mock.responses)
+    con._log_response.assert_called_once_with(session_mock.responses[-1])
+    assert response == session_mock.responses[-1]
+
+
+@patch.dict("os.environ", {"DSP_TOOLS_TESTING": "false"})
```
jnussbaum, line 282: Strange. In the 2 tests above, `@patch.dict("os.environ", {})` works fine, but in this one, I need this (rather unrealistic) line for it to work.

test/unittests/utils/test_connection_live.py
```diff
@@ -1,4 +1,217 @@
...
+
+
+def test_log_in_bad_credentials() -> None:
+    con = ConnectionLive("http://example.com/")
+    con._log_response = Mock()
+    con.session.request = Mock(return_value=Mock(status_code=401))
+    with patch("dsp_tools.utils.connection_live.time.sleep") as sleep_mock:
+        with pytest.raises(UserError, match="Username and/or password are not valid"):
+            con.login("invalid@example.com", "wrong")
+        sleep_mock.assert_not_called()
```
Nora-Olivia-Ammann, line 56: What is the function of this?

test/unittests/utils/test_connection_live.py
```diff
@@ -1,4 +1,217 @@
...
+    con = ConnectionLive("http://example.com/")
+    con._try_network_action = Mock()
+    for method in (con.post, con.put, con.get, con.delete):
+        method = cast(Callable[..., Any], method)
+        method(route="/v2/resources", headers={"foo": "bar"})
+        expected_params: RequestParameters = con._try_network_action.call_args.args[0]
+        assert expected_params.headers == {"foo": "bar"}
+
+
+def test_custom_content_type() -> None:
```
Nora-Olivia-Ammann, line 177: It would mean a lot more writing, but in my opinion, it would give more readable results if you created a class for the functions on 177, 187, 197, 207 and then create a function  for the lines that you call within the loop (eg. 181-184) and for each method you call this mini function wrapped in a separate function:

```
def helper_with_assert(con) -> None:
   con.post(route="/v2/resources", headers={"Content-Type": "bar"})
        expected_params: RequestParameters = con._try_network_action.call_args.args[0]
        assert expected_params.headers == {"Content-Type": "bar"}

def con_post() -> None:
       con = ConnectionLive("http://example.com")
       con._try_network_action = Mock()
      helper_with_assert(con)
```

In case of an error, the result would be a lot cleaner and the function that has the error would be easier to identify.

test/unittests/utils/test_connection_live.py
jnussbaum: I want to make sure that when you log in with bad credentials, there is no retry, and no sleeping. Instead, a UserError should be raised immediately.

test/unittests/utils/test_connection_live.py
jnussbaum: Luckily, I could achieve a better feedback message with a simple trick: 

```
for method in (con.post, con.put, con.get, con.delete):
        assert x = y, f"Method '{method.__name__}' failed"
```

Nora-Olivia-Ammann (approved):

PR #752, opened 2024-01-19
chore(connection): content type doesn't have to be parametrized

jnussbaum:


Nora-Olivia-Ammann (commented):
I disagree that this is unnecessary (rdflib is sometimes iffy with jsonld). It is possible that in future we may wish to hand over the content type differently. As it already exists, I would leave it.

But we are in a democracy, if Balduin approves then you can merge if you like.

BalduinLandolt (changes requested):
to me it makes sense not to pass content-type as a separate parameter. However, I would say we should only set it to the default, if it's not already set in the provided headers. Like that we'd still have the option to pass a different content type, simply by passing it in the headers dict.

Would that be "konsensfähig"? @Nora-Olivia-Ammann what do you think?

Nora-Olivia-Ammann:
> to me it makes sense not to pass content-type as a separate parameter. However, I would say we should only set it to the default, if it's not already set in the provided headers. Like that we'd still have the option to pass a different content type, simply by passing it in the headers dict.
> 
> Would that be "konsensfähig"? @Nora-Olivia-Ammann what do you think?

That makes sense to me.

jnussbaum:
> > to me it makes sense not to pass content-type as a separate parameter. However, I would say we should only set it to the default, if it's not already set in the provided headers. Like that we'd still have the option to pass a different content type, simply by passing it in the headers dict.
> > Would that be "konsensfähig"? @Nora-Olivia-Ammann what do you think?
> 
> That makes sense to me. Would it be best you think to give a dict as default value, with that info, then deepcopy it in the function and change the rest there'

No, he meant that if someone wants to give a custom content type, it can be passed inside the "headers" parameter. the content type belongs into the header anyways, so it doesn't make sense to have a separate parameter for it. 

But for this to work, I must not overwrite an existing content type that is in the passed headers. I changed that, so the PR is now again ready for review.

BalduinLandolt (approved):
thanks!

PR #753, opened 2024-01-22
docs: DSP-API uses "just" instead of "make" (DEV-3212)

jnussbaum:


Nora-Olivia-Ammann:
Is this dead or are we waiting for something else?

jnussbaum:
> Is this dead or are we waiting for something else?

Sorry, I left this open for too long. I wanted to discuss with Balduin what `just` is doing behind the scenes, and if the approach of `dsp-tools start-stack` is still up-to-date. I feel that we could streamline `dsp-tools start-stack`, but I have to talk to someone from the backend about it. I have now created a follow-up ticket, so that we can merge this one.

Nora-Olivia-Ammann (approved):

PR #754, opened 2024-01-22
fix(xmlupload): retry on all non-OK response statuses (DEV-3214)

jnussbaum:


src/dsp_tools/utils/connection_live.py
```diff
@@ -352,10 +351,10 @@ def _mask(self, sensitive_info: str) -> str:
         else:
             return f"{sensitive_info[:unmasked_until]}[+{len(sensitive_info) - unmasked_until}]"
 
-    def _should_retry(self, response: Response) -> bool:
-        in_500_range = HTTP_SERVER_ERROR_LOWER <= response.status_code <= HTTP_SERVER_ERROR_UPPER
-        try_again_later = "try again later" in response.text
-        return try_again_later or in_500_range
+    def _in_ci(self) -> bool:
+        # https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
+        ci = os.getenv("GITHUB_ACTIONS")
```
jnussbaum, line 356: if we don't do that, the CI tests run very long, and it takes too much time until we know that a test failed

src/dsp_tools/utils/connection_live.py
```diff
@@ -352,10 +351,9 @@ def _mask(self, sensitive_info: str) -> str:
         else:
             return f"{sensitive_info[:unmasked_until]}[+{len(sensitive_info) - unmasked_until}]"
 
-    def _should_retry(self, response: Response) -> bool:
-        in_500_range = HTTP_SERVER_ERROR_LOWER <= response.status_code <= HTTP_SERVER_ERROR_UPPER
-        try_again_later = "try again later" in response.text
-        return try_again_later or in_500_range
+    def _in_testing_environment(self) -> bool:
+        in_testing_env = os.getenv("DSP_TOOLS_TESTING")  # set in .github/workflows/tests-on-push.yml
+        return in_testing_env == "true"
```
jnussbaum, line 356: if we don't do that, the CI tests run very long, and it takes too much time until we know that a test failed.

In `test_connection_live.py`, we test that calling a non-existing route leads to an error (`self.assertRaises(BaseError, self.con.get, "/doesNotExist")`). DSP-API responds with 404 on the route `/doesNotExist`, and in our tests, we don't want to retry then.

src/dsp_tools/utils/connection_live.py
```diff
@@ -291,17 +290,17 @@ def _try_network_action(self, action: Callable[[], Response]) -> Response:
                 continue
 
             self._log_response(response)
-            if self._should_retry(response):
-                msg = f"Server unresponsive: Try reconnecting to DSP server, next attempt in {2 ** i} seconds..."
+            if response.status_code == HTTP_OK:
+                return response
+            elif not self._in_testing_environment():
+                msg = f"Non-200 response code: Try reconnecting to DSP server, next attempt in {2 ** i} seconds..."
                 print(f"{datetime.now()}: {msg}")
```
Nora-Olivia-Ammann, line 297: why don't you use log an sleep here?

jnussbaum: Because we are not inside an exception handler here, then we shouldn't call https://docs.python.org/3/library/logging.html#logging.Logger.exception.

But in fact, it's probably better to refactor `_log_and_sleep()` so that it can be called from outside an exception handler. (see the commit I just committed)

Nora-Olivia-Ammann (approved):

PR #761, opened 2024-01-24
fix: fully mask passwords in logfile (DEV-3225)

jnussbaum:


src/dsp_tools/utils/connection_live.py
```diff
@@ -307,12 +307,12 @@ def _anonymize(self, data: dict[str, Any] | None) -> dict[str, Any] | None:
...
                 data["Authorization"] = f"Bearer {self._mask(match.group(1))}"
         if "password" in data:
-            data["password"] = self._mask(data["password"])
+            data["password"] = "*" * len(data["password"])
         return data
 
     def _mask(self, sensitive_info: str) -> str:
         unmasked_until = 5
-        if len(sensitive_info) <= unmasked_until:
+        if len(sensitive_info) <= unmasked_until * 2:
```
jnussbaum, line 315: this is for all non-password fields: it makes sure that secrets shorter than 10 are always fully masked.

Nora-Olivia-Ammann (approved):

PR #762, opened 2024-01-24
chore: tidy up exceptions

jnussbaum:


src/dsp_tools/models/exceptions.py
```diff
@@ -83,25 +79,8 @@ class UserError(BaseError):
...
-    Represents an error raised in the context of the XML import.
-    """
-
-    _message: str
-
-    def __init__(self, msg: str):
-        self._message = msg
-
-    def __str__(self) -> str:
-        return f"XML-ERROR: {self._message}"
```
jnussbaum, line 107: This entire brimborium in unnecessary, because it inherits the attribute, the __init__ and the __str__ from BaseError. And since no other code depends on this specific string representation, it can be safely deleted.

Nora-Olivia-Ammann (approved):

PR #764, opened 2024-01-25
chore: avoid mutable default values in class attributes (DEV-3234)

jnussbaum:
In some places, dsp-tools uses defaults for mutable class attributes, which is a bad idea: https://docs.astral.sh/ruff/rules/mutable-class-default/

### DIFF
--- a/src/dsp_tools/commands/id2iri.py
+++ b/src/dsp_tools/commands/id2iri.py
@@ -1,5 +1,6 @@
 import copy
 import json
+from collections.abc import Mapping
 from datetime import datetime
 from pathlib import Path
 
@@ -61,7 +62,7 @@ def _parse_json_file(json_file: Path) -> dict[str, str]:
 
 def _replace_resptrs(
     tree: etree._Element,
-    mapping: dict[str, str],
+    mapping: Mapping[str, str],
     used_mapping_entries: set[str],
 ) -> tuple[etree._Element, set[str]]:
     """
@@ -94,7 +95,7 @@ def _replace_resptrs(
 
 def _replace_salsah_links(
     tree: etree._Element,
-    mapping: dict[str, str],
+    mapping: Mapping[str, str],
     used_mapping_entries: set[str],
 ) -> tuple[etree._Element, set[str]]:
     """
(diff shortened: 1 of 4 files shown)


src/dsp_tools/commands/id2iri.py
```diff
@@ -61,7 +62,7 @@ def _parse_json_file(json_file: Path) -> dict[str, str]:
 
 def _replace_resptrs(
     tree: etree._Element,
-    mapping: dict[str, str],
+    mapping: Mapping[str, str],
```
jnussbaum, line 65: By doing this, I loosen the constraint on the input, which is okay - or even desirable, because input type annotations should be as loose as possible, while return type annotations should be as strict as possible.

test/unittests/commands/test_id2iri.py
```diff
@@ -16,12 +16,14 @@
...
-        "test_thing_with_ark_1": "http://rdfh.ch/082E/qwasddoiu8_6flkjh67dss",
-    }
+    mapping = MappingProxyType(
+        {
+            "test_thing_0": "http://rdfh.ch/082E/-lRvrg7tQI6aVpcTJbVrwg",
+            "test_thing_1": "http://rdfh.ch/082E/JK63OpYWTDWNYVOYFN7FdQ",
+            "test_thing_2": "http://rdfh.ch/082E/1l63Oasdfopiujlkmn78ak",
+            "test_thing_with_ark_1": "http://rdfh.ch/082E/qwasddoiu8_6flkjh67dss",
+        }
+    )
```
jnussbaum, line 26: This is the most pythonic way to make a dictionary immutable. Read more about it on https://adamj.eu/tech/2022/01/05/how-to-make-immutable-dict-in-python/

src/dsp_tools/commands/id2iri.py
Nora-Olivia-Ammann: why is this a default value?

jnussbaum: It isn't. The thing with the default value is further below. This here is just a follow-up change

Nora-Olivia-Ammann: Because the data type would not match?
Can we then still give a dict without mypy complaining?

jnussbaum: Further below, I replaced the mutable dict with the immutable MappingProxyType. But MappingProxyType is not a subtype of dict, so the type has to be annotated with a supertype, "collections.abc.Mapping" in our case. This is the overarching supertype of dict and MappingProxyType.

And yes, of course the actual input doesn't change, we still pass a dict to the method.

PR #765, opened 2024-01-25
fix: prevent crash when venv isn't activated correctly (DEV-3233)

jnussbaum:


Nora-Olivia-Ammann (approved):
Looks good, but I don't know it well enough to really judge. Could we test this?

jnussbaum:
> Looks good, but I don't know it well enough to really judge. Could we test this?

On my machine, it resolved the issue that I had, so I guess it passed the test :-)

PR #772, opened 2024-01-30
chore(project-create): clean up ruff PLR0912

Nora-Olivia-Ammann:


src/dsp_tools/commands/project/create/project_create.py
```diff
@@ -117,32 +107,30 @@ def _update_basic_info_of_project(
 
     Args:
         project: the project to be updated (must exist on the DSP server)
-        shortcode: shortcode of the project (must be the same as in the existing project)
-        shortname: shortname of the project (must be the same as in the existing project)
-        longname: longname of the project
-        descriptions: descriptions of the project in different languages
-        keywords: keywords of the project
+        project_definition: class with project info
```
jnussbaum, line 110: to be very precise, you cannot hand over a class to a function, but rather an object. 
(Occurs several times)

src/dsp_tools/commands/project/create/project_create.py
```diff
@@ -1120,17 +1089,55 @@ def create_project(
...
+    all_ontos = project_json["project"]["ontologies"]
+    # rectify the "hlist" of the "gui_attributes" of the properties
+    for onto in all_ontos:
+        if onto.get("properties"):
+            onto["properties"] = _rectify_hlist_of_properties(
+                lists=all_lists or [],
+                properties=onto["properties"],
+            )
+
+    return project_def, all_lists, all_ontos
```
jnussbaum, line 1143: I think this function does too many things. An indicator for this is the name: It prepares the project info, the ontos and the lists, so it does 3 things. Another indicator is the long list of returned objects (`tuple[ProjectDefinition, list[dict[str, Any]] | None, list[dict[str, Any]]]`). 

Some ideas how this could be improved:
- put the lists inside the ProjectDefinition
- take the schema validation and the hlist rectification out of this function

src/dsp_tools/commands/project/create/project_create.py
```diff
@@ -81,18 +72,21 @@ def _create_project_on_server(
...
-        shortcode=shortcode,
-        shortname=shortname,
-        longname=longname,
-        description=LangString(descriptions),  # type: ignore[arg-type]
-        keywords=set(keywords) if keywords else None,
+        shortcode=project_definition.shortcode,
+        shortname=project_definition.shortname,
+        longname=project_definition.longname,
+        description=LangString(project_definition.descriptions),  # type: ignore[arg-type]
+        keywords=set(project_definition.keywords) if project_definition.keywords else None,
```
jnussbaum, line 79: if you feel like, you could teach the constructor of `Project` how to deal with a `ProjectDefinition`

jnussbaum (approved):
This is definitively an improvement, LGTM.
I have made some comments how you could even further improve it.

src/dsp_tools/commands/project/create/project_create.py
Nora-Olivia-Ammann: It would be more elegant to read, but in my opinion it would increase the dependencies too much. I aimed to keep the functions as stupid as possible.

src/dsp_tools/commands/project/create/project_create.py
Nora-Olivia-Ammann: I have taken the function apart, but regarding the lists in the ProjectDefinition I had the following considerations:
There are three distinct parts of the json: project meta-data, lists and ontologies. In my opinion all of these should be an independent class. If I added the lists to the ProjectDefinition, it would make this class more complex and mix those sections. We already have an ontology class, which in my opinion is quite atrocious. A clean overhaul of this code is definitively a good idea, but I don't think this is that time. Because ideally we could combine it with an implementation of rdflib. So I would also like to wait with a list class until we do a rdflib implementation.
So I decided to do a minimally invasive cleaning.

jnussbaum: sure, that's fine

src/dsp_tools/commands/project/create/project_create.py
```diff
@@ -1120,17 +1093,62 @@ def create_project(
...
+    # validate against JSON schema
+    validate_project(project_json, expand_lists=False)
+    print("    JSON project file is syntactically correct and passed validation.")
+    logger.info("JSON project file is syntactically correct and passed validation.")
+
+    return project_def
+
+
+def _get_all_lists(project_json: dict[str, Any]) -> list[dict[str, Any]] | None:
+    # expand the Excel files referenced in the "lists" section of the project (if any), and add them to the project
```
jnussbaum, line 1136: This comment is not true any more

Nora-Olivia-Ammann: I will do it in the next PR (it is also cleaning)

PR #777, opened 2024-01-31
feat(excel2json): add optional column "subject" to properties.xlsx (DEV-3253)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -165,6 +152,33 @@ def _do_property_excel_compliance(df: pd.DataFrame, excelfile: str) -> None:
...
+        "comment_fr",
+        "comment_it",
+        "comment_rm",
+        "subject",
+    }
+    in_df_cols = set(df.columns)
+    if "subject" not in in_df_cols:
+        warnings.warn(
+            "It is possible to add a constraint to the subject of a property.\n"
+            'To do this you can enter a resource name in the column "subject".'
```
jnussbaum, line 173: This message is not ideal:
- It's a warning, but then the user is only told that it is possible to do something. For me, a warning only makes sense if we tell the user that he SHOULD do something, and that there are negative consequences if he doesn't. So we should also discuss in the RDU team if the subject constraint is something completely optional, or rather something that should be done if ever possible.
- If I understand correctly, the code at this place only checks if the column is present or not. So it doesn't make sense to say that the user `can enter a resource name in the column "subject"`, if the column is not there.

docs/assets/data_model_templates/rosetta (rosetta)/properties.xlsx
jnussbaum, line 1: In addition to this file, the file `testdata/excel2json/excel2json_files/test-name (test_label)/properties.xlsx` should also be updated.

test/unittests/commands/excel2json/test_properties.py
```diff
@@ -626,6 +628,38 @@ def test_invalid_gui_attrib_values(self) -> None:
...
+            }
+        )
+        returned_df = e2j._add_optional_columns(df=original_df)
+        # as the columns are extracted via a set, they are not sorted and may appear in any order,
+        # this would cause the validation to fail
+        returned_df = returned_df.sort_index(axis=1)
+        assert_frame_equal(expected_df, returned_df)
+        # if all columns exist, the df should be returned unchanged
+        unchanged_df = e2j._add_optional_columns(df=expected_df)
+        assert_frame_equal(expected_df, unchanged_df)
```
jnussbaum, line 661: To be very precise, this should be in a separate function

jnussbaum (commented):
See my comments

test/unittests/commands/excel2json/test_properties.py
```diff
@@ -626,6 +628,57 @@ def test_invalid_gui_attrib_values(self) -> None:
...
```
jnussbaum, line 661: But now, this is a duplicate? I guess you have to remove these 3 lines

testdata/excel2json/excel2json_files/test-name (test_label)/properties.xlsx
jnussbaum, line 1: I think it would  be good to fill the new column with some data, and then test if the entries are correctly transformed into JSON

Nora-Olivia-Ammann: Question, this does not correspond with the resource excel right? It is completely independent?

jnussbaum (approved):

PR #781, opened 2024-01-31
chore(project_validate): fix one ruff PLR0912

Nora-Olivia-Ammann:


jnussbaum (changes requested):
See my comment about the signature. Otherwise, it looks good.

PR #782, opened 2024-01-31
chore(model-project): fix one ruff PLR0912

Nora-Olivia-Ammann:


jnussbaum (approved):
My only suggestion is to remove the "action" from `rename _toJsonObj_action_update` and from `_toJsonObj_action_create`, otherwise LGTM

PR #788, opened 2024-02-01
chore(models-resourceclass): fix ruff PLR0912

Nora-Olivia-Ammann:


src/dsp_tools/commands/project/models/resourceclass.py
```diff
@@ -110,120 +110,75 @@ def ptype(self) -> Ptype:
...
+            con=con,
+            context=context,
+            ontology_id=ontology_id,
+            property_id=property_id,
+            cardinality=cardinality,
+            gui_order=gui_order,
+            ptype=ptype,
+        )
+
+    @classmethod
```
jnussbaum, line 142: In this file we have some `@classmethod`, some `@staticmethod`, and some which are neither of the two. Have you got a system behind this?

Nora-Olivia-Ammann: I followed to formalities that were suggested by the ide. I imagine other functions in the code may also be class / static methods but are not declared as such. But I did not systematically look for those, only declared the functions I touched anyway.

jnussbaum (approved):

PR #792, opened 2024-02-01
chore: delete dead code fragments

Nora-Olivia-Ammann:


src/dsp_tools/utils/create_logger.py
```diff
@@ -27,8 +27,6 @@ def _make_handler(
         handler instance
     """
     formatter = logging.Formatter(fmt="{asctime} {filename: <30} l. {lineno: >4} {levelname: <8} {message}", style="{")
-    formatter.default_time_format = "%Y-%m-%d %H:%M:%S"
-    formatter.default_msec_format = "%s.%03d"
```
jnussbaum, line 31: We need these 2 lines, because they specify the format of the timestamp in the log file. Without this, the default formatting would be applied, but the default differs from the format the DSP-API uses. 

Back then, when I added these 2 lines, my goal was that the logs of DSP-API and DSP-TOOLS use the same time stamp format, so that if you investigate an incident, you can easily find the corresponding lines in both log files.

src/dsp_tools/models/exceptions.py
```diff
@@ -42,29 +42,6 @@ def __init__(self, custom_msg: str | None = None, keep_default_msg: bool = True)
                 super().__init__(default_msg)
 
 
-class RetryError(BaseError):
```
jnussbaum, line 45: in `entry_point.py`, we have a reference to this error class (in a docstring), which should also be removed

PR #796, opened 2024-02-01
chore(models-ontology): delete actions object

Nora-Olivia-Ammann:


src/dsp_tools/commands/project/models/ontology.py
```diff
@@ -265,63 +265,48 @@ def allOntologiesFromJsonObj(cls, con: Connection, json_obj: Any) -> list["Ontol
...
-                raise BaseError("There must be a valid project given!")
-            tmp = {
-                knora_api + ":ontologyName": self._name,
-                knora_api + ":attachedToProject": {"@id": self._project},
-                rdfs + ":label": self._label,
-                "@context": self._context.toJsonObj(),
-            }
-            if self._comment is not None:
-                tmp[rdfs + ":comment"] = self._comment
-        elif action == Actions.Update:
```
jnussbaum, line 287: just to double-check: was updating used nowhere?

jnussbaum (approved):

src/dsp_tools/commands/project/models/ontology.py
jnussbaum: ok, fine

PR #802, opened 2024-02-01
chore(models-project): remove all project update code

Nora-Olivia-Ammann:


src/dsp_tools/commands/project/create/project_create.py
```diff
@@ -51,23 +49,13 @@ def _create_project_on_server(
     """
     all_projects = Project.getAllProjects(con=con)
     if project_definition.shortcode in [proj.shortcode for proj in all_projects]:
-        project_local = Project(con=con, shortcode=project_definition.shortcode)
-        project_remote: Project = project_local.read()
         msg = (
-            f"A project with shortcode {project_definition.shortcode} already exists on the DSP server. Updating it..."
```
Nora-Olivia-Ammann, line 57: @jnussbaum Last week we decided to raise an error here, but Balduin pointed out that the rest of the program also updates lists, ontologies, etc, which is a feature we want. Therefore I changed it back to a message.

test/unittests/commands/project/test_models_project.py
```diff
@@ -0,0 +1,54 @@
...
+def test_return_values(project: Project) -> None:
+    assert project.iri == "http://rdfh.ch/test"
+    assert project.shortcode == "0FF0"
+    assert project.shortname == "test_project"
+    assert project.longname == "Test Project"
+    assert project.description["en"] == "This is a test project"
+    assert project.description["de"] == "Das ist ein Testprojekt"
+    assert project.selfjoin is False
+    assert project.status is True
+    assert project.keywords == set()
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: The reason why I tested it, was because it originally was, and the variables are not accessed directly but through functions with `@property` decorators. Do you think this does not need to be tested?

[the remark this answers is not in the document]
BalduinLandolt: makes sense. To me this seems excessive, but admittedly, with properties, this could in theory break (unlike dataclasses), so it's also ok to have this test here. And it doesn't hurt us either.

src/dsp_tools/commands/project/create/project_create.py
jnussbaum: I'm not sure if this is a feature we want. I rather see it this way: If a user executes `dsp-tools create xyz.json`, and the JSON file contains some info that is already on the server (e.g. a list that is partially on the server, but some new nodes will be updated), then this should be an error. I dislike all this updating stuff.

Nora-Olivia-Ammann: That shouldn't happen, but it is possible to add a completely new list:
https://github.com/dasch-swiss/dsp-tools/blob/8c667a69d981d4dff0ca2f1fc3f8534c95d93ca0/src/dsp_tools/commands/project/create/project_create_lists.py#L120

BalduinLandolt: I also don't like the idea of updating anything. But as long as we don't have any way of creating an ontology without the `create` command, I find it crucial to be able to add ontologies to existing projects. And adding an ontology to a project is not the same as updating a project.

src/dsp_tools/commands/project/create/project_create.py
```diff
@@ -51,23 +49,13 @@ def _create_project_on_server(
...
+            f"The project with the shortcode'{project_definition.shortcode}' already exists on the server.\n"
```
jnussbaum, line 53:
```suggestion
            f"The project with the shortcode '{project_definition.shortcode}' already exists on the server.\n"
```

jnussbaum (approved):

test/unittests/commands/project/test_models_project.py
```diff
@@ -0,0 +1,54 @@
...
+        "longname": "Test Project",
+        "description": [
+            {"language": "en", "value": "This is a test project"},
+            {"language": "de", "value": "Das ist ein Testprojekt"},
+        ],
+        "selfjoin": False,
+        "status": True,
+    }
+    assert res_json == expected
+
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I agree, but I do not want to do this here, it is not related and this PR is large enough.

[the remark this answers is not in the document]
jnussbaum: That's fine for me

PR #804, opened 2024-02-05
chore(test-create-project): turn unittests into pytests

Nora-Olivia-Ammann:


BalduinLandolt (approved):
I didn't compare everything in detail, but from what I can see, fixtures look good, and if the test still passes, it should be good

PR #807, opened 2024-02-12
chore: rename value.py file into formatted_text_value.py

Nora-Olivia-Ammann:


BalduinLandolt (approved):
Hitting "Approve" so merging is not blocked.

IMO it is too fine-grained to have a file per value type, so in the long run I would hope to see more than just `FormattedTextValue` in `value.py`. And not having to have such long file names where not needed, keeps the imports looking tidier.  
So all in all I'm not entirely convinced by this change. But it's up to you - and ultimately also something that could easily be changed back in the future, if need be. So if you feel this makes more sense, then feel free to merge it

PR #808, opened 2024-02-12
chore: renaming functions in preparation for TextValue verification

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -14,7 +14,7 @@
 genericPrefixedOntology: Pattern[str] = regex.compile(r"^[\w\-]+:\w+$")
 
 
-def do_xml_consistency_check(onto_client: OntologyClient, root: etree._Element) -> None:
+def do_xml_consistency_check_with_ontology(onto_client: OntologyClient, root: etree._Element) -> None:
```
Nora-Olivia-Ammann, line 17: @BalduinLandolt 
I am not sure how I want to re-organise this. For the TextValue check we need the same response fromt the API, but we need to filter out different information. So I do want to combine it, but I am unsure if it is a good idea to add the lower level functions that do the specific operations into this file. I think it could get very confusing as these checks do similar but slightly different things.
What do you think?

BalduinLandolt: not super familiar with this code... could "low level stuff" be hidden inside the OntologyClient? Otherwise maybe we could discuss the issue in detail together, so I understand a bit better

BalduinLandolt (approved):

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
Nora-Olivia-Ammann: That would be nice thanks!

Nora-Olivia-Ammann:
@BalduinLandolt 
More work is needed to reorganise some functions but I want to do that separately. Here no actual functional changes should take place.

BalduinLandolt (approved):
Looks good. Naming can probably still improved in some places, but I can't think of anything better for now, so I don't think we should be bothered too much.

PR #810, opened 2024-02-13
feat: check if text encoding in data conforms to gui-element in ontology (DEV-3279)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -149,3 +164,66 @@ def _get_separate_prefix_and_iri_from_onto_prop_or_cls(
...
+
+def _get_prop_encoding_from_one_property(res_id: str, property: etree._Element) -> TextValueData:
+    prop_name = property.attrib["name"]
+    child_attrib = [x.attrib for x in property.iterchildren()]
+    encodings = {x.get("encoding") for x in child_attrib}
+    return TextValueData(res_id, prop_name, encodings)
+
+
+def _check_correctness_one_prop(text_val: TextValueData, text_prop_look_up: TextValuePropertyGUI) -> bool:
+    text_encoding = text_val.encoding
```
Nora-Olivia-Ammann, line 210: I am not a fan of either of the two options (one of them commented out).
Ruff is happy with the first one, but does not accept the second.
While the second is ugly I wonder if it is easier to understand?
What would your preference be?

docs/file-formats/json-project/ontologies.md
```diff
@@ -569,7 +569,7 @@ Example:
       "en": "Comment",
       "fr": "Commentaire"
     },
-    "gui_element": "SimpleText"
+    "gui_element": "Richtext"
```
Nora-Olivia-Ammann, line 572: The knora-api ontology specifies `hasComment` as `Richtext`

test/unittests/commands/xmlupload/test_ontology_problem_models.py
```diff
@@ -1,88 +1,149 @@
...
-    onto = InvalidOntologyElementsInData([("clsA", ["idA"], "wrong")], [], ontos_on_server=["test"])
-    msg = onto._compose_problem_string_for_cls()
-    assert msg == (
-        "The following resource(s) have an invalid resource type:\n\n"
-        "    Resource Type: 'clsA'\n"
-        "    Problem: 'wrong'\n"
-        "    Resource ID(s):\n"
-        "    - idA"
-    )
+class TestInvalidOntologyElementsInData:
```
Nora-Olivia-Ammann, line 11: I did not change any tests, just put them in a class but ruff seems to have reformatted a lot of things.

Nora-Olivia-Ammann, line 11: I did not change any tests, just put them in a class but ruff seems to have reformatted it a enough to show as a diff.

BalduinLandolt (commented):
I over all like this PR, but I remarked some things I would suggest changing. (Mostly cosmetic but some of it more regarding how to model stuff.)  
I'm happy to have another look when/if you address them

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
jnussbaum: Is the problem perhaps the underlying data structure you use? Does it make sense that the set can contain `None` AND `utf-8` at the same time? 

So if you go to the code where this set is constructed, perhaps you find a way to write only strings into this set.

Nora-Olivia-Ammann: Regarding mixed types:
Yes it is possible, it really should not be but in our test data we had a case of (wrong) mixed types: https://github.com/dasch-swiss/dsp-tools/blob/443195f68023eb267e34568e0fa9a215886d3944/testdata/xml-data/test-data-systematic.xml#L186

Regarding `None`:
I misremembered this, I knew we don't have to specify an encoding and then it is by default utf8, I though this default value would happen during the xmlupload and not the excel2xml.make_text_prop. My bad, I will fix that and remove it as an option. 

Regarding an Enum class:
I have thought about that but it seemed overkill for so little options. I look into doing it with a union type.

BalduinLandolt: Regarding mixed types:  
if it is no longer in the test data, could the xml schema be adjusted to enforce that it never happens?

Enum/Union:  
Yes... python enums are not great, too. But if there are only two possible states, you can simply capture the information with a boolean. That's the easiest, I suppose

Nora-Olivia-Ammann: I will look into fixing the xml schema now, and would do that first, so I can then simplify the logic here. Thanks for your input!

Nora-Olivia-Ammann:
I will let this one die, and migrate the pertinent changes into a new branch after the mixed encoding checks are done.

PR #814, opened 2024-02-16
chore: refactor validate xml in preparation for extension

Nora-Olivia-Ammann:


src/dsp_tools/utils/validate_data_xml.py
BalduinLandolt, line 1: I would propose renaming this file - maybe something like `xml_validation.py`?  
to me, the imperative form `validate_xyz` sounds like an instruction and therefor like a function/method name.

src/dsp_tools/utils/xml_utils.py
```diff
@@ -103,3 +103,22 @@ def _parse_xml_file(input_file: Union[str, Path]) -> etree._ElementTree[etree._E
     """
     parser = etree.XMLParser(remove_comments=True, remove_pis=True)
     return etree.parse(source=input_file, parser=parser)
+
+
+def remove_namespaces_from_xml(
+    data_xml: Union[etree._ElementTree[etree._Element], etree._Element],
+) -> Union[etree._ElementTree[etree._Element], etree._Element]:
```
BalduinLandolt, line 110: Not for this PR. But since we discussed changing the XML format - the existence of this function to me suggests, that the new XML format should not use any namespaces.

BalduinLandolt (approved):
I didn't compare the test code that's been moved around, but I trust this has not functionally changed. So looks good, thanks!

PR #817, opened 2024-02-20
chore: renaming functions of ontology checks in xmlupload

Nora-Olivia-Ammann:
- Name changes as suggested by Balduin in https://github.com/dasch-swiss/dsp-tools/pull/810/files
- Chaning the `UserError` in the the `InputError` as is the convention now if an error follows the new protocol, which this does.

### DIFF
--- a/src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
+++ b/src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
@@ -10,7 +10,7 @@ from dsp_tools.commands.xmlupload.models.ontology_problem_models import (
     InvalidOntologyElementsInData,
 )
 from dsp_tools.commands.xmlupload.ontology_client import OntologyClient
-from dsp_tools.models.exceptions import UserError
+from dsp_tools.models.exceptions import InputError
 
 defaultOntologyColon: Pattern[str] = regex.compile(r"^:\w+$")
 knoraUndeclared: Pattern[str] = regex.compile(r"^\w+$")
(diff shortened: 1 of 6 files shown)


src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -22,30 +22,30 @@ def do_xml_consistency_check_with_ontology(onto_client: OntologyClient, root: et
...
         default_ontology_prefix=onto_client.default_ontology,
         onto_lookup=onto_client.get_all_ontologies_from_server(),
     )
     classes_in_data, properties_in_data = _get_all_classes_and_properties_from_data(root)
-    _find_if_all_classes_and_properties_exist_in_onto(classes_in_data, properties_in_data, onto_check_info)
+    _find_all_classes_and_properties_exist_in_onto(classes_in_data, properties_in_data, onto_check_info)
 
 
-def _find_if_all_classes_and_properties_exist_in_onto(
+def _find_all_classes_and_properties_exist_in_onto(
```
jnussbaum, line 42: For me, this name grammatically not clear. It says: "Find all classes and properties exist in onto". The conjugated "exist" does not fit into this sentence.

Perhaps you mean one of the following:

- Find all classes and properties in the onto
- Check all classes and properties

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -96,17 +96,17 @@ def _get_all_property_names_and_resource_ids_one_resource(
     return prop_dict
 
 
-def _check_if_all_class_types_exist(
+def _check_all_class_types_exist(
```
jnussbaum, line 99: Again, the sentence "Check all class types exist" is grammatically not understandable. Perhaps you mean: "Check all class types"?

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -96,17 +96,17 @@ def _get_all_property_names_and_resource_ids_one_resource(
...
     problem_list = []
     for cls_type, ids in classes.items():
-        if problem := _check_if_one_class_type_exists(cls_type, onto_check_info):
+        if problem := _check_one_class_type_exists(cls_type, onto_check_info):
             problem_list.append((cls_type, ids, problem))
     return problem_list
 
 
-def _check_if_one_class_type_exists(cls_type: str, onto_check_info: ProjectOntosInformation) -> str | None:
+def _check_one_class_type_exists(cls_type: str, onto_check_info: ProjectOntosInformation) -> str | None:
```
jnussbaum, line 109: The old name was grammatically correct, but the new one isn't.

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
```diff
@@ -18,39 +18,39 @@ class ProjectOntosInformation:
     onto_lookup: dict[str, OntoInfo]
 
 
-def extract_classes_properties_from_onto(onto_graph: list[dict[str, Any]]) -> OntoInfo:
+def extract_classes_properties_from_onto(onto_json: list[dict[str, Any]]) -> OntoInfo:
```
jnussbaum, line 21: Grammatically, it is not correct to say: "Extract classes properties from onto". It is necessary to have an "and" between the 2 things that you extract from the onto.

jnussbaum (changes requested):
Some of the new names are grammatically not correct, and hence, not understandable. See my comments

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
Nora-Olivia-Ammann: Well for me the "if" clause was no problem, then it was clear...
I am not sure however, if "check" is a good choice, because we not only check it but find the problems and return those.

jnussbaum: So if you had to formulate 1 sentence that best describes what this function does, what would it be? If we restrain ourselves to 1 main sentence, without subclause, then I would vote for "Check all classes and properties". 

Of course, it also finds problems and returns them, but perhaps that's too much to pack into 1 function name.

jnussbaum (approved):
LGTM

PR #818, opened 2024-02-20
feat(xmlupload): check if the encoding in the text-prop is consistent (DEV-3296)

Nora-Olivia-Ammann:


src/dsp_tools/utils/xml_validation_models.py
```diff
@@ -0,0 +1,143 @@
...
+        """
+        This method composes an error message for the user.
+
+        Returns:
+            the error message and a dataframe with the errors if they exceed 50
+        """
+        msg = (
+            "\nSome <text-prop> elements contain <text> elements that use both 'xml' and 'utf8' encoding.\n"
+            "Only one encoding type can be used per <text-prop> element."
+        )
```
jnussbaum, line 117: The order how you do things in this method seems very unnatural to me. You do:

- compose a base message
- create a dataframe
- if the dataframe is big enough to be saved --> return it together with the message
    - on the caller side, you add the info that the file was saved. For me, this would belong here.
- compose an additional message from the dataframe 
    - here it gets really confusing: at that point, the dataframe will not be used, but you use it to create an "additional" message 
    - it is not clear what "additional" means in this context
- return the additional message.

My suggestion for a more natural flow of things: 

- compose a base message
- based on `problematic_resources`, determine whether to save a df or not
- if not:
    - based on `problematic_resources`, compose a print message
    - return it
- if yes:
    - create a dataframe
    - to the base message, add that a dataframe was saved, and where it was saved
    - return df + message

src/dsp_tools/utils/xml_validation_models.py
```diff
@@ -0,0 +1,143 @@
...
+        if len(df) > maximum_prints:
+            return msg, df
+        additional_msg = self._make_msg_from_df(df)
+        return msg + grand_separator + additional_msg, None
+
+    def _get_problems_as_df(self) -> pd.DataFrame:
+        df = pd.DataFrame(
+            {
+                "Resource ID": list(x.resource_id for x in self.problematic_resources),
+                "Property Name": list(x.property_name for x in self.problematic_resources),
```
jnussbaum, line 128:
```suggestion
                "Resource ID": [x.resource_id for x in self.problematic_resources],
                "Property Name": [x.property_name for x in self.problematic_resources],
```

src/dsp_tools/utils/xml_validation_models.py
```diff
@@ -0,0 +1,143 @@
...
+            }
+        )
+        return df.sort_values(by=["Resource ID", "Property Name"], ignore_index=True)
+
+    @staticmethod
+    def _make_msg_from_df(df: pd.DataFrame) -> str:
+        groups = df.groupby(by="Resource ID")
+        return medium_separator.join(
+            [InconsistentTextValueEncodings._make_msg_for_one_resource(str(_id), res_df) for _id, res_df in groups]
```
jnussbaum, line 137: it seems unnatural to me that these 2 methods are static. I would expect them both to be normal instance methods, and to call them with self.xy()

test/unittests/utils/test_xml_validation_models.py
```diff
@@ -0,0 +1,374 @@
...
+                <uri permissions="prop-default">https://dasch.swiss</uri>
+            </uri-prop>
+            <boolean-prop name=":hasBoolean">
+                <boolean permissions="prop-default">true</boolean>
+            </boolean-prop>
+        </resource>
+        """
+    )
+    res = _get_id_prop_encoding_from_one_resource(test_props)
+    assert res is None
```
jnussbaum, line 35: if you follow my recommendation to return an empty list instead of None, then you have to write:

```suggestion
    assert not res
```

test/unittests/utils/test_xml_validation_models.py
```diff
@@ -0,0 +1,374 @@
...
+
+
+def test_get_prop_encoding_from_all_properties_mixed() -> None:
```
jnussbaum, line 38: why mixed?

test/unittests/utils/test_xml_validation_models.py
```diff
@@ -0,0 +1,374 @@
...
+                </resource>
+                <resource label="resC" restype=":TestThing2" id="resC" permissions="res-default">
+                    <resptr-prop name=":hasResource2">
+                        <resptr permissions="prop-default">resB</resptr>
+                    </resptr-prop>
+                    <text-prop name=":hasSimpleText">
+                        <text encoding="utf8">Text</text>
+                    </text-prop>
+                </resource>
+                <resource label="resC" restype=":TestThing2" id="resD" permissions="res-default">
```
jnussbaum, line 193:
```suggestion
                <resource label="resD" restype=":TestThing2" id="resD" permissions="res-default">
```

test/unittests/utils/test_xml_validation_models.py
```diff
@@ -0,0 +1,374 @@
...
+                <resource label="resC" restype=":TestThing2" id="resC" permissions="res-default">
+                    <resptr-prop name=":hasResource2">
+                        <resptr permissions="prop-default">resB</resptr>
+                    </resptr-prop>
+                    <text-prop name=":hasSimpleText">
+                        <text encoding="utf8">Text 1</text>
+                        <text encoding="utf8">Text 2</text>
+                    </text-prop>
+                </resource>
+                <resource label="resC" restype=":TestThing2" id="resD" permissions="res-default">
```
jnussbaum, line 267:
```suggestion
                <resource label="resD" restype=":TestThing2" id="resD" permissions="res-default">
```

test/unittests/utils/test_xml_validation_models.py
```diff
@@ -0,0 +1,374 @@
...
+                <resource label="resC" restype=":TestThing2" id="resC" permissions="res-default">
+                    <resptr-prop name=":hasResource2">
+                        <resptr permissions="prop-default">resB</resptr>
+                    </resptr-prop>
+                    <text-prop name=":hasSimpleText">
+                        <text encoding="utf8">Text 1</text>
+                        <text encoding="xml">Text 2</text>
+                    </text-prop>
+                </resource>
+                <resource label="resC" restype=":TestThing2" id="resD" permissions="res-default">
```
jnussbaum, line 307:
```suggestion
                <resource label="resD" restype=":TestThing2" id="resD" permissions="res-default">
```

jnussbaum (changes requested):
I like the intention of the PR, and also most of how it is done, but there are 2 main reasons why I'd like to ask you to revise it:

- the present PR contains some changes from another pending PR
- there are many parts of the code that could be simplified / made more pythonic (see my comments)

src/dsp_tools/utils/xml_validation_models.py
```diff
@@ -0,0 +1,143 @@
...
+    wrong_props = [x for x in text_props if not _check_only_one_valid_encoding_used_one_prop(x.encoding)]
+    if len(wrong_props) == 0:
+        return None
+    return wrong_props
+
+
+def _check_only_one_valid_encoding_used_one_prop(text_encodings: set[str]) -> bool:
+    if text_encodings == {"xml"} or text_encodings == {"utf8"}:  # noqa: PLR1714
+        return True
+    return False
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I guess since the XML does not allow any other values we can scrap it.

src/dsp_tools/utils/xml_validation_models.py
Nora-Olivia-Ammann: They are static because they do not need to have access to any class variables at any point. In case of static functions I prefer to declare them so, because one it is correct and for another it makes testing simpler because you need not create a class to call the function.

src/dsp_tools/utils/xml_validation_models.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I am not entirely happy with this either. My plan for this file is that it will also host the check if the correct encoding with respect to the property is used. It seemed a bit like overkill to create two files, one with the check functions and another with the model, but I am open for suggestions.

src/dsp_tools/utils/xml_validation_models.py
```diff
@@ -0,0 +1,143 @@
+from dataclasses import dataclass
+
+import pandas as pd
+from lxml import etree
+
+list_separator = "\n    - "
+medium_separator = "\n----------------------------\n"
+grand_separator = "\n\n---------------------------------------\n\n"
+
+maximum_prints = 50
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I disagree:
- In other files that generate a error string (eg. src/dsp_tools/commands/xmlupload/models/ontology_problem_models.py) These are listed at the top, making it consistent.
- The idea is that the messages will be uniform, meaning other problems should have access to it. I would be happy to look into making these constants for the entire repo, but that should be another PR.
- I intend to add the problem model that generates the error string if the wrong encoding is used with respect to the property in this file. This will also need access to these variables.

src/dsp_tools/utils/xml_validation_models.py
Nora-Olivia-Ammann: Where the df is saved:
- I did it this way because in the past that is how we structured it. The rationale if I remember correctly is for better testing.
- Another reason for the location of the save message is, that it is in fact not saved in that function, therefore the message would be technically wrong in my opinion.

Composing the df:
- The reason why, I did it first, is because for a clean message, I want to group by resource id and then sort by prop name. This can both be done nicely in a df, since we may in any case create a df, I did not see a reason not to do it this way. In my opinion any other method I can think of that creates the same result is more complicated, so I don't think we should change this.

Additional msg:
- I will restructure as suggested below, then the problem with the variable name is solved.

src/dsp_tools/utils/xml_validation_models.py
```diff
@@ -0,0 +1,143 @@
...
+
+    Args:
+        root: root of the data xml document
+
+    Returns:
+          True and None if all the elements are consistent
+          False and a list of all the inconsistent <text-props>
+    """
+    text_props = _get_all_ids_prop_encoding_from_root(root)
+    if (checked_props := _check_only_one_valid_encoding_used_all_props(text_props)) is not None:
```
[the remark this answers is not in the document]
jnussbaum: Is it possible that you encountered this with pandas dataframe comparison? I'm almost 100% sure that in pure python, my suggestion is cleaner and will never raise a future warning. I don't think that your variant is a good habit. (Except for pandas, there it might be necessary)

src/dsp_tools/utils/xml_validation_models.py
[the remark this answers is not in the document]
jnussbaum: That sounds like a perfect use case for inheritance: If `Problem` was a class (not a protocol), and the `InconsistentTextValueEncodings` inherits from it, then you can add these fields to the base class, and all children will have the same values.

With the present solution, we have multiple files, each defines its own values, and if we want to change the values, we have to modify multiple files. 

So for right now, you can leave it as it is, but for the future, we should check if inheritance would be a better solution.

src/dsp_tools/utils/xml_validation_models.py
jnussbaum: Even if these methods don't access class variables, they access (static) class methods. And for this, I think that static is less elegant than non-static. For example, I think that on line 137, `InconsistentTextValueEncodings._make_msg_for_one_resource()` is slightly more difficult to understand than `self._make_msg_for_one_resource()`.

But perhaps it's a matter of taste, so if you prefer them to be static, it's okay for me.

src/dsp_tools/utils/xml_validation_models.py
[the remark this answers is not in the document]
jnussbaum: I would prefer to have to separate files, one called "models", one called "functions". But feel free to leave it as it is.

src/dsp_tools/utils/xml_validation.py
```diff
@@ -59,25 +68,16 @@ def validate_xml(input_file: Union[str, Path, etree._ElementTree[Any]]) -> bool:
...
         xmlschema = etree.XMLSchema(etree.parse(schema_file))
-    if isinstance(input_file, (str, Path)):
-        try:
-            data_xml = etree.parse(source=input_file)
-        except etree.XMLSyntaxError as err:
-            logger.error(f"The XML file contains the following syntax error: {err.msg}", exc_info=True)
-            raise InputError(f"The XML file contains the following syntax error: {err.msg}") from None
-    else:
-        data_xml = input_file
+    data_xml = parse_and_remove_comments_xml_file(input_file)
```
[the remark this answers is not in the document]
jnussbaum: I feel you, and I have done such things in the past myself, and I had to make the painful experience that mostly, you don't do yourself a favor. Because if the other PR is modified (as it is the case now), you have to migrate the  modifications. Plus, for the reviewer, it can be really confusing, because multiple things are mixed together.

But it's fine for now. I have approved the other PR. As soon as you have merged it and "tidied up" this one, I can give you an approve.

src/dsp_tools/utils/xml_validation_models.py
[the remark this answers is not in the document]
BalduinLandolt: In lxml you may also get these warnings, I think. But it's true that in pure python this behaviour guaranteed to work - the only case where you need `is not None` is when you want to distinguish actual `None` from some falsy value like empty string, empty list, `False`, etc.

src/dsp_tools/utils/xml_validation_models.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I added it as a point to the notion page that discusses the config file.
https://www.notion.so/dasch-swiss/New-global-config-object-be2645e1e9a14e5da185dd8b3c6abd8f

src/dsp_tools/utils/xml_validation_models.py
Nora-Olivia-Ammann: Thanks for the suggestion, I think I prefer Balduin's suggestion.

src/dsp_tools/utils/xml_validation.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: No worries, I tend not to do that either but as you both are not officially working, I did not wait for an approve. I do appreciate that you take a look even though you are ill. Otherwise the mess would be even greater.

jnussbaum (approved):
LGTM

src/dsp_tools/utils/xml_validation.py
```diff
@@ -119,3 +125,74 @@ def _find_xml_tags_in_simple_text_elements(
...
+        <text encoding="xml">Text 2</text>
+    </text-prop>
+    ```
+
+    Args:
+        root: root of the data xml document
+
+    Returns:
+          True and None if all the elements are consistent
+          False and a list of all the inconsistent <text-props>
```
jnussbaum, line 171: That's outdated now.

src/dsp_tools/utils/xml_validation.py
```diff
@@ -119,3 +125,74 @@ def _find_xml_tags_in_simple_text_elements(
...
+    return [_get_encodings_from_one_property(res_id, child) for child in list(resource.iterchildren(tag="text-prop"))]
+
+
+def _get_encodings_from_one_property(res_id: str, prop: etree._Element) -> TextValueData:
+    prop_name = prop.attrib["name"]
+    encodings = {x.attrib["encoding"] for x in prop.iterchildren()}
+    return TextValueData(res_id, prop_name, encodings)
+
+
+def _check_only_one_valid_encoding_used_all_props(text_props: list[TextValueData]) -> list[TextValueData]:
```
jnussbaum, line 197: I have troubles understanding the function name. And when I compare it to what the function actually does, I have even more question marks. 

If I look at the code, I would say something like: 

- "find illegal text props"
- "find text props that use multiple encodings"

PR #819, opened 2024-02-20
chore(xmlupload): improve parse xml function

Nora-Olivia-Ammann:
Currently for the validate xml, the comments were not removed. This caused problems when finding the `encodings` attributes. However, we need the namespaces for the first validation step. Therefore a function was created that only parses the file and removes the comments.

### DIFF
--- a/src/dsp_tools/utils/xml_utils.py
+++ b/src/dsp_tools/utils/xml_utils.py
@@ -30,8 +30,6 @@ def parse_and_clean_xml_file(input_file: Union[str, Path, etree._ElementTree[Any
         InputError: if the input is not of either the expected types
     """
 
-    # remove comments and processing instructions (commented out properties break the XMLProperty constructor)
-
     if isinstance(input_file, (str, Path)):
         tree = _parse_xml_file(input_file)
     else:
@@ -43,11 +41,13 @@ def parse_and_clean_xml_file(input_file: Union[str, Path, etree._ElementTree[Any
     return tree.getroot()
 
 
-def parse_and_remove_comments_xml_file(
+def parse_and_remove_comments_from_xml_file(
     input_file: Union[str, Path, etree._ElementTree[Any]],
 ) -> etree._Element:
     """
-    Parse an XML file with DSP-conform data
+    Parse an XML file with DSP-conform data,
+    and remove the comments and processing instructions
+    (commented out properties break the XMLProperty constructor)
 
     Args:
         input_file: path to the XML file, or parsed ElementTree
@@ -58,8 +58,6 @@ def parse_and_remove_comments_xml_file(
     Raises:
         InputError: if the input is not of either the expected types
     """
-
-    # remove comments and processing instructions (commented out properties break the XMLProperty constructor)
 
     if isinstance(input_file, (str, Path)):
         tree = _parse_xml_file(input_file)
--- a/src/dsp_tools/utils/xml_utils.py
+++ b/src/dsp_tools/utils/xml_utils.py
@@ -6,6 +6,7 @@ from typing import Any, Union
 
 from lxml import etree
 
+from dsp_tools.models.exceptions import InputError
 from dsp_tools.utils.create_logger import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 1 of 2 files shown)


Nora-Olivia-Ammann:
> Now we have 2 functions that almost do the same:
> 
>     * `parse_and_remove_comments_xml_file()`
> 
>     * `parse_and_clean_xml_file()`
> 
> 
> This is confusing, because it is almost duplicating things, but with slight deviations. This makes it very hard to understand the paths that the program can take.
> 
> These code paths should be disentangled.

I agree, the reason why I did not do this here is because it opens a typing can of worms. If you would like to combine the functions, I am happy to do that, but I suggest another PR for that.

jnussbaum:
> > Now we have 2 functions that almost do the same:
> > ```
> > * `parse_and_remove_comments_xml_file()`
> > 
> > * `parse_and_clean_xml_file()`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is confusing, because it is almost duplicating things, but with slight deviations. This makes it very hard to understand the paths that the program can take.
> > These code paths should be disentangled.
> 
> I agree, the reason why I did not do this here is because it opens a typing can of worms. If you would like to combine the functions, I am happy to do that, but I suggest another PR for that.

Why would you like to do it in another PR? AFAIU, you introduced the duplication in the present PR. It doesn't seem natural to me to introduce a duplication/bad architecture in one PR, and then schedule a follow-up PR...?

Nora-Olivia-Ammann:
> > > Now we have 2 functions that almost do the same:
> > > ```
> > > * `parse_and_remove_comments_xml_file()`
> > > 
> > > * `parse_and_clean_xml_file()`
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This is confusing, because it is almost duplicating things, but with slight deviations. This makes it very hard to understand the paths that the program can take.
> > > These code paths should be disentangled.
> > 
> > 
> > I agree, the reason why I did not do this here is because it opens a typing can of worms. If you would like to combine the functions, I am happy to do that, but I suggest another PR for that.
> 
> Why would you like to do it in another PR? AFAIU, you introduced the duplication in the present PR. It doesn't seem natural to me to introduce a duplication/bad architecture in one PR, and then schedule a follow-up PR...?

Because I wanted to focus on PR that introduce the new feature and this stood in the way. But if it is a deal-breaker, then I can do it here.

src/dsp_tools/utils/xml_utils.py
Nora-Olivia-Ammann, line 1: @jnussbaum 
I now added the suggested changes, if we wanted to be very consistent we should go through the repo and replace all the parse functions with one of these.

However I have a concern about the changes. The new function `parse_and_remove_comments_from_xml_file` returns the root. The function `_remove_qnames_and_transform_special_tags` uses the `.iter()` function. This function is not quite the same on a `etree._ElementTree[etree._Element]` (https://lxml.de/apidoc/lxml.etree.html#lxml.etree._ElementTree.iter) and on a `etree._Element` (https://lxml.de/apidoc/lxml.etree.html#lxml.etree._Element.iter). The main difference is that the `etree._Element.iter()` only does first depth iteration. While `etree._ElementTree.iter()` loops over all elements. The tests do not fail, but this difference still makes me hesitant to implement the change.
We could omit the `getroot()` but I don't like that option either as this is what we want in most places.

Thoughts?

jnussbaum: As far as I understand the docs, both `iter()` functions do exactly the same. The only difference is the wording of the docs, but if I look at the meaning behind the wording, I don't see a difference:

- ElementTree: "The iterator loops over all elements in this tree, in document order."
- Element: "Iterate over all elements in the subtree in document order (depth first pre-order), starting with this element."

Both iterate over all elements in document order.

jnussbaum (approved):

src/dsp_tools/utils/xml_utils.py
Nora-Olivia-Ammann: You are right, I misunderstood the wording.

PR #821, opened 2024-02-21
feat(xmlupload): check if text encoding in data conforms to type specified in ontology (DEV-3279)

Nora-Olivia-Ammann:


test/unittests/commands/xmlupload/test_ontology_problem_models.py
```diff
@@ -1,88 +1,147 @@
...
-    msg = onto._compose_problem_string_for_cls()
-    assert msg == (
-        "The following resource(s) have an invalid resource type:\n\n"
-        "    Resource Type: 'clsA'\n"
-        "    Problem: 'wrong'\n"
-        "    Resource ID(s):\n"
-        "    - idA"
-    )
+class TestInvalidOntologyElementsInData:
+    def test_print_problem_string_cls(self) -> None:
```
Nora-Olivia-Ammann, line 14: I just put this into a class nothing else was changed.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -155,3 +162,85 @@ def _get_separate_prefix_and_iri_from_onto_prop_or_cls(
...
+
+
+def _get_id_prop_encoding_from_one_resource(resource: etree._Element) -> list[TextValueData]:
+    res_id = resource.attrib["id"]
+    return [_get_prop_encoding_from_one_property(res_id, child) for child in resource.iterchildren(tag="text-prop")]
+
+
+def _get_prop_encoding_from_one_property(res_id: str, property: etree._Element) -> TextValueData:
+    prop_name = property.attrib["name"]
+    encoding = cast(AllowedEncodings, next(x.attrib["encoding"] for x in property.iterchildren()))
```
Nora-Olivia-Ammann, line 230: The only way I managed to make mypy happy was by casting it. Otherwise I got the following error:
```
error: Argument 3 to "TextValueData" has incompatible type "str"; expected "Literal['utf8', 'xml']"  [arg-type]
```
If you know a better way to fix this I would be happy for suggestions.

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
```diff
@@ -75,3 +77,78 @@ def _get_all_properties_from_json(onto_json: list[dict[str, Any]]) -> list[str]:
...
+
+@dataclass
+class PropertyTextValueEncodingTypes:
+    """
+    This class contains the information
+    which properties have which type of encoding for a TextValue property in the ontology.
+    """
+
+    formatted_text: set[str] = field(default_factory=set)
+    unformatted_text: set[str] = field(default_factory=set)
```
jnussbaum, line 100: To make it more clear:

```suggestion
    formatted_text_props: set[str] = field(default_factory=set)
    unformatted_text_props: set[str] = field(default_factory=set)
```

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
```diff
@@ -75,3 +77,78 @@ def _get_all_properties_from_json(onto_json: list[dict[str, Any]]) -> list[str]:
...
+
+    Args:
+        onto_json_dict: dict with the project ontologies
+        default_onto: name of the default ontology
+
+    Returns:
+        Look-up containing the properties separated according to the formatting
+    """
+    all_props = []
+    for onto_json in onto_json_dict.values():
```
jnussbaum, line 119: `onto_json` is a list of dicts. For me this doesn't make sense. A list can never be an ontology. So what does this list contain?

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
```diff
@@ -75,3 +77,78 @@ def _get_all_properties_from_json(onto_json: list[dict[str, Any]]) -> list[str]:
...
+        all_props.extend(_get_all_text_value_properties_and_types_from_json(onto_json))
+    return _make_text_value_property_type_lookup(all_props, default_onto)
+
+
+def _make_text_value_property_type_lookup(
+    prop_list: list[tuple[str, str]], default_onto: str
+) -> PropertyTextValueEncodingTypes:
+    re_pat = r"^" + default_onto + r":" + r".+$"
```
jnussbaum, line 127: Sourcery:

```suggestion
    re_pat = f"^{default_onto}:.+$"
```

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
```diff
@@ -75,3 +77,78 @@ def _get_all_properties_from_json(onto_json: list[dict[str, Any]]) -> list[str]:
...
+
+    def remove_default_prefix(prop_str: str) -> str:
+        if regex.search(re_pat, prop_str) is not None:
```
jnussbaum, line 130: I don't think you need a variable for this, you can use the pattern directly.
Plus, `is not None` is more verbose than necessary:

```suggestion
        if not regex.search(f"^{default_onto}:.+$", prop_str):
```

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
```diff
@@ -75,3 +77,78 @@ def _get_all_properties_from_json(onto_json: list[dict[str, Any]]) -> list[str]:
...
+        all_props.extend(_get_all_text_value_properties_and_types_from_json(onto_json))
+    return _make_text_value_property_type_lookup(all_props, default_onto)
+
+
+def _make_text_value_property_type_lookup(
+    prop_list: list[tuple[str, str]], default_onto: str
+) -> PropertyTextValueEncodingTypes:
+    re_pat = r"^" + default_onto + r":" + r".+$"
+
+    def remove_default_prefix(prop_str: str) -> str:
```
jnussbaum, line 129: I'm not a huge fan of nested functions, unless there is a clear advantage. In the present case, I don't see a good reason to nest it.

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
```diff
@@ -75,3 +77,78 @@ def _get_all_properties_from_json(onto_json: list[dict[str, Any]]) -> list[str]:
...
+            return prop_str.replace(default_onto, "", 1)
+        return prop_str
+
+    formatted_text = {remove_default_prefix(p) for p, _type in prop_list if _type == "salsah-gui:Richtext"}
+    formatted_text.add("hasComment")  # this is a knora-api property that can be used directly
+    unformatted_text = {remove_default_prefix(p) for p, _type in prop_list if _type != "salsah-gui:Richtext"}
+    return PropertyTextValueEncodingTypes(formatted_text, unformatted_text)
+
+
+def _get_all_text_value_properties_and_types_from_json(onto_json: list[dict[str, Any]]) -> list[tuple[str, str]]:
```
jnussbaum, line 140: again, I don't like the "json" in the function name.

jnussbaum: And the same for `onto_json`.

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
```diff
@@ -75,3 +77,78 @@ def _get_all_properties_from_json(onto_json: list[dict[str, Any]]) -> list[str]:
...
+    def check_correct_val_type(onto_ele: dict[str, Any]) -> bool:
```
jnussbaum, line 141: ...and the same for the nested function

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
```diff
@@ -75,3 +77,78 @@ def _get_all_properties_from_json(onto_json: list[dict[str, Any]]) -> list[str]:
...
+            return False
+        match object_type["@id"]:
+            case "knora-api:TextValue":
+                return True
+            case _:
+                return False
+
+    prop_id_list = [elem["@id"] for elem in onto_json if check_correct_val_type(elem)]
+    type_list = [elem["salsah-gui:guiElement"]["@id"] for elem in onto_json if check_correct_val_type(elem)]
+    return [(p_id, _type) for p_id, _type in zip(prop_id_list, type_list)]
```
jnussbaum, line 154:
```suggestion
    return list(zip(prop_id_list, type_list))
```

src/dsp_tools/commands/xmlupload/models/ontology_problem_models.py
```diff
@@ -113,3 +115,54 @@ def _format_prop(prop_tup: tuple[str, list[str], str]) -> str:
...
         else:
             return None
+
+
+@dataclass
+class InvalidTextValueEncodings:
+    """
+    This class contains information about resources and the respective properties that have invalid text encodings.
+
+    An invalid encoding would be a property that specifies an xml encoding in the ontology,
```
jnussbaum, line 125: I know what you mean, but it's technically not correct: The onto cannot specify an xml encoding, but only the salsah-gui attributes "Richtext", "Textarea", "SimpleText"

src/dsp_tools/commands/xmlupload/models/ontology_problem_models.py
```diff
@@ -113,3 +115,54 @@ def _format_prop(prop_tup: tuple[str, list[str], str]) -> str:
...
+    def execute_problem_protocol(self) -> tuple[str, pd.DataFrame | None]:
+        """
+        This method composes an error message for the user.
+
+        Returns:
+            the error message and a dataframe with the errors if they exceed the maximum allowed print statements
+        """
+        msg = (
+            "\nSome text encodings used in the data is not conform with the gui-element specified in the ontology.\n"
+            "Please consult the ontology regarding the assigned gui-elements."
```
jnussbaum, line 140:
```suggestion
            "\nSome text encodings used in the XML data are not conform with the gui_element specified in the JSON ontology.\n"
            "Please consult the ontology regarding the assigned gui_elements."
```

src/dsp_tools/commands/xmlupload/models/ontology_problem_models.py
```diff
@@ -113,3 +115,54 @@ def _format_prop(prop_tup: tuple[str, list[str], str]) -> str:
...
+        )
+        df = self._get_problems_as_df()
+        if len(df) > maximum_prints:
+            return msg, df
+        additional_msg = _make_msg_from_df(df)
+        return msg + grand_separator + additional_msg, None
```
jnussbaum, line 146:
```suggestion
        return msg + grand_separator + _make_msg_from_df(df), None
```

src/dsp_tools/commands/xmlupload/models/ontology_problem_models.py
```diff
@@ -113,3 +115,54 @@ def _format_prop(prop_tup: tuple[str, list[str], str]) -> str:
...
+    problematic_resources: list[TextValueData]
+
+    def execute_problem_protocol(self) -> tuple[str, pd.DataFrame | None]:
+        """
+        This method composes an error message for the user.
+
+        Returns:
+            the error message and a dataframe with the errors if they exceed the maximum allowed print statements
+        """
+        msg = (
```
jnussbaum, line 138: renaming suggestion: base_msg

jnussbaum (commented):
Some comments for now, but I didn't look yet at the tests.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -155,3 +162,85 @@ def _get_separate_prefix_and_iri_from_onto_prop_or_cls(
...
+            return _check_correct(text_val.property_name, text_prop_look_up.formatted_text)
+        case "utf8":
+            return _check_correct(text_val.property_name, text_prop_look_up.unformatted_text)
+    return False
+
+
+def _check_correct(text_prop_name: str, allowed_properties: set[str]) -> bool:
+    if text_prop_name in allowed_properties:
+        return True
+    return False
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I intgrated it into the calling function. The only reason for its existence is readability.

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
Nora-Olivia-Ammann: This is the `@graph` section of the ontology serialised into json-ld (we get actually standard, valid json-ld from the api)

```
{
    "rdfs:label": "The knora-api ontology in the complex schema",
    "@graph": [
        { property } ,
        { class } ,
    ]
}
```

This is a list of dicts. Granted removing all the context that comes after the list makes it an invalid json-ld (no declared prefixes). 

```
    "@context": {
        "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
        "knora-api": "http://api.knora.org/ontology/knora-api/v2#",
        "owl": "http://www.w3.org/2002/07/owl#",
        "salsah-gui": "http://api.knora.org/ontology/salsah-gui/v2#",
        "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
        "xsd": "http://www.w3.org/2001/XMLSchema#"
    }
```

This we do because they are not actually relevant to us, because we do not treat it like rdf.
But it very much is the ontology in a rdf graph.

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
Nora-Olivia-Ammann: I can't very well make an f-string for a string that is used in a regex. But the separation of the last two elements is not needed. I will join them.

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
Nora-Olivia-Ammann: I will make it less complicated, but the `not` inverses the logic.

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
```diff
@@ -75,3 +77,78 @@ def _get_all_properties_from_json(onto_json: list[dict[str, Any]]) -> list[str]:
...
+
+    def remove_default_prefix(prop_str: str) -> str:
+        if regex.search(re_pat, prop_str) is not None:
+            return prop_str.replace(default_onto, "", 1)
+        return prop_str
+
+    formatted_text = {remove_default_prefix(p) for p, _type in prop_list if _type == "salsah-gui:Richtext"}
+    formatted_text.add("hasComment")  # this is a knora-api property that can be used directly
+    unformatted_text = {remove_default_prefix(p) for p, _type in prop_list if _type != "salsah-gui:Richtext"}
+    return PropertyTextValueEncodingTypes(formatted_text, unformatted_text)
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I "simplified" is a matter of taste. Yours certainly produces less lines of code. But I find it much harder to read...

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
Nora-Olivia-Ammann: I am fine with the changes in the function name. But the `_json` refers to the type of value we get, so I don't see the problem there.

src/dsp_tools/commands/xmlupload/models/ontology_problem_models.py
Nora-Olivia-Ammann: I know, I quite dislike the way we name this. Using encoding to specify this in the xml is quite weird as xml themselves are usually utf8 encoded.

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -155,3 +162,85 @@ def _get_separate_prefix_and_iri_from_onto_prop_or_cls(
...
+            csv_path = Path(Path.cwd(), csv_file)
+            df.to_csv(path_or_buf=csv_path, index=False)
+            msg += (
+                "\n\n---------------------------------------\n\n"
+                f"All the problems are listed in the file: '{csv_path.absolute()}'"
+            )
+        raise InputError(msg)
+
+
+def _get_all_ids_prop_encoding_from_root(root: etree._Element) -> list[TextValueData]:
```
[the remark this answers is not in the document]
jnussbaum: Ideally, we should strive for both, grammatical and factual correctness. So what do you think about: get all prop ids and encodings?

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -155,3 +162,85 @@ def _get_separate_prefix_and_iri_from_onto_prop_or_cls(
...
+    res_list = []
+    for res_input in root.iterchildren(tag="resource"):
+        res_list.extend(_get_id_prop_encoding_from_one_resource(res_input))
+    return res_list
+
+
+def _get_id_prop_encoding_from_one_resource(resource: etree._Element) -> list[TextValueData]:
```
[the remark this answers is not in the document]
jnussbaum: what about: get prop ids and encodings from one resource?

src/dsp_tools/commands/xmlupload/models/ontology_problem_models.py
jnussbaum: Me too, I dislike the distinction "xml" versus "utf8". Technically, that's an awful distinction.

Now, I think that the docstring is better 👍

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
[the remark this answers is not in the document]
jnussbaum: So do you seriously think that a nested 1-line function is better readable than inlining it? 

Compare this:

```python
def _check_correctness_one_prop(text_val: TextValueData, text_prop_look_up: PropertyTextValueEncodingTypes) -> bool:
    def _check_correct(text_prop_name: str, allowed_properties: set[str]) -> bool:
        return text_prop_name in allowed_properties

    match text_val.encoding:
        case "xml":
            return _check_correct(text_val.property_name, text_prop_look_up.formatted_text_props)
        case "utf8":
            return _check_correct(text_val.property_name, text_prop_look_up.unformatted_text_props)
        case _:
            return False
```

to this:

```python
def _check_correctness_one_prop(text_val: TextValueData, text_prop_look_up: PropertyTextValueEncodingTypes) -> bool:
    match text_val.encoding:
        case "xml":
            return text_val.property_name in text_prop_look_up.formatted_text_props
        case "utf8":
            return text_val.property_name in text_prop_look_up.unformatted_text_props
        case _:
            return False
```

test/unittests/commands/xmlupload/test_check_consistency_with_ontology_low_level.py
```diff
@@ -293,3 +315,312 @@ def test_get_all_classes_and_properties_from_data() -> None:
...
+            <text-prop name=":hasRichtext">
+                <text encoding="xml">Text</text>
+            </text-prop>
+            <text-prop name=":hasSimpleText">
+                <text encoding="utf8">Text</text>
+            </text-prop>
+        </resource>
+        """
+    )
+    res: list[TextValueData] = _get_id_prop_encoding_from_one_resource(test_props)
```
jnussbaum, line 387:
```suggestion
    res = _get_id_prop_encoding_from_one_resource(test_props)
```

test/unittests/commands/xmlupload/test_check_consistency_with_ontology_low_level.py
```diff
@@ -293,3 +315,312 @@ def test_get_all_classes_and_properties_from_data() -> None:
...
+                    <resptr-prop name=":hasResource2">
+                        <resptr permissions="prop-default">resB</resptr>
+                    </resptr-prop>
+                </resource>
+        </knora>"""
+    )
+    test_lookup = PropertyTextValueEncodingTypes(
+        formatted_text_props={":hasRichtext"}, unformatted_text_props={":hasSimpleText", ":hasText"}
+    )
+    _check_correctness_all_text_value_encodings(test_ele, test_lookup)
```
jnussbaum, line 573: Here you forgot to check the result. This test will always pass, even with wrong input data!

test/unittests/commands/xmlupload/test_ontology_lookup_models.py
```diff
@@ -166,5 +170,124 @@ def test_remove_prefixes_knora_properties() -> None:
...
+        "knora-api:objectType": {"@id": "knora-api:TextValue"},
+        "salsah-gui:guiElement": {"@id": "salsah-gui:Textarea"},
+        "@id": "onto:hasTextarea",
+    }
+    res = _check_correct_val_type(test_dict)
+    assert res
+
+
+def test_make_text_value_property_gui() -> None:
+    test_li = [
```
jnussbaum, line 270: it took me a moment to understand what "li" is. The only abbreviations of list that are known to me are "lst" or "list" (unabbreviated)

jnussbaum (commented):
Here is the rest of my review. Among all the comments, there is only 1 thing that must absolutely be fixed before merging: A test that always passes, without testing anything. (I commented it). The rest of my comments are just cosmetics / opinions.

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
```diff
@@ -50,28 +52,109 @@ def _extract_classes_and_properties_from_onto(onto_json: list[dict[str, Any]]) -
...
+    """
+    This class contains the information
+    which properties have which type of encoding for a TextValue property in the ontology.
+    """
+
+    formatted_text_props: set[str] = field(default_factory=set)
+    unformatted_text_props: set[str] = field(default_factory=set)
+
+
+def get_text_value_properties_and_formatting_from_onto(
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I don't think using encoding is the better choice. Because the type in fact has nothing to do with the encoding, it is just how we called it in the XML, and in the ontology it is referred to differently. But I changed it so that it is more conform with the naming in the ontology, as this deals with that and not the XML.

test/unittests/commands/xmlupload/test_check_consistency_with_ontology_low_level.py
Nora-Olivia-Ammann: Thanks for spotting it.

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
Nora-Olivia-Ammann: Did not know that was possible.

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: This replaces the whole string with `""` if the substring is found.

Just replacing it without running the regex is also not an option because:

`other_onto:hasRichtext` would become `other_:hasRichtext` if `onto` is the default.

test/unittests/commands/xmlupload/test_ontology_problem_models.py
jnussbaum: 👍

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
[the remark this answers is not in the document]
jnussbaum: You're right, I'm often not attentive enough. But I'm 100% sure it can and should be turned into a 1-liner, because the `regex` library is capable of doing such things in 1 shot. Another try from my side:

```
return regex.sub(fr"^{default_onto}:", ":", prop_str)
```

You could even make a capturing group, and then insert its content in the replaced string:

```
regex.sub(fr"^{default_onto}:(.+)$", r":\1", prop_str)
```

src/dsp_tools/commands/xmlupload/models/ontology_lookup_models.py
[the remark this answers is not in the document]
jnussbaum: 👍

jnussbaum (approved):
LGTM

PR #828, opened 2024-02-23
chore(test): turn xml_utils.py into pytest

Nora-Olivia-Ammann:


test/unittests/utils/test_xml_utils.py
```diff
@@ -1,56 +1,67 @@
...
+import pytest
 import regex
 from lxml import etree
 
 from dsp_tools.utils.xml_utils import parse_and_clean_xml_file
 
-# ruff: noqa: PT009 (pytest-unittest-assertion) (remove this line when pytest is used instead of unittest)
+
+@pytest.fixture()
+def data_systematic_unclean() -> Any:
```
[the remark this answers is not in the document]
jnussbaum: I think this can be fixed with `from __future__ import annotations`

test/unittests/utils/test_xml_utils.py
```diff
@@ -1,56 +1,67 @@
...
-        self.assertEqual(
-            len(comments),
-            0,
-            msg="properties that are commented out would break the the constructor of the class XMLProperty, "
-            "if they are not removed in the parsing process",
-        )
+def test_parse_and_clean_xml_file_same_regardless_of_input(
+    data_systematic_unclean: Any, data_systematic_cleaned: etree._Element
+) -> None:
+    from_tree = parse_and_clean_xml_file(data_systematic_unclean)
```
jnussbaum, line 28: Now, you mix 2 things:
 - cleaned versus uncleaned
 - from tree versus from file

For me it is now very unclear what exactly is tested, and what is not tested

Nora-Olivia-Ammann: I did not change the content of the test, the difference was that I used the fixture. But I understand that when we want to test the parser function it is clearer if it is actually called, even if it is redundant. So I changed that.

jnussbaum: Yes, now it's clear. This might be the well-known exception from the DRY-rule :-)

test/unittests/utils/test_xml_utils.py
```diff
@@ -1,56 +1,68 @@
...
+    assert (
+        cleaned_from_file == cleaned_from_tree
+    ), "The output must be equal, regardless if the input is a path or parsed."
+
+
+def test_annotations_regions_links_before(data_systematic_unclean: etree._ElementTree[etree._Element]) -> None:
+    annotations_regions_links_before = [
+        e for e in data_systematic_unclean.iter() if regex.search("annotation|region|link", str(e.tag))
+    ]
+    assert len(annotations_regions_links_before) == 5
```
jnussbaum, line 41: Good decision to compare against the real number 👍

jnussbaum (approved):
LGTM

PR #832, opened 2024-02-26
refactor(connection): call specific error response codes (DEV-3339)

jnussbaum:


src/dsp_tools/utils/connection_live.py
```diff
@@ -271,7 +271,7 @@ def _try_network_action(self, params: RequestParameters) -> Response:
                 return response
             elif "v2/authentication" in params.url and response.status_code == HTTP_UNAUTHORIZED:
                 raise BadCredentialsError("Bad credentials")
-            elif "OntologyConstraintException" in response.text:
+            elif any(x in response.text for x in ["OntologyConstraintException", "DuplicateValueException"]):
```
BalduinLandolt, line 274: I'm personally not a big fan of these kinds of logical statements, they simply don't look very readable to me. On the other side, I fully acknowledge that it's exactly the logic we need. How would you feel about extracting `any(x in response.text for x in ["OntologyConstraintException", "DuplicateValueException"])` into a little function `_isNotRetryable()` or `_isNoRetryException()` or `_noRetryNeeded()` or something like that, so on the caller side we would have something more speaking, and the implementation detail is hidden away?

Nora-Olivia-Ammann (commented):
What I am most unhappy with, is that the error handling that lives in the connection object is too specific for such a generic class. 
In my opinion what could work is that we do 
- a network action without
- the first types of responses are generic as they can happen with any input, for these responses we have the loop automatically
- the second types of error handling are specific depending on when the function is used (xmlupload, create, etc.) 
-  There we work out the problem
- communicate that
- if it is an error that we want to retry, we enter a function that executes the loop (for simplicity's sake no matter what type of response it gets, it will retry the set amount of times.
- if we don't want to retry we continue

Do you think something like that is feasible?

src/dsp_tools/utils/connection_live.py
```diff
@@ -11,14 +11,15 @@
...
-from dsp_tools.models.exceptions import BadCredentialsError, BaseError, PermanentConnectionError, UserError
+from dsp_tools.models.exceptions import BadCredentialsError, BaseError, InputError, PermanentConnectionError, UserError
 from dsp_tools.utils.create_logger import get_logger
 from dsp_tools.utils.set_encoder import SetEncoder
 
 HTTP_OK = 200
 HTTP_UNAUTHORIZED = 401
 
 logger = get_logger(__name__)
+LOGFILES = ", ".join([handler.baseFilename for handler in logger.handlers if isinstance(handler, FileHandler)])
```
BalduinLandolt, line 22: IMO this should be a function somewhere in a util, preferably in a `log_util` together with `get_logger()`. It is very low level and has nothing to do with connections

BalduinLandolt (approved):
LGTM

Nora-Olivia-Ammann:
@jnussbaum I was not able to request a review from you, probably because you opened the PR, but I would be glad for your input.

src/dsp_tools/utils/connection_live.py
Nora-Olivia-Ammann: I agree that it is not nice. so you would do something like this?

```
from log_util import get_log_info

logger, logfiles = get_log_info()
```

BalduinLandolt: I'd rather keep them separate:

```python
from log_util import get_logger, get_log_files

logger = get_logger(__name__)
logfiles = get_log_files()
```

but by all means, this should not block this PR

jnussbaum:
> @jnussbaum I was not able to request a review from you, probably because you opened the PR, but I would be glad for your input.

@nora: LGTM

PR #835, opened 2024-02-27
chore: remove docker as dependency

jnussbaum:


Nora-Olivia-Ammann (approved):
Curious: Why was this here / or is not used anymore?

jnussbaum:
> Curious: Why was this here / or is not used anymore?

It's funny: Yesterday, when I removed the fast-xmlupload stuff, you asked me if I could also remove some of the python dependencies, and I answered no. Today, by accident I stumbled upon this, and it was only used by the fast-xmlupload :-)

PR #839, opened 2024-02-29
chore: Update poetry.lock

Nora-Olivia-Ammann:


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.
```
jnussbaum, line 1: could you `poetry self update`, and then do it again?

PR #843, opened 2024-02-29
chore(test): make integration test ingest-xmlupload

Nora-Olivia-Ammann:


test/integration/commands/ingest_xmlupload/test_ingest_xmlupload.py
```diff
@@ -18,15 +20,12 @@ def _retrieve_mapping_file() -> Iterator[None]:
...
-    expected_msg = (
-        "The upload cannot continue as there are problems with the multimedia files referenced in the XML.\n"
-        "    The data XML file does not reference the following multimedia files "
-        "which were previously uploaded through dsp-ingest:\n"
-        "    - extra.jpg\n"
-        "    The data XML file contains references to the following multimedia files "
-        "which were not previously uploaded through dsp-ingest:\n"
-        "    - Resource ID: 'GoodGirlImage' | Filepath: 'images/GoodGirl.jpg'"
-    )
+    expected_msg = """The upload cannot continue as there are problems with the multimedia files referenced in the XML.
```
Nora-Olivia-Ammann, line 23: For some reason, the message was not checked. Meaning that I deleted some letters but the test still passed. With the multiline strings the test now also fails if the message was changed.

jnussbaum: I fixed it with `regex.escape()`

PR #848, opened 2024-03-01
chore(test): escape all regex string properly

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/models/input_error.py
```diff
@@ -150,7 +150,6 @@ def execute_error_protocol(self) -> str:
             message for the error
         """
         msg = (
-            "The excel file 'resources.xlsx' has problems.\n"
             "The names of the excel sheets must be 'classes' "
```
Nora-Olivia-Ammann, line 153: We repeated this line when the error was raised, meaning it appeared double.

PR #853, opened 2024-03-04
feat: resume an interrupted xmlupload (DEV-3323)

jnussbaum:


docs/cli-commands.md
```diff
@@ -164,6 +164,33 @@ dsp-tools xmlupload -s https://api.dasch.swiss -u 'your@email.com' -p 'password'
...
+
+The following options are available:
+
+- `-s` | `--server` (optional, default: `0.0.0.0:3333`): URL of the DSP server where DSP-TOOLS sends the data to
+- `-u` | `--user` (optional, default: `root@example.com`): username (e-mail) used for authentication with the DSP-API 
+- `-p` | `--password` (optional, default: `test`): password used for authentication with the DSP-API
+
+For this command to work,
+the pickle file `~/.dsp-tools/xmluploads/[server]/resumable/latest.pkl` must exist. 
+Currently, only one interrupted upload can be resumed at a time.
```
Nora-Olivia-Ammann, line 193: **comment**: Maybe add what happens if the same xmlupload is interupted and then resumed multiple times.

And based on the file name, should it not be possible to be able to resume one upload per server?

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
```diff
@@ -37,7 +37,7 @@ def upload_stashed_resptr_props(
     print(f"{datetime.now()}: Upload the stashed resptrs...")
     logger.info("Upload the stashed resptrs...")
     not_uploaded: list[LinkValueStashItem] = []
-    for res_id, stash_items in stashed_resptr_props.res_2_stash_items.items():
+    for res_id, stash_items in stashed_resptr_props.res_2_stash_items.copy().items():
```
Nora-Olivia-Ammann, line 40: **question**: 
Would it not also be necessary to add this exception handling here if the upload were interrupted here?

Nora-Olivia-Ammann (commented):
In the file: `src/dsp_tools/commands/xmlupload/resource_create_client.py`
line: 48.
What happens if the post is already done but we do not have the answer from the server yet but it is a sucessful upload?

If there is no interrupted upload to finish i get a non-200 error, which it retries... that is for a future PR but we should keep in mind to clean that up.

**Bad news**: The number of triples in the db is not the same after an interrupted upload than a normal one:
- Regular xmlupload project systematic: 3735
- Test 1: Interrupted 2 times: 3763
- Test 2: interrupted 1 times: 3766
Both interruptions were not during the stash upload.
I have not done more precise investigations what happened, but clearly something is wrong.

jnussbaum:
> In the file: `src/dsp_tools/commands/xmlupload/resource_create_client.py` line: 48. What happens if the post is already done but we do not have the answer from the server yet but it is a sucessful upload?
> 
> If there is no interrupted upload to finish i get a non-200 error, which it retries... that is for a future PR but we should keep in mind to clean that up.
> 
> **Bad news**: The number of triples in the db is not the same after an interrupted upload than a normal one:
> 
>     * Regular xmlupload project systematic: 3735
> 
>     * Test 1: Interrupted 2 times: 3763
> 
>     * Test 2: interrupted 1 times: 3766
>       Both interruptions were not during the stash upload.
>       I have not done more precise investigations what happened, but clearly something is wrong.

Thanks a lot for your valuable feedback. We noted these points and created a follow-up ticket: https://linear.app/dasch/issue/DEV-3392/tidy-up-resume-xmlupload-follow-up

BalduinLandolt (approved):

PR #855, opened 2024-03-04
fix(xmlupload): don't crash if the project has more than 50 errors (DEV-3364)

Nora-Olivia-Ammann:


jnussbaum (approved):
LGTM. Regarding the PR title: Perhaps it's just me, but I always try to choose a title that reads well in a changelog. So I imagine I scroll through another project's changelog. Then I don't want to see "program crashes if this happens". Rather, I'd like to read "don't crash if this happens".

In this sense, how do you feel about "fix(xmlupload): don't crash if the project has more than 50 errors (DEV-3364)" ?

PR #859, opened 2024-03-05
docs(excel2xml): code examples in docstrings: prepend function names with their namespace

jnussbaum:


Notheturtle (approved):
Vielen lieben Dank !

PR #862, opened 2024-03-07
feat: Add `interrupt-after` to `xmlupload` and `resume-xmlupload` (DEV-3390)

BalduinLandolt:


src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -397,6 +398,9 @@ def _upload_resources(
     )
 
     total_res = len(resources)
+    # if the interrupt_after value is not set, the upload will not be interrupted
+    interrupt_after = config.interrupt_after or total_res + 1
+
```
jnussbaum, line 403: i think these lines should be at the very beginning of the for loop, i.e. on line 405

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -407,6 +411,8 @@ def _upload_resources(
 
         res = None
         try:
+            if i >= interrupt_after:
+                raise KeyboardInterrupt(f"Interrupted: Maximum number of resources was reached ({interrupt_after})")
```
jnussbaum, line 415: EDIT: i think these lines should be at the very beginning of the for loop, i.e. on line 405

PR #863, opened 2024-03-08
fix(xmlupload): exit with code 0 if xmlupload finishes with batch (DEV-3396)

jnussbaum:


src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -497,34 +498,39 @@ def _handle_upload_error(
...
     """
-    logfiles = ", ".join([handler.baseFilename for handler in logger.handlers if isinstance(handler, FileHandler)])
-    msg = (
-        f"\n==========================================\n"
-        f"{datetime.now()}: xmlupload must be aborted because of an error.\n"
-        f"Error message: '{err_msg}'\n"
-        f"For more information, see the log file: {logfiles}\n"
-    )
+    if isinstance(err, XmlUploadInterruptedError):
+        msg = "\n==========================================\n" + err.message + "\n"
```
Nora-Olivia-Ammann, line 510: I would not consider this an error, because the interruption is wanted. Therefore this message may not be optimal

Nora-Olivia-Ammann (approved):

src/dsp_tools/commands/xmlupload/xmlupload.py
jnussbaum: You mean the variable names are not optimal? I mean, because the message that is displayed to the user doesn't call it an error, it looks like this: 

```
==========================================
Interrupted: Maximum number of resources was reached (15)
Saved the current upload state to /Users/nussbaum/.dsp-tools/xmluploads/localhost/resumable/latest.pkl.
```

jnussbaum: I could consider renaming `XmlUploadInterruptedError` to `XmlUploadInterruptedException`, then it doesn't have the connotation of error.

jnussbaum: `xmlupload must be aborted because of an error` is not printed in case of an `XmlUploadInterruptedError`

PR #865, opened 2024-03-08
fix(resume-xmlupload): make it impossible to resume after all resources have been created (DEV-3398)

jnussbaum:


Nora-Olivia-Ammann (approved):
Wehre is the pkl file deleted?

jnussbaum:
> Wehre is the pkl file deleted?

on this line: `config.diagnostics.save_location.unlink(missing_ok=True)`

PR #880, opened 2024-03-13
fix(xmlupload): don't stop xmlupload when resource is wrong (DEV-3419)

Nora-Olivia-Ammann:


src/dsp_tools/utils/connection_live.py
```diff
@@ -314,7 +318,7 @@ def _log_and_raise_timeouts(self, error: TimeoutError | ReadTimeout) -> None:
         msg = f"A '{error.__class__.__name__}' occurred during the connection to the DSP server."
         print(f"{datetime.now()}: {msg}")
         logger.exception(msg)
-        raise PermanentConnectionError(msg) from None
```
Nora-Olivia-Ammann, line 317: Because the function `handle_non_ok_responses` also raises the `PermanentConnectionError` now the upload stopped with the new message of the resume upload even if for example a cardinality of a resource is wrong. This re-established the old way of dealing with those errors.

src/dsp_tools/utils/connection_live.py
```diff
@@ -294,7 +295,10 @@ def _handle_non_ok_responses(self, response: Response, request_url: str, retry_c
             self._log_and_sleep("Transient Error", retry_counter, exc_info=False)
             return None
         else:
-            msg = f"Permanently unable to execute the network action. See logs for more details: {LOGFILES}"
+            msg = "Permanently unable to execute the network action. "
+            if original_str := regex.search(r'{"knora-api:error":"dsp\.errors\.(.*)","@context', str(response.content)):
+                msg += f"\n{' '*37}Original Message: {original_str.group(1)}\n"
```
Nora-Olivia-Ammann, line 300: Noemi came to me with this bug. She told me that it would be helpful to her if this message is not only found in the logs.

BalduinLandolt (commented):
LGTM

**nit-pick:** could you change the PR title?

```diff
- make so that xmlupload doesn't stop when resource is wrong
+ don't stop xmlupload when resource is wrong
```

I find `make so that XYZ doesn't XYZ` a bit clunky

BalduinLandolt (approved):
(meant to click "approve" the first time ^^)

PR #883, opened 2024-03-14
feat: save warnings and errors in second file with loguru (DEV-3406)

Nora-Olivia-Ammann:


src/dsp_tools/utils/logger_config.py
```diff
@@ -0,0 +1,43 @@
...
+    - warnings.log in the cwd only with level warning and higher for the user (no stack-trace)
+    - print output on the terminal, formatted the same as the warnings.log
+
+    Returns:
+        None
+    """
+    # If this is not removed, the default formatting is also printed out on the terminal
+    logger.remove()
+
+    text_format = "{time:YYYY-MM:DD HH:mm:ss.SSSS} | <level>{level} | {message}</level>"
```
jnussbaum, line 21:
```suggestion
    text_format = "{time:YYYY-MM-DD HH:mm:ss.SSS} | <level>{level} | {message}</level>"
```

jnussbaum: **thought**: In the old logging, I found it useful to have the file name and line number, to know where the message was emitted.

src/dsp_tools/resources/logger_config.yml
```diff
@@ -0,0 +1,19 @@
+handlers:
+  - sink: ext://sys.stdout
+    format: "{time:YYYY-MM:DD HH:mm:ss.SSSS} | <level>{level} | {message}</level>"
```
jnussbaum, line 3: And for the others, accordingly:

```suggestion
    format: "{time:YYYY-MM-DD HH:mm:ss.SSS} | <level>{level} | {message}</level>"
```

src/dsp_tools/utils/logger_config.py
Nora-Olivia-Ammann: The line and filename are included in the logs

src/dsp_tools/commands/project/create/project_create.py
```diff
@@ -230,13 +228,13 @@ def _get_group_iris_for_user(
...
                 print(f"    WARNING: {err_msg}")
-                logger.warning(err_msg, exc_info=True)
+                logger.warning(err_msg)
                 success = False
                 continue
             existing_group = [g for g in remote_groups if g.project == current_project.iri and g.name == group_name]
             if not existing_group:
                 print(f"    WARNING: {inexisting_group_msg}")
-                logger.warning(inexisting_group_msg, exc_info=True)
+                logger.warning(inexisting_group_msg)
```
Nora-Olivia-Ammann, line 237: the exc_info should not be necessary anymore
https://github.com/Delgan/loguru/issues/30

pyproject.toml
```diff
@@ -17,6 +17,7 @@ homepage = "https://www.dasch.swiss/"
 repository = "https://github.com/dasch-swiss/dsp-tools"
 include = [
     "src/dsp_tools/resources/*",
+    "src/dsp_tools/resources/logger_config.yml",
```
BalduinLandolt: this should no longer be needed, right?

src/dsp_tools/cli/entry_point.py
```diff
@@ -284,4 +283,5 @@ def _derive_sipi_url(
 
 
 if __name__ == "__main__":
+    logger_config()
```
BalduinLandolt: ... as discussed :)

BalduinLandolt (approved):

src/dsp_tools/cli/entry_point.py
```diff
@@ -67,12 +67,12 @@ def run(args: list[str]) -> None:
         )
         success = call_requested_action(parsed_arguments)
     except BaseError as err:
-        logger.exception("The process was terminated because of an Error:")
+        logger.error("The process was terminated because of an Error:")
```
jnussbaum, line 70: why don't you log the exception information any more? In loguru, it's the same as in the standard library:
 - logger.error(): without stack trace
 - logger.exception(): with stack trace 

cf. https://loguru.readthedocs.io/en/stable/api/logger.html#loguru._logger.Logger.exception

src/dsp_tools/cli/entry_point.py
```diff
@@ -67,12 +67,12 @@ def run(args: list[str]) -> None:
...
         print("\nThe process was terminated because of an Error:")
         print(err.message)
         sys.exit(1)
-    except Exception as err:
-        logger.exception(err)
+    except Exception as err:  # noqa: BLE001 (blind-except)
+        logger.error(err)
```
jnussbaum, line 75: why don't you log the exception information any more?

src/dsp_tools/commands/project/create/project_create.py
jnussbaum: That's not how I understand it. Rather, we should replace `logger.xyz(msg, exc_info=True)` with `logger.opt(exception=True).xyz(msg)`. See https://loguru.readthedocs.io/en/stable/resources/migration.html#replacing-exc-info-argument

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -38,11 +39,10 @@
 from dsp_tools.models.projectContext import ProjectContext
 from dsp_tools.utils.connection import Connection
 from dsp_tools.utils.connection_live import ConnectionLive
-from dsp_tools.utils.create_logger import get_log_filename_str
-from dsp_tools.utils.create_logger import get_logger
 from dsp_tools.utils.json_ld_util import get_json_ld_context_for_project
+from dsp_tools.utils.logger_config import logger_savepath
 
-logger = get_logger(__name__)
+LOGFILES = str(logger_savepath)
```
jnussbaum, line 45: This conversion is not necessary. You can use the Path directly.
POC:

```
> from pathlib import Path
> print(f"Path is {Path('foo/bar')}")
Path is foo/bar
```

src/dsp_tools/utils/connection_live.py
```diff
@@ -22,15 +23,12 @@
...
-from dsp_tools.utils.create_logger import get_logger
+from dsp_tools.utils.logger_config import logger_savepath
 from dsp_tools.utils.set_encoder import SetEncoder
 
 HTTP_OK = 200
 HTTP_UNAUTHORIZED = 401
-
-logger = get_logger(__name__)
-LOGFILES = get_log_filename_str(logger)
+LOGFILES = str(logger_savepath)
```
jnussbaum, line 31: same as above

src/dsp_tools/utils/logger_config.py
```diff
@@ -0,0 +1,45 @@
...
+    This function configures the log files.
+    Currently, there are three sinks:
+    - logger.log in the home directory for development purposes
+    - warnings.log in the cwd only with level warning and higher for the user (no stack-trace)
+    - print output on the terminal, formatted the same as the warnings.log
+    """
+    # If this is not removed, the default formatting is also printed out on the terminal
+    logger.remove()
+
+    text_format = "<level>{time:YYYY-MM:DD HH:mm:ss.SSS} | {level: <8} | {message}</level>"
```
jnussbaum, line 21: Sorry that I bring this up a second time, but I didn't see a reaction on my first comment. Are you sure you want the format `YYYY-MM:DD`? For me, `YYYY-MM-DD` would be more intuitive.

src/dsp_tools/utils/logger_config.py
```diff
@@ -0,0 +1,45 @@
...
+    rotation_size = "500 MB"
+    retention_number = 30
```
jnussbaum, line 23: This would retain 15 GB of logs, which seems too much for me. I guess that 5 x 0.5 GB = 2.5 GB would be enough.

src/dsp_tools/utils/logger_config.py
```diff
@@ -0,0 +1,45 @@
...
+    timestamp_str = datetime.now().strftime("%Y%m%d-%H%M")
+
+    logger.add(sink=sys.stdout, format=text_format, level="INFO", backtrace=False, diagnose=False)
```
jnussbaum, line 26: Cool. But shouldn't we now remove all the print statements which are just a duplication of what is already logged?

jnussbaum (changes requested):
I'm happy that we switch to loguru, thanks for digging into the subject 👍 

Regarding the concrete implementation, I have 2 main concerns:
 - I think we shouldn't remove all these `exc_info=True`. We need those stack traces. https://loguru.readthedocs.io/en/stable/resources/migration.html#replacing-exc-info-argument describes how the stack trace info can be logged in loguru.
 - I think we should now remove all the print statements which are just a duplication of the log statement, e.g. https://github.com/dasch-swiss/dsp-tools/blob/83b3ddc194ff54be0b55cad4d7efed4ae5d79291/src/dsp_tools/commands/xmlupload/xmlupload.py#L462

.gitignore
```diff
@@ -63,3 +63,4 @@ testdata/preprocessed_files/
 testdata/bitstreams/nested
 processing_result_*.pkl
 processed_files.txt
+/warnings-*.log
```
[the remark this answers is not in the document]
BalduinLandolt: no, in .gitignore, leading slash binds the pattern to the folder where the .gitignore file is located (e.g. https://sebastiandedeyne.com/leading-slashes-in-gitignore/)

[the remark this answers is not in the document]
jnussbaum: That's what I love about my job: I can learn something new every day. Thanks for the correction :-)

src/dsp_tools/utils/logger_config.py
Nora-Olivia-Ammann: Hey, I did not notice that comment then. Thanks for spotting this, this is wrong.

src/dsp_tools/utils/logger_config.py
Nora-Olivia-Ammann: I looked at the current logfiles. They are 104 MB, so I changed the number now to 100 and kept the retention number. They are big enough as it is...

src/dsp_tools/utils/logger_config.py
Nora-Olivia-Ammann: True, we should decide either way. I spoke about it with balduin and he argues for not having any log statements printed out. I understand the argument, therefore in fact, we could remove this line and just keep the new logile

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -525,12 +521,11 @@ def _handle_upload_error(
         msg = "\n==========================================\n" + err.message + "\n"
         exit_code = 0
     else:
-        logfiles = get_log_filename_str(logger)
         msg = (
             f"\n==========================================\n"
             f"{datetime.now()}: xmlupload must be aborted because of an error.\n"
             f"Error message: '{err}'\n"
-            f"For more information, see the log file: {logfiles}\n"
+            f"For more information, see the log file: {LOGFILES}\n"
```
jnussbaum, line 528:
```suggestion
            f"For more information, see the log file: {logger_savepath}\n"
```

src/dsp_tools/utils/connection_live.py
```diff
@@ -22,15 +23,12 @@
...
-from dsp_tools.utils.create_logger import get_logger
+from dsp_tools.utils.logger_config import logger_savepath
 from dsp_tools.utils.set_encoder import SetEncoder
 
 HTTP_OK = 200
 HTTP_UNAUTHORIZED = 401
-
-logger = get_logger(__name__)
-LOGFILES = get_log_filename_str(logger)
+LOGFILES = logger_savepath
```
jnussbaum, line 31:
```suggestion
```

src/dsp_tools/utils/logger_config.py
BalduinLandolt: To me, there is a fundamental difference between logging and user feedback: One is for a good user experience, the other is for easy diagnostics and tracing back what happened in exceptional cases. And for that reason I would like to keep them separate both conceptually and in code.  
But that is something we already have on our notion page, and that definitely should not block this PR; and if you like/feel the need, we can also discuss it again in our next meeting.

For now, I'm ok with either logging these messages, or printing them (the way it currently is set up, there is no conceptual difference anyways), but if they are being logged to stdout, they of course should not be printed, otherwise they show up twice.

Nora-Olivia-Ammann: I think I will remove the sink here, because it is easier to add this than to re-introduce the print statements and then we can discuss this later.

jnussbaum: That sounds reasonable, then we can still discuss later how we want to proceed in the future

PR #888, opened 2024-03-22
chore: remove timestamp from warnings.log

Nora-Olivia-Ammann:


src/dsp_tools/utils/logger_config.py
```diff
@@ -32,7 +30,7 @@ def logger_config() -> None:
     )
 
     logger.add(
-        sink=Path(f"warnings-{timestamp_str}.log"),
+        sink=Path("warnings.log"),
```
Nora-Olivia-Ammann, line 33: I find the number of logs extremely annoying and not helpful. The timestamp can be seen in the logs.

jnussbaum: Question: With this new configuration, would the old file be overwritten, or would the new entries be appended to the old ones? I find both ones problematic, but would prefer overwrite over appending.

Nora-Olivia-Ammann: This should work like a normal logger. The thought was to use the timestamps to differentiate the different times it was called, but I find it extremely annoying.

jnussbaum (approved):

PR #891, opened 2024-03-27
chore: bump start-stack to 2024.03.02

jnussbaum:


src/dsp_tools/resources/start-stack/docker-compose.yml
```diff
@@ -4,31 +4,37 @@ version: "3.7"
 
 services:
 
-  app:
```
jnussbaum, line 7: I changed the order, so that it corresponds to https://github.com/dasch-swiss/ops-deploy/blob/main/roles/dsp-deploy/files/RELEASE.json. So, in the future, the copy-pasting should become easier.

Nora-Olivia-Ammann (approved):

PR #895, opened 2024-04-02
chore(mypy): disallow untyped defs

jnussbaum:
By default the bodies of functions without annotations are not type checked. So, to be more strict with typing, it would be better to disallow functions that are not annotated, see https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-untyped-defs.


Nora-Olivia-Ammann:
This is already checked...?
Screenshot 2024-04-03 at 09 09 11

PR #896, opened 2024-04-02
chore: add type annotations for pandas and networkx

jnussbaum:


src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -166,7 +167,7 @@ def _add_optional_columns(df: pd.DataFrame) -> pd.DataFrame:
     in_df_cols = set(df.columns)
     if not optional_col_set.issubset(in_df_cols):
         additional_col = list(optional_col_set.difference(in_df_cols))
-        additional_df = pd.DataFrame(columns=additional_col, index=df.index, data=pd.NA)
+        additional_df = pd.DataFrame(columns=additional_col, index=df.index, data=None)
```
Nora-Olivia-Ammann, line 170: What is the advantage here?

src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -181,13 +182,13 @@ def _check_missing_values_in_row(df: pd.DataFrame) -> None | list[MissingValuesI
     if missing_gui_attributes is not None:
         missing_dict.update(missing_gui_attributes)
     if missing_dict:
-        missing_dict = get_wrong_row_numbers(wrong_row_dict=missing_dict, true_remains=True)
-        return [MissingValuesInRowProblem(column=col, row_numbers=row_nums) for col, row_nums in missing_dict.items()]
+        missing_dict_int = get_wrong_row_numbers(wrong_row_dict=missing_dict, true_remains=True)
```
Nora-Olivia-Ammann, line 185: If we are renaming perhaps `missing_int_dict` is clearer. I find it easiest if the data type is at the end...

src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -208,15 +209,17 @@ def _check_compliance_gui_attributes(df: pd.DataFrame) -> dict[str, pd.Series] |
...
-            final_series = pd.Series(np.logical_or(mandatory_check, no_attribute_check))  # type: ignore[arg-type]
+            mandatory_check = cast("pd.Series[bool]", mandatory_check)
+            no_attribute_check = cast("pd.Series[bool]", no_attribute_check)
+            final_series: pd.Series[bool] = pd.Series(np.logical_or(mandatory_check, no_attribute_check))
         case pd.Series(), None:
-            final_series = mandatory_check
+            final_series = cast("pd.Series[bool]", mandatory_check)
         case None, pd.Series:
-            final_series = no_attribute_check
+            final_series = cast("pd.Series[bool]", no_attribute_check)
```
Nora-Olivia-Ammann, line 218: I understand in theory why more typing is good, but this makes the code extremely verbose and in my opinion harder to understand, I appreciate this form of extreme typing if the datatype is very unclear but in this case there is no other datatype possible for `no.logical_or` therefore that is not an advantage. So all in all, in my opinion in this case, this does not improve the code.

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -270,7 +272,7 @@ def _validate_excel_file(classes_df: pd.DataFrame, df_dict: dict[str, pd.DataFra
             f"Please use {[f'label_{lang}' for lang in languages]}"
         )
     problems: list[Problem] = []
-    if missing_super_rows := [int(index) + 2 for index, row in classes_df.iterrows() if not check_notna(row["super"])]:
+    if missing_super_rows := [int(str(i)) + 2 for i, row in classes_df.iterrows() if not check_notna(row["super"])]:
```
Nora-Olivia-Ammann, line 275: I don't understand why it has to become a string first?

src/dsp_tools/commands/excel2json/utils.py
```diff
@@ -168,13 +171,13 @@ def get_wrong_row_numbers(wrong_row_dict: dict[str, pd.Series], true_remains: bo
     Returns:
         Dictionary with the column name as key and the row number as a list.
     """
-    wrong_row_dict = {
+    wrong_row_dict_int = {
```
Nora-Olivia-Ammann, line 174: Same as above with `dict` at the end.

src/dsp_tools/commands/excel2json/properties.py
jnussbaum: There is no advantage, I did this to please mypy

src/dsp_tools/commands/excel2json/resources.py
jnussbaum: Because a `Hashable` cannot be casted to an `int` directly. mypy wants to have the detour via str.

Nora-Olivia-Ammann: Why does it work then?

src/dsp_tools/commands/excel2json/properties.py
jnussbaum: I agree it's ugly. The only alternative I see is to add `# type: ignore` comments, which I did now.

Nora-Olivia-Ammann: Much appreciated thanks.

Nora-Olivia-Ammann (approved):

src/dsp_tools/commands/excel2json/resources.py
Nora-Olivia-Ammann: Ok, thanks for the explanation. I am not sure if the mypy logic makes sense to me, because not every string can be converted to an int, but that is besides the point.

PR #901, opened 2024-04-05
feat: interrupt xmlupload if a 404 occurs (DEV-3444)

jnussbaum:


src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -432,33 +424,49 @@ def _upload_resources(
...
+            )
+            logger.error(msg)
+            raise XmlUploadInterruptedError(msg)
+
+        try:
+            _tidy_up_resource_creation_idempotent(
+                res, resources, resource, failed_uploads, iri_resolver, current_res, total_res
+            )
+        except KeyboardInterrupt:
+            _tidy_up_resource_creation_idempotent(
```
Nora-Olivia-Ammann, line 440: Maybe print here as we discussed, that tidying up is in progress.

PR #904, opened 2024-04-08
feat: add models (DEV-3464)

Nora-Olivia-Ammann:


Nora-Olivia-Ammann:
> Beautiful code! Two ideas that you could implement either in this PR or later:
> 
>     * add tests for de/fr/it/rm labels, and negative tests for not supported labels like "es".
>       
>       * currently, there is no protection against invalid language tags, since `typing.Literal` is only checked by external type checkers, not by the python interpreter.
> 
>     * instead of `ListRoot.make()`, you could name this `to_json()`

Thanks! I will rename it.
All conformity checks will be done on the excel before the code even gets this far...

PR #907, opened 2024-04-09
fix: excel2properties no longer crashes if optional columns are missing (DEV-3468)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -166,7 +166,7 @@ def _add_optional_columns(df: pd.DataFrame) -> pd.DataFrame:
     in_df_cols = set(df.columns)
     if not optional_col_set.issubset(in_df_cols):
         additional_col = list(optional_col_set.difference(in_df_cols))
-        additional_df = pd.DataFrame(columns=additional_col, index=df.index, data=None)
+        additional_df = pd.DataFrame(columns=additional_col, index=df.index, data=pd.NA)  # type: ignore[call-overload]
```
Nora-Olivia-Ammann, line 169: If this is not `pd.NA` (`np.nan` is likely to work also) then it is not recognised as empty. This causes a problem in the code in utils.

src/dsp_tools/commands/excel2json/utils.py
```diff
@@ -190,7 +190,7 @@ def get_labels(df_row: pd.Series[Any]) -> dict[str, str]:
     Returns:
         A dictionary with the language tag and the content of the cell
     """
-    return {lang: df_row[f"label_{lang}"] for lang in languages if df_row[f"label_{lang}"] is not pd.NA}
+    return {lang: df_row.get(f"label_{lang}") for lang in languages if not pd.isna(df_row[f"label_{lang}"])}
```
Nora-Olivia-Ammann, line 193: This change is now no longer necessary, but I think it is more stable in case in the future a change to `np.nan` or similar would be made.

src/dsp_tools/commands/excel2json/properties.py
jnussbaum: I'm not enough knowledgeable to complain, but I assume that the typing system knows better than we do. If we introduce something that the type system doesn't like, it should make us suspicious. if So perhaps it's worth digging deeper here, to make sure that we don't introduce another bug.
But if you're confident enough, I don't want to block you.

jnussbaum (approved):
LGTM. Would it be worth the effort to add a test/test file to prevent regression?

jnussbaum:
Concerning the PR title: "no longer crashes if optional columns are missing" is a sentence without subject, which is a bit fishy. If I read a changelog, I want to read a list of actions that were taken, e.g. "prevent crash if optional columns are missing"

Nora-Olivia-Ammann:
> LGTM. Would it be worth the effort to add a test/test file to prevent regression?

What would you like to have tested?

jnussbaum:
> > LGTM. Would it be worth the effort to add a test/test file to prevent regression?
> 
> What would you like to have tested?

Currently, there is no test that prevents us from accidentally fall back into a worse state than we have now (=regression). Ideally, you would write a test that fails on the current main branch, but succeeds on this branch here.

Nora-Olivia-Ammann:
> > > LGTM. Would it be worth the effort to add a test/test file to prevent regression?
> > 
> > 
> > What would you like to have tested?
> 
> Currently, there is no test that prevents us from accidentally fall back into a worse state than we have now (=regression). Ideally, you would write a test that fails on the main branch, but succeeds on this branch here.

Right I can do that.

src/dsp_tools/commands/excel2json/properties.py
jnussbaum: So do you think there is a bug in the pandas stubs? Because the `data=pd.NA` is not accepted by the stubs, which makes me suspicious.

jnussbaum: What was Daniela's problem? Could it be solved in another way?

src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -166,7 +166,7 @@ def _add_optional_columns(df: pd.DataFrame) -> pd.DataFrame:
     in_df_cols = set(df.columns)
     if not optional_col_set.issubset(in_df_cols):
         additional_col = list(optional_col_set.difference(in_df_cols))
-        additional_df = pd.DataFrame(columns=additional_col, index=df.index, data=None)
+        additional_df = pd.DataFrame(columns=additional_col, index=df.index)
```
Nora-Olivia-Ammann, line 169: The problem is now fixed.
https://chat.openai.com/share/e/f7229856-04fb-407b-bee5-c6d6b986be60
I didn't realise that if you give an index, you are also allowed to leave data empty...

src/dsp_tools/commands/excel2json/properties.py
Nora-Olivia-Ammann: Honestly, I don't really understand this either. Because if you print out `pd.isna(None)` you get `True`. But pandas really is super difficult with nan values. That I can say for sure. Maybe the problem was that the nan values in the df were now mixed, because while they both are nan they are not identical, see screenshot. But giving `data=None` or `pd.NA` seem like it is a bad idea so, at least we fixed something we understand. I checked this cannot be found anywhere else in the code.

Screenshot 2024-04-09 at 10 58 05

Nora-Olivia-Ammann:
> > > > LGTM. Would it be worth the effort to add a test/test file to prevent regression?
> > > 
> > > 
> > > What would you like to have tested?
> > 
> > 
> > Currently, there is no test that prevents us from accidentally fall back into a worse state than we have now (=regression). Ideally, you would write a test that fails on the main branch, but succeeds on this branch here.
> 
> Right I can do that.

I just checked, we do test that and the test never failed. So further confusion abounds, but I am 99% sure we did not make it worse...

jnussbaum:
> > > > > LGTM. Would it be worth the effort to add a test/test file to prevent regression?
> > > > 
> > > > 
> > > > What would you like to have tested?
> > > 
> > > 
> > > Currently, there is no test that prevents us from accidentally fall back into a worse state than we have now (=regression). Ideally, you would write a test that fails on the main branch, but succeeds on this branch here.
> > 
> > 
> > Right I can do that.
> 
> I just checked, we do test that and the test never failed. So further confusion abounds, but I am 99% sure we did not make it worse...

Ui, that doesn't sound good. This should be investigated. But apart from that, I rather thought about preventing **future** regression. It can easily happen in some months that someone changes the code and accidentally introduces a new bug.

PR #908, opened 2024-04-09
fix(excel2json): if gui_order is empty no warning is longer displayed (DEV-3469)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -118,14 +118,18 @@ def _row2resource(
     """
```
jnussbaum, line 118: PR title: the word "no" is duplicated

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -167,12 +171,17 @@ def _create_all_cardinalities(class_name: str, class_df_with_cardinalities: pd.D
...
     detail_problem_msg = ""
+    attempt_conversion = False
     if "gui_order" not in class_df_with_cardinalities:
-        detail_problem_msg = "the column 'gui_order' does not exist."
+        pass
+    elif class_df_with_cardinalities["gui_order"].isna().all():
+        pass
     elif class_df_with_cardinalities["gui_order"].isna().any():
         detail_problem_msg = "some rows in the column 'gui_order' are empty."
+    elif not class_df_with_cardinalities["gui_order"].isna().all():
```
jnussbaum, line 181: technically, it would be more correct to turn this into an `else`. But I understand that sometimes, for the human brain it is more helpful to have another elif.

jnussbaum (approved):

src/dsp_tools/commands/excel2json/resources.py
Nora-Olivia-Ammann: yes, but then I would have to do another if clause in the else, I don't like to do that...

jnussbaum: No, you don't need an if clause. The exact opposite of this if clause is already tested on line 177, so line 181 will always return True.

Nora-Olivia-Ammann: I meant the wrong elif, but yes, I preferred it for expressiveness...

PR #910, opened 2024-04-11
fix: allow slashes and semicola in query part of URL (DEV-3480)

jnussbaum:


Nora-Olivia-Ammann (approved):
I didn't take a close look, but it looks good.

PR #917, opened 2024-04-22
refactor(xmlupload): refactor resource creation

jnussbaum:
## Problem statement

The processes of xmlupload during resource creation are not tested, because they are not testable. It is unclear if the error handling, the preservation of the current state, and the user communication are correct.

## Approach

Make the `UploadState` mutable, and create it at the very beginning at a high level. Then, modify it in the lower level functions. This allows testing the higher level functions, using mocking.

## Challenges

The things stored in `UploadState`  must be slightly rearranged. So the new DSP-TOOLS version won't be able to continue an xmlupload aborted with an older version. -> Don't forget to inform RDU.

Because of logging/printing, `_tidy_up_resource_creation_idempotent()` is not really idempotent any more.


src/dsp_tools/models/custom_warnings.py
```diff
@@ -13,6 +13,15 @@ def showwarning(cls, message: str) -> None:
         """Functionality that should be executed when a warning of this class is emitted"""
 
 
+class GeneralDspToolsWarning(Warning):
+    """Class for user-facing deprecation warnings"""
```
jnussbaum, line 17: - no deprecation here
- check with https://github.com/dasch-swiss/dsp-tools/pull/930, and take the warning defined there, instead of defining a new one here

src/dsp_tools/commands/xmlupload/project_client.py
```diff
@@ -19,6 +19,10 @@ class ProjectInfo:
 class ProjectClient(Protocol):
     """Interface (protocol) for project-related requests to the DSP-API."""
 
+    con: Connection
+    shortcode: str
+    project_info: ProjectInfo | None
+
```
jnussbaum, line 25: I just added these for better typing support. The real implementation of the ProjectClient already had these fields before.

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -343,129 +291,138 @@ def _extract_resources_from_xml(root: etree._Element, default_ontology: str) ->
...
+    if resource in upload_state.pending_resources:
+        upload_state.pending_resources.remove(resource)
 
 
 def _create_resource(
     resource: XMLResource,
     bitstream_information: BitstreamInfo | None,
     resource_create_client: ResourceCreateClient,
-) -> tuple[str, str] | tuple[None, None]:
+) -> str | None:
```
jnussbaum, line 425: I simplified the return values of the ResourceCreateClient. It's useless that it returns the label, because we already know the label.

Nora-Olivia-Ammann (approved):

PR #919, opened 2024-04-22
feat: add classes for list nodes (DEV-3524)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/models/list_node.py
```diff
@@ -0,0 +1,103 @@
+from __future__ import annotations
```
Nora-Olivia-Ammann, line 1: I ended up not going for pydantic, as there is no way to validate it without raising an error.
When I create the lists, I don't want to catch errors but match the class type that was returned. And if it is of type "...Problem" then I will append it to a problem list, which I will check in the end. Copilot told me that a class method the easiest way to create this functionality.

test/unittests/commands/excel2json/test_list_node.py
```diff
@@ -0,0 +1,174 @@
...
+
+from dsp_tools.commands.excel2json.models.input_error import ListNodeProblem
+from dsp_tools.commands.excel2json.models.input_error import ListProblem
+from dsp_tools.commands.excel2json.models.list_node import ListNode
+from dsp_tools.commands.excel2json.models.list_node import ListRoot
+
+
+class TestListRoot:
+    def test_make_myself_comments(self) -> None:
+        sub_1 = ListNode(id_="Node_1", labels={"en": "Node_1_label_en"}, row_number=1)
```
Nora-Olivia-Ammann, line 12: Here I did not use the `.create()` method because of typing when the function that is tested is called.

src/dsp_tools/commands/excel2json/models/input_error.py
```diff
@@ -240,3 +241,44 @@ def execute_error_protocol(self) -> str:
...
+@dataclass(frozen=True)
+class ListNodeProblem:
+    node_id: str
+    problems: dict[str, str]
+
+    def execute_error_protocol(self) -> str:
+        msg = [
+            f"The node '{self.node_id}' has the following problem(s):",
+        ]
+        msg.extend([f"Field: '{key}' Problem: {value}" for key, value in self.problems.items()])
```
jnussbaum, line 255: and similarly in the other occurrences

```suggestion
        msg.extend([f"Field: '{key}', Problem: {value}" for key, value in self.problems.items()])
```

src/dsp_tools/commands/excel2json/models/input_error.py
```diff
@@ -240,3 +241,44 @@ def execute_error_protocol(self) -> str:
...
         return separator.join(msg)
+
+
+@dataclass(frozen=True)
+class ListNodeProblem:
+    node_id: str
+    problems: dict[str, str]
+
+    def execute_error_protocol(self) -> str:
+        msg = [
```
jnussbaum, line 252: If the `problems` dict has no entries, this will generate an ugly message. Perhaps you should check that before.

src/dsp_tools/commands/excel2json/models/input_error.py
jnussbaum: Not sure if I got this correctly, but what do you mean with "Field"? Is it the Excel column name? Then I would call it like that.

src/dsp_tools/commands/excel2json/models/list_node.py
```diff
@@ -0,0 +1,103 @@
...
+        if comments:
+            if not set(comments).issubset({"en", "de", "fr", "it", "rm"}):
+                user_problem["comments"] = "Only the following languages are supported: 'en', 'de', 'fr', 'it', 'rm'."
+        if len(nodes) == 0:
+            user_problem["list nodes"] = "At least one node per list is required."
+        if user_problem:
+            return ListProblem(id_, user_problem)
+        return cls(id_=id_, labels=labels, comments=comments, nodes=nodes)
+
+    def to_json(self) -> dict[str, Any]:
```
jnussbaum, line 91: Perhaps, `as_dict` would be more precise, because AFAIU, JSON is the serialized string, whereas the thing that you produce is a python dictionary, which has nothing to do with a particular string format like JSON.

test/unittests/commands/excel2json/test_list_node.py
```diff
@@ -0,0 +1,174 @@
...
+        assert isinstance(root, ListProblem)
+        assert root.root_problems == {
+            "name": "The name of the list may not be empty.",
+            "labels": "At least one label per list is required.",
+            "list nodes": "At least one node per list is required.",
+        }
+
+    def test_float(self) -> None:
+        nd = ListNode(id_="NodeID", labels={"en": "node_label_en"}, row_number=1)
+        root = ListRoot.create(id_=1.123, labels={"en": "node_label_en"}, comments={}, nodes=[nd])  # type: ignore[arg-type]
```
jnussbaum, line 95: It seems a bit odd to type `id_` as string, but then handling ints and floats, too. Either, you should allow ints and floats in the typing (because your code is able to handle them), or you should raise an error if they are passed in.

jnussbaum (approved):

src/dsp_tools/commands/excel2json/models/input_error.py
Nora-Olivia-Ammann: Well it is not quite the excel column name, for example the labels are in several columns, so I can't call it column...

src/dsp_tools/commands/excel2json/models/input_error.py
Nora-Olivia-Ammann: For the excel and the Node, problems is never empty, but it is true that List itself may not have any problems.

src/dsp_tools/commands/excel2json/models/list_node.py
```diff
@@ -0,0 +1,103 @@
...
+    def to_dict(self) -> dict[str, Any]:
+        node = self._make_myself()
+        if self.sub_nodes:
+            node["nodes"] = self._make_subnodes()
+        return node
+
+    def _make_subnodes(self) -> list[dict[str, Any]]:
+        return [nd.to_dict() for nd in self.sub_nodes]
+
+    def _make_myself(self) -> dict[str, Any]:
```
BalduinLandolt, line 50: I don't really like this name... maybe `_as_dict`?

BalduinLandolt (approved):
LGTM

PR #920, opened 2024-04-22
deprecated: deprecate isSequenceOf (DEV-3525)

jnussbaum:


src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -83,6 +86,19 @@ def excel2properties(
     return props, True
 
 
+def _check_for_deprecated_syntax(df: pd.DataFrame) -> None:
+    _check_for_deprecated_isSequenceOf(df)
```
jnussbaum, line 90: my idea is to leave `_check_for_deprecated_syntax()` in place when I remove the `isSequenceOf` definitively. Kind of as infrastructure for future deprecation warnings.

Nora-Olivia-Ammann: I would even leave it in for a little longer and raise an error instead.

Nora-Olivia-Ammann (commented):
Even though it is only temp code, I would like to have at least one test.

Nora-Olivia-Ammann (approved):

PR #926, opened 2024-04-24
feat: nodes from df (DEV-3467)

Nora-Olivia-Ammann:


jnussbaum (commented):
Here comes the first part of my comments. More is to follow...

test/unittests/commands/excel2json/test_list_node.py
```diff
@@ -62,37 +64,36 @@ def test_to_json(self) -> None:
...
-        root = ListRoot.create(id_="", labels={}, comments={}, nodes=[])
-        assert isinstance(root, ListProblem)
-        assert root.root_id == ""
+        root = ListRoot.create(id_="", sheet_name="sheet", labels={}, comments={}, nodes=[])
+        assert isinstance(root, ListSheetProblem)
+        assert root.sheet_name == "sheet"
         assert root.root_problems == {
-            "name": "The name of the list does not contain any characters.",
             "labels": "At least one label per list is required.",
-            "list nodes": "At least one node per list is required.",
```
jnussbaum, line 71: Why has this warning disappeared? I think it would make sense to require at least 1 node.

test/unittests/commands/excel2json/test_lists.py
```diff
@@ -0,0 +1,417 @@
...
+def test_handle_ids() -> None:
+    test_df = pd.DataFrame(
+        {
+            "ID (optional)": [pd.NA, pd.NA, "2", "3", pd.NA, "5", "6", pd.NA],
+            "en_list": ["list_en", "list_en", "list_en", "list_en", "list_en", "list_en", "list_en", "list_en"],
+            "en_1": ["list_en", "nd_en_1", "nd_en_2", "nd_en_3", "nd_en_4", "nd_en_5", "nd_en_6", "nd_en_7"],
+        }
+    )
+    res = _fill_id_and_parent_id_columns(test_df, "en")
+    assert res["id"].tolist() == ["list_en", "nd_en_1", "2", "3", "nd_en_4", "5", "6", "nd_en_7"]
```
jnussbaum, line 165: 2 columns are created/filled, so it would make sense to test both of them

test/unittests/commands/excel2json/test_lists.py
```diff
@@ -0,0 +1,417 @@
...
+        {
+            "id": [
+                "list_en",
+                "1",
+                "2",
+                "3",
+                "3.1",
+                "3.2",
+                "3.2.1",
+                "3.2.2",
```
jnussbaum, line 179: if you remove the trailing comma and then run auto-format, this will fit on 1 line

jnussbaum (approved):
LGTM. I have a lot of suggestions that would make the code more understandable, but nothing severe that would prevent merging.

Nora-Olivia-Ammann:
General comment: There is no place where users can put a comment for a list in the excel. I would ask RDU if they want to have it at all. Untill we cleared that up, I would leave it as it is easier to remove than add.

Nora-Olivia-Ammann:
> ct, or check with PM/the project lead if you're unsure. Development is not the best time for discovering requirements. :)

Thanks, I'll ask Daniela, and I did not mean, to delete it from the json schema generally, just in the excel2json.

BalduinLandolt:
> > ct, or check with PM/the project lead if you're unsure. Development is not the best time for discovering requirements. :)
> 
> Thanks, I'll ask Daniela, and I did not mean, to delete it from the json schema generally, just in the excel2json.

My plead was mainly towards simplicity/pragmatism: That's not your main focus here, so you shouldn't cause yourself too much work with it. Or in other words: If it's there, leave it there unless you have good reason to touch it (e.g. it causes you much extra work); and if it's not useable, that can be discussed some other time.

src/dsp_tools/commands/excel2json/new_lists.py
```diff
@@ -0,0 +1,155 @@
...
+    node_dict, problems = _make_list_nodes(df)
+    cols = [x for x in df.columns if regex.search(r"^(en|de|fr|it|rm)_list$", x)]
+    root = ListRoot.create(
+        id_=df.at[0, "id"],
+        labels=_get_labels(df.iloc[0], cols),
+        sheet_name=sheet_name,
+    )
+    match (root, problems):
+        case (ListRoot(), list(ListNodeProblem())):
+            return ListSheetProblem(sheet_name, root_problems={}, nodes=problems)
```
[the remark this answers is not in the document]
jnussbaum: Sad that the type deduction doesn't properly work here, but then I suggest this:

```
match (root, node_problems):
        case (ListRoot(), list(ListNodeProblem())):
            return ListSheetProblem(sheet_name, root_problems={}, node_problems=node_problems)
        case (ListSheetProblem(), _):
            root = cast(ListSheetProblem, root)
            root.node_problems = node_problems
            return root
    root = cast(ListRoot, root)
    root.nodes = _add_nodes_to_parent(node_dict, df.at[0, "id"])
    return root
```

This would make the code better understandable (at least for me). The price to pay would be two manual castings.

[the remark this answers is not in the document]
BalduinLandolt: from what I understand, the following should get the typing right:

```python
match (root, problems):
    case (ListRoot(), []): # only matches empty list, so the happy path
        root.nodes = _add_nodes_to_parent(node_dict, df.at[0, "id"])
        return root
    case (ListRoot(), _): # non-empty, so unhappy path
        return ListSheetProblem(sheet_name, root_problems={}, nodes=problems)
    case (ListSheetProblem(), _):
        root.nodes = problems
        return root
```

PR #930, opened 2024-04-25
feat: formal compliance checks of Excel (shape only) (DEV-3544)

Nora-Olivia-Ammann:
- [x] DF: at least two rows: list name and one node
- [x] Columns: cols that do not fit the expected format: print a warning that these will not be included
- [x] Columns: one "lang_list", at least one "lang_num"
- [x] Columns: in each hierarchy group (eg. "1_en", "1_de"), all languages used elsewhere must be present


src/dsp_tools/commands/excel2json/models/input_error.py
```diff
@@ -284,3 +284,29 @@ def execute_error_protocol(self) -> str:
...
+@dataclass(frozen=True)
+class ListSheetComplianceProblem:
+    sheet_name: str
+    problems: dict[str, str]
+
+    def execute_error_protocol(self) -> str:
+        msg = [
+            f"The excel sheet '{self.sheet_name}' has the following problem(s):",
+        ]
+        msg.extend([f"Field: '{key}', Problem: {value}" for key, value in self.problems.items()])
```
jnussbaum, line 298: What is "Field" in this context? In the code, you provide a short description as the key for the dict, and not a "field". So I wonder if the final error message will be understandable.

jnussbaum: Follow up question: Have you checked that the final error message looks like you intended?

src/dsp_tools/commands/excel2json/models/input_error.py
```diff
@@ -284,3 +284,29 @@ def execute_error_protocol(self) -> str:
...
+
+@dataclass(frozen=True)
+class ListSheetComplianceProblem:
+    sheet_name: str
+    problems: dict[str, str]
+
+    def execute_error_protocol(self) -> str:
+        msg = [
+            f"The excel sheet '{self.sheet_name}' has the following problem(s):",
+        ]
```
jnussbaum, line 297:
```suggestion
        msg = [f"The excel sheet '{self.sheet_name}' has the following problem(s):"]
```

src/dsp_tools/commands/excel2json/models/input_error.py
```diff
@@ -284,3 +284,29 @@ def execute_error_protocol(self) -> str:
...
+
+@dataclass
+class ListExcelComplianceProblem:
+    excel_name: str
+    problems: list[ListSheetComplianceProblem]
+
+    def execute_error_protocol(self) -> str:
+        msg = [
+            f"The excel '{self.excel_name}' has the following problem(s):",
+        ]
```
jnussbaum, line 310:
```suggestion
        msg = [f"The excel '{self.excel_name}' has the following problem(s):"]
```

test/unittests/commands/excel2json/test_lists_compliance_checks.py
```diff
@@ -0,0 +1,122 @@
...
+            "'[lang]_[column_number]'",
+        }
+        warning_msg = regex.escape(
+            "The following columns do not conform to the expected format "
+            "and will not be included in the output: additional_1"
+        )
+        with pytest.warns(DspToolsUserWarning, match=warning_msg):
+            res = _df_shape_compliance(test_df, "sheet")
+            res = cast(ListSheetComplianceProblem, res)
+            assert res.problems == expected
```
jnussbaum, line 41:
```suggestion
            assert res and res.problems == expected
```

test/unittests/commands/excel2json/test_lists_compliance_checks.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class TestCheckMinNumColPresent:
+    def test_good(self) -> None:
+        cols = pd.Index(["ID (optional)", "en_list", "en_2"])
+        assert not _check_min_num_col_present(cols)
+
+    def test_good_no_id(self) -> None:
+        cols = pd.Index(["ID (optional)", "en_list", "en_2"])
```
jnussbaum, line 50:
```suggestion
        cols = pd.Index(["en_list", "en_2"])
```

test/unittests/commands/excel2json/test_lists_compliance_checks.py
```diff
@@ -0,0 +1,122 @@
...
+        )
+        with pytest.warns(DspToolsUserWarning, match=warning_msg):
+            res = _df_shape_compliance(test_df, "sheet")
+            res = cast(ListSheetComplianceProblem, res)
+            assert res.problems == expected
+
+
+class TestCheckMinNumColPresent:
+    def test_good(self) -> None:
+        cols = pd.Index(["ID (optional)", "en_list", "en_2"])
```
jnussbaum, line 46: variable renaming suggestion: `col_titles`

test/unittests/commands/excel2json/test_lists_compliance_checks.py
```diff
@@ -0,0 +1,122 @@
...
+            "en_2"
+        }
+        assert _check_all_expected_translations_present(test_cols) == expected
+
+    def test_missing_translations_list_columns(self) -> None:
+        test_cols = pd.Index(["ID (optional)", "en_list", "de_1", "en_1", "de_2", "en_2"])
+        expected = {
+            "missing translations": "All the nodes must be translated into the same languages, "
+            "the translations for the following column(s): "
+            "de_list"
```
jnussbaum, line 120: as already commented upon in the production code, I don't think that this message is understandable for the user.

src/dsp_tools/commands/excel2json/models/input_error.py
jnussbaum: That's fine, but I still wonder if "field" is the right term here.

PR #933, opened 2024-04-30
feat!: add support for video and audio segments, remove isSequenceOf and standalone interval value (DEV-3439)

jnussbaum:


test/unittests/commands/excel2xml/test_excel2xml_lib.py
```diff
@@ -406,11 +407,14 @@ def test_make_integer_prop(self) -> None:
         self.run_test(prop, method, [int(x) for x in different_values], invalid_values)
 
     def test_make_interval_prop(self) -> None:
```
jnussbaum, line 409: `<interval-prop>`s can now only have one value, while all other props can have multiple values. Thus, the usual testing pattern cannot be applied any more.

docs/file-formats/xml-data-file.md
```diff
@@ -980,6 +945,47 @@ Technical note: A `<link>` is in fact a `<resource restype="LinkObj">`. But it i
...
+
+- `isSegmentOf` (1)
+- `hasSegmentBounds` (1)
+- `hasTitle` (0-1)
+- `hasDescription` (0-n)
+- `hasComment` (0-n)
+- `hasKeyword` (0-n)
+- `relatesTo` (0-n)
+
+The bounds of the segment are defined by two floating point numbers, which represent the seconds.
```
Nora-Olivia-Ammann, line 961: say what floating point is. What happens with milliseconds?

Perhaps you could write:
The segment bounds must be entered in seconds. Decimal values for milliseconds are allowed. eg. 1.4:2

docs/file-formats/xml-data-file.md
```diff
@@ -980,6 +945,47 @@ Technical note: A `<link>` is in fact a `<resource restype="LinkObj">`. But it i
...
+
+Example:
+
+```xml
+<video-segment label="Segment of video_thing_1" id="video_segment_0" permissions="res-default">
+    <resptr-prop name="isSegmentOf">
+        <resptr permissions="prop-default">video_thing_1</resptr>
+    </resptr-prop>
+    <interval-prop name="hasSegmentBounds">
+        <interval permissions="prop-default">10:20</interval>  # from 0h 00min 10s to 0h 00min 20s
```
Nora-Olivia-Ammann, line 972: I would remove the inline comment.

docs/file-formats/xml-data-file.md
```diff
@@ -980,6 +945,47 @@ Technical note: A `<link>` is in fact a `<resource restype="LinkObj">`. But it i
...
+    </interval-prop>
+    <text-prop name="hasComment">
+        <text encoding="xml" permissions="prop-default">Optional comment</text>
+    </text-prop>
+</video-segment>
+```
+
+Technical notes:
```
Nora-Olivia-Ammann, line 980: I don't think this is the right place for this information. It is basically only relevant for dsp-tools devs and will confuse everyone else.

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -1071,80 +1070,77 @@ def make_integer_prop(
...
+        - If the name is not "hasSegmentBounds"
 
     Returns:
         an etree._Element that can be appended to the parent resource with resource.append(make_*_prop(...))
 
     Examples:
-        >>> excel2xml.make_interval_prop(":testproperty", "61:3600")
-                <interval-prop name=":testproperty">
-                    <interval permissions="prop-default">61:3600</interval>
+        >>> interval = excel3xml.create_interval_value(start="0:01:00", end="0:02:00")  # result: "60:120"
```
Nora-Olivia-Ammann, line 1094: not excel3xml

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -1071,80 +1070,77 @@ def make_integer_prop(
...
                 </interval-prop>
-        >>> excel2xml.make_interval_prop(":testproperty", excel2xml.PropertyElement("61:3600", permissions="prop-restricted", comment="example"))
-                <interval-prop name=":testproperty">
-                    <interval permissions="prop-restricted" comment="example">61:3600</interval>
-                </interval-prop>
-        >>> excel2xml.make_interval_prop(":testproperty", ["61:3600", "60.5:120.5"])
-                <interval-prop name=":testproperty">
-                    <interval permissions="prop-default">61:3600</interval>
-                    <interval permissions="prop-default">60.5:120.5</interval>
+        >>> interval = excel3xml.create_interval_value(start="0:30:00", end="1:00:00")  # result: "1800:3600"
```
Nora-Olivia-Ammann, line 1099: also excel3xml

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -1761,6 +1753,107 @@ def make_link(  # noqa: D417 (undocumented-param)
     )
 
 
+def make_audio_segment(  # noqa: D417 (undocumented-param)
```
Nora-Olivia-Ammann, line 1756: I would include the params because then you see the information in hover.

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -1761,6 +1753,107 @@ def make_link(  # noqa: D417 (undocumented-param)
...
+    return etree.Element(
+        "{%s}audio-segment" % xml_namespace_map[None],
+        label=label,
+        id=id,
+        permissions=permissions,
+        nsmap=xml_namespace_map,
+    )
+
+
+def make_video_segment(  # noqa: D417 (undocumented-param)
```
Nora-Olivia-Ammann, line 1788: same as above

test/unittests/commands/excel2xml/test_excel2xml_lib.py
```diff
@@ -406,11 +407,14 @@ def test_make_integer_prop(self) -> None:
...
-        prop = "interval"
-        method = excel2xml.make_interval_prop
         different_values = ["+.1:+.9", "10:20", "1.5:2.5", "-.1:5", "-10.0:-5.1"]
```
Nora-Olivia-Ammann, line 410: Does it make sense that we allow negative values, if we only use it with segment bounds?

Nora-Olivia-Ammann (approved):

docs/file-formats/xml-data-file.md
jnussbaum: I see your point. The same applies for the `<link>` and `<region>` and so on, some lines above. We could consider placing this information somewhere else, but I don't know where. I feel it should be documented somewhere, and the docs just seemed the most convenient place (at least for me).

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
jnussbaum: Haha, that's a funny one. Thanks for spotting it 🤣

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
jnussbaum: I see the params on hover.

test/unittests/commands/excel2xml/test_excel2xml_lib.py
jnussbaum: No, it doesn't make sense. I moved the negative values down to the `self.assertWarns(Warning)` section. In fact, the code remains the same, because we only issue warnings for fishy values, no errors any more. But it's good to assert that this issues a warning, because it doesn't make sense.

BalduinLandolt (approved):
LGTM

PR #941, opened 2024-05-06
feat: integrate new list section (DEV-3545)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/lists_compliance_checks.py
Nora-Olivia-Ammann, line 1: Due to circular imports, this file was merged with the new_lists.py

BalduinLandolt: as a rule of thumb, in case of circular dependencies, I would generally advise against merging code (because this doesn't inherently address the issue of the circular dependency, but just widens the code unit to include the entire circle), but instead try to extract the shared logic into a third unit, upon which both other units may depend without forming a circular reference. It may be ok in this instance, hard to say spontaneously, but just as a general remark.

src/dsp_tools/commands/excel2json/project.py
```diff
@@ -86,7 +87,64 @@ def _get_validate_list_folder(data_model_files: str, folder: list[Path]) -> tupl
...
 
-def _get_validate_onto_folder(data_model_files: str, folder: list[Path]) -> tuple[list[Path], list[str]]:
+def excel2json(
+    data_model_files: str,
+    path_to_output_file: str,
+) -> bool:
+    """
+    Converts a folder containing Excel files into a JSON data model file. The folder must be structured like this:
+
+    ::
```
BalduinLandolt, line 97: is this double colon there intentionally?

BalduinLandolt (commented):
I only had a brief look. From what I can see in code, it looks ok.

However it is not exactly what we agreed upon in today's meeting: We said we'd accept the shortcut of not attempting to follow all quality assurance we should normally do, in order to get this finished out there, so you can prioritise other work without having to keep the branch open and the feature half-finished; and we agreed that this is acceptable, if we mark this as an experimental feature and consider the old workflow to be still the stable and default one. Now, renaming the old workflow to "old-xyz" and changing the behaviour of an established workflow sends a very different message, makes this a breaking change (which it would not be otherwise) and therefor goes against the idea of trying to improve the perceived quality of our software.  
In the backend, we embraced the policy of not implementing any breaking changes without deprecating the old behaviour and giving users a grace period of some weeks.

I think I would like this PR a lot better, if it left `excel2lists` and `excel2json` unchanged, and exposed either `excel2lists-experimental` or `excel2lists-new` or `excel2lists --experimental` or something like this.  
And just to be clear: The naming in code, I'm absolutely fine with. I'm really only referring to the naming used in the parser, which defines the CLI that is exposed to the users.

src/dsp_tools/commands/excel2json/lists_compliance_checks.py
Nora-Olivia-Ammann: Thanks, that is very valid.

Nora-Olivia-Ammann:
> I only had a brief look. From what I can see in code, it looks ok.
> 
> However it is not exactly what we agreed upon in today's meeting: We said we'd accept the shortcut of not attempting to follow all quality assurance we should normally do, in order to get this finished out there, so you can prioritise other work without having to keep the branch open and the feature half-finished; and we agreed that this is acceptable, if we mark this as an experimental feature and consider the old workflow to be still the stable and default one. Now, renaming the old workflow to "old-xyz" and changing the behaviour of an established workflow sends a very different message, makes this a breaking change (which it would not be otherwise) and therefor goes against the idea of trying to improve the perceived quality of our software. In the backend, we embraced the policy of not implementing any breaking changes without deprecating the old behaviour and giving users a grace period of some weeks.
> 
> I think I would like this PR a lot better, if it left `excel2lists` and `excel2json` unchanged, and exposed either `excel2lists-experimental` or `excel2lists-new` or `excel2lists --experimental` or something like this. And just to be clear: The naming in code, I'm absolutely fine with. I'm really only referring to the naming used in the parser, which defines the CLI that is exposed to the users.
> 
> Apart from this one complaint: Great that this could be implemented so quickly! Thanks for your work!

Hey! Thanks for your comment, yesterday, this exactly occurred to me too and I was about to write a message about it.
I propose the following:
- Deprecation warning for regular `excel2json`
- Renaming into `new-excel2json`
- Opening up another project: `excel2json lists` refinement
    - look at removing complicated pandas logic
    - Possible simplification of the `Problem` classes
    - improving test coverage for the more higher level function
    - removing the old code and adapting the e2e tests
    - basically all the things I would have liked to do now before it was decided to do an emergency landing

What do you think?

src/dsp_tools/commands/excel2json/project.py
Nora-Olivia-Ammann: This is the same way it appears on main, I didn't write it but I assume it serves a purpose.

BalduinLandolt: sorry, missed the fact that this is one of the moved functions, not something you wrote.

BalduinLandolt:
> > I only had a brief look. From what I can see in code, it looks ok.
> > However it is not exactly what we agreed upon in today's meeting: We said we'd accept the shortcut of not attempting to follow all quality assurance we should normally do, in order to get this finished out there, so you can prioritise other work without having to keep the branch open and the feature half-finished; and we agreed that this is acceptable, if we mark this as an experimental feature and consider the old workflow to be still the stable and default one. Now, renaming the old workflow to "old-xyz" and changing the behaviour of an established workflow sends a very different message, makes this a breaking change (which it would not be otherwise) and therefor goes against the idea of trying to improve the perceived quality of our software. In the backend, we embraced the policy of not implementing any breaking changes without deprecating the old behaviour and giving users a grace period of some weeks.
> > I think I would like this PR a lot better, if it left `excel2lists` and `excel2json` unchanged, and exposed either `excel2lists-experimental` or `excel2lists-new` or `excel2lists --experimental` or something like this. And just to be clear: The naming in code, I'm absolutely fine with. I'm really only referring to the naming used in the parser, which defines the CLI that is exposed to the users.
> > Apart from this one complaint: Great that this could be implemented so quickly! Thanks for your work!
> 
> Hey! Thanks for your comment, yesterday, this exactly occurred to me too and I was about to write a message about it. I propose the following:
> 
>     * Deprecation warning for regular `excel2json`
> 
>     * Renaming into `new-excel2json`
> 
>     * Opening up another project: `excel2json lists` refinement
>       
>       * look at removing complicated pandas logic
>       * Possible simplification of the `Problem` classes
>       * improving test coverage for the more higher level function
>       * removing the old code and adapting the e2e tests
>       * basically all the things I would have liked to do now before it was decided to do an emergency landing
> 
> 
> What do you think?

Good to hear that wasn't just me! :)

I would mainly focus on doing the renaming for now.  
If we add the deprecation warning already, then that still shifts the "default" to the new implementation, so I would hold off with that for the time being.  
And the follow-up project makes a lot of sense, I'll bring it up in today's PM meeting. Thanks!

src/dsp_tools/commands/excel2json/project.py
```diff
@@ -136,6 +203,36 @@ def _create_project_json(
     return overall_success, project
 
 
+def _new_create_project_json(
+    data_model_files: str, onto_folders: list[Path], list_folder: Path | None
+) -> tuple[bool, dict[str, Any]]:
+    overall_success = True
+    lists = None
```
jnussbaum, line 210: This line is not necessary:

```suggestion
```

test/unittests/commands/excel2json/test_new_lists.py
```diff
@@ -34,15 +34,15 @@ def test_resolve_duplicates_in_all_excels(self) -> None:
                 "id": ["0", "1", "2"],
                 "en_list": ["List1", "List1", "List1"],
                 "en_1": [pd.NA, "Node1", "Node2"],
-                "ID (optional)": [pd.NA, pd.NA, pd.NA],
+                "id (optional)": [pd.NA, pd.NA, pd.NA],
```
jnussbaum, line 37: As mentioned, I would be okay if you lowercase the column titles for ease of use inside the python code. But I would prefer to have uppercase ID in the Excel

jnussbaum (approved):
LGTM. Only one of my comments is really important: I think that you made a mistake in this line of code, where you write the labels into the comments field:

```
return {"name": self.id_, "labels": self.labels, "comments": self.labels}
```

src/dsp_tools/commands/excel2json/models/list_node.py
```diff
@@ -104,4 +104,6 @@ def _make_list_root(self) -> dict[str, Any]:
         lst = {"name": self.id_, "labels": self.labels}
         if self.comments:
             lst.update({"comments": self.comments})
+        else:
+            lst.update({"comments": self.labels})
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: This is not an accident but I agree is pretty dumb. If we do not add a comment, then the validation of the json fails. I looked at how it is done now and this is the way. But I just noticed that the comments are not necessary for the nodes, so I will remove them there.

src/dsp_tools/commands/excel2json/project.py
Nora-Olivia-Ammann: I did it because of a reference before assignement.

test/unittests/commands/excel2json/test_new_lists.py
Nora-Olivia-Ammann: The excel remains unchanged, this is because when we clean the excel we make everything lower case so that this is not something that will raise a keyerror. Therefore the dfs that are used after the cleaning must also be lower case.

src/dsp_tools/commands/excel2json/new_lists.py
```diff
@@ -1,18 +1,120 @@
...
+        msg = all_lists.execute_error_protocol()
+        logger.error(msg)
+        raise InputError(msg)
+    return [list_.to_dict() for list_ in all_lists]
+
+
+def _make_all_lists(excel_dfs: dict[str, dict[str, pd.DataFrame]]) -> list[ListRoot] | ListCreationProblem:
+    good_lists = []
+    problem_lists: list[Problem] = []
+    for filename, sheets in excel_dfs.items():
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Thanks for spotting it, it is actually another error, I need the file name.

src/dsp_tools/commands/excel2json/new_lists.py
```diff
@@ -69,16 +171,16 @@ def _remove_duplicate_ids_in_all_excels(
         for df in sheets.values():
             preferred_lang = _get_preferred_language(df.columns)
             for i, row in df.iterrows():
-                if row["id"] in duplicate_ids and pd.isna(row["ID (optional)"]):
+                if row["id"] in duplicate_ids and pd.isna(row["id (optional)"]):
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: See other comment.

PR #946, opened 2024-05-08
feat: add IIIF image support (DEV-3507)

Nora-Olivia-Ammann:


src/dsp_tools/cli/call_action.py
```diff
@@ -181,11 +181,12 @@ def _call_xmlupload(args: argparse.Namespace) -> bool:
         return validate_xml_file(Path(args.xmlfile))
     else:
         interrupt_after = args.interrupt_after if args.interrupt_after > 0 else None
+        iiif_validation = args.no_iiif_uri_validation
         return xmlupload(
             input_file=Path(args.xmlfile),
             creds=_get_creds(args),
             imgdir=args.imgdir,
-            config=UploadConfig(interrupt_after=interrupt_after),
+            config=UploadConfig(interrupt_after=interrupt_after, skip_iiif_validation=iiif_validation),
```
BalduinLandolt, line 189: cosmetic detail: if you don't do any extra check on `args.no_iiif_uri_validation` I'm not sure there is too much point in storing it in an intermediate variable, just to pass that to function later on.

src/dsp_tools/commands/xmlupload/iiif_client.py
BalduinLandolt, line 1: remark (non-blocking): to me it seems strange to call the file different from the class, if the file only contains a single class.

src/dsp_tools/commands/xmlupload/iiif_client.py
```diff
@@ -0,0 +1,62 @@
...
+        """Validate the URI and return a list of problems if any."""
+        iiif_uri_problems = [res for uri in self.uri_list if (res := self._validate_one_uri(uri)) is not None]
+        if iiif_uri_problems:
+            return AllIIIFUriProblems(problems=iiif_uri_problems)
+        return None
+
+    def _validate_one_uri(self, uri: str) -> IIIFUriProblem | None:
+        """Check if the IIIF-server is reachable. If not, it returns information for error message."""
+        regex_has_passed = is_iiif_uri(uri)
+        response = self._make_network_call(uri)
```
BalduinLandolt, line 26: question: if it doesn't match the regex, does it actually make sense to attempt the network call? and if so, if that succeeds, is it really still a problem?

src/dsp_tools/commands/xmlupload/models/input_problems.py
```diff
@@ -0,0 +1,64 @@
...
+
+@dataclass(frozen=True)
+class IIIFUriProblem:
+    """Information about a problem with an IIIF URI."""
+
+    uri: str
+    regex_has_passed: bool
+    status_code: int | None = None
+    raised_exception_name: str | None = None
+    original_text: str | None = None
```
BalduinLandolt, line 30: remark: it's ok to leave this as is now, but just because it caught my eye:  
the types of this class suggest that it would be more accurate to model this as two separate kinds of problems:

```python
type IIIFUriProblem = IIIFUriFormatProblem | IIIFServerNotReachableProblem

@dataclass(frozen=True)
class IIIFUriFormatProblem:
    uri: str
    # no boolean needed, this type indicates that the regex failed

@dataclass(frozen=True)
class IIIFServerNotReachableProblem:
    uri: str
    status_code: int
    exception_name: str
    original_text: str
    # no optionality needed, the type only gets instantiated when these params are non-optional
```

Not sure if there is a real advantage - but in terms of keeping separate concepts separate, this might make sense.  
Also, this would make it more natural to have `IIIFServerNotReachableProblem | None` be returned by `_make_network_call()`, which might not be a bad thing because you wouldn't have to pass around unspecified exceptions.

BalduinLandolt (approved):
I did some "loud thinking" during the review, but nothing that needs to be addressed.

src/dsp_tools/commands/xmlupload/models/input_problems.py
Nora-Olivia-Ammann: Principally I like the suggestion to separate this cleanly, but there are some reasons why I kept i
You asked later if it makes sense to do a network call if the regex did not pass. I did this because we cannot be 100% sure that our regex does not fail valid URIs. If this were the case, then I would like to get notified, so that we can take a look and perhaps change the regex. I would also like to pass this info to the user, so that they don't get confused, why something valid is failing. Because this information is relevant to the message, keeping both information, in the same class is easier.

src/dsp_tools/commands/xmlupload/iiif_client.py
Nora-Olivia-Ammann: See explanation below.

jnussbaum (approved):
LGTM

PR #947, opened 2024-05-08
feat: IIIF URI Regex (DEV-3614)

Nora-Olivia-Ammann:


src/dsp_tools/utils/uri_util.py
```diff
@@ -12,3 +12,55 @@ def is_uri(s: str) -> bool:
...
+    https://iiif.io/api/annex/notes/design_principles/#define-success-not-failure
+    This may need to be updated in case of a valid version which is not in the regex.
+
+    Args:
+        uri: The URI to be checked.
+
+    Returns:
+        True if the URI is a valid IIIF URI, False otherwise.
+    """
+    # {scheme}://{server}{/prefix}/{identifier}/{region}/{size}/{rotation}/{quality}.{format}
```
jnussbaum, line 32:
```suggestion
    # {scheme}://{server}/{prefix}/{identifier}/{region}/{size}/{rotation}/{quality}.{format}
```

src/dsp_tools/utils/uri_util.py
```diff
@@ -12,3 +12,55 @@ def is_uri(s: str) -> bool:
...
+    split_uri = uri.lower().split("/")
+    if len(split_uri) < 8:
```
jnussbaum, line 34: If the example above is splitted by `/`, I count 9 parts, not 8

src/dsp_tools/utils/uri_util.py
```diff
@@ -12,3 +12,55 @@ def is_uri(s: str) -> bool:
...
+        return False
+    # %5E is the URL encoded version of ^ ->
+    # because we do change the uri to lower case we need to change that in the regex
+    # (\d+(\.\d+)?) -> number, can be integer or float
+    # everything needs to encapsulated in a group
```
jnussbaum, line 39:
```suggestion
    # everything needs to be encapsulated in a group
```

src/dsp_tools/utils/uri_util.py
```diff
@@ -12,3 +12,55 @@ def is_uri(s: str) -> bool:
...
+    # otherwise the ^ and $ only apply respectively to the first and last listed in the alternative options
```
jnussbaum, line 40: I don't understand this sentence

src/dsp_tools/utils/uri_util.py
```diff
@@ -12,3 +12,55 @@ def is_uri(s: str) -> bool:
...
+    Returns:
+        True if the URI is a valid IIIF URI, False otherwise.
+    """
+    # {scheme}://{server}{/prefix}/{identifier}/{region}/{size}/{rotation}/{quality}.{format}
+    split_uri = uri.lower().split("/")
+    if len(split_uri) < 8:
+        return False
+    # %5E is the URL encoded version of ^ ->
+    # because we do change the uri to lower case we need to change that in the regex
+    # (\d+(\.\d+)?) -> number, can be integer or float
```
jnussbaum, line 38: To make it easier: store that in a variable, then the regexes below become easier

src/dsp_tools/utils/uri_util.py
```diff
@@ -12,3 +12,55 @@ def is_uri(s: str) -> bool:
...
+    # everything needs to encapsulated in a group
+    # otherwise the ^ and $ only apply respectively to the first and last listed in the alternative options
+    region_ex = (
+        r"^((full)|(square)|"  # full | square
+        r"((pct:)?(\d+(\.\d+)?),(\d+(\.\d+)?),(\d+(\.\d+)?),(\d+(\.\d+)?)))$"  # x,y,w,h | pct:x,y,w,h
```
jnussbaum, line 43: repetitions can be simplified with non-capturing groups followed by the number of repetitions: `(?:group){4}`

jnussbaum: and if you want to inject variables with a f-string, you probably have to use % style formatting, because `{}` are already occupied by the f-string syntax

src/dsp_tools/utils/uri_util.py
```diff
@@ -12,3 +12,55 @@ def is_uri(s: str) -> bool:
...
+    )
+    if not regex.search(region_ex, split_uri[-4]):
```
jnussbaum, line 45: the code would become easier if you assign the contents of the `split_uri` tuple to variables, like so:
`scheme, server, identifier, ... = split_uri`

src/dsp_tools/utils/uri_util.py
```diff
@@ -12,3 +12,55 @@ def is_uri(s: str) -> bool:
...
+    # {scheme}://{server}{/prefix}/{identifier}/{region}/{size}/{rotation}/{quality}.{format}
+    split_uri = uri.lower().split("/")
+    if len(split_uri) < 8:
+        return False
+    # %5E is the URL encoded version of ^ ->
+    # because we do change the uri to lower case we need to change that in the regex
+    # (\d+(\.\d+)?) -> number, can be integer or float
+    # everything needs to encapsulated in a group
+    # otherwise the ^ and $ only apply respectively to the first and last listed in the alternative options
+    region_ex = (
```
jnussbaum, line 41: renaming suggestion: `region_regex`

src/dsp_tools/utils/uri_util.py
```diff
@@ -12,3 +12,55 @@ def is_uri(s: str) -> bool:
...
+        r"^((full)|(square)|"  # full | square
+        r"((pct:)?(\d+(\.\d+)?),(\d+(\.\d+)?),(\d+(\.\d+)?),(\d+(\.\d+)?)))$"  # x,y,w,h | pct:x,y,w,h
+    )
+    if not regex.search(region_ex, split_uri[-4]):
+        return False
+    size_ex = (
```
jnussbaum, line 47: renaming suggestion: size_regex
and similarly below

src/dsp_tools/utils/uri_util.py
```diff
@@ -12,3 +12,55 @@ def is_uri(s: str) -> bool:
...
+        r"(\^|%5e)?,(\d+(\.\d+)?)|"  # ,h | ^,h
+        r"(\^|%5e)?!?(\d+(\.\d+)?),(\d+(\.\d+)?))$"  # w,h | ^w,h | !w,h | ^!w,h
+    )
+    if not regex.search(size_ex, split_uri[-3]):
+        return False
+    # rotation -> floating point number 0-360 -> n | !n (positive and negative are allowed)
+    rotation_ex = r"^(!?[+-]?(\d+(\.\d+)?))$"
+    if not regex.search(rotation_ex, split_uri[-2]):
+        return False
+    # quality -> color | colour | gray |grey | bitonal | default | native
```
jnussbaum, line 61:
```suggestion
    # quality -> color | colour | gray | grey | bitonal | default | native
```

src/dsp_tools/utils/uri_util.py
```diff
@@ -12,3 +12,55 @@ def is_uri(s: str) -> bool:
...
+
+
+def is_iiif_uri(uri: str) -> bool:
+    """
+    Checks if the given URL is a valid IIIF URL.
+    It should support all versions of IIIF servers,
+    but was constructed with the syntax described in: https://iiif.io/api/image/3.0/#2-uri-syntax
+    IIIF is open to modifications which are not described in the official documentation.
+    https://iiif.io/api/annex/notes/design_principles/#define-success-not-failure
+    This may need to be updated in case of a valid version which is not in the regex.
```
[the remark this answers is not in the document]
jnussbaum: I think I would delete this line. Because every single line of code, in every codebase, must be adapted in the future if the requirements change!

src/dsp_tools/utils/uri_util.py
Nora-Olivia-Ammann: Everything before region is not regulated strictly, so this is just an approximation. This is the general problem with this regex, they have a specification, but also state that they are open for other implementations.

src/dsp_tools/utils/uri_util.py
jnussbaum: I was able to remove all these brackets (see screenshot), and the tests still pass:

Screenshot 2024-05-17 at 10 07 24

jnussbaum: let me know if I shall commit that, then you don't have to do it manually

Nora-Olivia-Ammann: Added more tests, now try again.

jnussbaum: still passing

Nora-Olivia-Ammann: Not for me

Nora-Olivia-Ammann: Ok, there were a couple too many, but not all of them. The outer most still have to stay for the tests to pass. Except for the last

jnussbaum: Strange, because we run the same tests. I was so free and committed some more bracket removals. In a minute, we'll see what GitHub CI says. We can then still revert my commit if there is a problem with it.

jnussbaum: The tests pass, so my bracket removal was valid. But feel free to revert my commit if you prefer to have some more brackets (e.g. if you think it's more readable to have some more brackets)

PR #955, opened 2024-05-13
refactor(xmlupload): add serialise value for URI (DEV-3624)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/models/serialise/serialise_value.py
```diff
@@ -0,0 +1,54 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+from typing import Optional
+from typing import Protocol
+
+
+@dataclass
```
BalduinLandolt, line 9: could these dataclasses be frozen?

src/dsp_tools/commands/xmlupload/models/serialise/serialise_value.py
```diff
@@ -0,0 +1,54 @@
...
+class SerialiseProperty:
+    property_name: str
+    values: list[SerialiseValue]
+
+    def serialise(self) -> dict[str, Any]:
+        """Serialise the property and all its values."""
+        return {self.property_name: [value.serialise() for value in self.values]}
+
+
+@dataclass
```
BalduinLandolt, line 19: is it actually meaningful to mark a protocol as dataclass? to me this sounds contradictory, but I may be wrong

src/dsp_tools/commands/xmlupload/models/serialise/serialise_value.py
```diff
@@ -0,0 +1,54 @@
...
+class SerialiseValue(Protocol):
+    """A value to be serialised."""
+
+    value: str
+    permissions: str
+    comment: Optional[str]
```
BalduinLandolt, line 25: you can just use `str | None`, no need to use `Optional` I htink

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -351,14 +362,26 @@ def _make_time_value(value: XMLValue) -> dict[str, Any]:
...
-        "@type": "knora-api:UriValue",
-        "knora-api:uriValueAsUri": {
-            "@type": "xsd:anyURI",
-            "@value": value.value,
-        },
-    }
+def _serialise_uri_prop(prop: XMLProperty, permissions_lookup: dict[str, Permissions]) -> dict[str, Any]:
+    uri_values: list[SerialiseValue] = []
+    for v in prop.values:
+        value_permission = cast(str, v.permissions)
```
BalduinLandolt, line 368: why is this cast needed?

BalduinLandolt (approved):
Looks good, I added some minor remarks, but I like the general direction of this!

src/dsp_tools/commands/xmlupload/models/serialise/serialise_value.py
Nora-Olivia-Ammann: mypy doesn't accept the duck-typing if I don't declare this as a dataclass too...

src/dsp_tools/commands/xmlupload/resource_create_client.py
Nora-Olivia-Ammann: Because in the class XMLValue permissions are optional, then it does not accept me to retrieve it from the dict. This is a problem that would be solved when we separate the classes.

src/dsp_tools/commands/xmlupload/models/serialise/serialise_value.py
BalduinLandolt: fair enough

Nora-Olivia-Ammann: Now with explicit subtyping à la Christian it works without adding the decorator

jnussbaum (approved):
LGTM

src/dsp_tools/commands/xmlupload/models/serialise/serialise_value.py
```diff
@@ -0,0 +1,52 @@
...
+    comment: str | None
+
+    def serialise(self) -> dict[str, Any]:
+        """Serialise the URI value."""
+        serialised = self._make_value()
+        if self.comment:
+            serialised["knora-api:valueHasComment"] = self.comment
+        return serialised
+
+    def _make_value(self) -> dict[str, Any]:
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I did it this way because it followed the structure of the existing code. I can change it but this will disappear as soon as we move to rdflib...

PR #957, opened 2024-05-13
refactor(xmlupload): serialise int value with rdflib (DEV-3626)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/models/namespace_context.py
```diff
@@ -28,6 +18,12 @@ def get_default_json_ld_context() -> dict[str, str]:
     }
 
 
+def make_namespace_dict_from_onto_names(ontos: dict[str, str]) -> dict[str, Namespace]:
```
Nora-Olivia-Ammann, line 21: I decided it was unnecessarily complicated to have a class after all.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -58,11 +68,13 @@ def _make_resource_with_values(
         resource: XMLResource,
         bitstream_information: BitstreamInfo | None,
     ) -> dict[str, Any]:
+        res_bnode = BNode()
+        namespaces = make_namespace_dict_from_onto_names(self.project_onto_dict)
```
jnussbaum, line 72: AFAIU, the namespaces remain the same for the entire lifetime of the class, so it would be more efficient to have it as class field

test/unittests/commands/xmlupload/models/test_jsonld_serialiser.py
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        result = serialise_property(two_int_values, str(MY_ONTO.hasInteger))
+        assert result == expected
+
+
+def test_frame_property() -> None:
+    json_graph = [
+        {
+            "@id": "_:Nc4ababd236d441e094f2928089079629",
```
jnussbaum, line 98: if you use human-readable, short strings instead of those hex-strings, it would be easier to understand

jnussbaum (approved):
I admit that I don't understand everything that goes on here, but I find this PR interesting, and I think I can learn a lot in the future from this. 

I have some remarks, but generally, this PR looks good to me.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -276,12 +300,36 @@ def _make_geoname_value(value: XMLValue) -> dict[str, Any]:
...
+
+
+def _make_integer_prop(
+    prop: XMLProperty, res_bn: BNode, prop_name: URIRef, permissions_lookup: dict[str, Permissions]
+) -> Graph:
+    g = Graph()
+    for value in prop.values:
+        single_val_bn = BNode()
+        g.add((res_bn, prop_name, single_val_bn))
+        g += _make_integer_value(value, single_val_bn, permissions_lookup)
```
BalduinLandolt, line 317: Am I understanding correctly, that the only interger-specific thing in this function is the call to `_make_integer_value()`? In that case, wouldn't it make more sense to have a generalized function `_make_value_prop()` that gets the property name and a list of graphs which represent all the nodes you get from repeatedly calling of `_make_integer_value()`?

In case I'm not making much sense here, here's some pseudo code (which is absolutely not correct like this):

```python
def _serialise_integer_prop(prop, res_bn, prop_name, permissions_lookup):
    bn = BNode()
    vals = [_make_integer_value(v, bn, permissions_lookup) for v in prop.values]
    prop_graph = _make_prop(res_bn, prop_name, vals)
    return serialise_property(prop_graph, str(prop_name))

def _make_property(res_bn, prop_name, vals):
    g = Graph()
    for v in vals:
        g.add((res_bn, prop_name, single_val_bn))
        g += v
```

Like this, you wouldn't need two functions for each value type, but just one.

BalduinLandolt (approved):
Looks very good!  
I share Johannes' feelings regarding some potential for simplification, I added one suggestion in that regard too.  
Generally I'm all in favour of having may specific models/functions but where it's possible to have a function fit all value types rather than just one, by simply injecting the bit that actually is specific, then I wouldn't neglect that possibility either.

src/dsp_tools/commands/xmlupload/models/serialise/jsonld_serialiser.py
```diff
@@ -0,0 +1,35 @@
+import json
+from typing import Any
+
+from pyld import jsonld
+from rdflib import Graph
+
+
+def serialise_property(rdf_graph: Graph, prop_name: str) -> dict[str, Any]:
+    """
+    This function makes the serialised json-ld conform to the format expected by the DSP-API.
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: No it is an rdf graph, there are two steps to serialise it so that the api accepts it.
The api does not accept all json-ld formats (unfortunately there are variations). We cannot force rdflib to make the json-ld in the format we want, but in a second step we can change the format. Because the serialisation itself needs a little transformation, I decided that this function should take a graph and not a json-ld.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -58,11 +68,13 @@ def _make_resource_with_values(
         resource: XMLResource,
         bitstream_information: BitstreamInfo | None,
     ) -> dict[str, Any]:
+        res_bnode = BNode()
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: the reason is that in the future, we need this bnode for all the properties. So if we added change more properties into serialisation with rdflib we would need to create it at this point anyway. This way we do not need to change a function, that we don't actually want to work on in a future PR.

src/dsp_tools/commands/xmlupload/resource_create_client.py
Nora-Olivia-Ammann: The namespaces really only belong into this step of the code. I did not want to have to add another field in the class that should not actually be responsible for serialisation. Since we don't actually hand over the resource class into the functions that serialise the properties, I didn't see the advantage.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -110,6 +122,14 @@ def prop_name(p: XMLProperty) -> str:
             else:
                 return f"{p.name}Value"
 
+        def get_absolute_prop_iri(prefixed_prop: str) -> URIRef:
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Thanks I did not know that about the inner functions. I took it outside.

src/dsp_tools/commands/xmlupload/resource_create_client.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Let's discuss this quickly today.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -276,12 +300,36 @@ def _make_geoname_value(value: XMLValue) -> dict[str, Any]:
     }
 
 
-def _make_integer_value(value: XMLValue) -> dict[str, Any]:
+def _serialise_integer_prop(
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Not true, the only difference is that the other function can be generic for the other properties. (make_values). For me these are two very distinct steps that I want separate. As soon as we move completely to rdflib we can do away with this complication.

src/dsp_tools/commands/xmlupload/resource_create_client.py
Nora-Olivia-Ammann: In the code you don't actually use the correct bns, but that is not your point. I mistakenly believed, that you have to actually enumerate all the properties so that the framing worked, in that case, it would be inelegant. In the end when everything is rdflib we won't be framing the properties anymore but the resources, then all this piecemeal code will vanish anyway.
But for some reason (I don't get it entirely why) it works as long as you list one valid property name
Screenshot 2024-05-15 at 08 35 41

I will change this so we make one graph for all the properties and then frame that.

src/dsp_tools/commands/xmlupload/models/serialise/jsonld_serialiser.py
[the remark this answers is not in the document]
jnussbaum: But then the docstring is not correct. The input is a graph, but the docstring suggests that the input is a "serialised json-ld"

src/dsp_tools/commands/xmlupload/resource_create_client.py
BalduinLandolt: > In the code you don't actually use the correct bns, but that is not your point. 

Yes... I realized this while writing the code, that's why I double stressed its incorrectness :)

> I mistakenly believed, that you have to actually enumerate all the properties so that the framing worked, in that case, it would be inelegant. In the end when everything is rdflib we won't be framing the properties anymore but the resources, then all this piecemeal code will vanish anyway. But for some reason (I don't get it entirely why) it works as long as you list one valid property name 
> 
> I will change this so we make one graph for all the properties and then frame that.

👍

PR #959, opened 2024-05-14
chore: Use dsp-ingest instead of sipi for upload of assets  (DEV-3620)

seakayone:
- **Introduce IngestClient protocol and DspIngestClient implementation**
- **Replace Sipi with DspIngestClient** 
- **refactor: Move XMLResource code into xml_upload.py. Simplify implementation**
- **refactor: Make structural subtyping explicit for `Connection`**
- **Add and update tests. (Adds `requests-mock` dependency)**
- **Configure ingest url for dsp-api in [start-stack/docker-compose.yml](https://github.com/dasch-swiss/dsp-tools/pull/959/files#diff-caac8e3e5f1e92403f8db3a38703f567bbc6fcfe2566d0d11594a49663a76a98R18)**
- **Use `0.9.1` ingest container version because of bugfix https://github.com/dasch-swiss/dsp-ingest/pull/216**


src/dsp_tools/commands/xmlupload/models/ingest.py
```diff
@@ -0,0 +1,87 @@
...
+                    return IngestResponse(internal_filename=res.json()["internalFilename"])
+                elif res.status_code == 401:
+                    raise BadCredentialsError("Bad credentials")
+                else:
+                    user_msg = f"{err} See logs for more details: {logger_savepath}"
+                    print(user_msg)
+                    log_msg = f"{err}. Response status code {res.status_code} '{res.json()}'"
+                    logger.error(log_msg)
+                    raise PermanentConnectionError(log_msg)
+        except FileNotFoundError:
```
seakayone, line 84:
```suggestion
        except FileNotFoundError as e:
            logger.error(e)
```

jnussbaum:
@seakayone We were told to use "feat" only for user-facing new features, thus I changed the PR title to "chore"

src/dsp_tools/commands/xmlupload/models/ingest.py
```diff
@@ -0,0 +1,91 @@
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Optional
```
Nora-Olivia-Ammann, line 3: Stylistic comment I would replace this with `| None = None`

src/dsp_tools/commands/xmlupload/models/ingest.py
```diff
@@ -0,0 +1,91 @@
...
+
+from dsp_tools.models.exceptions import BadCredentialsError
+from dsp_tools.models.exceptions import PermanentConnectionError
+from dsp_tools.models.exceptions import UserError
+from dsp_tools.utils.logger_config import logger_savepath
+
+
+@dataclass(frozen=True)
+class IngestResponse:
+    internal_filename: str
```
Nora-Olivia-Ammann, line 20: Curiosity question: 
You use this class only once and only to retrieve the string. What is the advantage of wrapping that string in a class?

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -60,8 +60,7 @@ def create_resource(
             f"Attempting to create resource {resource.res_id} (label: {resource.label}, iri: {resource.iri})..."
         )
         resource_dict = self._make_resource_with_values(resource, bitstream_information)
-        headers = {"X-Asset-Ingested": "true"} if self.media_previously_ingested else None
-        res = self.con.post(route="/v2/resources", data=resource_dict, headers=headers)
+        res = self.con.post(route="/v2/resources", data=resource_dict, headers={"X-Asset-Ingested": "true"})
```
Nora-Olivia-Ammann, line 63: So it is ok, to give this header with all requests even if there are no assets that were ingested?

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -2,23 +2,25 @@
...
 
 
 def handle_media_info(
     resource: XMLResource,
     media_previously_uploaded: bool,
-    sipi_server: Sipi,
+    ingest_client: IngestClient,
     imgdir: str,
     permissions_lookup: dict[str, Permissions],
+    shortcode: str,
```
Nora-Olivia-Ammann, line 23: Why is the shortcode not a class variable of the `IngestClient`?

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -29,98 +31,44 @@ def handle_media_info(
...
+    if bitstream.permissions is None:
         return None
-    img = sipi_server.upload_bitstream(Path(imgdir) / Path(resource.bitstream.value))
-    internal_file_name_bitstream = img["uploadedFiles"][0]["internalFilename"]
-    return resource.get_bitstream_information(
-        internal_file_name_bitstream=internal_file_name_bitstream,
-        permissions_lookup=permissions_lookup,
-    )
+    else:
+        return permissions_lookup.get(bitstream.permissions)
```
Nora-Olivia-Ammann, line 74: It was not in the original code, but should we perhaps raise an error if the permissions don't exist as we do it in the `ResourceCreateClient`?

Nora-Olivia-Ammann (approved):

test/unittests/commands/xmlupload/models/test_ingest.py
```diff
@@ -0,0 +1,57 @@
...
+def shortcode() -> str:
+    return "0001"
+
+
+@pytest.fixture()
+def tmp_file(tmp_path: Path) -> Path:
+    return tmp_path / "filename.xml"
+
+
+def test_ingest_success(dsp_ingest_url, ingest_client, requests_mock, shortcode, tmp_file):  # type: ignore[no-untyped-def]
```
jnussbaum, line 32: wherever possible, I prefer proper typing:

```suggestion
def test_ingest_success(dsp_ingest_url: IngestClient, ingest_client: IngestClient, requests_mock: Mocker, shortcode: str, tmp_file: Path) -> None:
```

test/unittests/commands/xmlupload/models/test_ingest.py
```diff
@@ -0,0 +1,57 @@
...
+    tmp_file.write_text("<xml></xml>")
+    requests_mock.post(
+        f"{dsp_ingest_url}/projects/{shortcode}/assets/ingest/{tmp_file.name}", json={"internalFilename": tmp_file.name}
+    )
+    res = ingest_client.ingest(shortcode, tmp_file)
+    assert res.internal_filename == tmp_file.name
+
+
+def test_ingest_failure_when_file_not_found(ingest_client, shortcode, tmp_file):  # type: ignore[no-untyped-def]
```
jnussbaum, line 41: same as above (typing)

jnussbaum (changes requested):
Thanks for this PR. I could learn some new technologies, which is always fun :-)

The reason why I don't approve yet is: I identified 2 issues that I think are critical:

- Critical: You deleted 2 tests in test/integration/commands/xmlupload/test_resource_creation.py
- Critical: DspIngestClient.ingest() should not raise UserError
- Suggestion: You could make the diff much shorter if you don't pass around the shortcode everywhere. In many cases, you could retrieve the shortcode from the `UploadState` which is already available.

For more details, see my comments

src/dsp_tools/commands/xmlupload/models/ingest.py
seakayone: I am used to type safety and it for me it is not uncommon to wrap a single `String` type in a wrapper. Also the actual response from the API contains more information. It would be easy to expand the class to incorporate more information if need may be.

src/dsp_tools/commands/xmlupload/resource_create_client.py
seakayone: Yes.

src/dsp_tools/commands/xmlupload/resource_multimedia.py
seakayone: I tried to keep the behaviour the same. AFAIU, the case of non existing permissions in the lookup is valid. Did I miss something?

src/dsp_tools/commands/xmlupload/resource_multimedia.py
seakayone: So you can reuse the same instance for multiple uploads, passing in the `shortcode` for each ingest.

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -29,98 +31,44 @@ def handle_media_info(
...
+        return True, None
+    elif not media_previously_uploaded:
+        try:
+            res = ingest_client.ingest(shortcode, Path(imgdir) / Path(bitstream.value))
+            msg = f"Uploaded file '{bitstream.value}'"
+            print(f"{datetime.now()}: {msg}")
+            logger.info(msg)
+            return True, BitstreamInfo(
+                bitstream.value, res.internal_filename, _permissions(bitstream, permissions_lookup)
+            )
```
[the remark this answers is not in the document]
seakayone: Sure the IngestClient could also have the `start_bulk_ingest` as a method no need for different protocols.

src/dsp_tools/commands/xmlupload/models/ingest.py
```diff
@@ -0,0 +1,91 @@
...
+                    raise BadCredentialsError("Bad credentials")
+                else:
+                    user_msg = f"{err} See logs for more details: {logger_savepath}"
+                    print(user_msg)
+                    log_msg = f"{err}. Response status code {res.status_code} '{res.json()}'"
+                    logger.error(log_msg)
+                    raise PermanentConnectionError(log_msg)
+        except FileNotFoundError as e:
+            logger.error(e)
+            raise UserError(f"{err}. File {filepath} not found.")
```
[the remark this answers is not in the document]
seakayone: It is not a connection problem at this point. If the file is defined in xml and not present on the specified path this is IMHO a user error.

test/integration/commands/xmlupload/test_resource_creation.py
```diff
@@ -8,19 +8,23 @@
...
 from dsp_tools.commands.xmlupload.stash.stash_models import Stash
 from dsp_tools.commands.xmlupload.upload_config import UploadConfig
-from dsp_tools.models.exceptions import PermanentTimeOutError
 from dsp_tools.models.exceptions import XmlUploadInterruptedError
 from dsp_tools.utils.connection import Connection
 from dsp_tools.utils.connection_live import ConnectionLive
 
 
+@pytest.fixture()
+def ingest_client_mock():  # type: ignore[no-untyped-def]
```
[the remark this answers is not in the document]
seakayone: This will produce an error on the lines where I pass in the mock into functions which expect an `IngestClient`.

test/unittests/commands/xmlupload/models/test_ingest.py
seakayone: IMHO for test code this is too noisy.

I like my types for production code, but for the test fixtures it is not really much of a help to me as these functions are only ever called by the test runner. WDYT?

src/dsp_tools/commands/xmlupload/models/ingest.py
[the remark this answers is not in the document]
BalduinLandolt: I agree that it is not a "permanent connection error" if a file was specified that does not exist. So how about adding a `NonExistentFileError` which has the same effect of not aborting the upload, but recording the issue to report it to the user.

And on a more fundamental note: If this check was done before starting the upload, then it would be ok not to have to continue an upload that clearly isn't valid. (If there is no File uploaded, the resource can not be created; and then all resources having links to this resource will fail subsequently, etc.)

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -29,98 +31,44 @@ def handle_media_info(
...
-        print(f"{datetime.now()}: {msg}")
-        logger.info(msg)
-        return resource_bitstream
-    except PermanentConnectionError as err:
-        msg = f"Unable to upload file '{bitstream.value}' of resource '{resource.label}' ({resource.res_id})"
-        print(f"{datetime.now()}: WARNING: {msg}: {err.message}")
-        logger.opt(exception=True).warning(msg)
+def _permissions(
+    bitstream: Optional[XMLBitstream], permissions_lookup: dict[str, Permissions]
+) -> Optional[Permissions]:
```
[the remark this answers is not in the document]
seakayone: => acc0ed487fbdf176f2a62f0811d99cb212108b9a

src/dsp_tools/commands/xmlupload/models/ingest.py
seakayone: Thanks for the hint, fixed in acc0ed487fbdf176f2a62f0811d99cb212108b9a

src/dsp_tools/commands/xmlupload/resource_multimedia.py
```diff
@@ -29,98 +31,44 @@ def handle_media_info(
...
-def _handle_media_upload(
-    resource: XMLResource,
-    bitstream: XMLBitstream,
-    permissions_lookup: dict[str, Permissions],
-    sipi_server: Sipi,
-    imgdir: str,
-) -> BitstreamInfo | None:
-    """
-    Upload a bitstream file to SIPI
+        return True, BitstreamInfo(bitstream.value, bitstream.value, _permissions(bitstream, permissions_lookup))
```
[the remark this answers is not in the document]
seakayone: Thanks, fixed in acc0ed487fbdf176f2a62f0811d99cb212108b9a

test/unittests/commands/xmlupload/models/test_ingest.py
```diff
@@ -0,0 +1,57 @@
...
+
+@pytest.fixture()
+def tmp_file(tmp_path: Path) -> Path:
+    return tmp_path / "filename.xml"
+
+
+def test_ingest_success(dsp_ingest_url, ingest_client, requests_mock, shortcode, tmp_file):  # type: ignore[no-untyped-def]
+    tmp_file.write_text("<xml></xml>")
+    requests_mock.post(
+        f"{dsp_ingest_url}/projects/{shortcode}/assets/ingest/{tmp_file.name}", json={"internalFilename": tmp_file.name}
```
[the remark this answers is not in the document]
seakayone: Thanks, extracted that function in 9cea2bcee352d34dec4e857380da8f97b0e68cda

src/dsp_tools/commands/xmlupload/resource_multimedia.py
Nora-Olivia-Ammann: If the permissions string used in the xml is not defined we do not know what the user intended for the property. 

```
            <text permissions="prop-unknown" encoding="utf8">blah</text>
</text-prop>
```

Permissions dict:

```
{ "prop-default": Permission }
```

Maybe it was a typo, so we do not quietly just ignore the invalid input, as we do here (because permissions are allowed to be `None` if nothing is specified), but we raise an user-facing error that communicates there is something wrong. I know this was not in the original, but if we see a small mistake such as this maybe we could fix it now. Or do a follow up PR.

src/dsp_tools/commands/xmlupload/models/ingest.py
```diff
@@ -0,0 +1,91 @@
...
+                elif res.status_code == 401:
+                    raise BadCredentialsError("Bad credentials")
+                else:
+                    user_msg = f"{err} See logs for more details: {logger_savepath}"
+                    print(user_msg)
+                    log_msg = f"{err}. Response status code {res.status_code} '{res.json()}'"
+                    logger.error(log_msg)
+                    raise PermanentConnectionError(log_msg)
+        except FileNotFoundError as e:
+            logger.error(e)
```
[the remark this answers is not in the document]
seakayone: I naturally agree, IMHO an exception should either be handled or augmented with useful information and be passed on. Logging it when passing on might lead to duplicate logging. 

However, I have seen the `logger.error ...; raise Error ...;` pattern quite a lot and that lead me to think that we might not have a catch all higher up the call stacks which would log them and followed the established pattern.

The `FileNotFoundError` for example is not catched on the line you have linked.

test/unittests/commands/xmlupload/models/test_ingest.py
seakayone: Same as above, resolving this conversation, lets continue the discussion above.

src/dsp_tools/commands/xmlupload/models/ingest.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I agree with Balduin's input, maybe as part of the restructuring of the classes we could also analyse which information we could implement as a fast fail and move that to the beginning.
One problem if we move this check to the beginning is though that if we do a `resume-xmlupload` then we cannot guarantee that we are still in the correct directory and the files were not changed... But I think this needs to be part of a larger discussion.

src/dsp_tools/commands/xmlupload/resource_multimedia.py
jnussbaum: The xmlupload CLI command is always only called for 1 project, so I'm not sure if this makes sense

test/unittests/cli/test_helper_functions.py
```diff
@@ -34,95 +34,91 @@ def setUpClass(cls) -> None:
...
             "admin.08F4-test-server.dasch.swiss": [
                 "https://api.08F4-test-server.dasch.swiss",
-                "https://iiif.08F4-test-server.dasch.swiss",
+                "https://ingest.08F4-test-server.dasch.swiss",
             ],
             "app.08F4-test-server.dasch.swiss": [
                 "https://api.08F4-test-server.dasch.swiss",
-                "https://iiif.08F4-test-server.dasch.swiss",
-            ],
-            "iiif.E5bC-test-server.dasch.swiss": [
```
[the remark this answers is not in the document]
seakayone: Yes, this behaviour is not supported by the current code anymore because iiif was replaced by ingest. 

I have added a test for ingest instead in d9fed83c87701b1e637868ffe1b02a704f2e37c8

src/dsp_tools/commands/xmlupload/resource_multimedia.py
seakayone: IMHO: I see the client as a service which itself should know about how to contact the api but nothing more. This client is able to interact with the dsp-ingest api and all parameters for a single call should be parameters on the respective function calling that api endpoint.  For me it is confusing to put the shortcode into the instance while the filename for the same api call/client funciton stays a parameter.

It would also be simpler to reuse the `IngestClient` later should that single project requirement change at some point. 

The only "drawback" is that the project code parameter must be available when calling the api which requires us to pass it down the call stack currently, which in most cases is done by the `UploadState` so the additional parameters are kept to minimum.

src/dsp_tools/commands/xmlupload/models/ingest.py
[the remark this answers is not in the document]
jnussbaum: I quickly checked, and the existence of the files is already checked at the very beginning in `_check_if_bitstreams_exist()`. So in fact, you should just remove the `except FileNotFoundError` block. It's not necessary here.

test/unittests/commands/xmlupload/models/test_ingest.py
jnussbaum: That's a valid point. But then you could accidentally pass objects of the wrong type into a function under test. Most of our production code doesn't verify the types of their input, because we assume that mypy prevents us from passing in wrong types. Or what's your opinion on that?

seakayone: `_ test_ingest_success` is the test function, i.e. the function that is doing the testing, I am not changing the type hints of the function under test which in this case is the function on the IngestClient.

src/dsp_tools/commands/xmlupload/models/ingest.py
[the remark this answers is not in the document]
seakayone: done in b8317266d20bd312081e816f048b6954a3247e98

src/dsp_tools/commands/xmlupload/models/ingest.py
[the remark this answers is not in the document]
seakayone: Remove the block in https://github.com/dasch-swiss/dsp-tools/commit/b8317266d20bd312081e816f048b6954a3247e98 as discussed in https://github.com/dasch-swiss/dsp-tools/pull/959#discussion_r1603260984

test/unittests/commands/xmlupload/models/test_ingest.py
seakayone: IMHO it is save for a function to assume the passed in arguments are of the proper type.
I do not think that every function should then do their type checks, that would too much code to maintain, mypy is doing this for us using the type hints.

jnussbaum: My point was a bit different: If the arguments of the test function are untyped, big parts of the body of the test function will also be untyped. And then the test function could accidentally pass wrong objects into the well-typed prod code, and noone will notice. This will deteriorate the quality of the tests.

IMHO, typing only protects us if we do it everywhere (?) As soon as parts of the (test) code are untyped, nasty things can find its way through the well-typed codebase incognito. And then we think that we tested something, but actually we didn't test it, because the tests are flawed.

jnussbaum: P.S. While I very much enjoy the current conversation, I want to stress that this is non-blocking for your PR ;-)

test/integration/commands/xmlupload/test_resource_creation.py
```diff
@@ -172,53 +176,7 @@ def test_2_resources_with_stash() -> None:
     assert not upload_state.pending_stash or upload_state.pending_stash.is_empty()
 
 
-def test_2_resources_with_stash_interrupted_by_timeout() -> None:
```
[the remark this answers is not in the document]
seakayone: That was an accident, I have reverted that in e19308dec505cb6a83e4fdd1946f1fea721ae2e2

src/dsp_tools/commands/xmlupload/resource_multimedia.py
jnussbaum: @Nora-Olivia-Ammann since you are the original author of the comment: could you create a ticket for this? I have already resolved the conversation.

src/dsp_tools/commands/xmlupload/models/ingest.py
```diff
@@ -0,0 +1,86 @@
...
+import requests
+from loguru import logger
+from requests import Session
+from requests.adapters import HTTPAdapter
+from requests.adapters import Retry
+
+from dsp_tools.models.exceptions import BadCredentialsError
+from dsp_tools.models.exceptions import PermanentConnectionError
+from dsp_tools.utils.logger_config import logger_savepath
+
```
siers, line 14: Are all of these imports necessary?

jnussbaum: ruff (our linter) warns us when there are unused imports, so they are certainly used, yes

PR #965, opened 2024-05-16
feat: integrate into xmlupload (DEV-3616)

Nora-Olivia-Ammann:


Nora-Olivia-Ammann:
The display in the app does not work yet as the version of the app we use does not support it, but it can be uploaded successfully.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -141,6 +142,9 @@ def make_values(p: XMLProperty) -> list[dict[str, Any]]:
                     last_prop_name = int_prop_name
                 case _:
                     properties_serialised.update({prop_name(prop): make_values(prop)})
+        if resource.iiif_uri:
+            properties_graph += _make_iiif_uri_value(resource.iiif_uri, res_bnode, self.permissions_lookup)
+            last_prop_name = KNORA_API.hasStillImageFileValue
```
jnussbaum, line 147: What is `last_prop_name`? I don't understand what it is doing. Could you perhaps write a comment into the code?

jnussbaum (approved):
LGTM

src/dsp_tools/commands/xmlupload/resource_create_client.py
Nora-Olivia-Ammann: I added a comment does it make sense now?

src/dsp_tools/commands/xmlupload/check_consistency_with_ontology.py
```diff
@@ -100,12 +100,15 @@ def _get_all_property_names_and_resource_ids_one_resource(
...
-                prop_dict[prop_name] = [resource.attrib["id"]]
+        match prop.tag:
+            case "bitstream" | "iiif-uri":
+                pass
+            case _:
+                prop_name = prop.attrib["name"]
+                if prop_name in prop_dict:
+                    prop_dict[prop_name].append(resource.attrib["id"])
+                else:
+                    prop_dict[prop_name] = [resource.attrib["id"]]
```
BalduinLandolt, line 111: by passing an empty list as default to the `get()` method, this can be simplified:

```suggestion
                prop_name = prop.attrib["name"]
                previous = prop_dict.get(prop_name, [])
                prop_dict[prop_name] = previous.append(resource.attrib["id"])
```

Or you can inline this to be a 1 or 2 liner.  
It might even work (but I'm not sure) like this:

```python
                prop_name = prop.attrib["name"]
                prop_dict.get(prop_name, []).append(resource.attrib["id"])
```

Which then could definitely be turned into

```python
                prop_dict.get(prop.attrib["name"], []).append(resource.attrib["id"])
```

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -188,6 +193,19 @@ def _make_value(self, value: XMLValue, value_type: str) -> dict[str, Any]:
...
+def _make_iiif_uri_value(iiif_uri: IIIFUriInfo, res_bnode: BNode, permissions_lookup: dict[str, Permissions]) -> Graph:
+    g = Graph()
+    iiif_bn = BNode()
+    g.add((res_bnode, KNORA_API.hasStillImageFileValue, iiif_bn))
+    g.add((iiif_bn, RDF.type, KNORA_API.StillImageExternalFileValue))
+    g.add((iiif_bn, KNORA_API.fileValueHasExternalUrl, Literal(iiif_uri.value)))
+    if iiif_uri.permissions:
+        if not (per := permissions_lookup.get(iiif_uri.permissions)):
+            raise PermissionNotExistsError(f"Could not find permissions for value: {iiif_uri.permissions}")
+        g.add((iiif_bn, KNORA_API.hasPermissions, Literal(str(per))))
```
BalduinLandolt, line 205: these three lines will probably be needed on all value types, right? And they don't depend on anything specific to the value type. If so, you could extract them to a function.

BalduinLandolt (approved):
Looks good, I added some very minor stylistic remarks.

PR #967, opened 2024-05-17
refactor: reduce number of arguments in some functions

BalduinLandolt:


jnussbaum (approved):
Very much like this one, thank you a lot :-)

PR #973, opened 2024-05-21
docs: IIIF image support (DEV-3618)

Nora-Olivia-Ammann:


docs/file-formats/xml-data-file.md
```diff
@@ -198,6 +198,7 @@ Example of a property element of type integer with two values:
 The following property elements exist:
 
 - `<bitstream>`: contains a path to a file (if the resource is a multimedia resource)
+- `<iiif-uri>`: contains a URI to an IIIF image server (resource must be a `StillImageRepresentation`)
```
jnussbaum, line 201: the part in brackets should be identical to the line above. You can change either the line above, or this line, but I find it confusing if the same thing is described in 2 different ways

docs/file-formats/xml-data-file.md
```diff
@@ -267,6 +269,38 @@ Example of a public image inside a `StillImageRepresentation`:
...
 
+### `<iiif-uri>`
+
+The `<iiif-uri>` element is used for IIIF image server URIs. 
+All versions of the IIIF-API standards are supported.
+
+Notes:
+
+- Only _one_ `<iiif-uri>` element is allowed per `StillImageRepresentation`. 
+- If a `<iiif-uri>` element is present, the `<bitstream>` element must be omitted.
```
jnussbaum, line 280:
```suggestion
- If a resource has a `<iiif-uri>` element, it cannot have a `<bitstream>` element.
```

docs/file-formats/xml-data-file.md
```diff
@@ -223,6 +224,7 @@ Notes:
 
 - There is only _one_ `<bitstream>` element allowed per representation.
 - The `<bitstream>` element must be the first element.
+- If a `<bitstream>` element is present, the `<iiif-uri>` element must be omitted.
```
jnussbaum, line 227: There is no `<iiif-uri>` element in the first place, so it cannot be omitted. 

```suggestion
- If a resource has a `<bitstream>` element, it cannot have a `<iiif-uri>` element.
```

jnussbaum (approved):

jnussbaum:
I think you should also document the new CLI flag `--no-iiif-uri-validation`, and explain what it does / when to use it.
This should best be done in this file: `docs/cli-commands.md`

docs/file-formats/xml-data-file.md
jnussbaum: Sorry, my fault. Sure, this is correct, you can leave it as it is 👍

PR #974, opened 2024-05-21
chore(excel2json): stop crashing if optional columns are missing (DEV-3516)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/properties.py
jnussbaum, line 1: 2 comments regarding the PR title:

- perhaps "fix" would be the better category than "chore"
- instead of "stop crashing" --> don't crash
- only the resources are affected --> this should be reflected in the PR title

jnussbaum (approved):
LGTM

PR #978, opened 2024-05-24
test: set up testcontainers for improved test setup (DEV-3686)

BalduinLandolt:


test/e2e/conftest.py
BalduinLandolt, line 1: not sure if that is a problem we'll ever really encounter. But I'm always sceptical about something that automatically deletes stuff in the background. I tend to think "what if there is a scenario where I want to keep this there while I work on it?"  
So I would generally rather opt for such clean-up to be manual tasks (maybe have a make/just/proetry command to automate it). But I don't feel strongly about it.

In the long run, we may consider having certain assets there (but they could be copied over in the test startup) for tests that expect there to be assets available. But we can cross this bridge when we get there.

test/e2e/test_create_and_xmlupload.py
```diff
@@ -0,0 +1,129 @@
...
+ONTO_IRI = f"{CREDS.server}/ontology/{PROJECT_SHORTCODE}/{ONTO_NAME}/v2"
+PROPS_IN_ONTO_JSON = 1
+RESCLASSES_IN_ONTO_JSON = 2
+
+
+@pytest.fixture()
+def _create_project() -> Iterator[None]:
+    with get_containers():
+        success = create_project(Path("testdata/json-project/test-project-e2e.json"), CREDS, verbose=True)
+        assert success
```
BalduinLandolt, line 26: to me it feels a bit strange to have assertions in fixtures

test/e2e/test_create_and_xmlupload.py
```diff
@@ -0,0 +1,129 @@
...
+    _analyze_pdf_resources(pdf_resources)
+
+
+def _get_resources(resclass_iri: str, auth_header: dict[str, str], project_iri: str) -> list[dict[str, Any]]:
+    resclass_iri_encoded = urllib.parse.quote_plus(resclass_iri)
+    get_resources_route = f"{CREDS.server}/v2/resources?resourceClass={resclass_iri_encoded}&page=0"
+    headers = auth_header | {"X-Knora-Accept-Project": project_iri}
+    response = requests.get(get_resources_route, timeout=3, headers=headers).json()
+    resources: list[dict[str, Any]] = response.get("@graph", [response])
+    return resources
```
BalduinLandolt, line 109: just a detail: to me it seems a bit strange to have helper methods both before and after the test case(s). I would normally expect these "categories" of code to be grouped separately.

test/e2e/conftest.py
jnussbaum: thanks for the input. After thinking about it, I prefer leaving it as it is.

test/e2e/test_create_and_xmlupload.py
jnussbaum: That's indeed particular. For me, it is a way to test our `success` variables that we pass around in the code. A very superficial black-box test, but still.

test/e2e/test_create_and_xmlupload.py
jnussbaum: Good point, I rearranged them.

Nora-Olivia-Ammann (approved):
To really do a complete analysis I would have to do a deep dive. If both you and @BalduinLandolt are confident, I am ok with it. Otherwise I need more time.

PR #979, opened 2024-05-28
test: reorganize tests (preparation for testcontainers) (DEV-3685)

jnussbaum:


.github/actions/setup-from-wheel/action.yml
```diff
@@ -18,24 +18,24 @@ runs:
...
+        echo ########### STEP: CHECK WHEEL IN DEDICATED VENV ###########
         .dist-test-venv/bin/check-wheel-contents dist/
       shell: bash
-    - name: Install pytest
+    - name: Install pytest globally
       run: |
-        echo ########### STEP: INSTALL PYTEST ###########
-        .dist-test-venv/bin/pip install pytest pytest-unordered
+        echo ########### STEP: INSTALL PYTEST GLOBALLY ###########
+        pip install pytest pytest-unordered
```
jnussbaum, line 40: Originally, the venv that was intended for the `check-wheel-contents` was reused for installing pytest and running the distribution tests. But now I better like the idea to install pytest globally, and then run the distribution tests globally.

Because the intention of the distribution tests is to see if DSP-TOOLS has been installed correctly (globally). So I wonder how it worked until now, that the distribution tests were run from this venv, where dsp-tools and its dependencies were not installed...

.github/workflows/tests-on-push.yml
```diff
@@ -78,7 +78,7 @@ jobs:
         run: poetry run pytest -s test/benchmarking/
 
 
-  e2e-tests:
+  legacy-e2e-tests:
```
jnussbaum, line 81: By renaming the e2e into legacy-e2e, there is room for the new e2e tests with testcontainers

test/e2e/commands/xmlupload/test_list_client_live.py
jnussbaum, line 1: After comparing this to https://github.com/dasch-swiss/dsp-tools/blob/9473a7ef6448f25f25143c17c6717b6470cff0f5/test/integration/commands/xmlupload/test_list_client_live.py, I think this e2e test is not really necessary and should be deleted

test/e2e/commands/xmlupload/test_project_client_live.py
jnussbaum, line 1: After comparing this to https://github.com/dasch-swiss/dsp-tools/blob/9473a7ef6448f25f25143c17c6717b6470cff0f5/test/integration/commands/xmlupload/test_project_client_live.py, I think this e2e test is not really necessary and should be deleted

BalduinLandolt (approved):
Nice!

PR #981, opened 2024-05-31
chore(start-stack): remove network from docker-compose.yml

jnussbaum:
As per https://docs.docker.com/compose/networking/, `docker compose` sets up a default network, and each container joins that default network.

### DIFF
--- a/src/dsp_tools/resources/start-stack/docker-compose.yml
+++ b/src/dsp_tools/resources/start-stack/docker-compose.yml
@@ -9,8 +9,6 @@ services:
       - db
     ports:
       - "3333:3333"
-    networks:
-      - knora-net
     environment:
       - TZ=Europe/Zurich
       - KNORA_AKKA_LOGLEVEL=DEBUG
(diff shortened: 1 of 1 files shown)


Nora-Olivia-Ammann (commented):
From what is written on the webpage, the new network is the same name as the directory in which the file lives. Does it have any consequences that the network is now called by a different name?

jnussbaum:
> From what is written on the webpage, the new network is the same name as the directory in which the file lives. Does it have any consequences that the network is now called by a different name?

AFAIU, the containers just need to be in the same network, so that they can communicate with each other. But apart from that, the specifications of the network are not of importance. Apparently, the containers can communicate well, otherwise the `dsp-tools create` and `dsp-tools xmlupload` would not work. 

P.S: And right now I quickly checked: DSP-APP is also working.

Nora-Olivia-Ammann (approved):

BalduinLandolt (approved):

PR #995, opened 2024-06-06
fix(xmlupload): OntologyConstraintException should not interrupt an xmlupload (DEV-3724)

jnussbaum:


.github/dependabot.yml
```diff
@@ -3,9 +3,6 @@
 # Use Dependabot to automatically keep dependencies updated to the latest version, even when they don’t have any known vulnerabilities
 # see https://docs.github.com/en/code-security/dependabot/dependabot-version-updates
 
-env:
-  CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-
```
jnussbaum, line 8: This was a misunderstanding on my part. The codecov docs (https://github.com/codecov/codecov-action?tab=readme-ov-file#dependabot) say that I shall add the token to the dependabot secrets. But apparently, this is not the way to do it. Because I get an complaint: https://github.com/dasch-swiss/dsp-tools/pull/995/checks?check_run_id=25891093755.

Rather, I should have added the token not only as a repository secret (https://github.com/dasch-swiss/dsp-tools/settings/secrets/actions), but also as a dependabot secret (https://github.com/dasch-swiss/dsp-tools/settings/secrets/dependabot). 

But meanwhile, I wonder if it is important that the Dependabot PRs upload a coverage report. I think it's actually not necessary, so I will leave it now.

src/dsp_tools/utils/connection_live.py
```diff
@@ -298,6 +304,9 @@ def _handle_non_ok_responses(self, response: Response, request_url: str, retry_c
             msg = "Permanently unable to execute the network action. "
             if original_str := regex.search(r'{"knora-api:error":"dsp\.errors\.(.*)","@context', str(response.content)):
                 msg += f"\n{' '*37}Original Message: {original_str.group(1)}\n"
+                msg += f"See logs for more details: {logger_savepath}"
```
gNahcab, line 307: In the case where the following if-statement is false, this line is executed again on line 310 or am I mistaken here? And if so is this a desired behavior?

jnussbaum: Good catch, thank you. You're right, that's not intended.

src/dsp_tools/utils/connection_live.py
```diff
@@ -298,6 +304,9 @@ def _handle_non_ok_responses(self, response: Response, request_url: str, retry_c
             msg = "Permanently unable to execute the network action. "
             if original_str := regex.search(r'{"knora-api:error":"dsp\.errors\.(.*)","@context', str(response.content)):
                 msg += f"\n{' '*37}Original Message: {original_str.group(1)}\n"
+                if original_str.group(1).startswith("OntologyConstraintException"):
+                    msg += f"See logs for more details: {logger_savepath}"
+                    raise InvalidInputError(msg)
```
BalduinLandolt, line 309: It's ok for this one, but generally I think we should work towards having structured information in representations of error state, not just "prose" strings describing the error.

BalduinLandolt (approved):

PR #997, opened 2024-06-06
chore(start-stack): don't print that stack is ready before it is ready

jnussbaum:


gNahcab (approved):

.github/dependabot.yml
```diff
@@ -3,9 +3,6 @@
 # Use Dependabot to automatically keep dependencies updated to the latest version, even when they don’t have any known vulnerabilities
 # see https://docs.github.com/en/code-security/dependabot/dependabot-version-updates
 
-env:
-  CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-
```
jnussbaum, line 8: This was a misunderstanding on my part. The codecov docs (https://github.com/codecov/codecov-action?tab=readme-ov-file#dependabot) say that I shall add the token to the dependabot secrets. But apparently, this is not the way to do it. Because I get an complaint: https://github.com/dasch-swiss/dsp-tools/pull/995/checks?check_run_id=25891093755.

Rather, I should have added the token not only as a repository secret (https://github.com/dasch-swiss/dsp-tools/settings/secrets/actions), but also as a dependabot secret (https://github.com/dasch-swiss/dsp-tools/settings/secrets/dependabot). 

But meanwhile, I wonder if it is important that the Dependabot PRs upload a coverage report. I think it's actually not necessary, so I will leave it now.

PR #1001, opened 2024-06-10
feat: add flag to suppress warning when using old dsp-tools version (DEV-3707)

jnussbaum:
Add a new flag, --suppress-update-prompt, to the command-line interface. When this flag is set, the version check will be skipped, which is useful in contexts without an interactive shell (e.g. when Terminal output is piped to file, it is not possible to answer the prompt).


gNahcab (approved):

docs/cli-commands.md
```diff
@@ -61,6 +61,8 @@ The following options are available:
 - `-V` | `--validate-only` (optional): validate the JSON file without creating it on the DSP server
 - `-l` | `--lists-only` (optional): create only the lists (prerequisite: the project exists on the server)
 - `-v` | `--verbose` (optional): print more information about the progress to the console
+- `--suppress-update-prompt` (optional): don't prompt when using an outdated version of DSP-TOOLS
```
BalduinLandolt, line 64: why not just `-y | --yes` ?

jnussbaum: I think that would not be specific enough. What would "yes" refer to? (I know, it refers to the update prompt, but that's not obvious).

BalduinLandolt (approved):
In terms of implementation it looks nice and clean. In terms of the user experience, I would have opted for a more concise flag... 3 words IMO is 2 too many.  
(And all the CLIs I know, that do this either have `-y | --yes | --assume-yes` (apt-get, etc.) or `-f | --force` (git, etc.) so even by convention something shorter would make sense.)

docs/cli-commands.md
jnussbaum: Good point. I just searched the internet, and came to the same conclusion than you: `-f` and `-y` seem to be the most widespread. 

There is just one hindrance: We already have 2 places in the code where the user is asked for yes/no:

 - https://github.com/dasch-swiss/dsp-tools/blob/569c161cf2c286765734e7f690212d160a2058ac/src/dsp_tools/commands/start_stack.py#L333
 - https://github.com/dasch-swiss/dsp-tools/blob/569c161cf2c286765734e7f690212d160a2058ac/src/dsp_tools/commands/resume_xmlupload/resume_xmlupload.py#L76

So if we use `-f` or `-y`, this should also override the above 2, for consistency reasons. But I'm unsure if we want that. 

So we have these options:

- use `-f` or `-y`, but leave the existing prompts untouched 
    - pro: that's the behaviour we want
    - con: inconsistent
- use `-f` or `-y` and override the existing prompts
    - pro: consistent
    - con: we don't want this behaviour
- use `--suppress-update-prompt`
    - pro: that's the behaviour we want
    - con: inconsistent

What's your opinion, @gNahcab and @BalduinLandolt ?

BalduinLandolt: arguably, it might be nice to allow for bypassing the other prompts too. But admittedly, it's not what this task is about...

But at the end of the day, I think all three options would be ok

PR #1002, opened 2024-06-11
fix(ci): use GITHUB_TOKEN instead of GH_TOKEN, move CHANGELOG.md back (DEV-3730)

jnussbaum:


.github/workflows/release-please.yml
```diff
@@ -17,6 +17,6 @@ jobs:
     steps:
       - uses: googleapis/release-please-action@v4
         with:
-          token: ${{ secrets.GH_TOKEN }}
+          token: ${{ secrets.GITHUB_TOKEN }}
```
jnussbaum, line 20: This was the culprit. There is a fundamental difference between these 2:

- `GH_TOKEN` is an organisation-wide secret that was once stored in github.com/dasch-swiss. Its permissions cannot be changed by lines 10-12.
- `GITHUB_TOKEN` is created automatically by GitHub (https://docs.github.com/en/actions/security-guides/automatic-token-authentication). Its permissions are modified by lines 10-12.

The problem: All the time, lines 10-12 did not modify the permissions of `GH_TOKEN`, because that's not possible. Therefore, the release-please always failed.

gNahcab (approved):

PR #1009, opened 2024-06-12
chore: add yamlfmt to pre-commit

jnussbaum:


gNahcab (approved):

BalduinLandolt (approved):
I personally hate pre-commit hooks and would opt for removing them entirely. But if it works for you, sure

jnussbaum:
> I personally hate pre-commit hooks and would opt for removing them entirely. But if it works for you, sure

Probably we have different workflows, because in my workflows, pre-commit hooks are the best invention ever. It would be interesting to discuss our preferences while having a cup of coffee :-)

BalduinLandolt:
> > I personally hate pre-commit hooks and would opt for removing them entirely. But if it works for you, sure
> 
> Probably we have different workflows, because in my workflows, pre-commit hooks are the best invention ever. It would be interesting to discuss our preferences while having a cup of coffee :-)

sure! ☕

PR #1021, opened 2024-06-14
feat: new command `ingest-files` to monitor mass ingest on a server (DEV-3743)

jnussbaum:


src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
```diff
@@ -82,3 +84,32 @@ def upload_file(
...
+    def retrieve_mapping(self) -> str | None:
+        """Try to retrieve the mapping CSV from the server."""
+        url = f"{self.dsp_ingest_url}/projects/{self.shortcode}/bulk-ingest/mapping.csv"
+        res = self.session.get(url, timeout=5)
+        if res.status_code == STATUS_CONFLICT:
+            print("Ingest process is still running. Wait until it completes...")
+            logger.info("Ingest process is still running. Wait until it completes...")
+            return None
+        elif not res.ok or not res.text.startswith("original,derivative"):
+            print("Dubious error")
```
gNahcab, line 110: If I read "Dubious error" I wouldn't have a clue what just happened, could the error message be more specific?

jnussbaum: better now?

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
```diff
@@ -82,3 +84,36 @@ def upload_file(
...
+    def kick_off_ingest(self) -> None:
+        """Start the ingest process on the server."""
+        url = f"{self.dsp_ingest_url}/projects/{self.shortcode}/bulk-ingest"
+        res = self.session.post(url, timeout=5)
+        if res.status_code == STATUS_CONFLICT:
+            msg = f"Ingest process on the server {self.dsp_ingest_url} is already running. Wait until it completes..."
+            print(msg)
+            logger.info(msg)
+        if res.json().get("id") != self.shortcode:
+            raise UserError("Failed to kick off the ingest process.")
```
Nora-Olivia-Ammann, line 97: Why not InputError? From what I remember we use InputError when we create a new error in the code.

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
```diff
@@ -82,3 +84,36 @@ def upload_file(
...
+        """Try to retrieve the mapping CSV from the server."""
+        url = f"{self.dsp_ingest_url}/projects/{self.shortcode}/bulk-ingest/mapping.csv"
+        res = self.session.get(url, timeout=5)
+        if res.status_code == STATUS_CONFLICT:
+            print("Ingest process is still running. Wait until it completes...")
+            logger.info("Ingest process is still running. Wait until it completes...")
+            return None
+        elif not res.ok or not res.text.startswith("original,derivative"):
+            msg = (
+                "Dubious error while polling for the mapping CSV. "
```
Nora-Olivia-Ammann, line 111: What action makes it dubious? User behaviour or ingest behaviour?

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
```diff
@@ -82,3 +84,36 @@ def upload_file(
...
+                f"If this happens again at the next polling, please check the logs at {logger_savepath}."
```
Nora-Olivia-Ammann, line 112: What do you mean with "polling"

src/dsp_tools/commands/ingest_xmlupload/ingest_files/ingest_files.py
```diff
@@ -0,0 +1,44 @@
...
+        success status
+    """
+    con: Connection = ConnectionLive(creds.server)
+    con.login(creds.user, creds.password)
+    bulk_ingest_client = BulkIngestClient(creds.dsp_ingest_url, con.get_token(), shortcode)
+    bulk_ingest_client.kick_off_ingest()
+    sleep(5)
+
+    while not (mapping := bulk_ingest_client.retrieve_mapping()):
+        sleep(10)
```
Nora-Olivia-Ammann, line 32: Assuming that the bulk ingest is done with a lot of files should we maybe make this longer, so that we don't spam the server?

test/e2e/commands/ingest_xmlupload/test_ingest_xmlupload.py
```diff
@@ -1,15 +1,26 @@
...
 import pytest
+from pytest_unordered import unordered
 
 from dsp_tools.cli.args import ServerCredentials
+from dsp_tools.commands.ingest_xmlupload.ingest_files.ingest_files import ingest_files
 from dsp_tools.commands.ingest_xmlupload.upload_files.upload_files import upload_files
 from dsp_tools.commands.project.create.project_create import create_project
 from test.e2e.setup_testcontainers import get_containers
 
 CREDS = ServerCredentials("root@example.com", "test", "http://0.0.0.0:3333")
```
Nora-Olivia-Ammann, line 14: Why not a fixture?

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
```diff
@@ -82,3 +84,36 @@ def upload_file(
         print(f"{datetime.now()}: {msg}")
         logger.info(msg)
         return None
+
+    def kick_off_ingest(self) -> None:
```
BalduinLandolt, line 88: nitpick: not that it matters, but maybe `start_ingest` or `trigger_ingest_process` would be less metaphorical?

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
BalduinLandolt, line 97: I'm not sure under what scenario this would happen, but is it really a User/Input problem when the response from the server does not contain the ID provided?

But am I seeing it right that you do not check for any other status code (404, 500, etc.)? What happens if you kick off an ingest of an unknown shortcode? shouldn't that result in 404? And if so, I think `.json()` will not give a user-friendly message...

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
BalduinLandolt, line 111: Is it even meaningful to do this check that it starts with `original,derivative`? can it ever start with anything else?

Also, I don't like the word dubious... maybe "unexpected"? however, I think 404 might be something we get at times?

I think I would split this test up into two distinct cases:
1. `if not res.ok:` some error message indicating that the server did not return a mapping
2. `if not res.text:` some error indicating that the server returned an empty mapping

src/dsp_tools/commands/ingest_xmlupload/ingest_files/ingest_files.py
BalduinLandolt: I think 10s is easily justifiable, refusing the mapping is a very cheap operation on the server

src/dsp_tools/commands/ingest_xmlupload/ingest_files/ingest_files.py
```diff
@@ -0,0 +1,44 @@
...
+    Returns:
+        success status
+    """
+    con: Connection = ConnectionLive(creds.server)
+    con.login(creds.user, creds.password)
+    bulk_ingest_client = BulkIngestClient(creds.dsp_ingest_url, con.get_token(), shortcode)
+    bulk_ingest_client.kick_off_ingest()
+    sleep(5)
+
+    while not (mapping := bulk_ingest_client.retrieve_mapping()):
```
BalduinLandolt, line 31: you could use tqdm to indicate that it's still working, how often it has already polled and how quickly it does so - obviously not with a real progress bar (so it won't look as good), but it might be better than just no output at all.  
For how to do it, see https://stackoverflow.com/a/67249498/11914562 or https://stackoverflow.com/a/73319148/11914562

BalduinLandolt (approved):
I think the error handling is not yet bullet proof, but other than that it looks good

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
jnussbaum: If I compare the docstring of `UserError`, it seems more appropriate than `InputError`. And also the names of these two suggest that this here is rather a UserError. 

I don't think that every new error that we raise in the code should be an InputError. I rather think that we should tidy up our exception hierarchy in `src/dsp_tools/models/exceptions.py`, that's a big mess, and it lacks a clear concept. I have added it to our Topics To Discuss list

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
jnussbaum: https://en.wikipedia.org/wiki/Polling_(computer_science).
But perhaps I should choose another word. It's probably not well understood.

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
jnussbaum: ingest behaviour

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
jnussbaum: Concerning the UserError: The underlying problem is that our exception hierarchy lacks a clear concept. This point is already noted on Notion, and I have added it to the Topics to discuss: https://www.notion.so/dasch-swiss/New-error-hierarchy-45fe23b41f254555843d7228552ac18a.

Concerning the other cases: Thanks for pointing it out. I have now handled all conceivable cases (https://github.com/dasch-swiss/dsp-tools/commit/97cea18b4600d6865731ac44f84ea26492bf21bf), and written unit tests for them (https://github.com/dasch-swiss/dsp-tools/commit/b937331a432542c820f2ea34633cd8ef8730dabd).

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
jnussbaum: > Is it even meaningful to do this check that it starts with original,derivative? can it ever start with anything else?
It should never start with anything else, but I think it is meaningful to check even for events that should not happen.

> Also, I don't like the word dubious
I adapted the wording.

> I think I would split this test up into two distinct cases
I only see 3 major cases:
- 409 Conflict: Do nothing, wait for next poll.
- non-200, or content of response is not a mapping: inform the user, wait for next poll. (Perhaps, the server recovers)
- 200: return the mapping

src/dsp_tools/commands/ingest_xmlupload/ingest_files/ingest_files.py
jnussbaum: > but it might be better than just no output at all

I can look at you 2 links after my holidays, and if possible, I will gladly implement something with tqdm. But there is already a print output now.

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
BalduinLandolt: in that case I would only check these 3 cases. For the "non-200" I would build in some kind of maximum number of attempts because if the Server doesn't recover, it should not wait forever.

siers (approved):

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
```diff
@@ -70,3 +79,52 @@ def upload_file(
...
+        if res.status_code in [STATUS_INTERNAL_SERVER_ERROR, STATUS_SERVER_UNAVAILABLE]:
+            raise UserError("Server is unavailable. Please try again later.")
+        try:
+            returned_shortcode = res.json().get("id")
+            failed: bool = returned_shortcode != self.shortcode
+        except JSONDecodeError:
+            failed = True
+        if failed:
+            raise UserError("Failed to trigger the ingest process. Please check the server logs, or try again later.")
+        print(f"Kicked off the ingest process on the server {self.dsp_ingest_url}. Wait until it completes...")
```
Nora-Olivia-Ammann, line 108: This is cosmetic, but could you put this string in a variable since it is identical with the one in the line below?

Nora-Olivia-Ammann (approved):

src/dsp_tools/cli/create_parsers.py
```diff
@@ -236,6 +237,23 @@ def _add_upload_files(
...
+    root_user_pw: str,
+) -> None:
+    subparser = subparsers.add_parser(
+        name="ingest-files",
+        help="Experimental: ingest uploaded files on a DSP server and retrieve the mapping CSV file",
+    )
+    subparser.set_defaults(action="ingest-files")
+    subparser.add_argument("-s", "--server", default=default_dsp_api_url, help=dsp_server_text)
+    subparser.add_argument("-u", "--user", default=root_user_email, help=username_text)
+    subparser.add_argument("-p", "--password", default=root_user_pw, help=password_text)
```
seakayone, line 253: *thought [security]*: Passing a password as argument is not safe

Providing credentials as a command line argument is convenient but is considered to be security risk. This is because the arguments end up in the history of a shell, which is generally a bad design.

Other ways of providing the credentials might be:
* Request it to from the user in interactive mode
* Use a dot file for configuration
* Environment variables

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
```diff
@@ -70,3 +79,52 @@ def upload_file(
...
+        res = self.session.post(url, timeout=5)
+        if res.status_code in [STATUS_UNAUTHORIZED, STATUS_FORBIDDEN]:
+            raise BadCredentialsError("Unauthorized to start the ingest process. Please check your credentials.")
+        if res.status_code == STATUS_NOT_FOUND:
+            raise UserError(
+                f"No assets have been uploaded for project {self.shortcode}. "
+                "Before using the 'ingest-files' command, you must upload some files with the 'upload-files' command."
+            )
+        if res.status_code == STATUS_CONFLICT:
+            msg = f"Ingest process on the server {self.dsp_ingest_url} is already running. Wait until it completes..."
```
seakayone, line 95: The function returns exactly as in the case when a new bulk-ingest was started. Does it make sense to ask the user to `[w]ait until it completes`?

seakayone (approved):

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
jnussbaum: > The function returns exactly as in the case when a new bulk-ingest was started. 

Yes, because it doesn't matter. As long as the mapping CSV is not ready, we have to wait.

> Does it make sense to ask the user to [w]ait until it completes?

I designed this command to wait until the mapping has been retrieved. It is documented that this might take very long: https://github.com/dasch-swiss/dsp-tools/blob/39898e83c7c4ac3ecb9541698bf6dd7cdf902ab1/docs/cli-commands.md#L271

In the meanwhile, the user sees a print output every 10 seconds: "Ingest process is still running. Wait until it completes..."

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
jnussbaum: My habit is to do it this way as long as the string fits 1 line. Because then, I need 2 LOC over all (as opposed to 3 LOC if I put it into a variable). 

As soon as the message is 2 lines long, I put it into a variable, which results in less LOC.

src/dsp_tools/commands/ingest_xmlupload/ingest_files/ingest_files.py
jnussbaum: Done with [d86d04f51f9328f2fe0c03f5c51925047a8f1725](https://github.com/dasch-swiss/dsp-tools/commit/d86d04f51f9328f2fe0c03f5c51925047a8f1725)

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
jnussbaum: Done with https://github.com/dasch-swiss/dsp-tools/commit/3285fbfe175e3fa3d3b1fd25c40347ba763fbed2

PR #1022, opened 2024-06-17
feat: display progress in xmlupload as progress bar (DEV-3708)

BalduinLandolt:


src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -254,15 +255,17 @@ def _upload_resources(clients: UploadClients, upload_state: UploadState) -> None
...
-        for creation_attempts_of_this_round, resource in enumerate(upload_state.pending_resources.copy()):
+        for creation_attempts_of_this_round, resource in enumerate(progress_bar):
             _upload_one_resource(
                 upload_state=upload_state,
                 resource=resource,
                 ingest_client=clients.asset_client,
                 resource_create_client=resource_create_client,
                 creation_attempts_of_this_round=creation_attempts_of_this_round,
             )
+            progress_bar.set_description(f"Creating Resources (failed: {len(upload_state.failed_uploads)})")
```
Nora-Olivia-Ammann, line 268: After how much time is this displayed?
The user may want to stop the upload as soon as one image fails...

Nora-Olivia-Ammann (approved):
I didn't try it out because I didn't want to destroy my prod db. But looks good.

src/dsp_tools/commands/xmlupload/xmlupload.py
BalduinLandolt: this is updated each loop, so ideally it will always display `Creating Resources (failed: 0)`, but if something goes wrong it should update pretty much immediately

BalduinLandolt: in case you're wondering how it looks:

image

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -254,15 +255,17 @@ def _upload_resources(clients: UploadClients, upload_state: UploadState) -> None
         media_previously_ingested=upload_state.config.media_previously_uploaded,
     )
 
+    progress_bar = tqdm(upload_state.pending_resources.copy(), leave=False)
```
jnussbaum, line 258: I would vote for leaving the bar after termination. Otherwise, the Terminal history is less understandable, because you see nowhere that the resources have been created. 

Also if an error happens, it might be better understandable if the progress bar is still visible, like so:

```
python -m dsp_tools.cli.entry_point xmlupload testdata/xml-data/test-data-systematic.xml 
You are using DSP-TOOLS version 8.1.0, but version 8.1.0.post18 is available. Consider upgrading via 'pip3 install --upgrade dsp-tools'.

Using DSP server 'http://0.0.0.0:3333' and ingest server 'http://0.0.0.0:3340'
2024-06-18 09:18:56.353154: The XML file is syntactically correct and passed validation.
2024-06-18 09:18:56.923257: Checking resources for circular references...
2024-06-18 09:18:56.982776: Stashing circular references...
Creating Resources (failed: 0):  70%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████                                                      | 40/57 [00:06<00:02,  5.72it/s]

==========================================
Lost connection to DSP server, probably because the server is down. Please continue later with 'resume-xmlupload'. Reason for this failure: Permanently unable to execute the network action. 
                                     Original Message: DuplicateValueException: IRI: \'http://rdfh.ch/4123/5d5d1FKaUC2Wfl4zicggfg\' already exists, try another one.
See logs for more details: /Users/nussbaum/.dsp-tools/logging.log
Saved the current upload state to /Users/nussbaum/.dsp-tools/xmluploads/localhost/resumable/latest.pkl.
```

jnussbaum (approved):

src/dsp_tools/commands/xmlupload/xmlupload.py
BalduinLandolt: Good point. Should we leave the bar, or should I add a print statement indicating that the resources were uploaded?

jnussbaum: I would just leave the bar.

BalduinLandolt: done (https://github.com/dasch-swiss/dsp-tools/pull/1022/commits/f236e3dfe36685727aa0c5ca41253cce0e1eff23) plus added an initial description for consistency

gNahcab (approved):

PR #1025, opened 2024-06-19
chore: write tests for xmlupload step / add logging of requests (DEV-3781)

jnussbaum:


src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
```diff
@@ -113,3 +113,14 @@ def retrieve_mapping(self) -> str | None:
         print("Ingest process completed.")
         logger.info("Ingest process completed.")
         return res.text
+
+    def finalize(self) -> bool:
+        """Delete the mapping file and the temporary directory where the unprocessed files were stored."""
+        route = f"{self.dsp_ingest_url}/projects/{self.shortcode}/bulk-ingest/finalize"
```
Nora-Olivia-Ammann, line 119: Could you state the purpose of this?

src/dsp_tools/commands/ingest_xmlupload/create_resources/upload_xml.py
```diff
@@ -67,15 +69,21 @@ def ingest_xmlupload(
...
+        success = False
 
-def _parse_xml(xml_file: Path) -> tuple[str, etree._Element, str]:
+    return success
+
+
+def _parse_xml_and_replace_filepaths(xml_file: Path) -> tuple[str, etree._Element, str]:
     """
-    Validate and parse an upload XML file, when preprocessing has already been done.
+    Validate and parse an upload XML file, when ingest has already been done.
```
Nora-Olivia-Ammann, line 83: Maybe say that ingest is a prerequisite. Otherwise I could be lead to believe that it is an automated step that occurs before.

test/e2e/commands/ingest_xmlupload/test_ingest_xmlupload.py
```diff
@@ -59,3 +62,24 @@ def _test_ingest_step(caplog: pytest.LogCaptureFixture, mapping_file: Path) -> N
...
+            "All multimedia files referenced in the XML file were uploaded through dsp-ingest.\n"
+            "All multimedia files uploaded through dsp-ingest were referenced in the XML file."
+        ),
+    ]
+    assert all(expected_log in logs for expected_log in expected_logs)
+    expected_log_regexes = [
+        "Created resource 1/3",
+        "Created resource 2/3",
+        "Created resource 3/3",
+    ]
```
BalduinLandolt, line 83: I'm not a fan of testing against logs.  
If you feel it really helps, then I would suggest making it a very flexible test, something like `assert "mapping-4125.csv" in x`.  
But as it is currently, I fear this is more hindering than helpful.

BalduinLandolt (approved):

Nora-Olivia-Ammann (approved):

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
jnussbaum: This is to keep the server tidied up: https://ingest.dev.dasch.swiss/docs/#/bulk-ingest/postProjectsShortcodeBulk-ingestFinalize

seakayone (commented):
I am not sure that I am really fit to review this PR because this change is based on a wip branch which is nearly a month old and contains many other changes which I am not aware of.

For example the wip branch introduced the `BulkIngestClient`:
Now we have `BulkIngestClient` and `DspIngestClientLive(AssetClient)` which both hold the `dsp_ingest_url` and access the ingest api for different reasons. When I introduced the `AssetClient` the idea was to make this the only component accessing the dsp-ingest rest api. I wanted it to remain low level and build other services on top of it.

Maybe it makes sense to have two different rest clients in our code base. But having these also in very different places makes it hard for developers to find theses components:
```
src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
src/dsp_tools/commands/xmlupload/models/ingest.py
```

Other than that I am not entirely familiar with the existing code base/workflows and would not be able to assess if the code is added at the right places.

src/dsp_tools/commands/ingest_xmlupload/create_resources/upload_xml.py
```diff
@@ -67,15 +69,21 @@ def ingest_xmlupload(
     clients = _get_live_clients(con, config)
     state = UploadState(resources, stash, config, permissions_lookup)
 
-    return execute_upload(clients, state)
+    if not execute_upload(clients, state):
+        success = False
 
+    if not _tidy_up_on_server(con, shortcode, creds.dsp_ingest_url):
+        success = False
```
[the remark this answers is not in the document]
jnussbaum: If at a later point of time, someone wants to purge & recreate all resources, but doesn't have the mapping CSV any more. Then it will be very handy if it is still available on the server.

Until now, I only rarely called the clean up route, because I thought "Better keep it than delete it too early". 

But there is an important argument for including it here: If we make it a separate command, no one will ever use it. And then, the servers will become cluttered after a while. So we have to decide which evil is worse:

1. Cluttering the server with old CSVs. If a new ingest for the same project is triggered, new entries would be added to the existing file (?), which would lead to confusion.
2. Deleting the CSV every time, with the danger of permanently losing the CSV if a user accidentally deletes it on their machine.

While writing this, I wonder if in the backend, you should implement an archive for old CSVs? This would contain a history of every bulk-ingest that has ever taken place on that particular server. WDYT, @seakayone @BalduinLandolt @siers ?

src/dsp_tools/commands/ingest_xmlupload/create_resources/upload_xml.py
jnussbaum: Done with https://github.com/dasch-swiss/dsp-tools/commit/fb080db2e118ba1e1b66cd18cbcc8ed55222bcae

test/e2e/commands/ingest_xmlupload/test_ingest_xmlupload.py
jnussbaum: Good point. Now I don't test against logs any more. Rather, I test if the id2iri mapping file contains the correct resource IDs and IRIs: https://github.com/dasch-swiss/dsp-tools/commit/48d65acee20eb1b81dacfbe76f5637b9da0cfc4e

src/dsp_tools/commands/ingest_xmlupload/create_resources/upload_xml.py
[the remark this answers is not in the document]
BalduinLandolt: I would always pick clutter over data loss. But I'm not sure either will be a huge problem in reality, that I find very hard to judge. (How often do we actually do multiple xmluploads for the same project? does it hurt to have all asset mappings in the same file, or is it actually a good thing? How often will it happen that someone wants to purge and recreate a project _and_ didn't save the mapping file? ...)

In the long run we definitely should have a clear trace of everything that happened, including bulk ingests. But I don't think this falls under the scope of this project. So here we just need to make a call on calling cleanup: yes/no.

jnussbaum:
> I am not sure that I am really fit to review this PR because this change is based on a wip branch which is nearly a month old and contains many other changes which I am not aware of.
> 
> For example the wip branch introduced the `BulkIngestClient`: Now we have `BulkIngestClient` and `DspIngestClientLive(AssetClient)` which both hold the `dsp_ingest_url` and access the ingest api for different reasons. When I introduced the `AssetClient` the idea was to make this the only component accessing the dsp-ingest rest api. I wanted it to remain low level and build other services on top of it.
> 
> Maybe it makes sense to have two different rest clients in our code base. But having these also in very different places makes it hard for developers to find theses components:
> 
> ```
> src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
> src/dsp_tools/commands/xmlupload/models/ingest.py
> ```
> 
> Other than that I am not entirely familiar with the existing code base/workflows and would not be able to assess if the code is added at the right places.

Thanks for you comment. After considering it, I came to the conclusion that these clients differ in some major aspects.

The "old" client belonging to the xmlupload:
- no tracking of the entire filepath, since the internal filename is returned immediately
- `get_bitstream_info()` method, that uploads & ingests 1 single file, and then immediately returns the internal filename
- only 1 HTTP interaction (namely uploading & ingesting 1 single file)

My new client belonging to the ingest-xmlupload:
- entire filepath must be tracked, so that the final mapping contains the entire path
- no `get_bitstream_info()` method (doesn't fit into this scenario)
- a row of 4 subsequent HTTP interactions, that belong together:
    1. upload 1 file after the other
    2. trigger the ingest
    3. poll for the mapping
    4. finalize

Regarding the place where they live: They live in the directory of the command where they belong to:
- `src/dsp_tools/commands/xmlupload/models/ingest.py` belongs to the `xmlupload` command
- `src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py` belongs to the `ingest-xmlupload` command

I don't see a possibility to merge these two, but I'm of course open for suggestions.

src/dsp_tools/commands/ingest_xmlupload/create_resources/upload_xml.py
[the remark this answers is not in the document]
jnussbaum: > I would always pick clutter over data loss.

Me too. So I'd prefer not calling `/projects/{shortcode}/bulk-ingest/finalize` at all. Would you be okay with that, @seakayone ? It's you who had the effort of implementing it, so it might be disappointing that it will never be used.

PR #1028, opened 2024-06-20
fix: harmonize supported file extensions in docs, test data, and prod code (DEV-3788)

jnussbaum:


src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -234,7 +234,7 @@ def _make_bitstream_file_value(bitstream_info: BitstreamInfo) -> dict[str, Any]:
         case "jpg" | "jpeg" | "jp2" | "png" | "tif" | "tiff" | "jpx":
             prop = "knora-api:hasStillImageFileValue"
             value_type = "StillImageFileValue"
-        case "odd" | "rng" | "txt" | "xml" | "xsd" | "xsl" | "xslt" | "csv":
+        case "odd" | "rng" | "txt" | "xml" | "xsd" | "xsl" | "csv":
```
BalduinLandolt, line 237: if the API allows `xslt`, then we shouldn't remove that here, but add it everywhere else. This is essentially a breaking change without technical motivation, potentially being very annoying to users - and we should take all possible measures not to introduce these kinds of changes.

BalduinLandolt (commented):
I would prefer to add `xslt` everywhere rather than to remove it, if it is a legal file extension.

BalduinLandolt (approved):

PR #1029, opened 2024-06-21
feat(excel2json): creating classes without any cardinalities (DEV-3755)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -284,26 +280,53 @@ def _prepare_classes_df(resource_dfs: dict[str, pd.DataFrame]) -> tuple[pd.DataF
...
+
+
+def _validate_classes_excel_sheet(classes_df: pd.DataFrame) -> list[Problem]:
     if any(classes_df.get(lang) is not None for lang in languages):
         warnings.warn(
             f"The file 'resources.xlsx' uses {languages} as column titles, which is deprecated. "
             f"Please use {[f'label_{lang}' for lang in languages]}"
         )
     problems: list[Problem] = []
-    if missing_super_rows := [int(str(i)) + 2 for i, row in classes_df.iterrows() if not check_notna(row["super"])]:
```
Nora-Olivia-Ammann, line 302: This code is not actually gone, it just now uses a previously existing and tested function.

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -284,26 +280,53 @@ def _prepare_classes_df(resource_dfs: dict[str, pd.DataFrame]) -> tuple[pd.DataF
...
-    # check that all the sheets have an entry in the names column and vice versa
-    if (all_names := set(classes_df["name"].tolist())) != (all_sheets := set(df_dict)):
-        problems.append(ResourcesSheetsNotAsExpected(all_names, all_sheets))
     return problems
+
+
+def _validate_individual_class_sheets(class_df_dict: dict[str, pd.DataFrame]) -> list[Problem]:
+    required_cols = ["property", "cardinality"]
+    missing_required_columns = {
+        sheet: res
```
jnussbaum, line 318: renaming suggestion: res --> problem

src/dsp_tools/utils/shared.py
```diff
@@ -15,46 +15,6 @@
 from dsp_tools.models.exceptions import BaseError
 
 
-def prepare_dataframe(
```
jnussbaum, line 18: why is this not necessary any more?

test/integration/commands/excel2json/test_resources.py
```diff
@@ -183,63 +189,49 @@ class TestValidateWithSchema:
     lambda _: e2j._validate_resources([])
 
     def test_invalid_super(self) -> None:
-        expected_msg = (
```
jnussbaum, line 186: In fact you have now removed the `test_invalid_super()` test, and replaced it by a new test. I think we need both of them: The `test_invalid_super()` should remain untouched. Your new test should be added as a test on its own.

jnussbaum: Now I see that I have put several mistakes into the file `testdata/invalid-testdata/excel2json/resources-invalid-super.xlsx`. We should create 2 different Excel files, one for every mistake.

test/integration/commands/excel2json/test_resources.py
```diff
@@ -183,63 +189,49 @@ class TestValidateWithSchema:
...
+            "---------------------------------------\n\n"
+            "At the following locations mandatory values are missing:\n"
+            "    - Located at: Sheet 'Owner' | Column 'property' | Row 18\n"
+            "    - Located at: Sheet 'Owner' | Column 'cardinality' | Row 17"
         )
         with pytest.raises(InputError, match=expected_msg):
             e2j.excel2resources("testdata/invalid-testdata/excel2json/resources-invalid-cardinality.xlsx", "")
 
     def test_invalid_property(self) -> None:
-        expected_msg = (
```
jnussbaum, line 211: same here: you can just fix the invalid cardinalities in the "Owner" sheet. Then, the ':fan:ta:sy' error will again pop up.

test/unittests/commands/excel2json/test_resources.py
```diff
@@ -121,3 +126,36 @@ def test_make_one_property() -> None:
...
+    sheet_missing_card = pd.DataFrame({"property": ["p2", "p3"], "cardinality": [pd.NA, "1-n"]})
+    sheet_missing_prop = pd.DataFrame({"property": ["p4", pd.NA], "cardinality": ["0-1", "1"]})
+    test_dict = {
+        "sheet_good": sheet_good,
+        "sheet_missing_card": sheet_missing_card,
+        "sheet_missing_prop": sheet_missing_prop,
+    }
+    res = e2j._validate_individual_class_sheets(test_dict)
+    assert len(res) == 1
+    casted_res = cast(MissingValuesProblem, res[0])
```
jnussbaum, line 142: I would rather assert than cast: `assert isinstance(res[0], MissingValuesProblem)`

test/unittests/commands/excel2json/test_resources.py
```diff
@@ -121,3 +126,36 @@ def test_make_one_property() -> None:
...
+    assert prop_problem.sheet == "sheet_missing_prop"
+    assert prop_problem.column == "property"
+    assert prop_problem.row == 3
+
+
+def test_validate_individual_class_sheets_missing_column() -> None:
+    test_dict = {"sheet_missing_col": pd.DataFrame({"property": ["p5"]})}
+    res = e2j._validate_individual_class_sheets(test_dict)
+    assert len(res) == 1
+    problem = cast(InvalidContentInSheetProblem, res[0])
```
jnussbaum, line 157: I would rather assert than cast: `assert isinstance(res[0], InvalidContentInSheetProblem)`

jnussbaum (commented):
Overall, it looks good. There is only 1 thing that should be fixed before merging: Currently, some tests in `test/integration/commands/excel2json/test_resources.py` are just overwritten, instead of adding new tests. (This is because in the past, I had produced invalid files with more than 1 mistake in it, which is subpar).

These should be disentangled, i.e. you should fix the old invalid files, and create new invalid files, so that you can add new test cases for your new validations.

docs/file-formats/json-project/ontologies.md
```diff
@@ -946,8 +946,9 @@ A cardinality is defined as follows:
     - `"1-n"`: At least one value must be present, but multiple values may be present.
     - `"0-n"`: The value may be omitted, but may also occur multiple times.
 
-(*It is technically possible to have a resource without cardinalities,
-but in most cases it doesn't make sense to omit them.)
+(*A resource can have no cardinalities.
+If the `super` is a class of the own ontology, the child class will inherit the cardinalities.
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: For inheritance i think so.

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -141,17 +144,12 @@ def _row2resource(
...
-    class_df_with_cardinalities = prepare_dataframe(
-        df=class_df_with_cardinalities,
-        required_columns=["Property", "Cardinality"],
-        location_of_sheet=f"Sheet '{class_name}' in file 'resources.xlsx'",
-    )
+def _make_cardinality_section(
+    class_name: str, class_df_with_cardinalities: pd.DataFrame | None
+) -> list[dict[str, str | int]]:
+    if class_df_with_cardinalities is None:
+        return []
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: If you check `if pd.DataFrame` you will get an error: "The truth value of a dataframe is ambiguous" (or similar)

test/integration/commands/excel2json/test_resources.py
```diff
@@ -183,9 +189,8 @@ class TestValidateWithSchema:
     lambda _: e2j._validate_resources([])
 
     def test_invalid_super(self) -> None:
-        expected_msg = (
-            "\nThe Excel file 'testdata/invalid-testdata/excel2json/resources-invalid-super.xlsx' "
-            "did not pass validation.\n"
+        expected_msg = regex.escape(
+            "\nThe Excel file 'resources.xlsx' did not pass validation.\n"
```
jnussbaum, line 193: I wonder if it is a good idea to always say 'resources.xlsx', even if the actual name could be different. It would be better to use the real file name. (Even if currently, it will always be 'resources.xlsx', in the future it might be different).

PR #1031, opened 2024-06-24
chore: improve error messages according to consulting with RDU 

jnussbaum:


src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -389,7 +390,11 @@ def _make_link_value(value: XMLValue, iri_resolver: IriResolver) -> dict[str, An
     elif resolved_iri := iri_resolver.get(s):
         iri = resolved_iri
     else:
-        raise BaseError(f"Could not resolve ID {s} to IRI.")
+        msg = (
+            f"Could not resolve ID {s} to IRI. This is probably because the resource '{s}' could not be created. "
```
Nora-Olivia-Ammann, line 394: Or maybe you could say, could not find the IRI in the id2iri mapping file?

jnussbaum: For me, that would be too much from a insider perspective. I guess that anyone who doesn't know the code of DSP-TOOLS very well does not know what a "id2iri mapping" is.

src/dsp_tools/commands/xmlupload/models/ingest.py
```diff
@@ -107,7 +107,7 @@ def _ingest(self, filepath: Path) -> IngestResponse:
                 elif res.status_code == STATUS_UNAUTHORIZED:
                     raise BadCredentialsError("Bad credentials")
                 else:
-                    user_msg = f"{err} See logs for more details: {logger_savepath}"
+                    user_msg = f"{err} See {warnings_savepath} for more information."
```
jnussbaum, line 110: I have now replace logger_savepath by warnings_savepath in the entire codebase. Because on all occurrences, the message is logged on level warning/error. There is no reference any more to the log file. But I'm not 100% sure if that's what we agreed upon. 

Generally, do you think it makes sense to refer the users to `warnings.log` all the time, or rather to the log file, or perhaps to both? I think it makes most sense to refer only to warnings.log, but I wanted to check with you.

src/dsp_tools/utils/logger_config.py
```diff
@@ -3,6 +3,7 @@
 from loguru import logger
 
 logger_savepath = (Path.home() / ".dsp-tools" / "logging.log").absolute()
+warnings_savepath = Path("warnings.log")
```
jnussbaum, line 6: These 2 should probably be constants, in upper case. What do you think?

Nora-Olivia-Ammann: I am fine with either.

src/dsp_tools/commands/xmlupload/models/ingest.py
Nora-Olivia-Ammann: I think while we have the warnings file, it is ok to refer only to it. If we delete that, we just should not forget to do a search of the whole repo...

src/dsp_tools/commands/xmlupload/resource_create_client.py
Nora-Olivia-Ammann: Well they know the file, it is saved in their cwd and they know to keep it.

Nora-Olivia-Ammann (approved):

src/dsp_tools/commands/xmlupload/models/ingest.py
jnussbaum: Now that we have a variable that is imported, it's impossible to forget :-)

src/dsp_tools/commands/xmlupload/resource_create_client.py
gNahcab: I would also write "could not find ID in the id2iri mapping" , it makes it more clear what the error is about than writing "could not resolve ID to IRI". And maybe also from a general point of view:  it could be better to write exactly what the problem is than trying to write easier messages for 'outsiders' if we want to spread knowledge about code and practices in DaSCH as much as possible.

jnussbaum: That's true, they know the file name. I have adapted the error message.

gNahcab (approved):

PR #1032, opened 2024-06-25
fix(new-excel2json): don't crash when Excel files are upper-cased (DEV-3808)

jnussbaum:


src/dsp_tools/commands/excel2json/lists.py
```diff
@@ -352,7 +352,9 @@ def validate_lists_section_with_schema(
     Returns:
         True if the "lists" section passed validation
     """
-    if bool(path_to_json_project_file) == bool(lists_section):
```
Nora-Olivia-Ammann, line 355: This would be a lot easier to read if you did a match case

Nora-Olivia-Ammann (approved):

Nora-Olivia-Ammann:
> Should we adapt the glob patterns for the resources and properties excel files, too?

I missed this question, sorry.
1. I would not warn and definitively not raise an error, because not having lists is valid. We inform which excels are processed, if it is not on the list. Then the user has to act... 

2. I would not do the glob for resources and properties because this is only one excel that should have this exact name. With lists we do it because they can have several. I would not over correct this.

PR #1037, opened 2024-07-01
fix(xmlupload): don't omit any stashed links (DEV-3827)

gNahcab:


src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
```diff
@@ -31,25 +33,74 @@ def upload_stashed_resptr_props(
...
+
+    Args:
+        link_value_stash: the current state of the link value stash
+        iri_resolver: service for resolving internal IDs to IRIs
+        con: connection to DSP
+        context: the JSON-LD context of the resource
+    """
+    for res_id, stash_items in link_value_stash.res_2_stash_items.items():
+        res_iri = iri_resolver.get(res_id)
         if not res_iri:
```
Nora-Olivia-Ammann, line 63: I think this would be a good place for a new user warning (https://github.com/dasch-swiss/dsp-tools/blob/main/src/dsp_tools/models/custom_warnings.py).
We could tell the user, that the item could not be uploaded because a dependency failed. This may be helpful to identify follow up problems.

test/integration/commands/xmlupload/stash/test_upload_stash_with_mock.py
```diff
@@ -99,6 +99,33 @@ def test_upload_link_value_stash(self) -> None:
         _upload_stash(upload_state, ProjectClientStub(con, "1234", None))
         assert not upload_state.pending_stash or upload_state.pending_stash.is_empty()
 
+    def test_upload_link_value_stash_multiple(self) -> None:
```
Nora-Olivia-Ammann, line 102: Could you make a test where it fails to resolve the IRI. If you do the user warning you can also test that the function emitted a warning.

Nora-Olivia-Ammann (approved):

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
```diff
@@ -21,6 +22,7 @@ def upload_stashed_resptr_props(
 ) -> None:
     """
     After all resources are uploaded, the stashed resptr props must be applied to their resources in DSP.
+    Finally, the resource-ids should be removed if all their resptr props could be uploaded.
```
jnussbaum, line 25: removed from what?

Is this sentence relevant on this abstraction level? My impression is that this is a low-level detail that is relevant on the level of `_remove_empty_ids()`, but not on the level of `upload_stashed_resptr_props()`

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
```diff
@@ -31,25 +33,74 @@ def upload_stashed_resptr_props(
...
+def _upload_process(
+    link_value_stash: LinkValueStash,
+    iri_resolver: IriResolver,
+    con: Connection,
+    context: dict[str, str],
+) -> None:
+    """
+    Try to upload the stashed items. If the target_iri of the item can be found
+    and the item can then successfully be uploaded it is removed from link_value_stash.
+    The link_value_stash is updated accordingly, as a side effect.
```
jnussbaum, line 53: The sentence before says: "it is removed from link_value_stash". It doesn't make sense to repeat this information

```suggestion
```

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
```diff
@@ -31,25 +33,74 @@ def upload_stashed_resptr_props(
...
         print(f"{datetime.now()}:   Upload resptrs of resource '{res_id}'...")
         logger.info(f"  Upload resptrs of resource '{res_id}'...")
-        for stash_item in stash_items:
-            target_iri = upload_state.iri_resolver.get(stash_item.target_id)
+
+        start_position = len(stash_items) - 1
+        while start_position >= 0:
+            # taking values by index from the list of 'stash_items' should avoid the problem we encounter
+            # when looping over the same list we simultaneously remove from
+            # counting max-length to 0 avoids problem with diminishing list-length
```
jnussbaum, line 75:
```suggestion
            # Taking values by index from the list of 'stash_items' should avoid the problem we encounter
            # when looping over the same list we simultaneously remove from.
            # Counting from max-length to 0 avoids the problem of diminishing list-length.
```

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
```diff
@@ -31,25 +33,74 @@ def upload_stashed_resptr_props(
...
+    for res_id, stash_items in link_value_stash.res_2_stash_items.items():
+        res_iri = iri_resolver.get(res_id)
         if not res_iri:
             # resource could not be uploaded to DSP, so the stash cannot be uploaded either
             # no action necessary: this resource will remain in nonapplied_resptr_props,
             # which will be handled by the caller
             continue
         print(f"{datetime.now()}:   Upload resptrs of resource '{res_id}'...")
         logger.info(f"  Upload resptrs of resource '{res_id}'...")
-        for stash_item in stash_items:
```
seakayone, line 45: The simplest thing I can do to make your test pass is replacing this line with:

```
for stash_item in reversed(stash_items):
```

When iterating over a list and removing elements, the indices of the remaining elements are shifted. This can cause elements to be skipped or result in accessing invalid indices.

By reversing the list and iterating backwards, the indices of the elements that have not yet been processed are not affected by the removal of elements that have already been processed. This ensures that each element is properly accessed and handled.

All the other code changes in this PR seem to be unnecessary.

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
```diff
@@ -31,25 +33,74 @@ def upload_stashed_resptr_props(
...
+            ):
+                continue
+            link_value_stash.res_2_stash_items[res_id].remove(stash_item)
+
+
+def _remove_empty_ids(link_value_stash: LinkValueStash) -> None:
+    """
+    If the upload of resptr-props was successful, the list of resptr props are empty
+    but the resource-ids still exist as keys. If the list is empty, its resource-id is removed.
+    The link_value_stash is updated accordingly, as a side effect.
```
jnussbaum, line 94:
```suggestion
    If the upload of resptr-props was successful, the list of resptr props is empty,
    but the resource-ids still exist as keys. If the list is empty, its resource-id must be removed
    from the link_value_stash.
```

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
```diff
@@ -31,25 +33,74 @@ def upload_stashed_resptr_props(
...
+
+    Args:
+        link_value_stash: the current link value stash of upload state
+    """
+    # remove empty res_id from res_2_stash_items-dictionary
```
jnussbaum, line 99:
```suggestion
```

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
```diff
@@ -31,25 +33,74 @@ def upload_stashed_resptr_props(
...
                 continue
-            if _upload_stash_item(stash_item, res_iri, target_iri, con, context):
-                link_value_stash.res_2_stash_items[res_id].remove(stash_item)
-        if not link_value_stash.res_2_stash_items[res_id]:
-            del link_value_stash.res_2_stash_items[res_id]
+            if not _upload_stash_item(
+                stash=stash_item, res_iri=res_iri, target_iri=target_iri, con=con, context=context
+            ):
+                continue
+            link_value_stash.res_2_stash_items[res_id].remove(stash_item)
```
jnussbaum, line 87: Could be simplified to:

```suggestion
            if _upload_stash_item(
                stash=stash_item, res_iri=res_iri, target_iri=target_iri, con=con, context=context
            ):
                link_value_stash.res_2_stash_items[res_id].remove(stash_item)
```

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
```diff
@@ -31,25 +33,74 @@ def upload_stashed_resptr_props(
...
+    The link_value_stash is updated accordingly, as a side effect.
+
+    Args:
+        link_value_stash: the current link value stash of upload state
+    """
+    # remove empty res_id from res_2_stash_items-dictionary
+    for res_id, stash_items in link_value_stash.res_2_stash_items.copy().items():
+        if link_value_stash.res_2_stash_items[res_id]:
+            continue
+        del link_value_stash.res_2_stash_items[res_id]
```
jnussbaum, line 103: Could be simplified:

```suggestion
        if not link_value_stash.res_2_stash_items[res_id]:
            del link_value_stash.res_2_stash_items[res_id]
```

jnussbaum (approved):
Congrats that you found this bug, and managed to analyse it so thoroughly and exactly! I have some small feedback, but generally, it looks very good.

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
jnussbaum: I'm a bit unsure if emitting a warning is a good idea here. If a resource could not be created, the user was already alerted. And AFAIU, there will be no follow-up problems if a stashed link cannot be reapplied. 

But if you think that it would be helpful for the user, I don't want to block it.

Nora-Olivia-Ammann: Well we know that a resource could not be uploaded but we do not know if that is the source of the failure to upload this stash no? I think RDU prefers us to be extremely explicit and spell out the problems. That is the reason, they often mention that they do not understand if it is a follow up problem. They don't know about the stash and how that all links.

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
```diff
@@ -6,6 +6,7 @@
```
jnussbaum, line 6: Our PR titles are used to create the release notes, so we should try to convey helpful information about what was fixed. For our users, it doesn't matter that we had a problem with a shallow copy. Rather, they want to know what kind of problem was solved in which dsp-tools command. So I suggest:

### fix(xmlupload): don't omit any stashed links (DEV-3827)

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
gNahcab: But wouldn't that be part of a new PR with the goal of refactoring or adding a new feature?

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
gNahcab: Yes, you are right I did not understand what was going on, that is why I tried to split the method so that it makes more sense to me what was happening. But since it is about a fix, I guess something like that would be a refactoring task and better be discussed first if it should be done and how.

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
gNahcab: Makes sens, I will change the title

src/dsp_tools/commands/xmlupload/stash/upload_stashed_resptr_props.py
jnussbaum: > we know that a resource could not be uploaded but we do not know if that is the source of the failure to upload this stash no?

I don't agree. At this point (line 63), the only failure is that the target resource doesn't exist. The actual stash upload hasn't started yet, so there is no failure in the stash upload. 

There is a direct cause --> effect: resource could not be uploaded --> link pointing to failed resource will be skipped. There is no actual failure.

jnussbaum: > But wouldn't that be part of a new PR with the goal of refactoring or adding a new feature?

Yes. If we do this, then in a separate PR .

test/integration/commands/xmlupload/stash/test_upload_stash_with_mock.py
gNahcab: hm,  I would suggest, let's discuss it for another PR, I wouldn't write that now, because it doesn't affect the problem with the omitted stash items.

PR #1039, opened 2024-07-02
chore: Add db configuration to ingest container

seakayone:
Use a docker volume to store the ingest sqlite file persistently.

From the upcoming dsp-ingest release on the service requires an sqlite file to start.
The location is configured with the jdbc location.
The service will create and maintain that file by itself.
In order to retain the database's state it is stored in a docker volume.

This is the same change in the dsp-api setup:
https://github.com/dasch-swiss/dsp-api/pull/3304/files

### DIFF
--- a/src/dsp_tools/resources/start-stack/docker-compose.yml
+++ b/src/dsp_tools/resources/start-stack/docker-compose.yml
@@ -67,6 +67,7 @@ services:
     volumes:
       - ./sipi/images:/opt/images
       - ./sipi/tmp-dsp-ingest:/opt/temp
+      - ingest-db:/opt/db
     environment:
       - SERVICE_LOG_FORMAT=text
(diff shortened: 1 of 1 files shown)


siers (approved):

jnussbaum:
@seakayone @siers Thank you for thinking at us :-) This config should also be added to `test/e2e/setup_testcontainers.py`. 

A general thought: Now we are still in a transitional phase, where much internal knowledge about DSP-TOOLS is in the brains of the "old core-devs" Balduin, Nora, and me. What do you (generally) think about waiting for a review of one of the old core-devs?

siers:
@jnussbaum I think it's generally a good idea and I think Christian likely shares this view, though probably this is a simple enough change that it only didn't apply in this case.

PR #1040, opened 2024-07-03
fix(excel2json-lists): assignment of wrong parent ID if the node names are identical (DEV-3844)

Nora-Olivia-Ammann:


jnussbaum (approved):
I trust you :-)

Nora-Olivia-Ammann:
> I trust you :-)

Thanks, I made sure. The test failed before and now it is good.

PR #1043, opened 2024-07-10
chore: bump start-stack to 2024.07.01

gNahcab:


Nora-Olivia-Ammann (approved):
Perhaps you could correct the link on line 5:
https://github.com/dasch-swiss/ops-deploy/blob/main/versions/RELEASE.json

PR #1044, opened 2024-07-10
try-outs-sub classing edit conflict

Nora-Olivia-Ammann:


src/dsp_tools/commands/project/models/resourceclass.py
```diff
@@ -486,11 +486,14 @@ def getProperty(self, property_id: str) -> Optional[HasProperty]:
...
         self,
-        last_modification_date: DateTimeStamp,
+        # last_modification_date: DateTimeStamp,
         property_id: str,
         cardinality: Cardinality,
         gui_order: Optional[int] = None,
     ) -> DateTimeStamp:
+        encoded_res_iri = quote_plus(self.iri)
+        get_mod_date = self._con.get(f"{self.ROUTE}/{encoded_res_iri}")
+        latest_date = DateTimeStamp(get_mod_date["knora-api:lastModificationDate"]["@value"])
```
Nora-Olivia-Ammann, line 496: I assumed that the last_modification_date was outdated. So I requested it again at this stage. However, in the debug mode, I saw that the dates were identical. So I don't understand the problem... Any ideas?

testdata/json-project/sub-class-project.json
```diff
@@ -0,0 +1,67 @@
...
+                "gui_order": 1,
+                "cardinality": "1"
+              }
+            ]
+          },
+          {
+            "name": "SubClass1",
+            "super": [
+              ":SuperClass",
+              "cidoc-crm:E70_Thing"
```
Nora-Olivia-Ammann, line 57: If line 57 is removed then the upload works perfectly.
If the superclass `cidoc-crm:E70_Thing` is added to `SuperClass` then the upload works too. The only situation in which I could produce this bug, was when a subclass had a reference to an external ontology.

testdata/json-project/sub-class-success.json
```diff
@@ -0,0 +1,69 @@
...
+            },
+            "gui_element": "SimpleText"
+          }
+        ],
+        "resources": [
+          {
+            "name": "SuperClass",
+            "super": [
+              "Resource",
+              "cidoc-crm:E70_Thing"
```
Nora-Olivia-Ammann, line 43: same external super, sub used in the `SuperClass`, this file uploads sucessfully.

testdata/json-project/sub-class-success-other-card.json
```diff
@@ -0,0 +1,69 @@
...
+          {
+            "name": "SubClass1",
+            "super": [
+              ":SuperClass",
+              "cidoc-crm:E70_Thing"
+            ],
+            "labels": {
+              "en": "Sub Class 1"
+            },
+            "cardinalities": [
```
Nora-Olivia-Ammann, line 57: Here I moved the cardinality from the super class to the subclass. This file is also successful. But the code is so messy that I don't understand how this makes a difference.

PR #1050, opened 2024-07-15
test: enable project deletion & ingest db in testcontainers

jnussbaum:
The changes introduced by #1042 and #1039 should also be applied to the testcontainers.


Nora-Olivia-Ammann (approved):
I don't think project deletion must be enabled in the test container, but I assume it does not really make a difference.

BalduinLandolt (approved):

seakayone (approved):

seakayone:
@jnussbaum thanks for fixing this. I was not aware that we have testcontainer setup in dsp-tools; otherwise I would have also added the configuration there.

PR #1053, opened 2024-07-16
refactor(excel2json): resources.xlsx problem diagnosis

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/models/input_error.py
```diff
@@ -192,12 +192,52 @@ def execute_error_protocol(self) -> str:
...
+
+        return (
+            f"'classes' is a protected sheet name. Please delete all but one.\n"
+            f"The following sheets contain this in their name:"
+            f"{list_separator}{list_separator.join(self.name_classes)}"
+        )
+
+
+@dataclass
+class ResourceIndividualClasses:
```
jnussbaum, line 217:
```suggestion
class MissingEntryInClassesSheet:
```

src/dsp_tools/commands/excel2json/models/input_error.py
```diff
@@ -192,12 +192,52 @@ def execute_error_protocol(self) -> str:
         )
 
 
+@dataclass
+class ResourceClassSheet:
+    """This class contains information if the excel contains two sheets with 'classes' in the name."""
```
jnussbaum, line 197: So, "Englishclasses" or "Frenchclasses" would not be accepted? I think that the only thing which should be forbidden is "classes" itself.

src/dsp_tools/commands/excel2json/models/input_error.py
```diff
@@ -192,12 +192,52 @@ def execute_error_protocol(self) -> str:
         )
 
 
+@dataclass
+class ResourceClassSheet:
```
jnussbaum, line 196:
```suggestion
class IllegalSheetName:
```

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -247,24 +252,27 @@ def excel2resources(
     return resources, True
 
 
-def _prepare_classes_df(resource_dfs: dict[str, pd.DataFrame]) -> tuple[pd.DataFrame, dict[str, pd.DataFrame]]:
-    resource_dfs = {k.strip(): v for k, v in resource_dfs.items()}
+def _do_classes_excel_sheet_compliance(resource_dfs: dict[str, pd.DataFrame]) -> list[Problem]:
     sheet_name_list = list(resource_dfs)
+    problems: list[Problem] = []
+    cls_sheet_name = _find_all_classes(sheet_name_list)
```
jnussbaum, line 258: This is a list of strings, so the variable name cannot be singular

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -247,24 +252,27 @@ def excel2resources(
...
+    if not cls_sheet_name:
+        problems.append(MissingSheetProblem(missing_sheets=["classes"], existing_sheets=sheet_name_list))
+    elif len(cls_sheet_name) > 1:
+        problems.append(ResourceClassSheet(cls_sheet_name))
+    return problems
+
+
+def _find_all_classes(sheet_name_list: list[str]) -> list[str]:
```
jnussbaum, line 266: does this find all class sheets? Then it should be renamed to "_find_all_class_sheet_names"

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -247,24 +252,27 @@ def excel2resources(
...
     cls_sheet_name = [
```
jnussbaum, line 267: plural

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -247,24 +252,27 @@ def excel2resources(
...
         ok.group(0) for x in sheet_name_list if (ok := regex.search(r"classes", flags=regex.IGNORECASE, string=x))
```
jnussbaum, line 268: This is confusing. Why do you search for `r"classes"`? `r"^classes$"` would make more sense to me. I think it should be okay if the substring "classes" appears.

src/dsp_tools/commands/excel2json/models/input_error.py
Nora-Olivia-Ammann: Yes, I agree this is what i am going to change now. But this was actually already forbidden previously...

src/dsp_tools/commands/excel2json/resources.py
Nora-Olivia-Ammann: This already existed previously and will be deleted.

PR #1059, opened 2024-07-17
refactor(excel2json): serialise property with dataclass

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/models/ontology.py
```diff
@@ -0,0 +1,44 @@
...
+        if self.gui_attributes:
+            prop_dict["gui_attributes"] = self.gui_attributes.get()
+        return prop_dict
+
+
+@dataclass
+class LanguageDict:
+    lang_dict: dict[str, str]
+
+    def get(self) -> Any:
```
Nora-Olivia-Ammann, line 35: If I did not type with Any then mypy threw an assignment error

src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -226,32 +229,33 @@ def _get_final_series(
...
+        gui_attributes=gui_attrib,
+    )
+    return prop
 
 
 def _get_gui_attribute(
     df_row: pd.Series[Any],
     row_num: int,
-) -> dict[str, int | str | float] | InvalidExcelContentProblem | None:
+) -> GuiAttributes | InvalidExcelContentProblem | None:
```
Nora-Olivia-Ammann, line 258: I wanted to change the serialisation because in future I would like to do all the invalid checks during the construction, this way we can collect all the errors and communicate them at the end, right now the program stops if one error was found. Also better typing.

src/dsp_tools/commands/excel2json/models/ontology.py
jnussbaum: Then, the underlying problem should be investigated. Simply typing this with Any is not a good solution :-)

Nora-Olivia-Ammann: I have tried everything I can think of, including copilot and chatgpt, they did not suggest anything else that worked. If you could take a look, I would be glad.

src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -226,32 +229,33 @@ def _get_final_series(
...
         case InvalidExcelContentProblem():
             msg = f"There is a problem with the excel file: '{excelfile}'\n" + gui_attrib.execute_error_protocol()
             raise InputError(msg) from None
-
-    if comment := get_comments(df_row=df_row):
-        _property["comments"] = comment
-
-    return _property
+        case _:
+            pass
```
jnussbaum, line 241: The match case doesn't make sense any more, because you only want to take action in 1 case. Then, a simple `if` is more appropriate (and uses only 3 LOC instead of 6 LOC :-)

```
if isinstance(gui_attrib, InvalidExcelContentProblem):
        msg = f"There is a problem with the excel file: '{excelfile}'\n" + gui_attrib.execute_error_protocol()
        raise InputError(msg)
```

jnussbaum: In addition: `raise Exception() from None`: the `from None` is only meaningful if you are inside an `except` block. Here, it's superfluous.

src/dsp_tools/commands/excel2json/utils.py
```diff
@@ -218,7 +219,9 @@ def get_comments(df_row: pd.Series[Any]) -> dict[str, str] | None:
         A dictionary with the language tag and the content of the cell
     """
     comments = {lang: df_row[f"comment_{lang}"] for lang in languages if not pd.isna(df_row[f"comment_{lang}"])}
-    return comments or None
+    if comments:
+        return LanguageDict(comments)
+    return None
```
jnussbaum, line 224:
```suggestion
    return LanguageDict(comments) if comments else None
```

src/dsp_tools/commands/excel2json/models/ontology.py
```diff
@@ -0,0 +1,44 @@
...
+    name: str
+    super: list[str]
+    object: str
+    subject: str | None
+    labels: LanguageDict
+    comments: LanguageDict | None
+    gui_element: str
+    gui_attributes: GuiAttributes | None
+
+    def get(self) -> dict[str, Any]:
```
jnussbaum, line 18: Not sure, but perhaps this could be renamed to `serialize`? Just a thought, I don't have a strong opinion.

jnussbaum (approved):

src/dsp_tools/commands/excel2json/models/ontology.py
jnussbaum: Done with https://github.com/dasch-swiss/dsp-tools/commit/ea717e12ea5c613a5473a6a9787bea507a87fe4b

src/dsp_tools/commands/excel2json/properties.py
jnussbaum: The type annotations look much cleaner now, thanks for this improvement 👍

PR #1060, opened 2024-07-17
refactor(excel2json): reorder the functions in resources.py

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -29,76 +29,120 @@
 languages = ["en", "de", "fr", "it", "rm"]
```
Nora-Olivia-Ammann, line 31: No content was changed, the order of the functions now follow the flow of the main function better.

jnussbaum (approved):

PR #1063, opened 2024-07-18
refactor(excel2json): properties collect problems before raising error

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -229,13 +237,13 @@ def _get_final_series(
...
-def _row2prop(df_row: pd.Series[Any], row_num: int, excelfile: str) -> OntoProperty:
+def _row2prop(df_row: pd.Series[Any], row_num: int) -> OntoProperty | PropertyProblem:
     subj = df_row["subject"] if not pd.isna(df_row["subject"]) else None
     comment = get_comments(df_row=df_row)
     gui_attrib = _get_gui_attribute(df_row=df_row, row_num=row_num)
     if isinstance(gui_attrib, InvalidExcelContentProblem):
-        msg = f"There is a problem with the excel file: '{excelfile}'\n" + gui_attrib.execute_error_protocol()
-        raise InputError(msg)
+        problem: list[Problem] = [gui_attrib]
+        return PropertyProblem(df_row["name"], problem)
```
jnussbaum, line 246:
```suggestion
        problems: list[Problem] = [gui_attrib]
        return PropertyProblem(df_row["name"], problems)
```

or: 

```suggestion
        return PropertyProblem(df_row["name"], cast(list[Problem], [gui_attrib]))
```

test/unittests/commands/excel2json/test_properties.py
```diff
@@ -222,6 +224,33 @@ def test_get_final_series_two_None(self) -> None:
...
+                "object": "object_1",
+                "gui_element": "Simple",
+                "gui_attributes": "max:1.4 / min:1.2",
+            }
+        )
+        res = e2j._row2prop(test_series, row_num=0)
+        assert isinstance(res, PropertyProblem)
+        assert res.prop_name == "name_1"
+        assert len(res.problem) == 1
+        assert isinstance(res.problem[0], InvalidExcelContentProblem)
```
jnussbaum, line 252: At first, I didn't understand why there is a problem. Could you add some more assertions, so that it becomes clear? For example:

```
        assert res.problem[0].actual_content == "max:1.4 / min:1.2"
        assert res.problem[0].expected_content == "attribute: value, attribute: value"
```

PR #1065, opened 2024-07-19
feat(excel2json): add option to create header with Excel file (DEV-3898)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/json_header.py
```diff
@@ -0,0 +1,249 @@
...
+        JsonHeader object
+
+    Raises:
+        InputError: If the file does not conform to the specifics
+    """
+    if not excel_filepath.is_file():
+        print("No json_header.xlsx file found in the directory, empty header was added.")
+        return EmptyJsonHeader()
+    sheets_df_dict = read_and_clean_all_sheets(excel_filepath)
+    sheets_df_dict = {x.lower(): df for x, df in sheets_df_dict.items()}
```
BalduinLandolt, line 55: this seems rather low-level. Could this step simply be integrated in the function above?

src/dsp_tools/commands/excel2json/json_header.py
```diff
@@ -0,0 +1,249 @@
...
+    expected_sheets = ["prefixes", "project", "description", "keywords"]
+
+    def _check_df(sheet: str) -> str | None:
+        if (df := df_dict.get(sheet)) is None:
+            return sheet
+        if len(df) == 0:
+            return sheet
+        return None
+
+    if empty := [x for x in expected_sheets if _check_df(x)]:
```
BalduinLandolt, line 75: maybe I'm misremembering - but I think this logic can be achieved with the `any()` function

src/dsp_tools/commands/excel2json/json_header.py
```diff
@@ -0,0 +1,249 @@
...
+    if compliance_problem := _do_formal_compliance(sheets_df_dict):
+        raise InputError(compliance_problem.execute_error_protocol())
+    result = _process_file(sheets_df_dict)
+    if isinstance(result, ExcelFileProblem):
+        raise InputError(result.execute_error_protocol())
+    print("json_header.xlsx file used to construct the json header.")
+    return result
+
+
+def _do_formal_compliance(df_dict: dict[str, pd.DataFrame]) -> ExcelFileProblem | None:
```
BalduinLandolt, line 65: I always find "do" a red flag in function names, it's notoriously imprecise. What about "check" or "ensure" formal compliance? Or even "ensure expected sheets present", which is what it actually does?

src/dsp_tools/commands/excel2json/json_header.py
```diff
@@ -0,0 +1,249 @@
...
+    if not isinstance(project_result, Project):
+        problems.extend(project_result)
+    if problems:
+        return ExcelFileProblem("json_header.xlsx", problems)
+    project = cast(Project, project_result)
+    prefixes = cast(Prefixes, prefix_result)
+    return FilledJsonHeader(project, prefixes)
+
+
+def _do_prefixes(df: pd.DataFrame) -> ExcelSheetProblem | Prefixes:
```
BalduinLandolt, line 95: maybe "extract" or "read" prefixes?

src/dsp_tools/commands/excel2json/json_header.py
```diff
@@ -0,0 +1,249 @@
...
+    prefix_result = _do_prefixes(df_dict["prefixes"])
+    if isinstance(prefix_result, ExcelSheetProblem):
+        problems.append(prefix_result)
+    project_result = _do_project(df_dict)
+    if not isinstance(project_result, Project):
+        problems.extend(project_result)
+    if problems:
+        return ExcelFileProblem("json_header.xlsx", problems)
+    project = cast(Project, project_result)
+    prefixes = cast(Prefixes, prefix_result)
```
BalduinLandolt, line 91: there is a lot of type checking and casting going on. The reason for that is, that Python's type system is just not great... But I wonder, if many of those problems could be solved by separating validation from creation in these functions? Then you could first do all the validations; and when they are done, creation can be done less defensively because there should not be any errors anymore.

src/dsp_tools/commands/excel2json/json_header.py
```diff
@@ -0,0 +1,249 @@
...
+    if missing_cols := check_contains_required_columns(df, {"prefixes", "uri"}):
+        return ExcelSheetProblem("prefixes", [missing_cols])
+    if missing_vals := check_required_values_get_position_in_excel(df, ["prefixes", "uri"]):
+        return ExcelSheetProblem("prefixes", [missing_vals])
+    pref = dict(zip(df["prefixes"], df["uri"]))
+    pref = {k.rstrip(":"): v for k, v in pref.items()}
+    return Prefixes(pref)
+
+
+def _do_project(df_dict: dict[str, pd.DataFrame]) -> list[ExcelSheetProblem] | Project:
```
BalduinLandolt, line 105: as above

src/dsp_tools/commands/excel2json/json_header.py
```diff
@@ -0,0 +1,249 @@
...
+        problems.append(keywords_result)
+    all_users = None
+    if (user_df := df_dict.get("users")) is not None:
+        if len(user_df) > 0:
+            user_result = _do_users(user_df)
+            match user_result:
+                case ExcelSheetProblem():
+                    problems.append(user_result)
+                case _:
+                    all_users = user_result
```
BalduinLandolt, line 124: above, instance checking was done with `isinstance`, here you do pattern matching. Both options are viable, but I would keep it consistent.

src/dsp_tools/commands/excel2json/models/json_header.py
```diff
@@ -9,13 +9,13 @@
 
 @dataclass
 class JsonHeader(Protocol):
-    def make(self) -> dict[str, Any]:
+    def serialise(self) -> dict[str, Any]:
```
BalduinLandolt, line 12: technically, serialisation is the process of turning objects into strings, so a function that returns a dict is not really a function that serialises. I'm ok with it here because python dicts are so close to JSON - but it is a bit imprecise.

src/dsp_tools/commands/excel2json/utils.py
```diff
@@ -130,6 +132,31 @@ def check_column_for_duplicate(df: pd.DataFrame, to_check_column: str) -> None |
         return None
 
 
+def check_required_values_get_position_in_excel(
```
BalduinLandolt, line 135: If the `MissingValuesProblem` contains a location attribute, then `check_required_values` should be enough, and finding the position is an implementation detail that should not be in the function name.

test/integration/commands/excel2json/test_json_header.py
```diff
@@ -0,0 +1,62 @@
...
+def test_get_json_header_no_file() -> None:
+    test_path = Path("testdata/excel2json/excel2json_files")
+    result = get_json_header(test_path)
+    assert isinstance(result, EmptyJsonHeader)
+
+
+@pytest.fixture()
+def filled_json_header() -> JsonHeader:
+    test_path = Path("testdata/excel2json/new_excel2json_files/json_header.xlsx")
+    return get_json_header(test_path)
```
BalduinLandolt, line 23: just as a matter of taste: I would put all fixtures above the first test case. IMO it makes navigating the file easier.

BalduinLandolt (approved):
I added a bunch of stylistic remarks. Apart from those it looks good.

src/dsp_tools/commands/excel2json/json_header.py
Nora-Olivia-Ammann: No I would not do that by default. We use this function for all excel2json, including the classes where the names are relevant. So we would have to change some code there too...

src/dsp_tools/commands/excel2json/json_header.py
Nora-Olivia-Ammann: I want to use the empty, so I could do it like this:
`if any(empty := [x for x in expected_sheets if _check_df(x)])`
What do you think would be the advantage?

src/dsp_tools/commands/excel2json/json_header.py
Nora-Olivia-Ammann: I have thought about that too, the reason why I combined it is because if I do not i need to write the code that retrieves the information twice... It felt more cumbersome to me. But I don't think I would like either one version 100%

src/dsp_tools/commands/excel2json/models/json_header.py
Nora-Olivia-Ammann: Ok, thanks. What would you say is a better name?

src/dsp_tools/commands/excel2json/json_header.py
BalduinLandolt: fair enough, I was assuming this function was just for that purpose

src/dsp_tools/commands/excel2json/json_header.py
BalduinLandolt: probably there is none here, I think I missed the fact that you're re-using `empty` below (which, by the way is very confusingly named, because this evaluates true when the list is non-empty 😉 )

src/dsp_tools/commands/excel2json/json_header.py
BalduinLandolt: in the end it's up to you. But to me the retrieving code seems so minimal that I don't think the duplication would hurt.

src/dsp_tools/commands/excel2json/utils.py
Nora-Olivia-Ammann: I understand the comment, but this function name already exists. I will open a ticket to fix this later, so it doesn't pollute the diff here.

testdata/excel2json/expected_json_header.json
```diff
@@ -0,0 +1,76 @@
...
+  "prefixes": {
+    "foaf": "http://xmlns.com/foaf/0.1/",
+    "dcterms": "http://purl.org/dc/terms/",
+    "schema": "https://schema.org/",
+    "sdh": "https://ontome.net/ns/sdhss/",
+    "cidoc": "http://www.cidoc-crm.org/cidoc-crm/"
+  },
+  "$schema": "https://raw.githubusercontent.com/dasch-swiss/dsp-tools/main/src/dsp_tools/resources/schema/project.json",
+  "project": {
+    "shortcode": "11",
```
jnussbaum, line 11: A shortcode must always be 4 digits long, so I don't think it's a good idea to introduce wrong test data here.

jnussbaum (approved):
I very much like the architecture of this. The code is beautiful and elegant 😄

docs/file-formats/excel2json.md
```diff
@@ -77,6 +80,34 @@ Then, use the following command:
...
+The header section can be filled out with the "json_header.xlsx" file.
+The file must be in the main folder, and the name must be an exact match.
+
+It is recommended to use the [template](../assets/data_model_templates/json_header.xlsx).
+
+The sheets should be filled as follows:
+
+- **Prefixes**
+    - Both columns: prefixes and URI are mandatory.
+    - At least one prefix has to be declared
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: No I made prefixes mandatory, I discussed this with Daniela, because we want to strongly encourage references to external ontologies, we can remind people by making prefixes mandatory. But I would not go so far as to make them mandatory in the json, that would be a breaking change, here it is not.

src/dsp_tools/commands/excel2json/models/json_header.py
```diff
@@ -123,12 +115,19 @@ def get(self) -> dict[str, Any]:
...
+    project_admin: bool = False
+    sys_admin: bool = False
+
+    def to_dict(self) -> dict[str, list[str]]:
         if self.sys_admin:
-            usr_dict["groups"] = ["SystemAdmin"]
-            usr_dict["projects"] = [":admin", ":member"]
-        elif self.project_member:
-            usr_dict["projects"] = [":member"]
+            return {"groups": ["SystemAdmin"], "projects": [":admin", ":member"]}
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I did it this way because you never create a sys admin on prod during a `create` but through the app. So if more than one project is on a test server, then it is nice that the person from dasch can also find the project easily... So i saw no harm

testdata/excel2json/expected_json_header.json
Nora-Olivia-Ammann: Great that you noticed it! I just noticed, that the read function does not specify a datatype, which means it is interpreted as an int and the leading zeros are removed. This could have been a stupid bug later!

jnussbaum (approved):

docs/file-formats/excel2json.md
```diff
@@ -29,13 +29,15 @@ data_model_files
 `-- onto_name (onto_label)
     |-- properties.xlsx
     `-- resources.xlsx
+    `json_header.xlsx` (optional)
```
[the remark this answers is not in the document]
jnussbaum: @Nora-Olivia-Ammann Unfortunately, this was merged in an incorrect state. Could you fix that?

docs/file-formats/excel2json.md
```diff
@@ -68,6 +70,7 @@ data_model_files
...
```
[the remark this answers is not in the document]
jnussbaum: This, too, should be fixed

PR #1066, opened 2024-07-19
refactor: ignore ARG001 (unused function argument) in .pre-commit-config.yaml

Nora-Olivia-Ammann:
This keeps blocking me from pushing stuff, I would like to version even incomplete files.


jnussbaum:
> This keeps blocking me from pushing stuff, I would like to version even incomplete files.

FYI: You also have the possibility to circumvent the pre-commit-hooks with `git commit -m "message" --no-verify`

PR #1069, opened 2024-07-23
refactor: json header find problems first

Nora-Olivia-Ammann:


jnussbaum (approved):
Very good, I'm positively surprised!

PR #1070, opened 2024-07-24
refactor(excel2json): restructure and rename find missing values function (DEV-3909)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -133,10 +130,8 @@ def _validate_individual_class_sheets(class_df_dict: dict[str, pd.DataFrame]) ->
         return [ExcelSheetProblem(sheet, [missing]) for sheet, missing in missing_required_columns.items()]
     problem_list: list[PositionInExcel] = []
```
jnussbaum, line 131: Here again, I think that the naming is not optimal: These are actually not problems, but rather positions, so I would vote for naming them `positions_of_problems` or so

src/dsp_tools/commands/excel2json/resources.py
```diff
@@ -133,10 +130,8 @@ def _validate_individual_class_sheets(class_df_dict: dict[str, pd.DataFrame]) ->
...
     for sheet_name, df in class_df_dict.items():
-        if missing_dict := check_required_values(df, required_cols):
-            row_nums = get_wrong_row_numbers(missing_dict)
-            for col, nums in row_nums.items():
-                problem_list.extend([PositionInExcel(sheet_name, col, x) for x in nums])
+        if missing_values := find_missing_required_values(df, required_cols, sheet_name):
```
jnussbaum, line 133: renaming suggestion: `missing_vals_positions`

jnussbaum (approved):
LGTM

PR #1074, opened 2024-08-06
feat(excel2xml-lib): transform formatting into stand-off (DEV-3962)

Nora-Olivia-Ammann:


Nora-Olivia-Ammann:
This approach is being abandoned because of instability concerns.

PR #1081, opened 2024-08-07
test: add test file with special chars (DEV-4098)

jnussbaum:


src/dsp_tools/commands/xmlupload/models/ingest.py
```diff
@@ -94,7 +93,7 @@ def _ingest(self, filepath: Path) -> IngestResponse:
         Returns:
             IngestResponse: The internal filename of the uploaded file.
         """
-        filename = urllib.parse.quote(unicodedata.normalize("NFKD", filepath.name))
+        filename = urllib.parse.quote(filepath.name)
```
jnussbaum: @siers : Previously, I had tried out to do an NFKD normalization of files with special characters, before constructing the ingest URL. But now, I see that it also works without. What do you think is better / more future-proof: Doing it or not? 

(I'm refering to the POST /projects/<shortcode>/assets/ingest/filename route)

src/dsp_tools/resources/start-stack/docker-compose.yml
```diff
@@ -62,7 +62,7 @@ services:
   ingest:
     # on the verge of every deployment (fortnightly), update the "image" value from the "ingest" value of
     # https://github.com/dasch-swiss/ops-deploy/blob/main/versions/RELEASE.json
-    image: daschswiss/dsp-ingest:v0.12.2
+    image: daschswiss/dsp-ingest:v0.12.2-7-gf5cfb90
```
jnussbaum, line 65: Only merge this PR after the next deployment.

src/dsp_tools/commands/xmlupload/models/ingest.py
siers: Since it is now normalized in the backend, it is not strictly necessary.

jnussbaum: Thanks, then I won't do it.

Nora-Olivia-Ammann (commented):
While you're at it could you include all the german and french special characters?

Nora-Olivia-Ammann (approved):

siers (approved):

PR #1084, opened 2024-08-08
fix(cli): `resume-xmlupload --skip-first-resource` now works

Nora-Olivia-Ammann:


jnussbaum (approved):
LGTM. Have you also checked if there are other untested CLI commands?

Nora-Olivia-Ammann:
no because ideally that would just be adding tests and not a fix. But I can do that in another PR.

PR #1089, opened 2024-08-09
refactor(excel2json-lists): add excel name to lists problem classes

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/models/input_error_lists.py
```diff
@@ -32,7 +33,27 @@ def execute_error_protocol(self) -> str:
 
 
 @dataclass
-class ListSheetProblem:
+class CollectedSheetProblems:
```
jnussbaum, line 36: - This class is unused in the code.
- If you have unimplemented methods, shouldn't it then be a Protocol? Otherwise it doesn't make sense to even define these methods.
- Not sure if it is usual to `raise NotImplementedError` in an unimplemented Protocol method. I thought it would rather be done with `...`:

```
def execute_error_protocol(self) -> str: ...
```

src/dsp_tools/commands/excel2json/models/input_error_lists.py
```diff
@@ -32,7 +33,27 @@ def execute_error_protocol(self) -> str:
...
+
+    def execute_error_protocol(self) -> str:
+        raise NotImplementedError
+
+    def _sort_by_excel(self) -> dict[str, SheetProblem]:
+        raise NotImplementedError
+
+
+@dataclass
+class SheetProblem(Protocol):
```
jnussbaum, line 47: I would prefer defining this Protocol before its first usage. (It's used on line 37)

src/dsp_tools/commands/excel2json/models/input_error_lists.py
```diff
@@ -46,8 +67,37 @@ def execute_error_protocol(self) -> str:
...
+@dataclass
+class MultipleListPerSheetProblem(SheetProblem):
+    excel_name: str
+    sheet_name: str
+    list_names: list[str]
+
+    def execute_error_protocol(self) -> str:
+        return (
+            f"Per Excel sheet only one list is allowed.\n"
+            f"The following sheet: '{self.sheet_name}' has more than one list: {', '.join(self.list_names)}"
```
jnussbaum, line 94:
```suggestion
            f"The sheet '{self.sheet_name}' has more than one list, namely the following: {', '.join(self.list_names)}"
```

src/dsp_tools/commands/excel2json/models/input_error_lists.py
```diff
@@ -68,20 +119,35 @@ def execute_error_protocol(self) -> str:
...
+    sheet: str
+    node_problems: list[MissingNodeTranslationProblem]
 
     def execute_error_protocol(self) -> str:
         msg = [
-            f"The excel sheet '{self.sheet_name}' contains rows that are completely identical "
-            f"(excluding the column 'ID (optional)'). The following rows are duplicates:"
+            f"The excel sheet '{self.sheet}' has the following problem(s):\n"
+            "In one list, all the nodes must be translated into all the languages used. "
+            "The following nodes are missing translations:"
```
jnussbaum, line 132:
```suggestion
            "For the following nodes, the translations are missing:"
```

src/dsp_tools/commands/excel2json/models/input_error_lists.py
```diff
@@ -68,20 +119,35 @@ def execute_error_protocol(self) -> str:
...
         return list_separator.join(msg)
 
 
+@dataclass(frozen=True)
+class MissingNodeSheetProblem:
+    sheet: str
+    node_problems: list[NodesPerRowProblem]
+
+    def execute_error_protocol(self) -> str:
+        msg = "Each list node and each list must have its own row in the excel. " "The following rows are incorrect:"
```
jnussbaum, line 145:
```suggestion
        msg = "Each list node and each list must have its own row in the excel. The following rows are incorrect:"
```

jnussbaum (approved):

src/dsp_tools/commands/excel2json/models/input_error_lists.py
Nora-Olivia-Ammann: I can delete it, but in the next PR I will add it again...
I prefer to do the not implemented, so that you for sure do not use it directly.

jnussbaum: But then it should still be a Protocol, shouldn't it?

Nora-Olivia-Ammann: No it will be used directly. But I just did not write the functions yet. I'll delete it.

PR #1090, opened 2024-08-09
refactor(excel2json-lists): add filename to sheet class

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/new_lists.py
```diff
@@ -75,7 +75,10 @@ def _parse_files(excelfolder: Path | str) -> list[ExcelFile]:
     file_names = [file for file in Path(excelfolder).glob("*list*.xlsx", case_sensitive=False) if _non_hidden(file)]
     all_files = []
     for file in file_names:
-        sheets = [ExcelSheet(sheet_name=name, df=df) for name, df in read_and_clean_all_sheets(file).items()]
+        sheets = [
+            ExcelSheet(excel_name=file.stem, sheet_name=name, df=df)
```
jnussbaum, line 79: why only the stem and not the entire `file.name`?

src/dsp_tools/commands/excel2json/models/new_lists_deserialise.py
```diff
@@ -13,5 +13,6 @@ class ExcelFile:
 
 @dataclass
 class ExcelSheet:
+    excel_name: str
```
jnussbaum, line 16: it's a bit subpar that in the class `ExcelFile`, the same information is named `filename`. I would prefer `filename` for both classes

jnussbaum (approved):

src/dsp_tools/commands/excel2json/models/new_lists_deserialise.py
Nora-Olivia-Ammann: because the class ExcelFile will be removed in the end. So that difference will not be there in the end. And excel name feels more explicit to me.

src/dsp_tools/commands/excel2json/new_lists.py
Nora-Olivia-Ammann: Good point

PR #1093, opened 2024-08-09
refactor(excel2json-lists): remove file hierarchy from Problem classes

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/models/input_error_lists.py
```diff
@@ -40,6 +42,22 @@ def execute_error_protocol(self) -> str:
         raise NotImplementedError
 
 
+@dataclass
+class CollectedSheetProblems:
```
Nora-Olivia-Ammann, line 46: The serialisation result of this function is used in a unittest. It did not change the result to the previous serialisation

jnussbaum (approved):
I trust you on this one

PR #1094, opened 2024-08-09
refactor(excel2json-lists): remove file hierarchy from code

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/models/new_lists_deserialise.py
```diff
@@ -5,12 +5,6 @@
 import pandas as pd
 
 
-@dataclass
-class ExcelFile:
-    filename: str
-    sheets: list[ExcelSheet]
-
-
 @dataclass
```
Nora-Olivia-Ammann, line 8: In this file there is now only this class. But I want to keep that for the future. Currently checks are happening in the serialisation class. I want to change that so that it is more similar to the json header. So I think it is ok to keep this file as it is.

test/unittests/commands/excel2json/test_new_lists.py
```diff
@@ -48,17 +47,17 @@ def test_resolve_duplicates_in_all_excels(self) -> None:
...
         res = _remove_duplicate_ids_in_all_excels(["1"], all_excels)
         assert len(res) == 2
-        file_one = res[0]
-        assert file_one.filename == "file1"
-        assert file_one.sheets[0].df["id"].to_list() == ["0", "List1:Node1", "2"]
-        file_two = res[1]
-        assert file_two.filename == "file2"
-        assert file_two.sheets[0].df["id"].to_list() == ["00", "List2:Node1"]
+        sheet_one = res[0]
+        assert sheet_one.excel_name == "file1"
```
jnussbaum, line 56: You could consider renaming the attribute "excel_name" to "filename"

jnussbaum (approved):
LGTM

PR #1097, opened 2024-08-12
feat(excel2xml): add linebreak to tag converter (RDU-14)

Nora-Olivia-Ammann:


examples/xmllib/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Installation
+
+**Install Demo Version Of dsp-tools**
+
+```
+pip3 uninstall dsp-tools
+```
+
+```
+pip3 install git+https://github.com/dasch-swiss/dsp-tools.git@wip/rdu-14-replace-linebreak-with-tag
```
jnussbaum, line 12:
```suggestion
pip3 uninstall dsp-tools
pip3 install git+https://github.com/dasch-swiss/dsp-tools.git@wip/rdu-14-replace-linebreak-with-tag
```

examples/xmllib/README.md
```diff
@@ -0,0 +1,39 @@
+# Guide To Try Out the New Concept of XML Helper Functions
+
+## Installation
+
+**Install Demo Version Of dsp-tools**
```
jnussbaum, line 5:
```suggestion
**Install Demo Version Of DSP-TOOLS**
```

examples/xmllib/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Finishing Up
+
+**Re-Install dsp-tools**
+
+```
+pip3 uninstall dsp-tools
+```
+
+```
+pip3 install dsp-tools
```
jnussbaum, line 38:
```suggestion
pip3 uninstall dsp-tools
pip3 install dsp-tools
```

examples/xmllib/README.md
```diff
@@ -0,0 +1,39 @@
...
+- `:hasName` -> SimpleText (1-n)
+- `:hasStreetAdress` -> RichText (1)
+
+Optional:
+- Image or IIIF-URI
+
+
+## Finishing Up
+
+**Re-Install dsp-tools**
```
jnussbaum, line 31:
```suggestion
**Re-Install DSP-TOOLS**
```

examples/xmllib/example_excel2xml.py
```diff
@@ -0,0 +1,53 @@
...
+    res = excel2xml.make_resource(id=row["id"], restype=row["restype"], label=row["label"])
+
+    # Check if there are files or iiif-links
+    if not pd.isna(row["file"]):
+        bitstream = excel2xml.make_bitstream_prop(path=row["file"])
+        res.append(bitstream)
+
+    if not pd.isna(row["iiif-uri"]):
+        iiif = excel2xml.make_iiif_uri_prop(iiif_uri=row["iiif-uri"])
+        res.append(iiif)
```
jnussbaum, line 21: The example is already quite long. I suggest simplifying it, e.g. by removing the IIIR-URI (which is quite rare, anyways)

examples/xmllib/example_excel2xml.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+        dog_names.append(private_name)
+
+    # Make and append the <text-prop>
+    name_prop = excel2xml.make_text_prop(name=":hasName", value=dog_names)
+    res.append(name_prop)
+
+    # Add street address and keep the linebreaks
+    street_prop = excel2xml.PropertyElement(value=row["hasStreetAddress"], encoding="xml")
+    street_address = excel2xml.make_text_prop(name=":hasStreetAddress", value=street_prop)
```
jnussbaum, line 46: "prop" is the abbreviation for property, but here we have a PropertyElement, which is misleading.

```suggestion
    street_propelem = excel2xml.PropertyElement(value=row["hasStreetAddress"], encoding="xml")
    street_address = excel2xml.make_text_prop(name=":hasStreetAddress", value=street_propelem)
```

jnussbaum (commented):
Thanks for the usage example and the readme, that's very useful. See some remarks in my comments.

examples/xmllib/README.md
BalduinLandolt: sorry to be controversial... but is it really too much to ask, to assume that people always use virtual environments, so that uninstalling is not needed (and you can use `python` and `pip` instead of `python3` and `pip3`?  
If not, isn't that something we should address with people, rather than cluttering our code examples?

jnussbaum: Unfortunately, RDU is (partially) not aware of virtualenvs, and it would first require us some effort to even convince them why they should use them. And then, we would have to explain them how to. 

But actually, teaching venvs to them is on my wishlist: https://www.notion.so/dasch-swiss/RDU-Technical-Presentations-57adafcbd46049b8b8bf516f096d725f

Nora-Olivia-Ammann: I would currently leave it as it is. I did not intend for this readme to be a final version. I just copy pasted what I wrote.
Before we set up these examples, I think it is a good idea to discuss this. But I have two remarks:
- Installation should not be necessary for production examples as they should work with the latest published version.
- .venv: many prefer to have dsp-tools globally, which I understand when you think about how often they have to update it. It is unlikely that they have many other packages, so it would be a lot of maintenance work for them to use it. Additionally they may prefer to use the terminal app, and if they do we would also have to teach them how to activate a venv. That will be another hurdle. I understand why it is not ideal, but I think it is for most the best pragmatic solution.

examples/xmllib/example_excel2xml.py
Nora-Olivia-Ammann: I will not do that here because it would be a breaking change to the example I sent them. I consider this to be more of an experimental branch.

examples/xmllib/__init__.py
BalduinLandolt, line 1: correct me if I'm wrong... but for this example, the root directory would be this folder; and I think one should not have a `__init__.py` in the root folder, right?

examples/xmllib/README.md
Nora-Olivia-Ammann: I can remove that here.

examples/xmllib/__init__.py
Nora-Olivia-Ammann: one of the linters complained when i left it out.
As this is just an illustrative example and not one we will merge, I did not spend too much time on the problem.

examples/xmllib/README.md
BalduinLandolt: it doesn't hurt either... as long as we don't lose too much time on discussing setup questions, I'm happy :)

PR #1098, opened 2024-08-12
fix(xmlupload): allow non-prefixed custom groups, and custom groups with dash (DEV-3979)

jnussbaum:


testdata/xml-data/test-data-systematic.xml
jnussbaum, line 1: These 2 new lines produced an error before, now they work. I need this for Rita's LIMC upload.

Nora-Olivia-Ammann (approved):
I don't object to the code, but I am not sure if we even want to create those groups.

jnussbaum:
> I don't object to the code, but I am not sure if we even want to create those groups.

Until recently, I thought we don't need them, but for LIMC, we indeed need it. There, we have the following access levels:
- KnownUsers: view some resources
- ProjectMembers: view all resources
- limc-editors: modify + delete
- ProjectAdmins: admin

This cannot be solved without custom groups, unfortunately.

Nora-Olivia-Ammann:
Does it make sense to create these this way even though we will change it in the future? Is Rita aware of this and okayed it?

BalduinLandolt (approved):

PR #1099, opened 2024-08-12
refactor(xmlupload): write tqdm to stdout instead of stderr (DEV-3980)

jnussbaum:


Nora-Olivia-Ammann (approved):

BalduinLandolt (commented):
To be honest, I'm a bit sceptical about this one: I never gave any thought to the distinction between standard out and standard error, but after reading up on it a bit (and learning/understanding lots of new stuff), I think tqdm is right to use stderr for its output, and we shouldn't change that. Test output is a particularly bad reason for such a decision, because it would be changing user-facing behaviour (IMO for the worse) in order to mitigate issues with the development setup - which I would say is always a bad trade-off.  
Probably it doesn't actually matter too much, which output channel tqdm uses in dsp-tools, as it is not a "real CLI tool" in the sense that it doesn't produce console output. But even so, this change would break the possibility to pipe stdout without polluting the file with progress bar output.  
So, if the test reports really bother you (which I don't think they should), I would rather look for ways to conditionally silence tqdm in the test setup.

jnussbaum:
> To be honest, I'm a bit sceptical about this one: I never gave any thought to the distinction between standard out and standard error, but after reading up on it a bit (and learning/understanding lots of new stuff), I think tqdm is right to use stderr for its output, and we shouldn't change that. Test output is a particularly bad reason for such a decision, because it would be changing user-facing behaviour (IMO for the worse) in order to mitigate issues with the development setup - which I would say is always a bad trade-off. Probably it doesn't actually matter too much, which output channel tqdm uses in dsp-tools, as it is not a "real CLI tool" in the sense that it doesn't produce console output. But even so, this change would break the possibility to pipe stdout without polluting the file with progress bar output. So, if the test reports really bother you (which I don't think they should), I would rather look for ways to conditionally silence tqdm in the test setup.

Thanks for you valuable thoughts. I didn't think that much about it. The test output doesn't bother me that much. I think tqdm has its reasons for this choice, even if it surprised me at first sight. So I'll close this PR.

PR #1109, opened 2024-08-14
refactor(excel2json-lists): access columns differently

Nora-Olivia-Ammann:


test/unittests/commands/excel2json/new_lists/test_compliance_checks.py
```diff
@@ -87,11 +87,12 @@ def test_content_compliance(self, f1_s1_good_en: ExcelSheet, f2_s2_missing_trans
             "The excel sheet 'sheet2' has the following problem(s):\n"
             "In one list, all the nodes must be translated into all the languages used. "
             "For the following nodes, the translations are missing:\n"
+            "    - Row Number: 2 | Column(s): de_list\n"
             "    - Row Number: 3 | Column(s): en_1\n"
             "    - Row Number: 8 | Column(s): de_1"
         )
         with pytest.raises(InputError, match=expected):
-            _check_for_missing_translations_all_excels(all_sheets)
+            make_all_excel_compliance_checks(all_sheets)
```
Nora-Olivia-Ammann, line 95: I noticed that I erroneously called the function that does this check directly here. Although I do that elsewhere too.

src/dsp_tools/commands/excel2json/new_lists/compliance_checks.py
```diff
@@ -230,12 +233,12 @@ def _check_for_missing_translations_all_excels(sheet_list: list[ExcelSheet]) ->
...
     col_endings = [str(num) for num in get_hierarchy_nums(sheet.df.columns)]
-    col_endings.append("list")
     languages = get_all_languages_for_columns(sheet.df.columns)
     all_cols = _compose_all_combinatoric_column_titles(col_endings, languages)
     problems = []
-    for column_group in all_cols:
-        problems.extend(_check_for_missing_translations_one_column_level(column_group, sheet.df))
+    for column_group in all_cols.reverse_sorted_node_cols():
+        problems.extend(_check_for_missing_translations_one_column_level(column_group.columns, sheet.df))
+    problems.extend(_check_for_missing_translations_one_column_level(all_cols.list_cols.columns, sheet.df))
```
jnussbaum, line 241: I don't understand why this additional line (after the for loop) is necessary.

Nora-Olivia-Ammann: Because i loop over the columns of the nodes, and then I check the columns for the list itself. These are separate because of a change that I am planning on in the future. Basically the big picture plan here is that I want to change how the nodes are deserialised. This should then make the comments easier. But again this will be a big diff and i am trying to cut it up. So it is possible that some things seem a bit weird for the moment.

PR #1112, opened 2024-08-15
refactor(excel2json-lists): change find missing translations

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/new_lists/compliance_checks.py
```diff
@@ -43,9 +43,10 @@ def make_all_excel_compliance_checks(sheet_list: list[ExcelSheet]) -> None:
     # These functions must be called in this order,
     # as some of the following checks only work if the previous have passed.
     _check_duplicates_all_excels(sheet_list)
-    _check_for_unique_list_names(sheet_list)
     _make_shape_compliance_all_excels(sheet_list)
-    _make_all_content_compliance_checks_all_excels(sheet_list)
+    _check_for_missing_translations_all_excels(sheet_list)
+    _check_for_unique_list_names(sheet_list)
```
Nora-Olivia-Ammann, line 48: The order has to change because the unique list names assume that there is no missing translation. This problem was not caught because I missed it in a test.

src/dsp_tools/commands/excel2json/new_lists/compliance_checks.py
```diff
@@ -236,23 +231,31 @@ def _check_for_missing_translations_one_sheet(sheet: ExcelSheet) -> MissingTrans
...
-def _check_for_missing_translations_one_column_level(
-    columns: list[str], df: pd.DataFrame
-) -> list[MissingNodeTranslationProblem]:
-    # column level refers to the hierarchical level of the nodes. eg. ["en_1", "de_1", "fr_1", "it_1", "rm_1"]
-    problems = []
-    for i, row in df.iterrows():
-        if problem := _check_for_missing_translations_one_node(row, columns, int(str(i))):
-            problems.append(problem)
-    return problems
+def _check_missing_translations_one_row(
```
Nora-Olivia-Ammann, line 242: I changed it to row level because as a user it is nicest to have all the missing translations sorted by row and one row should not be there more than once. The comments will also be easy to integrate this way.

PR #1117, opened 2024-08-16
chore: send X-Asset-Ingested header only when needed

jnussbaum:


src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -62,7 +62,8 @@ def create_resource(
             f"Attempting to create resource {resource.res_id} (label: {resource.label}, iri: {resource.iri})..."
         )
         resource_dict = self._make_resource_with_values(resource, bitstream_information)
-        res = self.con.post(route="/v2/resources", data=resource_dict, headers={"X-Asset-Ingested": "true"})
+        headers = {"X-Asset-Ingested": "true"} if bitstream_information else None
```
Nora-Olivia-Ammann, line 65: Is it a problem to always provide it?

jnussbaum: Currently, it doesn't break anything, but it is a wrong information that is sent to the server.

Nora-Olivia-Ammann (approved):

BalduinLandolt (approved):

PR #1121, opened 2024-08-17
chore: small adaptions to docs + docstrings

jnussbaum:


docs/excel2xml-module.md
```diff
@@ -196,7 +196,7 @@ here](./file-formats/xml-data-file.md#dsp-base-resources-and-base-properties-to-
 #### Resource ID
 
 Special care is needed when the ID of a resource is created. Every resource must have an ID that is unique in the file,
-and it must meet the constraints of xsd:ID. You can simply achieve this if you use the method `make_xsd_id_compatible()`.
+and it must meet the constraints of xsd:ID. You can simply achieve this if you use the function `make_xsd_id_compatible()`.
```
Nora-Olivia-Ammann, line 199: You should mention here that it also adds a UUID based on the timestamp which cannot be reconstructed later.

docs/excel2xml-module.md
```diff
@@ -266,7 +266,7 @@ For `make_boolean_prop(cell)`, the following formats are supported:
 - false: False, "false", "False", "0", 0, "no", "No"
 
 N/A-like values will raise an Error. 
-So if your cell is empty, this method will not count it as false, 
+So if your cell is empty, this function will not count it as false,
```
Nora-Olivia-Ammann, line 269: Why do you change it from method to function here and above the reverse? According to this: https://www.shiksha.com/online-courses/articles/difference-between-methods-and-functions-in-python/ currently we do not at all offer methods in `excel2xml-lib`.

docs/file-formats/xml-data-file.md
```diff
@@ -986,15 +986,15 @@ shortcut, so that the XML file can be validated more precisely.
...
-| `hasTitle`         | 0-1         | text-prop     | utf8     |
-| `hasDescription`   | 0-n         | text-prop     | xml      |
-| `hasComment`       | 0-n         | text-prop     | xml      |
-| `hasKeyword`       | 0-n         | text-prop     | utf8     |
-| `relatesTo`        | 0-n         | resptr-prop   |          |
+| prop name          | cardinality | prop type     | text type   |
+| ------------------ | ----------- | ------------- | ----------- |
+| `isSegmentOf`      | 1           | resptr-prop   |             |
+| `hasSegmentBounds` | 1           | interval-prop |             |
+| `hasTitle`         | 0-1         | text-prop     | unformatted |
```
Nora-Olivia-Ammann, line 993: Perhaps write the encoding in the brackets. Till we made that distinction without them having to enter the encoding, I would mention it in the documentation.

Nora-Olivia-Ammann (changes requested):

docs/excel2xml-module.md
jnussbaum: The change I did was from method -> function, but not the reverse. Function is the correct term.

There is no change in this PR that turns function -> method.

Nora-Olivia-Ammann (approved):

docs/excel2xml-module.md
jnussbaum: May I delegate this work to you, in https://linear.app/dasch/issue/DEV-4026/featexcel2xml-new-function-make-xsd-id-compatible-without-uuids?

docs/file-formats/xml-data-file.md
jnussbaum: Here, they don't have to enter the encoding, so it's better to omit the encoding

PR #1122, opened 2024-08-19
chore: bump dependencies and fix new linter errors

jnussbaum:


pyproject.toml
```diff
@@ -230,4 +230,3 @@ convention = "google"
 
 [tool.ruff.lint.isort]
 force-single-line = true
-known-first-party = ["dsp_tools"]
```
jnussbaum, line 233: Not necessary any more in ruff >= 0.6.0

test/conftest.py
```diff
@@ -8,7 +8,7 @@
 from dsp_tools.utils.logger_config import logger_config
 
 
-@pytest.fixture()
+@pytest.fixture
```
jnussbaum, line 11: Rule change in ruff 0.6.0. As per the release notes (https://astral.sh/blog/ruff-v0.6.0#default-behavior-changed-for-flake8-pytest-style-rules), the new rule adheres more closely to the style recommendations made by the official pytest project

test/unittests/commands/excel2json/test_utils.py
```diff
@@ -200,24 +200,24 @@ def test_get_comments(self) -> None:
 def test_add_optional_columns_with_missing_cols() -> None:
     original_df = pd.DataFrame(
         {
-            "comment_en": ["text_en", np.NaN],
-            "comment_it": ["text_it", np.NaN],
-            "comment_rm": [np.NaN, np.NaN],
+            "comment_en": ["text_en", np.nan],
+            "comment_it": ["text_it", np.nan],
+            "comment_rm": [np.nan, np.nan],
```
jnussbaum, line 205: `np.NaN` was removed in the NumPy 2.0 release. Use `np.nan` instead.

pyproject.toml
```diff
@@ -64,7 +64,7 @@ pytest-sugar = "^1.0.0"
 pandas-stubs = "^2.2.1.240316"
 types-networkx = "^3.2.1.20240331"
 requests-mock = "^1.12.1"
-testcontainers =  "^4.4.1"
+testcontainers =  "4.7.2"
```
jnussbaum, line 67: There's a problem with the newest version which produces a fuseki container that looks healthy, but refuses any incoming HTTP connections. Once this PR is merged, I have to investigate this, so that later, we can also upgrade testcontainers.

jnussbaum: EDIT: It seems as if testcontainers >= 4.8.0 is buggy. My investigations showed a similar behavior than described in https://github.com/testcontainers/testcontainers-python/issues/633. But I'm unable to work around it, so I'm afraid we will have to stick with testcontainers == 4.7.2.

Nora-Olivia-Ammann (approved):

PR #1124, opened 2024-08-19
feat(xml-validate): base branch for the validation experiments (DEV-4024)

Nora-Olivia-Ammann:
The following types of errors are chosen for this POC based on feedback from RDU.

- [Maximum Cardinality Violation](https://github.com/dasch-swiss/dsp-tools/blob/99a4d3b852e01ab29a9f15efa15ce4d46877b709/testdata/xml-validate/invalid-data.xml#L37)
- [List node not found](https://github.com/dasch-swiss/dsp-tools/blob/99a4d3b852e01ab29a9f15efa15ce4d46877b709/testdata/xml-validate/invalid-data.xml#L74)
- [Target class of the link value is not as specified in the ontology](https://github.com/dasch-swiss/dsp-tools/blob/99a4d3b852e01ab29a9f15efa15ce4d46877b709/testdata/xml-validate/invalid-data.xml#L56)
- [Type of value (eg. ListValue) in the data is not the same as in the ontology](https://github.com/dasch-swiss/dsp-tools/blob/99a4d3b852e01ab29a9f15efa15ce4d46877b709/testdata/xml-validate/invalid-data.xml#L66)
- [One resource has the identical value twice.](https://github.com/dasch-swiss/dsp-tools/blob/99a4d3b852e01ab29a9f15efa15ce4d46877b709/testdata/xml-validate/invalid-data.xml#L83)


src/dsp_tools/cli/create_parsers.py
```diff
@@ -296,6 +298,24 @@ def _add_xmlupload(
...
 
+def _add_xml_validate(
+    subparsers: _SubParsersAction[ArgumentParser],
+    default_dsp_api_url: str,
+    root_user_email: str,
+    root_user_pw: str,
+) -> None:
+    subparser = subparsers.add_parser(name="xml-validate", help="Validate an XML, no upload of the data occurs.")
+    subparser.add_argument("-u", "--user", default=root_user_email, help=username_text)
+    subparser.add_argument("-p", "--password", default=root_user_pw, help=password_text)
```
BalduinLandolt, line 309: detail: user and password are probably not meaningful for a "validate" command

src/dsp_tools/cli/call_action.py
```diff
@@ -194,6 +199,15 @@ def _call_ingest_xmlupload(args: argparse.Namespace) -> bool:
     )
 
 
+def _call_xml_validate(
+    # args: argparse.Namespace
+) -> bool:
+    success = xml_validate(
+        # xml_file=Path(args.xml_file), creds=_get_creds(args), imgdir=Path(args.imgdir)
+    )
```
BalduinLandolt, line 207: suggestion: for a simple (and of course not useable) setup for comparison, why not just hard-code the xml file, so that you never have to provide any params in the CLI, simply `dsp-tools xml-validate`?

src/dsp_tools/commands/xml_validate/api_connection.py
BalduinLandolt, line 1: like discussed previously, for the sake of this comparison, I would leave the API out of the play entirely: the SHACL can be written by hand and provided as a file; the ontology json file can be kept from when you create the test project, and then also be provided as a file.  
For the context of this PoC you shouldn't have to deal with the API; and even more so, adding unnecessary code like that, will obfuscate how the two compared approaches influence the code.

src/dsp_tools/commands/xml_validate/deserialise_api_responses.py
BalduinLandolt, line 1: same as above

src/dsp_tools/commands/xml_validate/models/api_response_deserialised.py
BalduinLandolt, line 1: could this be moved to the `src/dsp_tools/commands/xml_validate/models/deserialised.py` file?

src/dsp_tools/commands/xml_validate/models/xml_deserialised.py
BalduinLandolt, line 1: what is the difference between a `ProjectXML` and a `ProjectDeserialised`? Parsing XML is just as much deserialisation as parsing JSON or RDF

src/dsp_tools/commands/xml_validate/prepare_input.py
```diff
@@ -0,0 +1,150 @@
...
+def parse_file(file: Path) -> ProjectXML:
+    """Returns an object which follows the structure of the XML closely"""
+    root = _parse_file_validate_with_schema(file)
+    return _transform_into_xml_deserialised(root)
+
+
+def _parse_file_validate_with_schema(file: Path) -> etree._Element:
+    root = parse_and_clean_xml_file(file)
+    validate_xml(root)
+    return root
```
BalduinLandolt, line 30: to me, this seems redundant: in this experiment, the XML schema validation should never fail, otherwise we're violationg the experiment setup; and if that's the case, why validate?

src/dsp_tools/commands/xml_validate/prepare_input.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def _transform_into_xml_deserialised(root: etree._Element) -> ProjectXML:
+    shortcode = root.attrib["shortcode"]
+    default_ontology = root.attrib["default-ontology"]
+    permissions = [_get_permissions(x) for x in root.iterdescendants(tag="permissions")]
```
BalduinLandolt, line 36: are permissions ever relevant in the experiment setup? if not, I'd simply drop them entirely from the model and here from the deserialisation

src/dsp_tools/commands/xml_validate/prepare_input.py
```diff
@@ -0,0 +1,150 @@
...
+    return ResourceXML(res_attrib=ele_attribs, values=values)
+
+
+def _get_permissions(permission_ele: etree._Element) -> PermissionsXML:
+    per_id = permission_ele.attrib["id"]
+    permissions = list(permission_ele.iterchildren())
+    return PermissionsXML(permission_id=per_id, permission_eles=permissions)
+
+
+def deserialise_xml_project(project: ProjectXML) -> ProjectDeserialised:
```
BalduinLandolt, line 55: this naming is misleading -`ProjectXML` is already deserialised. Also, I'm not yet really convinced that these two separate steps provide any benefits. But I'm happy to be convinced otherwise

src/dsp_tools/commands/xml_validate/prepare_input.py
```diff
@@ -0,0 +1,150 @@
...
+                prop_value=txt,
+                list_name=list_name,
+                permissions=val.attrib.get("permissions"),
+                comments=val.attrib.get("comment"),
+            )
+        )
+    return all_vals
+
+
+def _deserialise_resptr_prop(prop: etree._Element) -> list[ValueDeserialised]:
```
BalduinLandolt, line 137: I would be very happy if we could stop using the term `resptr` in any new code we write. It's a "Link (Value)" in our domain language, and code should follow domain terminology

test/unittests/commands/xml_validate/conftest.py
BalduinLandolt, line 1: what is this file about?

test/unittests/commands/xml_validate/input_data_fixtures.py
BalduinLandolt, line 1: I don't think these kinds of "on the edge of the application" models that do not contain any logic are worth testing.

testdata/xml-validate/invalid-data.xml
```diff
@@ -0,0 +1,99 @@
...
+    </resource>
+
+
+    <!-- Ontology Violations -->
+
+    <resource label="Link Value Target" restype=":NormalClass" id="link-value-target-wrong"/>
+
+    <resource label="Target Link Class Mismatch" restype=":NormalClass" id="link-value-wrong">
+        <text-prop name=":hasSimpleText">
+            <text encoding="utf8">The target of the link should be of the class CardOneResource</text>
```
BalduinLandolt, line 51: suggestion: how about moving the description of what is invalid to comments in the XML? this would work for non-text values as well as text values.

BalduinLandolt (commented):
Over all this looks promising!

I was rather detailed/strict when looking through the code. See what makes sense to you and feel free to ignore everything else.

In terms of test cases/violations, I'm still missing inheritance issues, but maybe I missed that (ran out of time in the end).

src/dsp_tools/cli/call_action.py
Nora-Olivia-Ammann: Sure I can do that. I wanted to simulate the real world implementation, but I see the point. Would you suggest I create this without any arguments in the parser?

src/dsp_tools/commands/xml_validate/models/api_response_deserialised.py
Nora-Olivia-Ammann: I am planning on adding the ontology deserialised here. But that will differ based on the approach, which is why I did not add the classes here. The lists will not, which is why I implemented it already.

src/dsp_tools/commands/xml_validate/models/xml_deserialised.py
Nora-Olivia-Ammann: It goes from `ProjectXML` (has the structure of the data XML still) -> `ProjectDeserialised` (Structure of the XML is largely removed and is more abstract) And the last classes in the chain would be the format used for validation. I did not write that because it will be different. But I am happy to rename them if you prefer, what do you suggest?

src/dsp_tools/commands/xml_validate/prepare_input.py
Nora-Olivia-Ammann: I added that to make sure that the test data is for sure correct and a schema mismatch will not be the cause. And since I did not have to write this code it did not cause any work. Do you want me to remove it?

src/dsp_tools/commands/xml_validate/prepare_input.py
Nora-Olivia-Ammann: I want to include them in the validation. No permissions should be entered that are not specified in the XML. But since I did not plan on violating that in the test data I can remove it. What do you think?

src/dsp_tools/commands/xml_validate/prepare_input.py
Nora-Olivia-Ammann: I understand why it seems redundant. I was thinking about doing this in one step too, but I did not want to do too much logic at the same time. What I could do is when I implement the two versions that I already do some changes here for example make the props into IRIs for SHACL etc. What do you think?

src/dsp_tools/commands/xml_validate/prepare_input.py
Nora-Olivia-Ammann: I agree in general and disagree in this specific case.
The reason why I included `resptr` in this function name is because it takes an element with that tag. This XML syntax is being removed here and the output corresponds to the domain language. This is what I wanted to reflect in the function name. But if you feel strongly about it I can rename it.

test/unittests/commands/xml_validate/conftest.py
Nora-Olivia-Ammann: This imports makes the `pytest.fixtures` in all the files of this folder available without having to write import statements. I like to have the fixtures in a separate file (easier to read and better if used in several test files). When I did that elsewhere I had trouble with the import statements and it could be solved this way. We already do this for `excl2json`

test/unittests/commands/xml_validate/input_data_fixtures.py
Nora-Olivia-Ammann: What do you mean by that? I would like a test to know that I deserialised the lists correctly etc.

src/dsp_tools/cli/call_action.py
BalduinLandolt: yes, keep it as simple as possible.

src/dsp_tools/commands/xml_validate/models/xml_deserialised.py
BalduinLandolt: I'm really sceptical about the `ProjectXML` part: Having classes holding `lxml.etree._Element` objects seems like an anti-pattern to me. These will never be nicely testable, and they do not add any value that I can see.

src/dsp_tools/commands/xml_validate/prepare_input.py
BalduinLandolt: I leave it up to you - but in the context of an experiment regarding validation, seeing validation that is not relevant to the experiment seems a bit confusing to me. But knowing this now, it's also easy enough to ignore

src/dsp_tools/commands/xml_validate/prepare_input.py
BalduinLandolt: I would prefer to have `ProjectDeserialized` be the first step in the pipeline and kick out the `ProjectXML` layer, so that we do not have any XML elements in code. Then it is consistent with the naming too, as it deserializes the input into an un-opinionated object. Any further step should then transform it into an object that indicates what was done (like IRI resolving etc.).

src/dsp_tools/commands/xml_validate/api_connection.py
Nora-Olivia-Ammann: I kept the class but read the files instead is that fine for you?

src/dsp_tools/commands/xml_validate/deserialise_api_responses.py
Nora-Olivia-Ammann: I would like to leave the deserialise api responses because based on the approach i will deserialise them differently... Do you mean that I should return the Deserialised list object directly?

src/dsp_tools/commands/xml_validate/api_connection.py
BalduinLandolt: in principal, sure - but then the naming makes little sense `api_connection.py` and `ProjectClient` are hardly what it is, and in fact, the two methods of the client can simply be top-level functions

src/dsp_tools/commands/xml_validate/deserialise_api_responses.py
BalduinLandolt: I think my main point here was that the PoC should not need the API at all - so at least the naming is misleading.

src/dsp_tools/commands/xml_validate/models/api_response_deserialised.py
BalduinLandolt: I'm just failing to understand what `models/api_response_deserialised.py` and `models/deserialised.py` are, and how they differ.

testdata/xml-validate/project.json
```diff
@@ -0,0 +1,140 @@
...
+            "cardinalities": [
+              {
+                "propname": ":hasSimpleText",
+                "cardinality": "1",
+                "gui_order": 0
+              }
+            ]
+          },
+          {
+            "name": "SubClass",
```
Nora-Olivia-Ammann, line 130: Regarding the inheritance I added no cardinalities, but in the data a resource of this class has the inherited cardinality. Is this not what you meant @BalduinLandolt ?

test/unittests/commands/xml_validate/input_data_fixtures.py
BalduinLandolt: Fair enough. I normally tend to find deserialisation something that is hardly worth testing; but I'm in favour of thorough testing, and if you feel it's worth doing, then I'm not going to object to that.

testdata/xml-validate/invalid-data.xml
```diff
@@ -0,0 +1,76 @@
...
+       default-ontology="onto">
+
+    <!-- Cardinality Violation -->
+
+    <resource label="Max Cardinality Violation" restype=":SubClass" id="max-card-violation">
+        <text-prop name=":hasSimpleText">
+            <text encoding="utf8">Text Value One</text>
+            <text encoding="utf8"
+                  comment="This property has a cardinality 1"
+            >Text Value Two</text>
```
BalduinLandolt, line 16:
```suggestion
            <text encoding="utf8">Text Value Two</text> <!-- Violation: Property has cardinality 1, but two values were supplied -->
```

I think you misunderstood my suggestion of "using comments for indicating what the violation is", sorry about that.

src/dsp_tools/commands/xml_validate/deserialise_api_responses.py
Nora-Olivia-Ammann: I agree but even if saved in a file it is still a response from the api...

testdata/xml-validate/invalid-data.xml
```diff
@@ -0,0 +1,76 @@
...
+            >Text</text>
+        </text-prop>
+    </resource>
+
+    <resource label="Target Link Class Correct" restype=":NormalClass" id="link-value-correct">
+        <resptr-prop name=":linkProp">
+            <resptr comment="This is correct and should not appear in the validation errors."
+            >link-value-target-correct</resptr>
+        </resptr-prop>
+    </resource>
```
BalduinLandolt, line 46: is it necessary to have "correct cases" here?

Nora-Olivia-Ammann: I wanted to have a class that was not itself problematic. The reason why I included it as correct, is because the `ObjectClassConstraint` of the property is the superclass of `:SubClass`. This is here to ensure that this will not fail in the validation.

testdata/xml-validate/project.json
Nora-Olivia-Ammann: Well having subclasses that inherit the cardinality of the superclass is quite common, so that needs to be included. I can have the sub-class with a correct cardinality, so that again we have an example that should not appear and then a cardinality violation of the superclass. What do you think?

```
    <!-- This property has a cardinality 1 -->
    <resource label="Max Cardinality Violation" restype=":CardOneResource" id="max-card-violation">
            <text encoding="utf8">Text Value One</text>
            <text encoding="utf8">Text Value Two</text>
        </text-prop>
    </resource>

    
    <!-- This cardinality restriction is inherited from the superclass. 
           The example is correct and should not appear on the error report. 
           -->
    <resource label="Max Cardinality Violation" restype=":SubClass" id="inherited-cardinality-correct">
            <text encoding="utf8">Text Value One</text>
        </text-prop>
    </resource>
```

src/dsp_tools/commands/xml_validate/models/api_response_deserialised.py
BalduinLandolt: yes, that would make it clearer - but I'm still not convinced about the `models/api_response_deserialised.py`. Maybe this could be renamed to `models/project.py` because it contains the project specific reference information.

src/dsp_tools/commands/xml_validate/deserialise_api_responses.py
BalduinLandolt: While it's not coming from the API, it's not an API response. And naming it that in the context of a setup that does not include an API seems confusing to me.

testdata/xml-validate/invalid-data.xml
BalduinLandolt: ok

testdata/xml-validate/project.json
Nora-Olivia-Ammann: It is a good point that it bloats the data. To ensure that "complex" things will pass, I will make a file that is correct and should never fail. You are right that a mixture is not good.

PR #1127, opened 2024-08-20
feat(exce2xml-lib): make xsd compatible string without uuid (DEV-4026)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -52,22 +52,38 @@ def make_xsd_id_compatible(string: str) -> str:
...
+    res = xsd_id_remove_illegal_characters(string)
+    # add uuid
+    _uuid = uuid.uuid4()
+    res = f"{res}_{_uuid}"
+    return res
+
+
+def xsd_id_remove_illegal_characters(string: str) -> str:
+    """
+    This method replaces the illegal characters by "_" and appends a random component to the string to make it unique.
```
jnussbaum, line 66: It's not a method (should also be corrected above)

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -52,22 +52,38 @@ def make_xsd_id_compatible(string: str) -> str:
         BaseError: if the input cannot be transformed to an xsd:ID
 
     Returns:
-        an xsd ID based on the input string
+        an xsd ID based on the input string with a UUID attached.
```
jnussbaum, line 55:
```suggestion
        an xsd ID based on the input string, with a UUID attached.
```

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -52,22 +52,38 @@ def make_xsd_id_compatible(string: str) -> str:
...
     """
+    res = xsd_id_remove_illegal_characters(string)
+    # add uuid
```
jnussbaum, line 58:
```suggestion
```

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -52,22 +52,38 @@ def make_xsd_id_compatible(string: str) -> str:
...
+    _uuid = uuid.uuid4()
+    res = f"{res}_{_uuid}"
+    return res
+
+
+def xsd_id_remove_illegal_characters(string: str) -> str:
```
jnussbaum, line 64: IMHO it is not clear enough how the old and the new function relate to each other. I suggest renaming the old into `make_xsd_id_compatible_with_uuid()`, and the new into `make_xsd_id_compatible()`.

The 2 should have similar names, so that people realize that they do similar things.

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -52,22 +52,38 @@ def make_xsd_id_compatible(string: str) -> str:
...
+    """
```
jnussbaum, line 65: Some more background info is necessary:

```suggestion
    """
    Make a string compatible with the constraints of xsd:ID, so that it can be used as "id" attribute of a <resource>
    tag. An xsd:ID must not contain special characters, and it must be unique in the document.
```

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
jnussbaum, line 66:
```suggestion
    This method replaces the illegal characters by "_".
```

jnussbaum (changes requested):

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
Nora-Olivia-Ammann: I will not rename the old one because that would be a breaking change.

jnussbaum: Then we should rename the new one, e.g. into `make_xsd_id_compatible_without_uuid()`

src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -52,22 +52,38 @@ def make_xsd_id_compatible(string: str) -> str:
...
-        an xsd ID based on the input string
+        an xsd ID based on the input string, with a UUID attached.
     """
+    res = make_xsd_id_compatible_string_without_uuid(string)
+    _uuid = uuid.uuid4()
+    res = f"{res}_{_uuid}"
+    return res
+
+
+def make_xsd_id_compatible_string_without_uuid(string: str) -> str:
```
jnussbaum, line 63:
```suggestion
def make_xsd_id_compatible_without_uuid(string: str) -> str:
```

jnussbaum (approved):

PR #1129, opened 2024-08-21
feat(xmllib): new functionality (RDU-38)

Nora-Olivia-Ammann:


examples/xmllib/main.py
```diff
@@ -0,0 +1,77 @@
...
+from pathlib import Path
+
+import pandas as pd
+
+from dsp_tools import xmllib
+
+data_folder = Path("examples/xmllib/data")
+
+
+def _make_books(res_collection: xmllib.ResourceCollection) -> None:
```
BalduinLandolt, line 13: this still assumes a "mutable" pattern

examples/xmllib/main.py
```diff
@@ -0,0 +1,77 @@
...
+    res = xmllib.Resource(res_id=row["id"], restype=":Book", label=row["label"])
+    titles = row["hasTitle"].split("||")
+    titles = [x.strip() for x in titles]
+    res.with_SeveralSimpleTexts(
+        prop_name=":hasTitle",
+        values=titles,
+    ).with_IntValue(
+        prop_name=":hasNumberOfPages",
+        value=row["hasNumberOfPages"],
+    ).with_SimpleText(prop_name=":hasAuthor", value=row["hasAuthor"])
```
BalduinLandolt, line 30: cosmetics: IMO this is far more readable like this:

```suggestion
    res.with_SeveralSimpleTexts(prop_name=":hasTitle", values=titles)
        .with_IntValue(prop_name=":hasNumberOfPages", value=row["hasNumberOfPages"])
        .with_SimpleText(prop_name=":hasAuthor", value=row["hasAuthor"])
```
(same goes for other cases below)

examples/xmllib/main.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def main() -> None:
+    """Creates an XML file from the csv files in the data folder."""
+    resources = xmllib.ResourceCollection()
+    _make_books(resources)
+    _make_chapters(resources)
+    _add_images(resources)
+    root = xmllib.XMLRoot(shortcode="0001", default_ontology="onto", resource_collection=resources)
+    root.write_file("examples/xmllib/data.xml")
```
BalduinLandolt, line 73: this still doesn't use the fluent interface pattern. I would say it should be like this:

```suggestion
    xml = xmllib.XMLRoot(shortcode="0001", default_ontology="onto")
        .add(_make_books())
        .add(_make_chapters())
        .add(_make_images())
    xml.write_file("examples/xmllib/data.xml") 
```

then you could even do 

```python
    xmllib.XMLRoot(shortcode="0001", default_ontology="onto")
        .add(_make_books())
        .add(_make_chapters())
        .add(_make_images())
        .write_file("examples/xmllib/data.xml") 
```

BalduinLandolt: but in terms of architecture (and testability) it would be better to have it like this:

```python
def main() -> None:
    books_data = _load_books_data()
    chapters_data = ...
    images_data = ...
    xmllib.XMLRoot(shortcode="0001", default_ontology="onto")
        .add(_make_books(books_data))
        .add(_make_chapters(chapters_data))
        .add(_make_images(images_data))
        .write_file("examples/xmllib/data.xml") 
```

src/dsp_tools/commands/xmllib/models/resource.py
```diff
@@ -0,0 +1,175 @@
...
+from dsp_tools.models.exceptions import InputError
+from dsp_tools.utils.uri_util import is_iiif_uri
+
+XML_NAMESPACE_MAP = {None: "https://dasch.swiss/schema", "xsi": "http://www.w3.org/2001/XMLSchema-instance"}
+DASCH_SCHEMA = "{https://dasch.swiss/schema}"
+
+
+@dataclass
+class ResourceCollection:
+    resources: dict[str, Resource] = field(default_factory=dict)
```
BalduinLandolt, line 25: I would simply model this as a list of resources

src/dsp_tools/commands/xmllib/models/resource.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def to_list(self) -> list[Resource]:
+        return list(self.resources.values())
+
+    def with_Resource(self, resource: Resource) -> ResourceCollection:
```
BalduinLandolt, line 30:
```suggestion
    def with_resource(self, resource: Resource) -> ResourceCollection:
```
python naming conventions :)

src/dsp_tools/commands/xmllib/models/resource.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def add_ValueToResource(self, res_id: str, value: Value) -> ResourceCollection:
+        self._raise_does_not_exist(res_id)
+        self.resources[res_id].values.append(value)
+        return self
+
+    def add_FileValueToResource(self, res_id: str, filepath: str) -> ResourceCollection:
+        self._raise_does_not_exist(res_id)
+        self.resources[res_id].filepath = filepath
+        return self
```
BalduinLandolt, line 48: I wouldn't expose this functionality - when something is added, it should not be modified anymore

src/dsp_tools/commands/xmllib/models/resource.py
```diff
@@ -0,0 +1,175 @@
...
+        return [self._combine_values(prop_values) for prop_values in grouped.values()]
+
+    @staticmethod
+    def _combine_values(prop_values: list[Value]) -> etree._Element:
+        prop_ = prop_values[0].make_prop()
+        prop_eles = [x.make_element() for x in prop_values]
+        prop_.extend(prop_eles)
+        return prop_
+
+    def with_IntValue(
```
BalduinLandolt, line 134:
```suggestion
    def with_int_value(
```
naming conventions...

BalduinLandolt (commented):
still not quite what I imagined... I hope my comments make it clearer.

additionally, there are some features that make this a very compelling pattern which are not yet in there, which I would like to see:

operator variants or overloads:

```python
# variants
one_book: Resource = ...
multiple_books: ResourceCollection = ...
xml.add_resource(one_book)
    .add_resources(multiple_books)

# overloads
one_book: Resource = ...
multiple_books: ResourceCollection = ...
xml.add_resource(one_book)
    .add_resource(multiple_books)
```

Either of those are good. They both allow to to handle different things in a consistent way.

Optionality:

```pyhton
book_optional: Resource | None = ...
xml.add_optional(book_optional) # if None is provided, the Root remains unchanged, otherwise the Resource is added
```

This allows the user to have functions returning things optionally, depending on the input data, without having to break the chaining to do None checks.

... probably there are more things, but I can't think of them right now.

examples/xmllib/main.py
Nora-Olivia-Ammann: I agree but when I ran ruff it was either this or something really weird...

examples/xmllib/main.py
Nora-Olivia-Ammann: what do you imagine `books_data` to look like?

Nora-Olivia-Ammann: I see that you would like the fluent interface pattern everywhere. I am concerned that RDU won't like that. For example some prefer to try out only one class, or they want a separate file for each class (because it is easier to read when the file is super big). So they have to be able to add data without being forced to chain (I just noticed that I don't provide that optionality, but that is a detail).

src/dsp_tools/commands/xmllib/models/resource.py
Nora-Olivia-Ammann: I understand the concern with modification. We can just add a class for the FileValue. But I think this functionality is particularly useful, because it allows you to add files later when all the resources were created.

src/dsp_tools/commands/xmllib/models/resource.py
Nora-Olivia-Ammann: I modeled it as a dictionary because I want to be able to easily add further values to an existing resource at any time in the code. If it were a list I belive the code to add a value to an existing resource is more complicated. It is also easier to check if an id already exists in my opinion. What is exactly the reason why you prefer a list?

src/dsp_tools/commands/xmllib/models/resource.py
Nora-Olivia-Ammann: Do I understand your comment that it should not be modified, in the sense that you would just add values to a list with the resource id and then combine them only when you serialise it but not actually append it to the resource?

examples/xmllib/main.py
jnussbaum: There is never an urge to chain. AFAIU, you are always free to either chain or not, that's the big advantage. Because the object is modified in place, plus it is returned. So either you rely on the returning, by chaining your functions, or you rely on in-place-modification, by NOT chaining.

BalduinLandolt: I am very sceptical about in-place modification. Normally I would be 100% against it, but I do see that this may lead to errors when people try to do that (though the pattern shouldn't be unfamiliar: `str.replace()` is also not modifying in place).

But I very much agree that it the implementation has to be done in such a way that no one is forced to do chaining.  
But in order to convey the concept of the fluent interface pattern - and from my understanding, that's what this PR is about - this pattern should be used as consistently and ubiquitously as possible, to make it clear what it is about.

BalduinLandolt: > what do you imagine `books_data` to look like?

that would be the dataframe loaded from the CSV.  
maybe something like:

```python
books_data = _load_data("books.csv")
```

the point of that being that `_make_books()` then becomes a pure function that has no side effects and doesn't depend on the file system, which makes it a breeze to test.

src/dsp_tools/commands/xmllib/models/resource.py
BalduinLandolt: to keep it simple and because I do not think it should be the concern of the `ResourceCollection` to modify resources. In fact, I very much believe that modifying existing data is an anti-pattern that should be avoided and discouraged as much as possible. In that sense, anything that should be added through the fluent interface should be "finished" and not need any modification.  
`ResourceCollection` as I envisioned it, should be an implementation detail of the fluent interface, that users ideally shouldn't ever notice it exists. Exposing any functionality beyond what the fluent interface needs, works against that, and then the users find themselves trying to understand what "the deal" is with this resource collection class, that in their eyes could just as well be a list/dict.

src/dsp_tools/commands/xmllib/models/resource.py
BalduinLandolt: no, my point is that no resource should end in a resource collection when it's not yet finished.

examples/xmllib/main.py
```diff
@@ -0,0 +1,64 @@
...
+        .add_simple_text(prop_name=":hasTitle", value=row["hasTitle"])
+        .add_integer(prop_name=":hasChapterNumber", value=row["hasChapterNumber"])
+        .add_link(prop_name=":isPartOfBook", value=row["isPartOfBook"])
+    )
+
+
+def main() -> None:
+    """Creates an XML file from the csv files in the data folder."""
+    resources = []
+    resources.extend(_make_books())
```
BalduinLandolt, line 57:
```suggestion
    resources = _make_books()
```

examples/xmllib/main.py
```diff
@@ -0,0 +1,64 @@
...
+    resources.extend(_make_chapters())
+    root = xmllib.XMLRoot(shortcode="0001", default_ontology="onto", resource_collection=resources)
+    root.write_file("examples/xmllib/data.xml")
```
BalduinLandolt, line 60: I understand the argument that you want to make it so that chaining is optional and no one has to use it. But if we allow it in some places, it should be possible everywhere - so `root.add_resource(some_resource)` should probably be possible too. What do you think?

BalduinLandolt (approved):
two minor comments from my side, otherwise looks good

examples/xmllib/main.py
Nora-Olivia-Ammann: That is a good point. I will implement that.

src/dsp_tools/commands/xmllib/different_approaches/factory_option.py
```diff
@@ -0,0 +1,72 @@
...
+Advantage: 
+    Creation and usage separated
+    Res ID is included in the warning
+Disadvantage: 
+    Value classes can be instantiated without check (otherwise we would get duplicate warnings)
+    Each value also needs a dedicated factory.
+"""
+
+
+class IntValueFactory:
```
jnussbaum, line 22: In its current form, I dislike this approach. But I think there is a way to make it better: 
 - I would not have a separate class `IntValueFactory`, but instead add the `create()` and `create_several()` methods as static methods to the `IntValue` class. 
 - Then, you can enforce the usage of the factory methods by setting all fields of the dataclass to `field(init=False)`. Like this, no one will be able to circumvent the factory.

src/dsp_tools/commands/xmllib/different_approaches/factory_option.py
```diff
@@ -0,0 +1,72 @@
...
+    def create(
+        self,
+        value: int | str,
+        prop_name: str,
+        res_id: str,
```
jnussbaum, line 27: So the user has to provide the res_id manually? That's cumbersome, and people won't do it.

Nora-Olivia-Ammann: The idea is that the users never create the classes themselves, but always use the function of the resource. This way only we have to provide that.

src/dsp_tools/commands/xmllib/different_approaches/regular_class.py
```diff
@@ -0,0 +1,58 @@
...
+"""
+
+
+class IntValue:
+    value: int | str
+    prop_name: str
+    permissions: str | None = "prop-default"
+    comment: str | None = None
+
+    def __init__(
```
jnussbaum, line 26: I would not use normal classes. If I'm not mistaken, the dataclass-approach with factory + `init=False` covers all advantages, without the disadvantage of the verbose __init__

src/dsp_tools/commands/xmllib/different_approaches/factory_option.py
Nora-Olivia-Ammann: That I don't want. Because the `IntValue` is not responsible about creating itself. This is very unclean to me.

src/dsp_tools/commands/xmllib/different_approaches/regular_class.py
Nora-Olivia-Ammann: I honestly don't see the difference between `init=False` and using a normal class. Because you have to write the the values yourself too.

src/dsp_tools/commands/xmllib/different_approaches/factory_option.py
jnussbaum: Static factory methods are in my tool belt, so I just spent some time to do some research, to find out if I should remove it from there ;-) But it seems to be a legitimate pattern: 
 - https://stackoverflow.com/questions/929021
 - https://www.baeldung.com/java-constructors-vs-static-factory-methods

Nora-Olivia-Ammann: I know it is, but I still dislike it because it doesn't separate the concerns cleanly to me.

jnussbaum (approved):
Overall, it looks very good. I have only found some typos, formulations that could be improved, and some minor software-design improvements that could be applied.

src/dsp_tools/commands/xmllib/models/dsp_base_resources.py
```diff
@@ -0,0 +1,269 @@
...
+    label: str
+    comments: list[str]
+    annotation_of: str
+    permissions: str = "res-default"
+
+    def __post_init__(self) -> None:
+        _check_and_warn_strings(self.res_id, self.res_id, "Resource ID")
+        _check_and_warn_strings(self.res_id, self.label, "Label")
+
+    def serialise(self) -> etree._Element:
```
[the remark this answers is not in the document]
BalduinLandolt: how about `to_xml()`? It's similarly not quite correct, but maybe it's closer to correct?

src/dsp_tools/commands/xmllib/models/dsp_base_resources.py
```diff
@@ -0,0 +1,269 @@
...
+        attribs = {"label": self.label, "id": self.res_id}
+        if self.permissions:
+            attribs["permissions"] = self.permissions
+        return etree.Element(f"{DASCH_SCHEMA}link", attrib=attribs, nsmap=XML_NAMESPACE_MAP)
+
+    def _serialise_links(self) -> etree._Element:
+        vals = [LinkValue(value=x, prop_name="hasLinkTo", resource_id=self.res_id) for x in self.link_to]
+        prop_ele = vals[0].make_prop()
+        for v in vals:
+            prop_ele.append(v.make_element())
```
BalduinLandolt, line 124: maybe case for a list comprehension?

src/dsp_tools/commands/xmllib/models/root.py
```diff
@@ -0,0 +1,52 @@
...
+
+XML_NAMESPACE_MAP = {None: "https://dasch.swiss/schema", "xsi": "http://www.w3.org/2001/XMLSchema-instance"}
+DASCH_SCHEMA = "{https://dasch.swiss/schema}"
+
+
+@dataclass
+class XMLRoot:
+    shortcode: str
+    default_ontology: str
+    resource_collection: list[Resource] = field(default_factory=list)
```
BalduinLandolt, line 21: the term `resource_collection` is a leftover of previous discussions, in the current state it no longer makes much sense. In terms of the `XMLRoot`, this should just be `resources`, so I would suggest renaming it to that

src/dsp_tools/commands/xmllib/models/root.py
```diff
@@ -0,0 +1,52 @@
...
+                schema_location_key: schema_location_value,
+                "shortcode": self.shortcode,
+                "default-ontology": self.default_ontology,
+            },
+            nsmap=XML_NAMESPACE_MAP,
+        )
+
+    def add_resources(self, resources: list[Resource]) -> XMLRoot:
+        self.resource_collection.extend(resources)
+        return self
```
BalduinLandolt, line 41: for convenience and consistency sake, this could also expose `add_resource` (singular) and `add_resource_optional`, which would add only a couple of lines, and shouldn't be confusing to the users, if the same three variants are exposed on all values

src/dsp_tools/commands/xmllib/models/root.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def serialise(self) -> etree._Element:
```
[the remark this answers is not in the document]
BalduinLandolt: I don't think it hurts to expose it, and it gives the users the liberty to access the etree rather than always having to write to a file.

src/dsp_tools/commands/xmllib/value_checkers.py
```diff
@@ -0,0 +1,185 @@
...
+    Args:
+        value: value to check
+
+    Returns:
+        True if it conforms
+    """
+
+    optional_calendar = r"((GREGORIAN|JULIAN|ISLAMIC):)?"
+    first_era = r"((CE|BCE|BC|AD):)?"
+    second_era = r"(:(CE|BCE|BC|AD))?"
```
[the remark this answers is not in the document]
BalduinLandolt: or maybe `second_area_optional`?

test/unittests/commands/xmllib/test_values.py
```diff
@@ -0,0 +1,278 @@
...
+    def test_warns(self) -> None:
+        with pytest.warns(DspToolsUserWarning):
+            BooleanValue("other", ":booleanProp", resource_id="res_id")
+
+    def test_serialise(self) -> None:
+        v = BooleanValue("0", ":booleanProp", resource_id="res_id")
+        expected = (
+            b'<boolean-prop xmlns="https://dasch.swiss/schema" '
+            b'xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name=":booleanProp">'
+            b'<boolean permissions="prop-default">false</boolean></boolean-prop>'
```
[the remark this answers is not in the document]
BalduinLandolt: If you want to stick with the XML string, a common thing you see in HTML formatting is to have each attribute on a new line with indentation. This could be emulated here:
```python
        b'<boolean-prop '
        b'    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" '
        b'    name=":booleanProp" '
        b'>'
```
I don't think it would semantically change anything, and to me it seems a lot more readable.

BalduinLandolt (approved):

src/dsp_tools/commands/xmllib/models/dsp_base_resources.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: The reason why I chose it, is because if we ever change the output format, we would not need to rename the function. The way we structured it, that would not constitute a breaking change with regard to usage...

src/dsp_tools/commands/xmllib/models/dsp_base_resources.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def _serialise_resource_element(self) -> etree._Element:
+        attribs = {"label": self.label, "id": self.res_id}
+        if self.permissions:
+            attribs["permissions"] = self.permissions
+        return etree.Element(f"{DASCH_SCHEMA}link", attrib=attribs, nsmap=XML_NAMESPACE_MAP)
+
+    def _serialise_links(self) -> etree._Element:
+        vals = [LinkValue(value=x, prop_name="hasLinkTo", resource_id=self.res_id) for x in self.link_to]
+        prop_ele = vals[0].make_prop()
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I don't understand what you propose? Do you want this function to take the vals as a parameter? from the order 122 cannot be above 121, because it does not exist yet...

src/dsp_tools/commands/xmllib/models/file_values.py
```diff
@@ -0,0 +1,78 @@
...
+
+@dataclass
+class FileValue(AbstractFileValue):
+    value: str | Path
+    permissions: str | None = "prop-default"
+    comment: str | None = None
+    resource_id: str | None = None
+
+    def __post_init__(self) -> None:
+        if not is_string(self.value):
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Long-term I would like to check if it is an actual path in case of a file value or a iiif uri... I am waiting for this, because the checks should be cleared with RDU this is somewhat of a place-holder...

src/dsp_tools/commands/xmllib/models/root.py
Nora-Olivia-Ammann: I agree mostly but I don't see how optional can be used?

src/dsp_tools/commands/xmllib/models/root.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I don't like to prescribe too much. While it is unlikely that they use it, I prefer to give flexibility now instead of making a big change later because we were too restrictive.

src/dsp_tools/commands/xmllib/value_checkers.py
```diff
@@ -0,0 +1,185 @@
...
+        True if conforms to the above-mentioned criteria.
+    """
+    if isinstance(value, str):
+        return bool(regex.search(r"^\d+$", value))
+    elif isinstance(value, int):
+        return True
+    return False
+
+
+def is_list(node: Any, listname: Any) -> bool:
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: The reason why I wrote it is because i like to have one checker for one value type. But I do understand your point.

test/unittests/commands/xmllib/test_value_converters.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@pytest.mark.parametrize("val", ["TRUE ", "  1", "1.0", "Yes", True])
+def test_convert_to_bool_true(val: Any) -> None:
+    assert convert_to_bool_string(val) == "true"
+
+
+@pytest.mark.parametrize("val", [pd.NA, None, 2.421, 10, "other", "  "])
+def test_convert_to_bool_failure(val: Any) -> None:
+    assert convert_to_bool_string(val) == str(val)
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Converting this is a wish from RDU. I separated conversion and checking strictly, so there is a warning raised in the `__post__init__`. So if the input is wrong, I want them to keep the input. Raising errors is unpopular, so I wont' do that.

test/unittests/commands/xmllib/test_values.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I like Balduin's suggestion and have done something like that, because I have not been able to add indentation, because it considers that to be part of the string and the tests fail.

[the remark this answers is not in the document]
jnussbaum: Haven't you considered comparing the etree elements, instead of comparing strings? If I remember correctly, Balduin would be rather in favor of comparing etree elements, because it was him who once taught me this ;-)

src/dsp_tools/commands/xmllib/models/dsp_base_resources.py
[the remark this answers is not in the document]
BalduinLandolt: It's a good consideration, but it's very unlikely that we will stop using XML all-together - and if so, that would break this lib anyways, because it should no longer be called `xmllib`... But I don't feel strongly about this change

test/unittests/commands/xmllib/test_value_converters.py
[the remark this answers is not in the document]
jnussbaum: So, RDU wishes that the scripts produce an XML file with `<boolean>2.421</boolean>`, and only a warning is emitted, instead of an error? I'm a bit surprised, but if this is intended, then okay.

src/dsp_tools/commands/xmllib/models/root.py
BalduinLandolt: if you have a function that looks at some data and may or may not return a resource. I am aware that you don't think it's likely that anyone from RDU would write such code, but as I said, it doesn't hurt to add it, so I think a small likelihood and API consistency are good enough reasons.

src/dsp_tools/commands/xmllib/models/dsp_base_resources.py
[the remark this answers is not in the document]
jnussbaum: No, I suggest you remove the `LinkValue.make_prop()` method. It is at the wrong place there. Instead, you do this:

```
def _serialise_links(self) -> etree._Element:
        vals = [LinkValue(value=x, prop_name="hasLinkTo", resource_id=self.res_id) for x in self.link_to]
        prop_ele = etree.Element(f"{DASCH_SCHEMA}resptr-prop", name="hasLinkTo", nsmap=XML_NAMESPACE_MAP)
        prop_ele.extend([v.make_element() for v in vals])
        return prop_ele
```

src/dsp_tools/commands/xmllib/models/dsp_base_resources.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: That is true.

src/dsp_tools/commands/xmllib/models/root.py
Nora-Olivia-Ammann: what would the input look like though?

test/unittests/commands/xmllib/test_value_converters.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Yes, it will not pass XML validation, so it does no harm.
I think it is because they are able to more quickly fix the errors if they have all of them collected and they see it concretely and not only on the console...

test/unittests/commands/xmllib/test_values.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: We can do either. But I would like to merge this soon. Since we are not yet finished anyway, we can always clean this up later.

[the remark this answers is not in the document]
jnussbaum: I clearly prefer comparing etree elements. These multi-line strings without semantic line breaks are ugly to look at, and you will always have troubles with XML-strings. But feel free to do it now or note it down for later, so that you are not blocked on your way forward!

src/dsp_tools/commands/xmllib/models/dsp_base_resources.py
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I could do that, but I prefer to reuse classes, so that changes are easier. but I can live with it.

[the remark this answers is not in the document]
jnussbaum: What do you mean with "reuse classes"?

src/dsp_tools/commands/xmllib/models/root.py
BalduinLandolt: the input to what?

PR #1134, opened 2024-08-23
feat(xmlupload): allow free ordering of properties in segments, by using XSD 1.1 (DEV-4037)

jnussbaum:


jnussbaum:
Unfortunately I have to close this PR. It turned out that with XSD 1.1, we would get significantly worse tooling and error messages. Concretly, it would not be possible to get the line number of the failing element, and it would be very difficult (if not impossible) to retrieve the info of which property in which resource has failed.

In addition, VS Code's LSP doesn't seem to support XSD 1.1.

We spoke with RDU today, and they said it's okay if the properties have a fixed order. I will document that in a separate PR.

PR #1140, opened 2024-08-27
test: add tests for XMLResource, XMLProperty, XMLValue

jnussbaum:


src/dsp_tools/commands/xmlupload/models/deserialise/deserialise_value.py
```diff
@@ -202,7 +204,7 @@ class XMLBitstream:
     """
 
     value: str
-    permissions: Optional[str]
+    permissions: Optional[str] = None
```
Nora-Olivia-Ammann, line 207: Since you are already changing code you could also change the typing here into `str | None`.

Nora-Olivia-Ammann (approved):
It looks good to me.

src/dsp_tools/commands/xmlupload/models/deserialise/xmlresource.py
```diff
@@ -112,10 +128,17 @@ def _get_properties(
                     ungrouped_properties.append(XMLProperty.from_node(subnode, "interval", default_ontology))
                 case "hasTitle" | "hasComment" | "hasDescription" | "hasKeyword":
                     ungrouped_properties.append(XMLProperty.from_node(subnode, "text", default_ontology))
-                case _:
+                case str() as x if x.endswith("-prop") and x.startswith(COMPOSITE_PROPS):
```
BalduinLandolt, line 131: wouldn't it be more straight forward if `COMPOSITE_PROPS` contained `boolean-prop` etc., and you'd just check `str() as x if x in COMPOSITE_PROPS`?

src/dsp_tools/commands/xmlupload/models/deserialise/xmlresource.py
```diff
@@ -112,10 +128,17 @@ def _get_properties(
...
                     # get the property type which is in format type-prop, p.ex. <decimal-prop>
                     prop_type, _ = subnode.tag.split("-")
                     ungrouped_properties.append(XMLProperty.from_node(subnode, prop_type, default_ontology))
+                case _:
+                    raise XmlUploadError(f"Unexpected tag '{subnode.tag}'")
+        grouped_properties = XMLResource._group_props(ungrouped_properties)
```
BalduinLandolt, line 137: I would prefer this to be a top level function rather than a static method, as it doesn't really have a direct connection to the class containing it. Also, private static methods aren't particularly meaningful as a concept, as "private" would indicate that it is only intended in the scope of the class, which means that you normally have an instance of the class at hand anyway, which renders some of the use cases for static-ness obsolete.  
And in Python it's entirely idiomatic to have top level functions - which renders practically all use cases for static methods obsolete.  
I would really only (if at all) use static methods when it has a very direct relation to the Class it lives on, like e.g. a factory method.

test/unittests/commands/xmlupload/models/deserialize/test_deserialise_value.py
```diff
@@ -0,0 +1,243 @@
...
+            ]
+        )
+        node = etree.fromstring(string)
+        value = XMLValue.from_node(node, "text")
+        assert isinstance(value.value, FormattedTextValue)
+        assert value.value.xmlstr == expected
+        assert value.resrefs == unordered(["test_thing_0", "test_thing_1"])
+        assert value.comment is None
+        assert value.permissions is None
+        assert value.link_uuid is None
```
BalduinLandolt, line 149: This is a somewhat brittle and rigid pattern for testing:
1. If you add another property to the class, you have to remember to add it to the test, otherwise it will not be tested
2. If you change the structure of the class, you manually have to change the test (which, if done too often in a code base, might demotivate people from refactoring)

Alternatively, now that it is possible to create `XMLValue` instances without using the factory method that is under test here, you could create an "expected" object and compare res against expected (like you did some lines above).  
This has the downside of not being quite as precise a test (if someone were to break the default for `comment` or `permission`, that would not be caught here - but since this is not really the aspect under test here, I think that's ok) but IMO the advantages of this outweighs the downside.

Additionally - but I'm aware that this is out of scope of this refactoring here - ideally, the factory method would have less logic, and things like resrefs and formatting tidying would be extracted into a specific function for deserializing formatted text values. Then it would be good enough to test here, that the factory correctly creates a `FormattedTextValue` and the intricacies of this creation (resrefs, formatting) could be left to tests testing this function in isolation.

test/unittests/commands/xmlupload/models/deserialize/test_xmlresource.py
```diff
@@ -0,0 +1,167 @@
...
+
+from dsp_tools.commands.xmlupload.models.deserialise.deserialise_value import IIIFUriInfo
+from dsp_tools.commands.xmlupload.models.deserialise.deserialise_value import XMLBitstream
+from dsp_tools.commands.xmlupload.models.deserialise.deserialise_value import XMLProperty
+from dsp_tools.commands.xmlupload.models.deserialise.deserialise_value import XMLValue
+from dsp_tools.commands.xmlupload.models.deserialise.xmlresource import XMLResource
+from dsp_tools.commands.xmlupload.models.formatted_text_value import FormattedTextValue
+
+
+class Test_XMLResource:
```
BalduinLandolt, line 12: having one big test class makes having test classes pointless. In pytest (as opposed to unittest), test classes only serve for organization. And having one organizational unit is just as good as having none.  
In this case, I would see logical organizational units to have test classes for the test cases testing `get_restype_...` and `get_properties_(of)_...` (though the naming isn't fully consistent there), and maybe have the rest either top level or in a "sammelbecken" test class

src/dsp_tools/commands/xmlupload/models/deserialise/deserialise_value.py
BalduinLandolt, line 1: in this module, the file naming is very inconsistent. Would this something you'd consider streamlining along with this refactoring?

BalduinLandolt (approved):
Over all a big improvement. I added some suggestions/comments

src/dsp_tools/commands/xmlupload/models/deserialise/deserialise_value.py
jnussbaum: The file name has really not much to do with what happens in the file. I've noted it for later. I don't want to obsfuscate the diff.

src/dsp_tools/commands/xmlupload/models/deserialise/xmlresource.py
jnussbaum: Indeed.

src/dsp_tools/commands/xmlupload/models/deserialise/xmlresource.py
jnussbaum: Good considerations. However, I'm unsure if it applies here. This private static method is a helper of a helper of the (static) factory. I made all helpers static as well, but private.

BalduinLandolt: I'm aware of that, but I would advocate making helpers top-level functions - simply because I don't see a benefit in calling a function like that _on a class_ when you could just as well simply _call the function_.  
But in the end, this is cosmetic and doesn't make a real difference, so I will leave it up to you.

src/dsp_tools/commands/xmlupload/models/deserialise/deserialise_value.py
BalduinLandolt: fair enough, thanks!

test/unittests/commands/xmlupload/models/deserialize/test_deserialise_value.py
jnussbaum: Thanks for these insightful considerations. I have changed it.

PR #1159, opened 2024-09-11
fix(start-stack): `--latest` flag ships latest versions of all backend components, instead of only API (DEV-4100)

jnussbaum:


docs/cli-commands.md
```diff
@@ -538,7 +538,7 @@ The following options are available:
 - `--max_file_size=int` (optional, default: `2000`): max. multimedia file size allowed, in MB (max: 100'000)
 - `--latest` (optional): 
   instead of the latest deployed version,
-  use the latest development version of DSP-API (from the `main` branch)
+  use the latest development version of DSP-API/DSP-APP (from the `main` branch)
```
jnussbaum, line 541: This is a user-friendly abstraction of what is really happening. Already on line 526, we don't tell the entire truth.

src/dsp_tools/cli/create_parsers.py
```diff
@@ -121,7 +121,7 @@ def _add_start_stack(subparsers: _SubParsersAction[ArgumentParser]) -> None:
     subparser.add_argument(
         "--latest",
         action="store_true",
-        help="use the latest dev version of DSP-API, from the main branch of the GitHub repository",
+        help="use the latest dev version of DSP-API/DSP-APP, from the main branch of the GitHub repository",
```
jnussbaum, line 124: This is a user-friendly abstraction of what is really happening. Already on line 110, we don't tell the entire truth.

src/dsp_tools/commands/start_stack.py
```diff
@@ -94,15 +94,8 @@ def _get_url_prefix(self) -> str:
...
-            return f"{url_prefix_base}/main/"
-
-        with config_file.open("r", encoding="utf-8") as f:
-            try:
-                start_stack_config = yaml.safe_load(f)
-            except yaml.YAMLError:
-                start_stack_config = {}
-        commit_of_used_api_version = start_stack_config.get("DSP-API commit", "main")
+        start_stack_config = yaml.safe_load(config_file.read_bytes())
+        commit_of_used_api_version = start_stack_config["DSP-API commit"]
```
jnussbaum, line 98: Here, I did 2 things:
1. I simplified the with-clause to `config_file.read_bytes()`
2. I made it fail-fast. Before, the code was way too forgiving, so that it was possible to ship erroneous code to our users, because in our PRs, the tests erroneously passed. IMHO, if there is something wrong with the config file, we should get an error already in our PR, before we even ship the code.

src/dsp_tools/commands/start_stack.py
```diff
@@ -293,11 +286,17 @@ def _start_remaining_docker_containers(self) -> None:
...
+                "resources/start-stack/docker-compose.override.yml"
             )
+            override_content = yaml.safe_load(override_file.read_bytes())
+            dockerhub_latest_tags = [x["image"] for x in override_content["services"].values()]
+            for tag in dockerhub_latest_tags:
+                subprocess.run(
+                    f"docker pull {tag}".split(),
+                    cwd=self.__docker_path_of_user,
+                    check=True,
+                )
```
jnussbaum, line 299: This is the actual change of this PR. This is intentionally coupled to `docker-compose.override.yml`, because 2 things are necessary in order to get the latest tag from a dockerhub-image:
1. The image needs to be pulled (this happens here)
2. In the `docker-compose.yml`, the "latest" tag must be present (this is done by `docker-compose.override.yml`)

Nora-Olivia-Ammann (approved):

src/dsp_tools/commands/start_stack.py
BalduinLandolt, line 299: this could be simplified with `docker compose pull` which automatically calls `pull` on all components - and is much faster because it does so in parallel

jnussbaum: Thanks, that's indeed a huge improvement.

jnussbaum:
> I'm very sceptical that this will work as intended, as we do not publish the "latest" tag for the DSP-APP. Furthermore, using "latest" often doesn't work as expected (unless you explicitly do `pull`, which I see is being done here!) so it is generally not considered very good practice. Imagine someone in the future removing this pull - this would not give an error, but suddenly RDU would have outdated versions running.

Apart from DSP-APP, it works like a charm. `docker pull` adds the "latest" tag by default, so it doesn't really matter if I use it or not. But I indeed have a problem: There are only 2 ways to pull a docker image: Either a specified version tag, or latest. For the other images, this works well. But for DSP-APP, the only way to find out what the latest tag is, is to do some acrobatics via dockerhub-api.

To make it worse, DSP-APP used the "latest" tag earlier, but stopped to do so 18 months ago. Now, if you do `docker pull daschswiss/dsp-app`, you get this 18 month old version (with or without `:latest` doesn't matter).

So pragmatically, there are 2 ways to solve my problem:

1. convince frontend that they reintroduce the latest tag
2. dsp-tools always uses the currently deployed version of DSP-APP, even when `--latest` was requested.

WDYT?

BalduinLandolt:
> > I'm very sceptical that this will work as intended, as we do not publish the "latest" tag for the DSP-APP. Furthermore, using "latest" often doesn't work as expected (unless you explicitly do `pull`, which I see is being done here!) so it is generally not considered very good practice. Imagine someone in the future removing this pull - this would not give an error, but suddenly RDU would have outdated versions running.
> 
> Apart from DSP-APP, it works like a charm. `docker pull` adds the "latest" tag by default, so it doesn't really matter if I use it or not. But I indeed have a problem: There are only 2 ways to pull a docker image: Either a specified version tag, or latest. For the other images, this works well. But for DSP-APP, the only way to find out what the latest tag is, is to do some acrobatics via dockerhub-api.
> 
> To make it worse, DSP-APP used the "latest" tag earlier, but stopped to do so 18 months ago. Now, if you do `docker pull daschswiss/dsp-app`, you get this 18 month old version (with or without `:latest` doesn't matter).
> 
> So pragmatically, there are 2 ways to solve my problem:
> 
> 1. convince frontend that they reintroduce the latest tag
> 2. dsp-tools always uses the currently deployed version of DSP-APP, even when `--latest` was requested.
> 
> WDYT?

Pragmatically, I would go for 2. And if we ever encounter a case where we actually need the "real" latest version of the APP, we can still add an extra flag `--latest --app=v11.20.5-7987345` or something like that. Though I would think that this would ever only be relevant to a DSP-TOOLS dev, and we should be able to manually change the version of the app when needed.

PR #1164, opened 2024-09-17
feat(validate): reformat input XML data (DEV-4124)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xml_validate/reformat_input.py
```diff
@@ -0,0 +1,213 @@
...
+            case "resource":
+                all_res.append(_deserialise_one_resource(res))
+            case "annotation" | "region" | "link" | "video-segment" | "audio-segment":
+                all_res.append(_deserialise_dsp_base_resource(res))
+            case _:
+                pass
+    return all_res
+
+
+def _deserialise_dsp_base_resource(resource: etree._Element) -> ResourceData:
```
Nora-Olivia-Ammann, line 51: The validation of the base resources is part of milestone 3. But it is possible that other resources link to them, so they must exist in the data graph.

src/dsp_tools/commands/xml_validate/reformat_input.py
```diff
@@ -0,0 +1,201 @@
...
+            return []
+
+
+def _deserialise_boolean_prop(prop: etree._Element) -> list[ValueData]:
+    prop_name = prop.attrib["name"]
+    all_vals: list[ValueData] = []
+    for val in prop.iterchildren():
+        txt = val.text if val.text is not None else ""
+        all_vals.append(BooleanValueData(prop_name=prop_name, prop_value=txt))
+    return all_vals
```
seakayone, line 106: *suggestion*: Reduce code duplication

There is a lot of code duplication here, most of the `_deserialize*` function only differ by the type they construct.

You could simplify this by extracting the duplicate code and pass in a function which constructs it, e.g.:

```python 
from typing import Callable

def _deserialise(prop: etree._Element, func: Callable[[str, str], ValueData]) -> list[ValueData]:
    prop_name = prop.attrib["name"]
    all_vals: list[ValueData] = []
    for val in prop.iterchildren():
        txt = val.text if val.text is not None else ""
        all_vals.append(func(prop_name=prop_name, prop_value=txt))
    return all_vals

def _deserialise_boolean_prop(prop: etree._Element) -> list[ValueData]:
    return _deserialise(prop, BooleanValueData)

def _deserialise_color_prop(prop: etree._Element) -> list[ValueData]:
    return _deserialise(prop, ColorValueData)

# and so on and so forth ...
```

You could then also inline `_deserialise(prop, BooleanValueData)` into the `match` statement above.

src/dsp_tools/commands/xml_validate/models/data_deserialised.py
```diff
@@ -0,0 +1,104 @@
...
+@dataclass
+class UriValueData(ValueData):
+    prop_name: str
+    prop_value: str
+
+
+@dataclass
+class LinkValueData(ValueData):
+    prop_name: str
+    prop_value: str
```
seakayone, line 104: *question*: Should `prop_values` from above be `str`?

To me it seems odd that e.g. the `IntValueData` does not contain an `int` but a `str` instead. Maybe there is a reason I do not understand.

seakayone (approved):

src/dsp_tools/commands/xml_validate/reformat_input.py
Nora-Olivia-Ammann: That look really nice thanks for pointing it out, I never did something like that. I will try it out.

src/dsp_tools/commands/xml_validate/models/data_deserialised.py
Nora-Olivia-Ammann: Yes there is a reason. We expect to have invalid data, eg. a string with characters where an int should be. If I construct a literal with `rdflib` the 'correct' way:
`Literal("value", datatype=XSD.integer)`
Then an error is raised from `rdflib`: `ValueError: invalid literal for int() with base 10: 'value'`
I don't want to find these errors during the construction of the graph but through SHACL.
Another way to solve this would be to construct the graph this way:

```
try:
    Literal("value", datatype=XSD.integer)
except ValueError:
    Literal("value", datatype=XSD.string)
```

Then we could specify that it must be an `xsd:integer` in the SHACL. If the SHACL were exposed through the API then I would agree with you that this problem is best solved in DSP-TOOLS. This is not the case currently so I chose to do it like this. But I am happy to change it. However in any case the typing here would stay the same, by default if you get the content of XML it comes back as a string or `None`.

seakayone: I appreciate you explanation and it makes sense to me.

src/dsp_tools/commands/xml_validate/reformat_input.py
seakayone: I am too much influence by functional programming ;)

It is often quite easy to pass in a function if only a small portion of the code differs. Btw: I have tested out the code myself and it should work.

Nora-Olivia-Ammann: It does, Updated it, now I just have to fight with mypy about the typing and then I'll merge it.

src/dsp_tools/commands/xml_validate/models/data_deserialised.py
Nora-Olivia-Ammann: I have thought about it and now I think I prefer your suggestion. Because we actually violate `knora-api` with this: https://github.com/dasch-swiss/dsp-tools/blob/6712bd9fd3935e3c044da55403211039c1c5c550/src/dsp_tools/resources/xml_validate/knora-api.ttl#L5043

Because we expect a string in the SHACL: https://github.com/dasch-swiss/dsp-tools/blob/6712bd9fd3935e3c044da55403211039c1c5c550/src/dsp_tools/resources/xml_validate/knora-api-shapes.ttl#L176

If we solve it with `try:` and `except` we will still get the information in the SHACL graph... I just have to verify that I get the correct errors with `rdflib`. I will look into it during the other PRs.

test/unittests/commands/xml_validate/xml_fixtures.py
```diff
@@ -0,0 +1,362 @@
...
+
+@pytest.fixture
+def geometry_value_corr() -> etree._Element:
+    return etree.fromstring("""
+        <geometry-prop name="hasGeometry">
+            <geometry>
+                {
+                    "status": "active",
+                    "type": "polygon",
+                    "lineColor": "#ff1100",
```
BalduinLandolt, line 138: we should not use `lineColor` anywhere, see https://github.com/dasch-swiss/dsp-tools/pull/1166

BalduinLandolt (approved):

test/unittests/commands/xml_validate/xml_fixtures.py
Nora-Olivia-Ammann: I missed that thanks!

test/unittests/commands/xml_validate/data_rdf_fixtures.py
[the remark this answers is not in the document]
jnussbaum: For 1: That's a good reason, of course. But for 2: Good editors can split screen with both views in the same file ;-)

src/dsp_tools/commands/xml_validate/models/data_rdf.py
```diff
@@ -0,0 +1,211 @@
...
+        g = Graph()
+        g.add((self.res_id, RDF.type, self.res_class))
+        g.add((self.res_id, RDFS.label, self.label))
+        for v in self.values:
+            g += v.make_graph(self.res_id)
+        return g
+
+
+@dataclass
+class ValueRDF:
```
jnussbaum, line 47: You could mark this as abstract base class (https://docs.python.org/3/library/abc.html#abc.ABC):

```suggestion
class ValueRDF(ABC):
```

src/dsp_tools/commands/xml_validate/reformat_input.py
```diff
@@ -0,0 +1,194 @@
...
+from dsp_tools.commands.xml_validate.models.data_rdf import SimpleTextRDF
+from dsp_tools.commands.xml_validate.models.data_rdf import TimeValueRDF
+from dsp_tools.commands.xml_validate.models.data_rdf import UriValueRDF
+from dsp_tools.commands.xml_validate.models.data_rdf import ValueRDF
+
+
+def transform_into_data_deserialised(root: etree._Element) -> DataRDF:
+    """
+    Takes the root of an XML
+    Extracts the metadata of the project and transforms all its resources.
```
jnussbaum, line 29:
```suggestion
    Extracts the data of the project and transforms all its resources.
```

src/dsp_tools/commands/xml_validate/reformat_input.py
```diff
@@ -0,0 +1,194 @@
...
+def _deserialise_all_resources(root: etree._Element) -> list[ResourceRDF]:
+    all_res = []
+    for res in root.iterchildren():
+        match res.tag:
+            case "resource":
+                all_res.append(_deserialise_one_resource(res))
+            case "annotation" | "region" | "link" | "video-segment" | "audio-segment":
+                all_res.append(_deserialise_dsp_base_resource(res))
+            case _:
+                pass
```
jnussbaum, line 52: I would rather raise an error here. This is a critical error, which we shouldn't miss.
The permissions-elements can be passed, of course:

```suggestion
            case "permissions":
                pass
            case _:
                raise BaseError()
```

src/dsp_tools/commands/xml_validate/reformat_input.py
```diff
@@ -0,0 +1,194 @@
...
+    return all_res
+
+
+def _deserialise_dsp_base_resource(resource: etree._Element) -> ResourceRDF:
+    return ResourceRDF(
+        res_id=URIRef(resource.attrib["id"]),
+        res_class=URIRef(resource.attrib["restype"]),
```
jnussbaum, line 59: The special tags don't have a "restype" attribute. Have you tried this out? I think it wouldn't work

test/integration/commands/xml_validate/test_reformat_input.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+from lxml import etree
+
+from dsp_tools.commands.xml_validate.models.data_rdf import DataRDF
+from dsp_tools.commands.xml_validate.reformat_input import transform_into_data_deserialised
+
+
+def test_transform_into_data_deserialised(data_xml: etree._Element) -> None:
```
jnussbaum, line 8: I find this hard to understand. How can one easily find out where `data_xml` comes from? Actually, you only use it in this single file. So why doesn't it live here? It would be much easier to understand.

Now, it makes its way across 3 files: 

1. It is defined in xml_fixtures.py
2. Then, it is imported into conftest.py
3. Then, it magically appears here

jnussbaum (approved):

src/dsp_tools/resources/xml_validate/replace_namespace.xslt
jnussbaum, line 1: Cool, I'm curious to learn more about XSLT. But if I see correctly, then this is not tested. How have you made sure that this does what it is expected to do?

src/dsp_tools/commands/xml_validate/reformat_input.py
Nora-Olivia-Ammann: No I don't want to do that, because there are permissions we want to ignore. Since it passed our xml validation, we can trust the file

src/dsp_tools/commands/xml_validate/reformat_input.py
Nora-Olivia-Ammann: They do. In the xmlupload preparation step, we add the restype to the elements that are for example `<region>`. This is something that should be changed in the future, in the xmlupload too. When we implement the dsp-base resources in SHACL we will have a separate class for each. So this will be deleted, which is why I am ok having this sub-par handling here. Until then, this is the easiest way, because we need them as targets

src/dsp_tools/resources/xml_validate/replace_namespace.xslt
Nora-Olivia-Ammann: You were right I forgot to add a test, that is done now and it does work!

test/integration/commands/xml_validate/test_reformat_input.py
Nora-Olivia-Ammann: The reason it is in a separate file is because I plan on using it more in the future.

src/dsp_tools/commands/xml_validate/models/data_rdf.py
Nora-Olivia-Ammann: Mypy is not amused with that. It used to be a protocol, but I changed it. Because of that.

src/dsp_tools/commands/xml_validate/reformat_input.py
jnussbaum: But wait, I think this entire function is wrong and will actually never be called. Because in the preparation step you mentioned, not only the restypes are added, but the special tag names are transformed to `<resource>`. See https://github.com/dasch-swiss/dsp-tools/blob/45af782b15c9d0703daf6ee0e76066331632b95c/src/dsp_tools/utils/xml_utils.py#L35

That means:
 - This line is never reached: https://github.com/dasch-swiss/dsp-tools/blob/45af782b15c9d0703daf6ee0e76066331632b95c/src/dsp_tools/commands/xml_validate/reformat_input.py#L49
 - the function `_deserialise_dsp_base_resource()` is never called, and can be deleted.

(Anyway, `_deserialise_dsp_base_resource()` ignores the values, which is not okay. But this doesn't matter any more).

Or do I miss something?

test/integration/commands/xml_validate/test_reformat_input.py
jnussbaum: But then, I would not add it to conftest.py, rather import it directly from xml_fixtures.py. Our future selves will be grateful for this.

Nora-Olivia-Ammann: Yes, I prefer to do that too, but that did not work in the past. We looked at it together and I don't remember why it did not work, but we only managed to make it work with the config file. Maybe it was a linter issue... not sure.

PR #1168, opened 2024-09-20
refactor(xmlvalidate): change qname transformation

Nora-Olivia-Ammann:


Nora-Olivia-Ammann:
This way we are more independent of the xmlupload.

src/dsp_tools/utils/xml_utils.py
```diff
@@ -66,6 +67,22 @@ def remove_qnames_and_transform_special_tags(input_tree: etree._Element) -> etre
     return input_tree
 
 
+def transform_into_qnames(root: etree._Element) -> etree._Element:
```
jnussbaum, line 70: the name is wrong. Rather, the tags are transformed into localnames, but not qnames.

jnussbaum, line 70: I dislike the code redundancy that this introduces. In fact, it is a good idea to have this new function. But now it should be integrated in the existing code. This means, the qname transformation should be removed from the function`remove_qnames_and_transform_special_tags()`, and instead, your new function should be used.

src/dsp_tools/commands/xml_validate/reformat_input.py
```diff
@@ -42,21 +45,28 @@ def to_data_rdf(root: etree._Element) -> DataRDF:
...
+        "video-segment": KNORA_API.VideoSegment,
+        "audio-segment": KNORA_API.AudioSegment,
+    }
     for res in root.iterchildren():
         match res.tag:
             case "resource":
                 all_res.append(_deserialise_one_resource(res))
             case "annotation" | "region" | "link" | "video-segment" | "audio-segment":
-                all_res.append(_deserialise_dsp_base_resource(res))
+                all_res.append(_deserialise_dsp_base_resource(res, restype_mapper[res.tag]))
```
jnussbaum, line 60: have you tested in the debugger that this line is actually reachable?

Nora-Olivia-Ammann: Yes `test_reformat_input.py` Checks that the number of resources that were extracted is correct. And `test_xml_validate.py` Checks that the transformation was done correctly.

src/dsp_tools/utils/xml_utils.py
```diff
@@ -29,13 +30,12 @@ def parse_and_clean_xml_file(input_file: Path) -> etree._Element:
...
-    return remove_qnames_and_transform_special_tags(root)
+    return transform_special_tags_make_localname(root)
 
 
-def remove_qnames_and_transform_special_tags(input_tree: etree._Element) -> etree._Element:
+def transform_special_tags_make_localname(input_tree: etree._Element) -> etree._Element:
     """
-    This function removes the namespace URIs from the elements' names
-    and transforms the special tags `<annotation>`, `<region>`, `<link>`, `<video-segment>`, `<audio-segment>`
+    Transforms the special tags `<annotation>`, `<region>`, `<link>`, `<video-segment>`, `<audio-segment>`
```
jnussbaum, line 38: This should be reverted

jnussbaum (approved):

PR #1169, opened 2024-09-23
fix(xmlupload): increase timeout for file uploads to 10 min

jnussbaum:


Nora-Olivia-Ammann (approved):
Why don't we set this as a parameter in the cli command? I understand that it solves the current problem, but it is likely that in future this timeout will not solve a problem and is overkill

PR #1170, opened 2024-09-23
fix(excel2json): write newlines correctly into the "description" field of the JSON file (RDU-45)

jnussbaum:
@Nora-Olivia-Ammann @Notheturtle  This is a rather naive fix of the problem. It really only takes care about the "description" field. Is this already enough, or should this go further?


src/dsp_tools/commands/excel2json/models/json_header.py
```diff
@@ -76,6 +76,10 @@ def to_dict(self) -> dict[str, Any]:
 class Descriptions:
     descriptions: dict[str, str]
 
+    def __post_init__(self) -> None:
+        for k, v in self.descriptions.items():
+            self.descriptions[k] = v.replace("\n", "<br/>")
```
Nora-Olivia-Ammann, line 81: would be nicer:
```
splt = [x.strip() for x in v.split("\n")]
self.descriptions[k] = "<br/>".join(splt)
```

Nora-Olivia-Ammann (approved):

PR #1171, opened 2024-09-23
feat(xmlvalidate): get ontologies from API (DEV-4150)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xml_validate/api_connection.py
BalduinLandolt, line 1: In terms of naming: the file is called `api_connection` but the class is called `OntologyConnection`, so there is a discrepancy there.  
And furthermore, to me this almost looks more like an `OntologyClient` than like a `OntologyConnection`. But maybe I'm lacking the context of how you intend to use this.

src/dsp_tools/commands/xml_validate/api_connection.py
```diff
@@ -0,0 +1,73 @@
...
+from requests import Response
+
+from dsp_tools.models.exceptions import InternalError
+from dsp_tools.models.exceptions import UserError
+
+
+@dataclass
+class OntologyConnection:
+    api_url: str
+    shortcode: str
```
BalduinLandolt, line 18: After understanding what the class does, I see an issue in terms of separation of concerns and keeping the abstraction level consistent in related code:

`OntologyConnection` exposes two methods `get(url, headers)` and `get_ontologies()`. The latter is a high level concept that would fit an `OntologyClient`, while the first is a low-level concept that is specific to the HTTP protocol and absolutely fits a `OntologyConnection`.  
It would be absolutely fair to say that you call the class `OntologyClient` and make `get()` private. But in terms of separation of concerns and in terms of testability, it might even be the "cleaner" way, to separate this into two classes, where `OntologyClient` has a field `OntologyConnection`. But to be honest, I'm not sure that's worth the effort - because then in principal the `OntologyConnection` should be a protocol, and the concrete implementation would have to be `ApiOntologyConnection` (which in theory is nice because then you could have a `FileOntologyConnection` that can read ontologies from turtle files, and that would be perfectly interchangeable within the client... but realistically, you'll never need that).

BalduinLandolt (approved):

src/dsp_tools/commands/xml_validate/api_connection.py
Nora-Olivia-Ammann: With regard to usage, this class should only get all the project ontologies. In the future the file will also have the connection that gets the lists. I have not yet decided if I want to implement this in a separate class or add this to this class. But because it currently only gets the ontology, I put this in the name. I did not use the existing `OntologyClient` because that one gets the connection class as parameter. I would like to have this independent of the `Connection` class. I am hesitant to also call this `OntologyClient` because they are different from each other.

src/dsp_tools/commands/xml_validate/api_connection.py
Nora-Olivia-Ammann: I agree that with regard to separation and testability it would be cleaner to take the `get` out here. I did not because I don't want to create another connection monster that does everything. Additionally I only want to handle errors here that may happen in this particular situation. Which is why I did not use the `Connection` class. So if I separated it, the OntologyConnection that does the get request would only ever be used in this class. Do you think that is overkill?

src/dsp_tools/commands/xml_validate/api_connection.py
```diff
@@ -0,0 +1,75 @@
...
+
+        Returns:
+            Response of the request if it was a 200 response code
+
+        Raises:
+            InternalError: in case of errors raised by the requests library
+            UserError: If a non-200 response was given
+        """
+        try:
+            response = requests.get(url, headers=headers, timeout=100)
```
jnussbaum, line 36: One thing I completely forgot in my yesterday's review: For every new client that we introduce, we shouldn't forget to add logging statements for every request and every response. As an example how to do it:

https://github.com/dasch-swiss/dsp-tools/blob/e65c75315da8039fa9533ca3b79e3a4802768cd8/src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py#L71

jnussbaum (changes requested):

src/dsp_tools/commands/xml_validate/api_connection.py
```diff
@@ -0,0 +1,77 @@
...
+        Returns:
+            Response of the request if it was a 200 response code
+
+        Raises:
+            InternalError: in case of errors raised by the requests library
+            UserError: If a non-200 response was given
+        """
+        try:
+            logger.debug(f"REQUEST: GET to {url}, timeout: 100, headers: {headers}")
+            response = requests.get(url, headers=headers, timeout=100)
```
jnussbaum, line 37:
```suggestion
            timeout = 100
            logger.debug(f"REQUEST: GET to {url}, timeout: {timeout}, headers: {headers}")
            response = requests.get(url, headers=headers, timeout=timeout)
```

src/dsp_tools/commands/xml_validate/api_connection.py
```diff
@@ -0,0 +1,77 @@
...
+            logger.debug(f"RESPONSE: {response.status_code}")
```
jnussbaum, line 38: The text should also be logged, probably with `{response.text}`

jnussbaum (approved):

src/dsp_tools/commands/xml_validate/api_connection.py
BalduinLandolt: I absolutely don't advocate for re-using our existing `Connection` class: We have talked about this before, and agreed upon splitting that one up, not making it even more monstrous.  
As for naming: In this instance it doesn't matter much, but generally, we will always have different classes doing similar things in independent parts of our code base. We should of course keep them sparate in the interest of separation of concerns. But we can not shy away from calling them similarly or the same, if that is the best name for it. Otherwise we either end up with confusing names or we invent our own "namespacing" in class names like `ValidateXmlOntologyClient` which we do not want. And Python modules do give us namespacing for free.

src/dsp_tools/commands/xml_validate/api_connection.py
BalduinLandolt: I think in this instance, the pragmatic thing would be to make `get()` private `_get()` to hide the low level implementations from the caller that should just act on the high level. But it doesn't matter much here, the class is small enough to not become too confusing

PR #1176, opened 2024-09-25
fix(ingest-xmlupload): only convert absolute path from mapping CSV to relative path if necessary

jnussbaum:
Currently, absolute paths are always converted to relative ones, but in most cases, this is detrimental. The conversion should only happen if the absolute path cannot be found in the mapping CSV.

### DIFF
--- a/src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
+++ b/src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
@@ -1,4 +1,4 @@
-import urllib
+import urllib.parse
 from dataclasses import dataclass
 from dataclasses import field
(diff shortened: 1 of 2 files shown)


src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
```diff
@@ -1,4 +1,4 @@
-import urllib
+import urllib.parse
```
jnussbaum, line 1: Further down, `urllib.parse.quote()` is used. For this, I don't get code insights in my IDE with `import urllib`. Only when changing this to `import urllib.parse`, I get intellisense / code hover support

Nora-Olivia-Ammann (approved):

PR #1182, opened 2024-09-30
refactor(xmllib): make dsp built-in resources interface consistent (DEV-4186)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -132,14 +183,61 @@ class VideoSegmentResource:
     segment_start: float
```
Nora-Olivia-Ammann, line 183: The diff is already big enough so I will do that later.

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -132,14 +183,61 @@ class VideoSegmentResource:
...
-    description: list[str] = field(default_factory=list)
+    comments: list[str] = field(default_factory=list)
+    descriptions: list[str] = field(default_factory=list)
     keywords: list[str] = field(default_factory=list)
     relates_to: list[str] = field(default_factory=list)
     permissions: str = "res-default"
 
     def __post_init__(self) -> None:
         _validate_segment(self)
+        match self.comments:
```
Nora-Olivia-Ammann, line 194: I am undecided how I want to treat this. RDU may ignore type hints, therefore it is conceivable that it will not be a list...

PR #1184, opened 2024-09-30
refactor(xmllib): check and convert erroneous input datatypes (DEV-4190)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -50,11 +42,15 @@ def add_comments(self, comments: list[str]) -> AnnotationResource:
...
 
     def serialise(self) -> etree._Element:
+        self._check_for_and_convert_unexpected_input()
         res_ele = self._serialise_resource_element()
         res_ele.append(self._serialise_annotation_of())
         res_ele.append(_serialise_has_comment(self.comments, self.res_id))
         return res_ele
 
+    def _check_for_and_convert_unexpected_input(self) -> None:
+        self.comments = _transform_unexpected_input(self.comments, "comments", self.res_id)
```
BalduinLandolt, line 52: nitpick: I'm not sure it's worth extracting a one-liner here... I guess you did so for consistency, but to me it feels like over-kill.

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -100,13 +93,17 @@ def add_comments(self, comments: list[str]) -> RegionResource:
...
+        self._check_for_and_convert_unexpected_input()
         res_ele = self._serialise_resource_element()
         res_ele.append(self._serialise_geometry())
         res_ele.extend(self._serialise_values())
         if self.comments:
             res_ele.append(_serialise_has_comment(self.comments, self.res_id))
         return res_ele
 
+    def _check_for_and_convert_unexpected_input(self) -> None:
+        self.comments = _transform_unexpected_input(self.comments, "comments", self.res_id)
```
BalduinLandolt, line 105: as above

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -157,11 +141,18 @@ def add_comments(self, comments: list[str]) -> LinkResource:
...
         res_ele = self._serialise_resource_element()
         res_ele.append(_serialise_has_comment(self.comments, self.res_id))
         res_ele.append(self._serialise_links())
         return res_ele
 
+    def _check_for_and_convert_unexpected_input(self) -> None:
+        _check_strings(string_to_check=self.res_id, res_id=self.res_id, field_name="Resource ID")
+        _check_strings(string_to_check=self.label, res_id=self.res_id, field_name="Label")
+        self.comments = _transform_unexpected_input(self.comments, "comments", self.res_id)
+        self.link_to = _transform_unexpected_input(self.link_to, "link_to", self.res_id)
```
BalduinLandolt, line 154: here it does make sense :)

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -50,11 +42,15 @@ def add_comments(self, comments: list[str]) -> AnnotationResource:
         return self
 
     def serialise(self) -> etree._Element:
+        self._check_for_and_convert_unexpected_input()
```
BalduinLandolt, line 45: Are you sure you want to move these checks from the post_init to the serialise method? To me, validation should be done right away, not in the very end.

BalduinLandolt (approved):
I like the extraction of the logic rather than having the same match statement everywhere. But I'm not so sure it's better to do this in the serialise rather than the post init method. To me, object construction seems like the right place to do validation, and serialisation is potentially too late.  
But if you have strong reasons for this, feel free to merge it as is.

src/dsp_tools/xmllib/models/dsp_base_resources.py
Nora-Olivia-Ammann: I agree conceptually. But we allow people to add more information later. That will not be checked during post init (correct me if I am wrong). So we may not notice some mistakes unless we check at the very end.

src/dsp_tools/xmllib/models/dsp_base_resources.py
Nora-Olivia-Ammann: I can move it

src/dsp_tools/xmllib/models/dsp_base_resources.py
BalduinLandolt: Good point. I suppose the "clean" way would then be to also check in the methods that add this information. But maybe it's just simpler to go for checking in the end.

BalduinLandolt: Sounds reasonable

PR #1185, opened 2024-09-30
feat(xmllib): add factory method to resources (DEV-4189)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -201,6 +259,25 @@ def __post_init__(self) -> None:
             case _:
                 _warn_invalid_comments(self.comments, self.res_id)
 
+    def new(
```
Nora-Olivia-Ammann, line 262: I decided that the factory method only gets the mandatory fields. The optional ones they can add. I think this adds also clarity for the users with regard to mandatory fields.

BalduinLandolt: I'm not sure this is the nicest design choice, because it takes away options from the user, while I'm not sure it adds much benefit - if the optional fields have default values, then the users can simply omit them. But it's up to you.

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -175,13 +208,33 @@ def _serialise_links(self) -> etree._Element:
         return prop_ele
 
 
+@dataclass
+class SegmentBounds:
+    segment_start: float | int | str
+    segment_end: float | int | str
```
BalduinLandolt, line 214: suggestion: IMO this should be correctly typed as `float`, and the checks, warning and conversion could be done either in the calling factory methods, a factory method on `SegmentBounds` or in the `__post_init__` method.  
I don't think there is any real reason to hold on to other types here, eventually they will be turned into float anyways..

BalduinLandolt (approved):
Over all I like it, see some comments below.

src/dsp_tools/xmllib/models/dsp_base_resources.py
Nora-Olivia-Ammann: well it is not actually turned into a float is it? It is a string in the xml.
The reason why I allow it is because sometimes we read csv and force pandas to type everything datatype as string (except for empty). This is because pandas sometimes does weird things with dates. In those cases it is easier to specify that everything is a string. It is also possible that conversions may happen to floats that have a zero after the decimal point. It is really hard as a user to influence these datatype conversions. These are the reasons why I am open. In my opinion the most consistent would be to convert everything into a string and see if the regex matches... What do you think?

Nora-Olivia-Ammann: Right. Well this is a question of handling invalid input. They dislike if we crash everything if the input is invalid. If we do a conversion of something that cannot be turned into a float we will crash... Actually an argument against storing it as a float is this problem:

```
i = np.nan
match i:
    case float():
        print("Found")
    case _:
        print("Not found")
```

The `np.nan` is recognised as float. `pd.NA` is not but there is no guarantee you will not encounter `np.nan` when using pandas. If we always converted it into a string and checked with a regex we would never crash and cover all bases. What do you think?

PR #1186, opened 2024-10-01
feat(xmlvalidate): send validation request to API (DEV-4177)

Nora-Olivia-Ammann:


seakayone (commented):
I am not entirely sure if splitting different parts of the `dsp-api`'s API into instances of distinct classes  is a good thing or if having a stateless `DspApiClient` would be easier to use? 
-> how would you envision that?
If you envision that with a generic api client, we have the connection monster (https://github.com/dasch-swiss/dsp-tools/blob/main/src/dsp_tools/utils/connection_live.py) We previously discussed this and concluded that it would be better to have smaller more specialised clients, that also only handle the types of problems that can actually arise during this call...

seakayone (approved):

seakayone:
> I am not entirely sure if splitting different parts of the `dsp-api`'s API into instances of distinct classes is a good thing or if having a stateless `DspApiClient` would be easier to use? -> how would you envision that? If you envision that with a generic api client, we have the connection monster (https://github.com/dasch-swiss/dsp-tools/blob/main/src/dsp_tools/utils/connection_live.py) 

This is would probably be a lenghty discussion we should have another time in person. 
Long story short, the "connection monster" is mixing certain concerns from which its problems arise.

PR #1189, opened 2024-10-02
feat(xmllib): replace newline with XML tag (DEV-4195)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/value_converters.py
```diff
@@ -18,3 +21,57 @@ def convert_to_bool_string(value: Any) -> str:
...
+            return text
+        case NewlineReplacementTags.newline:
+            return replace_newlines_with_br_tags(text)
+        case NewlineReplacementTags.paragraph:
+            return replace_newlines_with_paragraph_tags(text)
+        case _:
+            raise InputError(f"The conversion option for the text does not exist: {converter_option}.")
+
+
+def replace_newlines_with_paragraph_tags(text: str) -> str:
```
Nora-Olivia-Ammann, line 52: @BalduinLandolt 
Ich erhalte hier ein linter problem:

./src/dsp_tools/xmllib/value_converters.py:52 (replace_newlines_with_paragraph_tags): DAR201 Missing "Returns" in Docstring: - return
./src/dsp_tools/xmllib/value_converters.py:52 (replace_newlines_with_paragraph_tags): DAR101 Missing parameter(s) in Docstring: - text
./src/dsp_tools/xmllib/value_converters.py:67 (replace_newlines_with_br_tags): DAR201 Missing "Returns" in Docstring: - return
./src/dsp_tools/xmllib/value_converters.py:67 (replace_newlines_with_br_tags): DAR101 Missing parameter(s) in Docstring: - text

Ich sehe nicht was ich am docstring anders machen kann. Siehst du den Fehler? Danke!

src/dsp_tools/xmllib/models/user_enums.py
```diff
@@ -0,0 +1,8 @@
+from enum import Enum
+from enum import auto
+
+
+class NewlineReplacementTags(Enum):
+    none = auto()
+    paragraph = auto()
+    newline = auto()
```
BalduinLandolt, line 8: Some suggestions:

- Naming: instead of `NewlineReplacementTags`, how about just `NewlineReplacement`?
- Naming: by convention, variants of an enum are - like global constants - normally all-caps, so the conventional thing to expect would be `NewlineReplacement.NONE`, `NewlineReplacement.PARAGRAPH`, `NewlineReplacement.NEWLINE`
- Naming: The HTML tag `<br/>` is typically "translated" to `linebreak` not `newline`
- Enum: I think this would be a case for using the enum subclass `StrEnum`, but I'm not 100% sure, and it probably doesn't make a difference anyways.

src/dsp_tools/xmllib/models/resource.py
```diff
@@ -311,21 +313,39 @@ def add_simpletext_optional(
...
     def add_richtexts(
-        self, values: list[str], prop_name: str, permissions: str | None = None, comment: str | None = None
+        self,
+        values: list[str],
+        prop_name: str,
+        permissions: str | None = None,
+        comment: str | None = None,
+        newline_replacement: NewlineReplacementTags = NewlineReplacementTags.newline,
     ) -> Resource:
+        values = [replace_newlines_with_tags(str(x), newline_replacement) for x in values]
```
BalduinLandolt, line 335: stylistic nitpick: I don't really like the use of variable names like `x` because they deliberately do not say what the variable means. In cases like this, I would prefer `[... for value in values]` or for brevity's sake even better `[... for v in values]`.

src/dsp_tools/xmllib/value_converters.py
```diff
@@ -18,3 +21,57 @@ def convert_to_bool_string(value: Any) -> str:
...
+    """
+    match converter_option:
+        case NewlineReplacementTags.none:
+            return text
+        case NewlineReplacementTags.newline:
+            return replace_newlines_with_br_tags(text)
+        case NewlineReplacementTags.paragraph:
+            return replace_newlines_with_paragraph_tags(text)
+        case _:
+            raise InputError(f"The conversion option for the text does not exist: {converter_option}.")
```
BalduinLandolt, line 49: I don't think this is even possible with enums?

BalduinLandolt (approved):
Functionally perfect! In terms of naming I added some thoughts, feel free to apply or ignore as you see fit

src/dsp_tools/xmllib/value_converters.py
Nora-Olivia-Ammann: That is true, it was just habit. I removed it.

src/dsp_tools/xmllib/value_converters.py
jnussbaum: Darglint is complaining because the docstring doesn't follow the syntax rules for reStructuredText (rst). rst is the language in which the docstrings are written. 

There are 2 problems here:
- the `\` is not escaped, it should be `\\`
- HTML/XML tags are not surrounded by backticks, so they are interpreted as markup. So they are wrongly rendered: 
  Screenshot 2024-10-07 at 11 30 18

I tried it out, if you fix these 2 things, it will work.

Read more about rst: 
- https://docutils.sourceforge.io/rst.html
- https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html

PR #1190, opened 2024-10-02
feat(xmllib): add migration metadata to resources (DEV-4194)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -53,6 +56,17 @@ def add_comments(self, comments: list[str]) -> AnnotationResource:
         self.comments.extend(comments)
         return self
 
+    def add_migration_metadata(
```
Nora-Olivia-Ammann, line 59: It is rare but Rita requested it, which is why I did not put it into the `new` function.
I plan on asking Noemi about these stylistic choices, she is closer to RDU than me, but will be able to give an opinion without seeing the options perfectly fleshed out.

BalduinLandolt: unfortunate that we need to add it... but I understand that it's a requirement.

how would you feel about adding the `@deprecated("Only used for Salsah migration. Will be removed eventually.")` decorator to it right away, so that no misunderstandings arise?

BalduinLandolt: (of course this goes for all of those, I'm not commenting on each of them)

src/dsp_tools/xmllib/models/resource.py
```diff
@@ -396,3 +398,18 @@ def add_iiif_uri(self, iiif_uri: str, permissions: str | None = None, comment: s
             )
         self.file_value = IIIFUri(iiif_uri, permissions, comment, self.res_id)
         return self
+
+    #######################
+    # Migration Metadata
+    #######################
```
BalduinLandolt, line 404: why add this here, but not everywhere?  
it's up to you, but I'm not a huge fan of these kinds of comments anymore. In a 400L file, you're unlikely to lose track of where things are.

BalduinLandolt (approved):
Looks good, thanks! See my comments for some details.

PR #1194, opened 2024-10-02
feat(start-stack): make api version necessary for validation accessible through a flag (DEV-4200)

Nora-Olivia-Ammann:


src/dsp_tools/commands/start_stack.py
```diff
@@ -285,9 +288,15 @@ def _start_remaining_docker_containers(self) -> None:
         Start the other Docker containers that are not running yet.
         (Fuseki is already running at this point.)
         """
+        compose_str = "docker compose -f docker-compose.yml"
         if self.__stack_configuration.latest_dev_version:
             subprocess.run("docker compose pull".split(), cwd=self.__docker_path_of_user, check=True)
-        subprocess.run("docker compose up -d".split(), cwd=self.__docker_path_of_user, check=True)
+            compose_str += " -f docker-compose.override.yml"
+        elif self.__stack_configuration.api_version_for_validate:
+            subprocess.run("docker compose pull".split(), cwd=self.__docker_path_of_user, check=True)
```
seakayone, line 296: *suggestion*: Remove needless `docker compose pull` call

When you run `docker compose up`, it automatically pulls any missing images. The `docker compose pull` command is mainly useful when you’re using tags like `latest`, which change frequently, and you want to ensure you’re pulling the most up-to-date images from the hub.

Here, the `docker compose pull` is not necessary because we use fixed version as tags for the validate setup.

seakayone (approved):

PR #1195, opened 2024-10-03
refactor(xmllib): improve decimal checks

Nora-Olivia-Ammann:


test/unittests/xmllib/test_value_checkers.py
```diff
@@ -78,22 +79,22 @@ def test_is_geoname_wrong(val: Any) -> None:
...
 
 
-@pytest.mark.parametrize("val", [1, "1"])
+@pytest.mark.parametrize("val", [1, "1", -1])
 def test_is_integer_correct(val: Any) -> None:
     assert is_integer(val)
 
 
-@pytest.mark.parametrize("val", [1.2, "1.2", "wdasd"])
+@pytest.mark.parametrize("val", [1.2, "1.2", "wdasd", True, False, "1e2"])
```
Nora-Olivia-Ammann, line 97: Even though `1e2` represents a whole number, python does not accept it. It is a float.

jnussbaum: The question is: Do we want to check if the input is an integer, or if the input can be casted to an integer? I guess that `int(float("1e2"))` would work.

jnussbaum (approved):

PR #1197, opened 2024-10-03
feat(xmlvalidate): value type checks (DEV-4122)

Nora-Olivia-Ammann:
**This does not include text values. This adds another complexity and will be implemented later.**


src/dsp_tools/commands/xml_validate/sparql/value_shacl.py
```diff
@@ -0,0 +1,135 @@
...
+        "knora-api:GeonameValue": "api-shapes:GeonameValue_ClassShape",
+        "knora-api:IntValue": "api-shapes:IntValue_ClassShape",
+        "knora-api:ListValue": "api-shapes:ListValue_ClassShape",
+        "knora-api:TimeValue": "api-shapes:TimeValue_ClassShape",
+        "knora-api:UriValue": "api-shapes:UriValue_ClassShape",
+    }
+    g = Graph()
+    for object_type, shacl_shape in property_type_mapper.items():
+        g += _construct_one_property_type_shape_based_on_object_type(onto, object_type, shacl_shape)
+    return g
```
seakayone, line 68: *thought*: There seems to be a naming convention hidden in the `property_type_mapper`

IMHO it is worth expressing this naming convention explicitly.

For example by adding an inner function:
```python
def _construct_property_type_shape_based_on_object_type(onto: Graph) -> Graph:
    def as_object_type_and_shacl_shape(property_type: str) -> tuple[str, str]:
        return "knora-api:" + property_type + "Value", "api-shapes:" + property_type + "Value_ClassShape"

    property_types = {"Boolean", "Color", "Date", "Decimal", "Geoname", "Int", "List", "Time", "Uri"}
    g = Graph()
    for t in property_types:
        object_type, shacl_shape = as_object_type_and_shacl_shape(t)
        g += _construct_one_property_type_shape_based_on_object_type(onto, object_type, shacl_shape)
    return g
```

seakayone (approved):

PR #1198, opened 2024-10-03
fix(xmlvalidate): duplicate cardinalities of LinkValues (DEV-4203)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xml_validate/sparql/resource_shacl.py
```diff
@@ -79,11 +80,13 @@ def _construct_1_cardinality(onto_graph: Graph) -> Graph:
     
       ?class a owl:Class ;
           knora-api:isResourceClass true ;
+          knora-api:canBeInstantiated true ;
           rdfs:subClassOf ?restriction .
       ?restriction a owl:Restriction ;
           owl:onProperty ?propRestriction ;
           salsah-gui:guiOrder ?order ;
           owl:cardinality 1 .
+      FILTER NOT EXISTS { ?propRestriction knora-api:isLinkValueProperty true }
```
Nora-Olivia-Ammann, line 89: If a property was a link property eg. `hasLinkTo` then we also get the `LinkValue` property `hasLinkToValue`, we do not want that in the cardinalities as we do not include it in the data.
Because of the in-built link properties for compound objects we need `knora-api` information.
Currently the validation does not like `knora-api` so it is only used for the construction of the SHACL but not during the validation. I am unsure if we have a problem in `knora-api` or if it simply does not conform to TopBraid's expectations...

src/dsp_tools/commands/xml_validate/xml_validate.py
```diff
@@ -39,10 +40,11 @@ def xml_validate(filepath: Path, api_url: str, dev_route: bool) -> bool:  # noqa
     _inform_about_experimental_feature()
     data_rdf, shortcode = _get_data_info_from_file(filepath, api_url)
     onto_con = OntologyConnection(api_url, shortcode)
-    ontologies = _get_project_ontos(onto_con)
-    data_graph = data_rdf.make_graph() + ontologies
+    ontologies, shapes = _get_shacl(onto_con)
+    data_graph = data_rdf.make_graph()
+    # data_graph += ontologies
```
Nora-Olivia-Ammann, line 45: This is a cheap fix and the basis for our meeting on monday, if we exclude the ontology from the val.
Currently our shapes are closed, so that no properties can be used that are not ennumerated. What I did not consider, this is a problem when a sub-class has properties that it's super-class does not have. At the moment not having inheritance solves the problem. The newly crated correct test data passes. This is not an option as soon as we validate the range of the link properties. As this is a future feature I decided to implement this.

seakayone (approved):

src/dsp_tools/commands/xml_validate/sparql/resource_shacl.py
seakayone: I guess we should take some time then investigating the knora-api ontology.

Nora-Olivia-Ammann: Yes, my plans are:
- Ask stackoverflow how inheritance with closed shapes should be approached, because I understand that it is actually contradictory. Perhaps we can hinder SHACL inheritance (shapes are also inherited, just as property cardinalities are)
- There is another weird behaviour regarding property inheritance, I am not sure if TopBraid is doing something weird or we are, I need to check further but I think in our data it is an edge case.
- Investiage `knora-api` because I don't think it is possible not to include it once we are implementing the built-in dsp classes. But that comes last so in the mean time I think it is ok like this.

We can discuss it further on Monday.

PR #1208, opened 2024-10-08
feat(validate-data): validate generic content of values (DEV-4225)

Nora-Olivia-Ammann:
This excludes ontology specific content from `ListValues` and `LinkValues`


test/unittests/commands/validate_data/fixtures_validation_result.py
```diff
@@ -0,0 +1,207 @@
...
+            sh:value <http://data/value-iri-geoname_not_number> ] .
+
+    _:bn_geoname_not_number a sh:ValidationResult ;
+    sh:focusNode <http://data/value-iri-geoname_not_number> ;
+    sh:resultMessage "The value must be a valid geoname code" ;
+    sh:resultPath knora-api:geonameValueAsGeonameCode ;
+    sh:resultSeverity sh:Violation ;
+    sh:sourceConstraintComponent sh:PatternConstraintComponent ;
+    sh:sourceShape api-shapes:geonameValueAsGeonameCode_Shape ;
+    sh:value "llllllllllllllllll" .
```
seakayone, line 118: *question* Is this the intended value?

This seems odd to me. I understand that this is "just" test data, but something more realistic would have not made me think. How about "this-is-not-a-valid-code"?

This comment also applies to the other parts of the code where you have used only `l`s.

Nora-Olivia-Ammann: That is a good point.

PR #1209, opened 2024-10-09
feat(validate-data): validate cardinality with `dash:closedByTypes` (DEV-4231)

Nora-Olivia-Ammann:


src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -20,6 +20,12 @@
 # RDFS:LABEL SHAPE
 ########################
 
+api-shapes:rdfsLabel_Shape
```
Nora-Olivia-Ammann, line 23: `sh:ignoredProperties` does not work with `dash`, which means that we have to explicitly mention the label now, which we planned on doing anyway.

PR #1211, opened 2024-10-09
feat(validate-data): make content validation conform with `dash` syntax (DEV-4233)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -92,25 +81,23 @@ def _inform_about_experimental_feature() -> None:
...
     api_shapes = Graph()
     api_shapes.parse("src/dsp_tools/resources/validate_data/api-shapes.ttl")
-    content_shapes += api_shapes
-    card_shapes += api_shapes
-    return RDFGraphs(
-        data=data_rdf.make_graph(), ontos=ontologies, cardinality_shapes=card_shapes, content_shapes=content_shapes
-    )
+    shapes_graph += api_shapes
+    data_onto = data_rdf.make_graph() + ontologies
+    return RDFGraphs(data_onto=data_onto, ontos=ontologies, shapes=shapes_graph)
```
seakayone, line 100: *question* Wouldn't it make sense to keep all graphs separated?

I understand that the current code only needs the data with the ontology.
That is probably why the `data_onto` and the `onto` property both contain the `ontologies` graph which introduces redundant information in the different graphs.
IMHO it might be conceptually "cleaner" and potentially less confusing to keep the `data` graph as a separate properties.
Then you can add a `data_onto` method on `RDFGraphs` to return the combined `data + ontologies` on demand and also a `shapes_onto` method which combines the shapes and the ontology which are used together as well.

src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -130,38 +117,25 @@ def _save_graphs(filepath: Path, rdf_graphs: RDFGraphs) -> Path:
...
     rdf_graphs.ontos.serialize(f"{generic_filepath}_ONTO.ttl")
-    rdf_graphs.cardinality_shapes.serialize(f"{generic_filepath}_SHACL_CARD.ttl")
-    rdf_graphs.content_shapes.serialize(f"{generic_filepath}_SHACL_CONTENT.ttl")
-    rdf_graphs.data.serialize(f"{generic_filepath}_DATA.ttl")
-    onto_data = rdf_graphs.data + rdf_graphs.ontos
+    shacl_onto = rdf_graphs.shapes + rdf_graphs.ontos
+    shacl_onto.serialize(f"{generic_filepath}_SHACL_ONTO.ttl")
+    rdf_graphs.shapes.serialize(f"{generic_filepath}_SHACL.ttl")
+    rdf_graphs.data_onto.serialize(f"{generic_filepath}_DATA.ttl")
+    onto_data = rdf_graphs.data_onto + rdf_graphs.ontos
```
seakayone, line 124: *issue* This is redunant because `data_onto` already contains the `ontos`.

```suggestion
    onto_data = rdf_graphs.data_onto 
```

Probably you can inline `onto_data` without loosing clarity of the rest of the code.

src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -130,38 +117,25 @@ def _save_graphs(filepath: Path, rdf_graphs: RDFGraphs) -> Path:
...
+    conforms = bool(next(validation_results.objects(None, SH.conforms)))
+    return ValidationReport(
         conforms=conforms,
-        content_validation=content_report,
-        cardinality_validation=card_report,
-        shacl_graphs=rdf_graphs.cardinality_shapes + rdf_graphs.content_shapes,
-        data_graph=rdf_graphs.data,
+        validation_graph=validation_results,
+        shacl_graphs=rdf_graphs.shapes,
+        data_graph=rdf_graphs.data_onto,
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: Yes it is, otherwise inheritance won't work. I will change the name. Thanks!

src/dsp_tools/commands/validate_data/validate_data.py
Nora-Olivia-Ammann: That was my initial intention, it just got a bit messy. I will change it, but I think it might also be worth it to sit together and discuss naming. I changed it back and dealt with the issue when serialising it into a string for the api.

seakayone (approved):

PR #1212, opened 2024-10-09
test(validate-data): add integration tests

Nora-Olivia-Ammann:


test/integration/commands/validate_data/test_reformat_validaton_result.py
```diff
@@ -3,9 +3,11 @@
 import pytest
 from rdflib import Graph
 
+from dsp_tools.commands.validate_data.models.input_problems import ContentRegexViolation
```
Nora-Olivia-Ammann, line 6: I have not previously tested the value type and content, because I need the data as ttl too. Otherwise I get a problem with the bn IDs that are different. I did not want to maintain two test data files. But now I think we can easily move the assertions to e2e tests so we can in the long term get rid of the files. For the meantime it gives us some security.

seakayone: I understand. The good thing is that you have already "modularized" the test data into separate files.

test/integration/commands/validate_data/test_reformat_validaton_result.py
```diff
@@ -26,31 +28,163 @@ def result_cardinality_violation() -> Graph:
...
+def result_card_every_constraint_once() -> Graph:
+    g = Graph()
+    g.parse("testdata/validate-data/validation_results/every_combination_once_result_cardinality.ttl")
+    return g
+
+
+@pytest.fixture
+def result_content_every_constraint_once() -> Graph:
+    g = Graph()
+    g.parse("testdata/validate-data/validation_results/every_constraint_once_result_content.ttl")
```
seakayone, line 76: *idea* Align fixture function names with file names.

The function names and the file names do not seem to follow a strict system for their naming. IMHO it would be beneficial to have the function named the same way the file is named. This would make it easy when reading the test to open the test file without having to navigate to the fixture method and read it first.

testdata/validate-data/data/value_type_violation.xml
```diff
@@ -14,75 +14,75 @@
         For the text-prop the encoding was switched
      -->
 
-    <resource label="Empty" restype=":ClassWithEverything" id="id_0"/>
+    <resource label="Empty" restype=":ClassWithEverything" id="empty_ok"/>
```
seakayone, line 17: **praise** I like the id strings which explain their intention, makes it so much easier to read.

seakayone (approved):

PR #1214, opened 2024-10-10
test(validate-data): turn integration tests into e2e

Nora-Olivia-Ammann:
- I copy-pasted the set-up we have with the test containers. The only change there is that the API version is taken from the correct docker-compose-override.
- It created a lot of redundancies, but I think this allows us to delete it easily in the end.
- The test assertions did not change with regard to content when I moved them from integration.
- In this PR only the set-up has to be reviewed.
- The ttl test files are no longer necessary because we now have live results.


test/e2e_validate_data/test_validate_data.py
```diff
@@ -0,0 +1,176 @@
...
+
+@pytest.fixture
+def _create_project() -> Iterator[None]:
+    with get_containers():
+        success = create_project(Path("testdata/validate-data/project.json"), CREDS, verbose=True)
+        assert success
+        yield
+
+
+@lru_cache(maxsize=None)
```
Nora-Olivia-Ammann, line 32: It is quite slow so I tried to solve this with a cache, please comment if you have any other ideas

test/e2e_validate_data/setup_testcontainers.py
```diff
@@ -0,0 +1,205 @@
...
+    _print_containers_are_ready(containers)
+    return containers
+
+
+def _get_image_version(docker_compose_content: str, component: str) -> str:
+    match = regex.search(rf"image: daschswiss/{component}:([^\n]+)", docker_compose_content)
+    return match.group(1) if match else "latest"
+
+
+def _get_api_image_version() -> str:
```
Nora-Olivia-Ammann, line 82: This is the only difference to the other set-up

test/e2e_validate_data/test_validate_data.py
```diff
@@ -0,0 +1,176 @@
...
+from dsp_tools.commands.validate_data.models.input_problems import NonExistentCardinalityViolation
+from dsp_tools.commands.validate_data.models.input_problems import ValueTypeViolation
+from dsp_tools.commands.validate_data.models.validation import ValidationReports
+from dsp_tools.commands.validate_data.reformat_validaton_result import reformat_validation_graph
+from dsp_tools.commands.validate_data.validate_data import _get_validation_result
+from test.e2e_validate_data.setup_testcontainers import get_containers
+
+CREDS = ServerCredentials("root@example.com", "test", "http://0.0.0.0:3333")
+LOCAL_API = "http://0.0.0.0:3333"
+SAVE_GRAPHS = False
```
seakayone, line 21: *idea* How about renaming this variable to `DONT_SAFE_GRAPHS`?

I assume the name of the function parameter is something like save_graph. However, when using this variable during the function call it is misleading as the name `SAVE_GRAPH` suggests the value to be true.

test/e2e_validate_data/test_validate_data.py
seakayone: I am not familiar enough with python testcontainers and pytest fixtures to give a hint here. 

Currently all validation reports in these fixtures really do not depend on the state of the api so caching them seems fine to me.

seakayone (approved):

test/e2e_validate_data/setup_testcontainers.py
seakayone: *question*: Would you be able to remove this by using the latest dsp-api release?

This is a short term solution and should be gone in a couple of days, so it is fine to me as long as we are removing it soon.

The latest api version [was released and contains the necessary endpoint](https://github.com/dasch-swiss/dsp-api/releases/tag/v30.20.0), are you able switch to it and do not need this additional setup at all?

Nora-Olivia-Ammann: I have thought about that, but I want to mimic the experience we get when we actually call the command. This is not the case if we use latest.

jnussbaum (approved):
LGTM. There is just 1 little thing where I think the setup is not entirely correct (see my comment).

test/e2e_validate_data/conftest.py
```diff
@@ -0,0 +1,19 @@
+import shutil
+from typing import Iterator
+
+import pytest
+
+from test.e2e.setup_testcontainers import INGEST_DB
+from test.e2e.setup_testcontainers import SIPI_IMAGES
+from test.e2e.setup_testcontainers import TMP_INGEST
+from test.e2e.setup_testcontainers import TMP_SIPI
```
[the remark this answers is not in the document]
jnussbaum: Well, I just realized that it probably doesn't matter.

PR #1215, opened 2024-10-11
feat(validate-data): validate link value class type (DEV-4237)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/reformat_validaton_result.py
```diff
@@ -205,31 +218,65 @@ def _reformat_one_with_detail(val_result: ResultWithDetail) -> InputProblem | Un
...
 
+def _reformat_detail_class_constraint_component(val_result: ResultWithDetail) -> InputProblem:
+    subject_id = _reformat_data_iri(str(val_result.res_iri))
+    prop_name = _reformat_onto_iri(str(val_result.property))
+    res_type = _reformat_onto_iri(str(val_result.res_class))
+    actual_type = _reformat_onto_iri(str(val_result.detail.value_type)).replace("knora-api:", "")
+    match val_result.detail.result_path:
+        case API_SHAPES.linkValueHasTargetID:
+            value = _reformat_data_iri(str(val_result.detail.value))
+            if val_result.detail.results_message == "Resource":
```
Nora-Olivia-Ammann, line 256: This is not a very good solution.
In my next PR I am planning on completely refactoring the queries for result details. The requirements for the different validation reports are too different to query all the information and do triage then. It is not feasible with a large graph. But I want to implement these changes in a separate PR that is not also a feature.

testdata/validate-data/data/cardinality_correct.xml
```diff
@@ -31,7 +31,7 @@
             <boolean>true</boolean>
         </boolean-prop>
         <resptr-prop name=":testHasLinkToCardOneResource">
-            <resptr>id_2</resptr>
+            <resptr>card_1</resptr>
```
Nora-Olivia-Ammann, line 34: Actually found a mistake in test data that should be correct. It works! 😄

PR #1217, opened 2024-10-11
fix(create): always respect admin status of user in JSON file (DEV-4239)

jnussbaum:


Nora-Olivia-Ammann (approved):
This is a semi-blind review. It looks feasible, but there is a lot of spaghetti which makes it hard to understand.

PR #1220, opened 2024-10-16
fix(xmllib): turn `Resource.new()` into `staticmethod` (DEV-4250)

Nora-Olivia-Ammann:


jnussbaum (approved):
LGTM. Just 2 minor remarks: 
- I would use the prefix "chore" instead of "fix", because this is not a bug.
- I would add the appropriate labels to the Linear ticket, and assign it to you.

PR #1224, opened 2024-10-16
refactor(validate-data): optimise queries for user results (DEV-4244)

Nora-Olivia-Ammann:


src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -169,14 +169,6 @@ api-shapes:valueAsString_Shape
   sh:datatype xsd:string ;
   sh:severity sh:Violation .
 
-api-shapes:SimpleTextValue_PropShape
```
Nora-Olivia-Ammann, line 172: Found that these two were duplicates...

seakayone (approved):

PR #1228, opened 2024-10-18
fix: add missing dependency in xmllib (DEV-4254)

jnussbaum:


Nora-Olivia-Ammann (approved):

test/distribution/test_xmllib.py
```diff
@@ -0,0 +1,22 @@
...
+    Indirect imports are okay, too, i.e. modules that are imported by modules that are imported here.
+    """
+    root = XMLRoot("0000", "my_onto")
+    my_res = Resource.new("my_res", "restype", "label")
+    if is_bool_like("0"):
+        my_res.add_bool(convert_to_bool_string("0"), "my_prop")
+    my_res.add_file("my_file")
+    annotation = AnnotationResource.new("res_id", "label", "annotation_of", ["comment1", "comment2"])
+    root.add_resource(annotation)  # type: ignore[arg-type]
+    root.add_resource(my_res)
```
jnussbaum, line 22: Not sure how smart this test is. It's just the only idea that I had how to solve the problem.

Nora-Olivia-Ammann: It is better than nothing. We will have to take care to test it better later. Perhaps we can take some of noemi's scripts for that

PR #1235, opened 2024-10-21
fix(validate-data): enable usage of other ontologies besides the default ontology (DEV-4263)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -141,27 +143,49 @@ def _validate(validator: ShaclValidator, rdf_graphs: RDFGraphs) -> ValidationRep
 
 
 def _get_data_info_from_file(file: Path, api_url: str) -> tuple[DataRDF, str]:
-    cleaned_root = _parse_and_clean_file(file, api_url)
-    deserialised: ProjectDeserialised = deserialise_xml(cleaned_root)
+    xml_project = _parse_and_clean_file(file, api_url)
+    deserialised: ProjectDeserialised = deserialise_xml(xml_project.root)
```
Nora-Olivia-Ammann, line 147: In the a next PR I will break up this function, to check if all the ontologies are uploaded, before I construct the `DataRDF`. I prepared this through a property in the `XMLProject` class (if you have renaming suggestions I am open, but I want to indicate that it is still an XML). But have not used it yet.

seakayone (approved):

PR #1236, opened 2024-10-22
feat(validate-data): add unique value constraint (DEV-4202)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/sparql/cardinality_shacl.py
```diff
@@ -88,7 +87,9 @@ def _construct_1_cardinality(onto_graph: Graph) -> Graph:
     
       ?propRestriction knora-api:isEditable true .
       FILTER NOT EXISTS { ?propRestriction knora-api:isLinkValueProperty true }
-    
+      FILTER (?class NOT IN (
```
Nora-Olivia-Ammann, line 90: Because we include knora-api now, these cause cardinality violations because we don't construct the triples for them yet. Once we implement these we can remove this filter.

src/dsp_tools/commands/validate_data/sparql/value_shacl.py
```diff
@@ -52,6 +53,61 @@ def _add_property_shapes_to_class_shapes(onto: Graph) -> Graph:
...
+
+                    $this ?path ?valueClass .
+                    
+                    {
+                        ?prop rdfs:subPropertyOf knora-api:valueHas .
+                        ?valueClass ?prop ?value .
+                    }
+                    UNION
+                    {
+                        ?valueClass knora-api:valueAsString|api-shapes:linkValueHasTargetID|api-shapes:listNodeAsString ?value .
```
Nora-Olivia-Ammann, line 88: I very much don't like this query. I have tried different approaches without success.
But I think this is an optimisation we can take a look at in the future.

seakayone (approved):

PR #1237, opened 2024-10-22
docs: fix documentation of `--no-iiif-uri-validation`

jnussbaum:


docs/cli-commands.md
```diff
@@ -144,8 +144,8 @@ The following options are available:
 - `-V` | `--validate` (optional): validate the XML file without uploading it
 - `--interrupt-after=int` (optional): interrupt the upload after `int` resources have been uploaded
 - `--suppress-update-prompt` (optional): don't prompt when using an outdated version of DSP-TOOLS 
-- `--no-iiif-uri-validation` (optional): don't check if the IIIF links are conform to the standard schema
   (useful for contexts without interactive shell, e.g. when the Terminal output is piped into a file)
+- `--no-iiif-uri-validation` (optional): don't check if the IIIF links are conform to the standard schema
```
Nora-Olivia-Ammann, line 148: This is actually not correct either, because if you validate them it makes a network call

jnussbaum: Can you elaborate? I don't understand your concern.

Background of my change: The thing with the interactive shell belongs to `--suppress-update-prompt`, not to `--no-iiif-uri-validation`.

I didn't really change `--no-iiif-uri-validation`, I just moved the line with the interactive shell to where it belongs.

Nora-Olivia-Ammann: Yes but actually the flag description is not correct. 
https://github.com/dasch-swiss/dsp-tools/blob/main/src/dsp_tools/commands/xmlupload/iiif_uri_validator.py
(not saying it is your fault, just that if you change it, it would be nice if you could correct it.)

jnussbaum: Ah, okay. What would be a better description? Perhaps something like this?

```suggestion
- `--no-iiif-uri-validation` (optional): don't check if the IIIF links are valid URLs that can be reached online.
```

PR #1238, opened 2024-10-23
refactor: Move authentication out of general purpose connection (DEV-3762)

BalduinLandolt:


src/dsp_tools/commands/project/models/group.py
```diff
@@ -224,10 +224,6 @@ def _toJsonObj_update(self) -> dict[str, Any]:
             tmp["selfjoin"] = self._selfjoin
         return tmp
 
-    def delete(self) -> Group:
```
Nora-Olivia-Ammann, line 227: Does this have any connection or did you just stuble upon dead code?

src/dsp_tools/utils/authentication_client.py
```diff
@@ -0,0 +1,12 @@
+from typing import Protocol
+
+# ruff: noqa: D102 (missing docstring in public method)
+
+
+class AuthenticationClient(Protocol):
+    """
+    Protocol for a client that can authenticate with a DSP server and return a token.
```
Nora-Olivia-Ammann, line 8: *returns

src/dsp_tools/utils/connection_live.py
```diff
@@ -335,49 +263,46 @@ def _log_and_raise_timeouts(self, error: TimeoutError | ReadTimeout) -> Never:
...
+            if "password" in data:
+                data["password"] = "***"
+            dumpobj["data"] = data
         if params.files:
             dumpobj["files"] = params.files["file"][0]
         logger.debug(f"REQUEST: {json.dumps(dumpobj, cls=SetEncoder)}")
+
+
+def _sanatize_headers(headers: dict[str, str | bytes]) -> dict[str, str]:
+    def _mask(k: str, s: str | bytes) -> str:
```
Nora-Olivia-Ammann, line 292: would be more readable if you used proper names not just one letter

test/unittests/utils/test_connection_live.py
```diff
@@ -348,9 +290,9 @@ def test_log_request() -> None:
     expected_output = {
         "method": "GET",
         "url": "http://example.com/",
-        "headers": {"Authorization": "Bearer my-ve[+13]", "request-header": "request-value"},
+        "headers": {"Authorization": "Bearer ***", "request-header": "request-value"},
         "timeout": 1,
-        "data": {"password": "************************", "foo": "bar"},
+        "data": {"password": "***", "foo": "bar"},
```
Nora-Olivia-Ammann, line 295: would the password still be in the request of connection? is that not the point to remove it?

Nora-Olivia-Ammann (approved):

test/unittests/utils/test_connection_live.py
BalduinLandolt: I was confused by that too: This is needed for the `create` command, when creating new users, because there we of course also have the password in the payload.  
I would have preferred not to have to do that check - but I suppose that would mean having a separate connection for the `create` command... that might actually be a good idea, but is beyond the scope of what I wanted to do here.

src/dsp_tools/commands/project/models/group.py
BalduinLandolt: when investigating the call hierarchy of `connection.logout()`, I came across some methods that were used in tests only, I think this was one of them

src/dsp_tools/utils/connection_live.py
BalduinLandolt: I tend to not mind short variables when they have such a small scope. But you're right, it's more readable otherwise: https://github.com/dasch-swiss/dsp-tools/pull/1238/commits/891bce8935808997e10cac4263e932e87ac5adb9

src/dsp_tools/utils/connection_live.py
```diff
@@ -335,49 +263,46 @@ def _log_and_raise_timeouts(self, error: TimeoutError | ReadTimeout) -> Never:
     def _log_response(self, response: Response) -> None:
         dumpobj: dict[str, Any] = {
             "status_code": response.status_code,
-            "headers": self._anonymize(dict(response.headers)),
+            "headers": _sanatize_headers(dict(response.headers)),
         }
         try:
-            dumpobj["content"] = self._anonymize(json.loads(response.text))
+            dumpobj["content"] = response.json()
         except json.JSONDecodeError:
```
jnussbaum, line 270:
```suggestion
        except requests.exceptions.JSONDecodeError:
```

src/dsp_tools/utils/connection_live.py
```diff
@@ -335,49 +263,46 @@ def _log_and_raise_timeouts(self, error: TimeoutError | ReadTimeout) -> Never:
...
+            data = params.data.copy()
+            if "password" in data:
+                data["password"] = "***"
+            dumpobj["data"] = data
         if params.files:
             dumpobj["files"] = params.files["file"][0]
         logger.debug(f"REQUEST: {json.dumps(dumpobj, cls=SetEncoder)}")
+
+
+def _sanatize_headers(headers: dict[str, str | bytes]) -> dict[str, str]:
```
jnussbaum, line 291:
```suggestion
def _sanitize_headers(headers: dict[str, str | bytes]) -> dict[str, str]:
```

test/unittests/utils/test_connection_live.py
```diff
@@ -157,11 +124,9 @@ def test_default_timeout() -> None:
...
         expected_params: RequestParameters = con._try_network_action.call_args.args[0]
         assert expected_params.timeout == con.timeout_put_post, f"Method '{method.__name__}' failed"
-    for method in (con.get, con.delete):  # type: ignore[assignment]
-        method = cast(Callable[..., Any], method)
-        method(route="/v2/resources")
-        expected_params = con._try_network_action.call_args.args[0]
-        assert expected_params.timeout == con.timeout_get_delete, f"Method '{method.__name__}' failed"
+    con.get(route="/v2/resources")
+    expected_params = con._try_network_action.call_args.args[0]
+    assert expected_params.timeout == con.timeout_get, f"Method '{method.__name__}' failed"
```
jnussbaum, line 129:
```suggestion
    assert expected_params.timeout == con.timeout_get, f"Method 'GET' failed"
```

jnussbaum (approved):
LGTM

src/dsp_tools/utils/authentication_client_live.py
```diff
@@ -0,0 +1,47 @@
...
+        if self._token:
+            return self._token
+        return self._get_token()
+
+    def _get_token(self) -> str:
+        url = f"{self.server}/v2/authentication"
+        payload = {"email": self.email, "password": self.password}
+        logger.info(f"Requesting token from url '{url}' for user '{self.email}'.")
+        try:
+            response: dict[str, Any] = requests.post(url, json=payload, timeout=10).json()
```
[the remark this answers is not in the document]
BalduinLandolt: Added in https://github.com/dasch-swiss/dsp-tools/pull/1238/commits/d80e124f73852d085e1044edfdd634676fe7a061 but as we discussed before, the whole setup with the user-agent header will eventually have to change

src/dsp_tools/utils/authentication_client_live.py
```diff
@@ -0,0 +1,47 @@
...
+        Returns a token. If no token is available, it will get one from the server.
+        """
+        if self._token:
+            return self._token
+        return self._get_token()
+
+    def _get_token(self) -> str:
+        url = f"{self.server}/v2/authentication"
+        payload = {"email": self.email, "password": self.password}
+        logger.info(f"Requesting token from url '{url}' for user '{self.email}'.")
```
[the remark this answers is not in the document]
BalduinLandolt: I see your point but I'm not really inclined to follow it. Part of why we wanted to separate authentication out of the general connection was that it made logging nasty, so I think it's the wrong default to "want to do it the same way" here.  
Additionally, the biggest reason why we always have to ask "what request was sent" is because there is such a wide variety of things that may be sent through the general purpose connection to the API. On the other hand, with this authentication client we know exactly what request has been sent, and all the moving parts (apart from the password which we wouldn't log anyways) are contained in the log I added; and for the response, if we get an OK, we have a pretty clear idea what happened, and if not, an error is propagated outward. So I do in fact think we have all the necessary information, as it stands now.  
So I'm not too happy to invest work to generate code that will need to be maintained, without it adding any extra value.

However, I'm absolutely willing to change the log message, if that helps you navigate the logs. What I could see right away, to align it with the existing:
- change log level to `debug`
- Add prefix `REQUEST: `
- Add a log line in the OK case, with the prefix `RESPONSE: `

Would that address your concerns?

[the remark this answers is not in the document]
jnussbaum: > So I do in fact think we have all the necessary information, as it stands now.
You're right.

> Would that address your concerns?
Yes :-)

PR #1242, opened 2024-10-24
fix(validate-data): revert to exclusion of properties through gui-order (DEV-4279)

Nora-Olivia-Ammann:
Excluding properties through the gui-order is not a good idea because it is optional.

Better is through the triple: `?propRestriction knora-api:isEditable true .`

However, file values are editable too but are not yet implemented. Until file values are implemented it is easiest exclude properties through the gui-order.

### DIFF
--- a/src/dsp_tools/commands/validate_data/sparql/cardinality_shacl.py
+++ b/src/dsp_tools/commands/validate_data/sparql/cardinality_shacl.py
@@ -38,7 +38,6 @@ def _construct_resource_nodeshape(onto_graph: Graph) -> Graph:
         ?class a owl:Class ;
                knora-api:isResourceClass true ;
                knora-api:canBeInstantiated true .
-
     }
     """
     if results_graph := onto_graph.query(query_s).graph:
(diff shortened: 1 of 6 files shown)


src/dsp_tools/commands/validate_data/sparql/cardinality_shacl.py
```diff
@@ -75,20 +70,16 @@ def _construct_1_cardinality(onto_graph: Graph) -> Graph:
...
-    
     } WHERE {
-    
       ?class a owl:Class ;
           knora-api:isResourceClass true ;
           knora-api:canBeInstantiated true ;
           rdfs:subClassOf ?restriction .
       ?restriction a owl:Restriction ;
           owl:onProperty ?propRestriction ;
+          salsah-gui:guiOrder ?order ;
```
Nora-Olivia-Ammann, line 80: I already created a new ticket to revert it once we added the file values:
https://linear.app/dasch/issue/DEV-4281/change-exclusion-of-properties-from-gui-order-to-editable

seakayone (approved):

PR #1244, opened 2024-10-29
feat(xmllib): add optional value to dsp base resource (DEV-4289)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -328,14 +356,24 @@ def add_description_multiple(self, descriptions: list[str]) -> VideoSegmentResou
...
         return self
 
-    def add_keyword(self, keywords: str) -> VideoSegmentResource:
-        self.keywords.append(keywords)
+    def add_description_optional(self, description: Any) -> VideoSegmentResource:
+        if not pd.isna(description):
+            self.descriptions.append(description)
+        return self
+
+    def add_keyword(self, keyword: str) -> VideoSegmentResource:
```
Nora-Olivia-Ammann, line 364: @Notheturtle I noticed that the variable was in the plural. I fixed this, but it may break a script of yours.

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -312,6 +328,13 @@ def add_title(self, title: str) -> VideoSegmentResource:
         self.title = title
         return self
 
+    def add_title_optional(self, title: Any) -> VideoSegmentResource:
+        if not pd.isna(title):
+            if self.title:
+                _warn_value_exists(self.title, title, "title", self.res_id)
```
jnussbaum, line 334: Calling a function with 4 different parameters of the same type, without using keywords, is prone to errors. Nothing will warn you if you mistakenly flip the order of the parameters.

Suggestion: make these parameters keyword-only. You can do this by prepending a `*` to the parameter list:
```
def _warn_value_exists(*, old_value: Any, new_value: Any, value_field: str, res_id: str | None) -> None
```

You will then be forced to always use keyword parameters

jnussbaum (approved):
LGTM

PR #1250, opened 2024-10-30
fix(validate-data): extract richtext as string if it starts with a tag (DEV-4280)

Nora-Olivia-Ammann:


testdata/validate-data/data/value_type_violation.xml
```diff
@@ -74,7 +72,7 @@
 
     <resource label="Richtext" restype=":ClassWithEverything" id="id_richtext">
         <text-prop name=":testRichtext">
-            <text encoding="utf8">Text</text>
+            <text encoding="utf8"><p>Text</p></text>
```
Nora-Olivia-Ammann, line 75: This should be richtext, this should not raise an error that the content is wrong.

src/dsp_tools/commands/validate_data/deserialise_input.py
```diff
@@ -129,7 +129,14 @@ def _deserialise_text_prop(prop: etree._Element) -> list[SimpleTextDeserialised
...
+                all_vals.append(SimpleTextDeserialised(prop_name, _get_text_as_string(val)))
             case "xml":
-                all_vals.append(RichtextDeserialised(prop_name, val.text))
+                all_vals.append(RichtextDeserialised(prop_name, _get_text_as_string(val)))
     return all_vals
+
+
+def _get_text_as_string(value: etree._Element) -> str | None:
+    if len(value):
+        return "".join(etree.tostring(child, encoding="unicode") for child in value)
```
jnussbaum, line 140: I think this could fail with nested tags. I guess you need something like `value.iterdescendants()`. But to be sure, it would be good to add some more test cases:

- no tags at all
- one tag at the start
- one tag in the middle
- several tags in the middle

jnussbaum (approved):

PR #1251, opened 2024-10-30
fix(validate-data): don't crash if list nodes have special characters (DEV-4299)

Nora-Olivia-Ammann:


test/e2e_validate_data/test_validate_data.py
```diff
@@ -268,6 +307,28 @@ def test_reformat_unique_value_violation(self, unique_value_violation: Validatio
...
+        assert not result.unexpected_results
+        assert len(result.problems) == len(expected_tuples)
+        sorted_problems = sorted(result.problems, key=lambda x: x.res_id)
+        for prblm, expected in zip(sorted_problems, expected_tuples):
+            if isinstance(prblm, GenericProblem):
+                assert prblm.res_id == expected[0]
+                assert prblm.problem == expected[1]
+                assert prblm.actual_content == expected[2]
+            elif isinstance(prblm, ContentRegexProblem):
+                assert prblm.res_id == expected[0]
```
Nora-Olivia-Ammann, line 330: Once we implemented the encoding problems, we can change this

src/dsp_tools/commands/validate_data/models/api_responses.py
```diff
@@ -26,10 +26,16 @@ def hlist(self) -> str:
...
     def shacl_list(self) -> str:
-        return f'( "{self.list_name}" )'
+        return f"( {self._escape(self.list_name)} )"
+
+    def escaped_list_name(self) -> str:
+        return self.list_name.replace("\\", "\\\\").replace("'", "\\'").replace('"', '\\"')
+
+    def _escape(self, in_str: str) -> str:
+        replace_problems = in_str.replace("\\", "\\\\").replace("'", "\\'").replace('"', '\\"')
+        return '"' + replace_problems + '"'
```
seakayone, line 41: **Issue**: Escaping should reusable and context aware

I propose we have dedicated functions for escaping rather than adding it as methods to a data class.

Having it on the data class is convenient to use but hard to reuse

Also, escaping in general is heavily depending on the context of where it is applied, i.e. it is probably different for Turtle as compared in a json context.

testdata/validate-data/special_characters/special_characters_correct.xml
```diff
@@ -0,0 +1,40 @@
...
+        </list-prop>
+    </resource>
+
+    <resource label="Single Quote" restype=":ClassWithEverything" id="list_single_quote">
+        <list-prop list="list \ ' space" name=":testListProp">
+            <list>single quote '</list>
+        </list-prop>
+    </resource>
+
+    <resource label="simpletext" restype=":ClassWithEverything" id="simpletext_non_ascii">
```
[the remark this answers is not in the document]
seakayone: I could be able to verify your guess once you provide me with a minimal example for our shacl validator.

seakayone (changes requested):

src/dsp_tools/commands/validate_data/models/api_responses.py
Nora-Olivia-Ammann: I agree, my intention was to put it at the only place where it is needed in this milestone, and look at the correct serialisation in the last milestone and analyse it in detail. But if you prefer to take a look now I am fine with that too.

src/dsp_tools/commands/validate_data/sparql/value_shacl.py
```diff
@@ -208,28 +220,43 @@ def _construct_list_shapes(onto: Graph, project_lists: AllProjectLists) -> Graph
...
+        list_iri=list_iri,
+        sh_path=API_SHAPES.listNodeAsString,
+        sh_message=f"Unknown list node for list: {one_list.list_name}.",
+        sh_in=one_list.nodes,
+    )
+    g += _construct_one_list_property_shape_with_collection(node_prop_info)
+    return g
+
+
+def _construct_one_list_property_shape_with_collection(shacl_info: SHACLListInfo) -> Graph:
```
Nora-Olivia-Ammann, line 244: With hindsight, this is the most obvious solution.

seakayone (approved):

PR #1257, opened 2024-11-05
docs(xmllib): add minimal docstrings (DEV-4297)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/resource.py
```diff
@@ -124,6 +163,28 @@ def add_bool_optional(
...
+            prop_name: name of the property
+            value: value to add
+            permissions: optional permissions of this value
+            comment: optional comment
+
+        Warnings:
+            If the value is not amongst the accepted formats.
+
+        Returns:
+            The original resource that this method has been called on, with the added value
```
Nora-Olivia-Ammann, line 186: This is not correct, only when it is non-empty @jnussbaum.

jnussbaum (approved):

PR #1259, opened 2024-11-07
docs(xmllib): make args consistent (DEV-4314)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -224,7 +224,7 @@ def add_comment(self, comment: str) -> RegionResource:
         Add a comment to the resource
 
         Args:
-            comment: Comment text
+            comment: text
```
Nora-Olivia-Ammann, line 227: This way it is consistent everywhere

mkdocs.yml
```diff
@@ -46,6 +46,7 @@ nav:
           - AudioSegmentResource: xmllib-api-reference/dsp-base-resources/audio-segment-resource.md
       - Configuration options: xmllib-api-reference/config-options.md
       - Value Checkers: xmllib-api-reference/value-checkers.md
+      - Value Converters: xmllib-api-reference/value-converters.md
```
Nora-Olivia-Ammann, line 49: I think it is helpful to have these here too, so they can see the source-code of the converters we apply if they want.

jnussbaum: Well, if it is really only about that, then they can look at the source code themselves. Adding this to the documentation conveys the message: "This is important for you, you can use these". If you don't want to convey this message, then I wouldn't add the converters to the docs.

src/dsp_tools/xmllib/value_converters.py
```diff
@@ -48,13 +47,15 @@ def replace_newlines_with_tags(text: str, converter_option: NewlineReplacement)
 
 def replace_newlines_with_paragraph_tags(text: str) -> str:
     """
-    Replace `Start\\nEnd` with `<p>Start</p><p>End</p>`
+    Replaces one newline `Start\\nEnd` with `<p>Start</p><p>End</p>`
+
+    Replaces multiple newlines `Start\\n\\nEnd` with `<p>Start</p><p>End</p>`
```
jnussbaum, line 52: https://peps.python.org/pep-0257/#one-line-docstrings:

> The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.

jnussbaum, line 52: For me, this is ambiguous. I think this would be clearer:

```suggestion
    Replace `Start\\nEnd` with `<p>Start</p><p>End</p>`
    Multiple consecutive newlines will be treated as one newline:
    `Start\\nMidde\\n\\nEnd` becomes `<p>Start</p><p>Middle</p><p>End</p>`
```

src/dsp_tools/xmllib/value_converters.py
```diff
@@ -63,12 +64,14 @@ def replace_newlines_with_paragraph_tags(text: str) -> str:
 
 def replace_newlines_with_br_tags(text: str) -> str:
     """
-    Replaces `Start\\nEnd` with `Start<br/>End`
+    Replaces one newline `Start\\nEnd` with `Start<br/>End`
+
+    Replaces multiple newlines `Start\\n\\nEnd` with `Start<br/><br/>End`
```
jnussbaum, line 69:
```suggestion
    Replace `Start\\nEnd` with `Start<br/>End`
    Multiple consecutive newlines will be replaced all:
    `Start\\nMidde\\n\\nEnd` becomes `Start<br/>Midde<br/><br/>End`
```

jnussbaum (approved):
LGTM

mkdocs.yml
Nora-Olivia-Ammann: Let's leave it for now and if they (or noemi) does not like it we can always remove it easily.

PR #1261, opened 2024-11-07
feat(xmllib): add helpers from `excel2xml` to `xmllib` (DEV-4316)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/helpers.py
```diff
@@ -56,3 +64,292 @@ def _name_label_mapper_iterator(
...
+        endyear = startyear // 10 * 10 + endyear
+    elif endyear // 100 == 0:
+        # endyear is only 2-digit: add the first 1-2 digits of startyear
+        endyear = startyear // 100 * 100 + endyear
+    if endyear <= startyear:
+        raise ValueError
+    return f"GREGORIAN:CE:{startyear}:CE:{endyear}"
+
+
+def make_xsd_compatible_id(input_value: str | float | int | bool) -> str:
```
Nora-Olivia-Ammann, line 304: I think it does not hurt to also accept other data types as long as they are not empty

src/dsp_tools/xmllib/__init__.py
```diff
@@ -1,4 +1,8 @@
-from .helpers import create_label_to_name_list_node_mapping as create_label_to_name_list_node_mapping
+from .helpers import create_label_to_name_list_node_mapping
```
Nora-Olivia-Ammann, line 1: is there a reason why you imported everything with an alias?

src/dsp_tools/xmllib/helpers.py
```diff
@@ -56,3 +64,342 @@ def _name_label_mapper_iterator(
         if "name" in node:
             yield node["labels"][language_of_label], node["name"]
             # the actual values of the name and the label
+
+
+def escape_reserved_xml_characters(text: str) -> str:
```
Nora-Olivia-Ammann, line 69: Noemi said this could be helpful for her

src/dsp_tools/xmllib/helpers.py
```diff
@@ -56,3 +64,342 @@ def _name_label_mapper_iterator(
...
+        raise ValueError
+    return f"GREGORIAN:CE:{startyear}:CE:{endyear}"
+
+
+def make_xsd_compatible_id(input_value: str | float | int | bool) -> str:
+    """
+    An xsd:ID may not contain all types of special characters,
+    additionally, it must start with a letter or underscore, if the provided string does not start with either,
+    an underscore is added.
+    This function replaces illegal characters with "_".
```
jnussbaum, line 359:
```suggestion
    An xsd:ID may not contain all types of special characters,
    and it must start with a letter or underscore.
    Replace illegal characters with "_", and prepend a leading "_" if necessary.
```

src/dsp_tools/xmllib/helpers.py
```diff
@@ -56,3 +64,342 @@ def _name_label_mapper_iterator(
...
+    return res
+
+
+def make_xsd_compatible_id_with_uuid(input_value: str | float | int | bool) -> str:
+    """
+    An xsd:ID may not contain all types of special characters,
+    additionally, it must start with a letter or underscore, if the provided string does not start with either,
+    an underscore is added.
+    This function replaces illegal characters with "_".
+    The UUID will be different each time the function is called.
```
jnussbaum, line 388:
```suggestion
    An xsd:ID may not contain all types of special characters,
    and it must start with a letter or underscore.
    Replace illegal characters with "_", and prepend a leading "_" if necessary.
    Additionally, add a UUID at the end.
    The UUID will be different each time the function is called.
```

src/dsp_tools/xmllib/__init__.py
jnussbaum: I thought aliasing would be necessary for the linters, but if it works without aliases, I don't care. You can remove the aliases everywhere, to be consistent.

src/dsp_tools/xmllib/helpers.py
```diff
@@ -56,3 +64,292 @@ def _name_label_mapper_iterator(
         if "name" in node:
             yield node["labels"][language_of_label], node["name"]
             # the actual values of the name and the label
+
+
+def find_date_in_string(string: str) -> str | None:
```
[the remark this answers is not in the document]
jnussbaum: It's fine to just take it over

src/dsp_tools/xmllib/helpers.py
jnussbaum: While I can see a potential use case for integers and floats, the boolean probably doesn't make sense.

jnussbaum (approved):

PR #1264, opened 2024-11-11
feat(validate-data): check file value cardinality (DEV-4338)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/make_data_rdf.py
```diff
@@ -197,3 +202,10 @@ def _transform_uri_value(val: ValueDeserialised, res_iri: URIRef) -> ValueRDF:
         else Literal("", datatype=XSD.string)
     )
     return UriValueRDF(URIRef(val.prop_name), content, res_iri)
+
+
+def _transform_file_value(val: AbstractFileValueDeserialised) -> AbstractFileValueRDF:
```
Nora-Olivia-Ammann, line 208: This function will be extended through the coming PRs.
Currently we only check the cardinalities, later on we identify which file type is given through the extension.
This will be implemented separately. I will likely keep `GenericFileValueRDF` for extensions we do not know.

src/dsp_tools/commands/validate_data/models/data_rdf.py
```diff
@@ -212,3 +212,25 @@ def make_graph(self) -> Graph:
         g.add((val_iri, KNORA_API.uriValueAsUri, self.object_value))
         g.add((self.res_iri, self.prop_name, val_iri))
         return g
+
+
+@dataclass
+class AbstractFileValueRDF(RDFTriples):
```
Nora-Olivia-Ammann, line 218: This will be sub-classed in the future with all the correct Value types.

src/dsp_tools/commands/validate_data/reformat_validaton_result.py
```diff
@@ -361,6 +356,24 @@ def _reformat_one_validation_result(validation_result: ValidationResult) -> Inpu
             raise BaseError(f"An unknown violation result was found: {validation_result.__class__.__name__}")
 
 
+def _reformat_min_cardinality_validation_result(validation_result: ResultMinCardinalityViolation) -> InputProblem:
+    iris = _reformat_main_iris(validation_result)
+    file_value_properties = ["shapes:hasGenericFileValue"]
```
Nora-Olivia-Ammann, line 361: I will change the identification of the problem to looking at the source shape. Because when a wrong file value is used, we will get two violations, one min card and one closed shapes.
But I may refactor this separately, I will have to look at how big the diff is. But for now this works.

src/dsp_tools/commands/validate_data/sparql/file_value_shacl.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Returns:
+        Graph with file cardinalities
+    """
+    g = Graph()
+    g += _construct_generic_file_value_cardinality(onto)
+    return g
+
+
+def _construct_generic_file_value_cardinality(onto: Graph) -> Graph:
```
Nora-Olivia-Ammann, line 22: This function will be deleted after we implement all the value types.

test/unittests/commands/validate_data/sparql/test_cardinality_shacl.py
```diff
@@ -19,7 +20,7 @@
 class TestCheckTripleNumbersOnto:
     def test_nodeshape(self, onto_graph: Graph) -> None:
         result = _construct_resource_nodeshape(onto_graph)
-        number_of_resource_classes = 6
+        number_of_resource_classes = 12
```
Nora-Olivia-Ammann, line 23: I updated the `onto.ttl` file, which is the cause of this.
I haven't done this after including a second onto. Which is why that is also added now.

test/unittests/commands/validate_data/test_make_data_rdf.py
```diff
@@ -264,5 +269,31 @@ def test_none(self, uri_value_deserialised_none: UriValueDeserialised) -> None:
...
+    assert _transform_file_value(value).value == Literal(expected)
+
+
+@pytest.mark.parametrize(
+    ("value", "expected"),
+    [
+        ("test.jpg", "jpg"),
+        ("test.JPG", "jpg"),
+        (None, "No file path was given."),
+        ("test", "This file is missing a valid extension, actual value: test"),
```
Nora-Olivia-Ammann, line 291: This ensures that even if a valid extension without a file is entered it will fail the validation.

testdata/validate-data/generic/project.json
```diff
@@ -505,21 +505,46 @@
...
+            "name": "TestMovingImageRepresentation",
+            "super": "MovingImageRepresentation",
+            "labels": {
+              "en": "MovingImageRepresentation"
+            }
+          },
+          {
+            "name": "TestStillImageRepresentation",
             "super": "StillImageRepresentation",
-            "cardinalities": [
```
Nora-Olivia-Ammann, line 513: I want to do seqnum and isPartOf separately

src/dsp_tools/commands/validate_data/sparql/file_value_shacl.py
```diff
@@ -0,0 +1,56 @@
...
+        ?class a owl:Class ;
+               rdfs:subClassOf ?superClass .
+        VALUES ?superClass {
+            knora-api:ArchiveRepresentation
+            knora-api:AudioRepresentation
+            knora-api:DocumentRepresentation
+            knora-api:MovingImageRepresentation
+            knora-api:StillImageRepresentation
+            knora-api:TextRepresentation
+        }
```
seakayone, line 51: The knora-base ontology also defines `DDDRepresentation` with `hasDDDFileValue`,
This represents a file containing 3D data.

Think we are missing this here.

seakayone (changes requested):
Looks good to me, except for the missing DDDRepresentation.

Nora-Olivia-Ammann:
> Looks good to me, except for the missing DDDRepresentation.

Yes it is missing, but I did not add it because it is not implemented in dsp-tools. So currently it is not possible to create one. 
I also did not know what extension it would have because it is not listed in the API docs. https://docs.dasch.swiss/latest/DSP-API/01-introduction/file-formats/

Should I add it anyway?

seakayone:
> Yes it is missing, but I did not add it because it is not implemented in dsp-tools. 
> 
> Should I add it anyway?

I guess in this case we can skip it.

PR #1270, opened 2024-11-13
fix(xmllib): resolve typing issues (RDU-57)

jnussbaum:


src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -960,7 +961,7 @@ def add_relates_to(self, relates_to: str) -> AudioSegmentResource:
         self.relates_to.append(relates_to)
         return self
 
-    def add_relates_to_multiple(self, relates_to: list[str]) -> AudioSegmentResource:
+    def add_relates_to_multiple(self, relates_to: Sequence[str]) -> AudioSegmentResource:
```
Nora-Olivia-Ammann, line 964: Looks good, you can now delete the line 1106 & 5.

src/dsp_tools/xmllib/__init__.py
```diff
@@ -1,30 +1,30 @@
...
-from .value_checkers import is_geoname
-from .value_checkers import is_integer
-from .value_checkers import is_nonempty_value
-from .value_checkers import is_string_like
-from .value_checkers import is_timestamp
-from .value_converters import convert_to_bool_string
-from .value_converters import replace_newlines_with_br_tags
-from .value_converters import replace_newlines_with_paragraph_tags
-from .value_converters import replace_newlines_with_tags
+from .helpers import create_label_to_name_list_node_mapping as create_label_to_name_list_node_mapping
```
Nora-Olivia-Ammann, line 1: Good to know that is it needed after all.

Nora-Olivia-Ammann (approved):

src/dsp_tools/xmllib/models/dsp_base_resources.py
jnussbaum: I'm not sure. AFAIU, `_transform_unexpected_input()` is now completely obsolete everywhere, if we assume that our users exclusively use our public API. I mean, if users only use the constructor and the public methods, there is no code path where there could be a wrong data type in one of these fields.

But as soon as someone manually modifies the class fields, there might be bad surprises. So we could also leave `_transform_unexpected_input()` in place, as an additional layer of security. WDYT?

jnussbaum:
@Nora-Olivia-Ammann I had to replace typing.Sequence by collections.abc.Collection (https://github.com/dasch-swiss/dsp-tools/pull/1270/commits/13b791ce6827035baf12e3b589c5f599e82789e0), because Sequence doesn't fit our needs, e.g. `isinstance(set(), Sequence) == False`.

src/dsp_tools/xmllib/models/dsp_base_resources.py
Nora-Olivia-Ammann: Sounds good, it was meant to make it more stable and if you hand in a string iteration will not raise an error...

PR #1271, opened 2024-11-14
refactor(xmlupload): serialise color prop in class

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -147,6 +161,7 @@ def make_values(p: XMLProperty) -> list[dict[str, Any]]:
                     bool_graph = _make_boolean_prop(prop, res_bnode, bool_prop_name, self.permissions_lookup)
                     properties_graph += bool_graph
                     last_prop_name = bool_prop_name
+                # serialised as dict
```
Nora-Olivia-Ammann, line 164: After we finish, this next section with `def _make_value` will be gone

src/dsp_tools/commands/xmlupload/models/serialise/serialise_value.py
```diff
@@ -25,27 +25,45 @@ class SerialiseValue(Protocol):
     def serialise(self) -> dict[str, Any]:
         """Serialise the value."""
 
+    def _get_optionals(self) -> dict[str, str]:
+        optionals = {"knora-api:valueHasComment": self.comment} if self.comment else {}
+        if self.permissions:
+            optionals["knora-api:hasPermissions"] = self.permissions
+        return optionals
+
```
jnussbaum, line 33: Here you mix up composition (protocols) and inheritance (abstract base classes). 

If you prefer composition (protocols), then you shouldn't implement anything on the protocol level. All implementation must go into the child classes. Protocols are pure structure (skeletons), no flesh.

If you prefer inheritance (abstract base classes), then you can leave your code as it is, but annotate it differently:
- SerialiseValue becomes an ABC (https://docs.python.org/3/library/abc.html#abc.ABC)
- SerialiseValue.serialise() should be decorated as abstract method (https://docs.python.org/3/library/abc.html#abc.abstractmethod). By doing so, you enforce child classes to implement it.

In your case, I would go for ABCs, because it's tiresome to repeat the implementation of _get_optionals() all over the place.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -452,28 +458,37 @@ def _make_time_value(value: XMLValue) -> dict[str, Any]:
...
-            SerialiseURI(
-                value=uri,
-                permissions=permission,
-                comment=v.comment,
-            )
-        )
+def _transform_into_serialise_property_with_string_value(
+    prop: XMLProperty,
+    permissions_lookup: dict[str, Permissions],
+    func: Callable[[str, str | None, str | None], SerialiseValue],
```
jnussbaum, line 464: This is not a function that you call, but a constructor of a class. I think "serialiser" is a good name.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -133,10 +135,22 @@ def make_values(p: XMLProperty) -> list[dict[str, Any]]:
...
+            "color": SerialiseColor,
+        }
+
         for prop in resource.properties:
             match prop.valtype:
-                case "uri":
-                    properties_serialised.update(_serialise_uri_prop(prop, self.permissions_lookup))
+                # serialised as dict
+                case "uri" | "color" as val_type:
+                    transformed_prop = _transform_into_serialise_property_with_string_value(
```
jnussbaum, line 147: The name _transform_into_serialise_property_with_string_value() is a bit complicated. What about `_serialise_prop()`?

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -452,28 +458,37 @@ def _make_time_value(value: XMLValue) -> dict[str, Any]:
...
-            )
-        )
+def _transform_into_serialise_property_with_string_value(
+    prop: XMLProperty,
+    permissions_lookup: dict[str, Permissions],
+    func: Callable[[str, str | None, str | None], SerialiseValue],
+) -> SerialiseProperty:
+    serialised_values: list[SerialiseValue] = [
+        _transform_into_serialise_value_with_string_value(v, permissions_lookup, func) for v in prop.values
+    ]
```
jnussbaum, line 468: 1. The name is a bit complicated. What about `_serialise_val()`?
2. The type hint is obsolete

```suggestion
    serialised_values = [_serialise_val(v, permissions_lookup, serialiser) for v in prop.values]
```

Nora-Olivia-Ammann: Regarding the name same as above.

src/dsp_tools/commands/xmlupload/resource_create_client.py
jnussbaum: You could use the nice name `_serialise_prop()` for now. And if later you need other functions for lists etc., you can name them accordingly, e.g. `_serialise_prop_special_case_1()`, `_serialise_prop_special_case_2()`, ...

PR #1272, opened 2024-11-14
refactor(xmlupload): serialise values in classes

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -136,20 +142,39 @@ def make_values(p: XMLProperty) -> list[dict[str, Any]]:
...
                 # serialised as dict
-                case "uri" | "color" as val_type:
+                case "uri" | "color" | "geoname" | "time" as val_type:
                     transformed_prop = _transform_into_serialise_prop(
                         prop=prop,
                         permissions_lookup=self.permissions_lookup,
                         seraliser=str_value_to_serialiser_mapper[val_type],
                     )
                     properties_serialised.update(transformed_prop.serialise())
+                case "decimal":
```
jnussbaum, line 161: I don't understand why it is necessary to have these special cases. Why can't they be handled like the others? As far as I see, there are some custom checks, for example for the decimal the lines 323/324: https://github.com/dasch-swiss/dsp-tools/blob/2ffbcc0f1315397732a6e8200204d89cae6bf7d2/src/dsp_tools/commands/xmlupload/resource_create_client.py#L323-L324

But if you move these special checks into the classes (e.g. the `SerialiseDecimal` class), then the code would become much easier and more readable.

Or do I miss something?

jnussbaum (commented):
See my comments

src/dsp_tools/commands/xmlupload/resource_create_client.py
Nora-Olivia-Ammann: I don't want the serialisation classes to have any logic or transformations. All they need to know is what and how to give the information to the API. Any kind of transformations (like with the permissions) etc. should happen outside. Another advantage if we move it into the transformation is, that once we move it out of the resource create client, the program would crash before even one resource was created in case of invalid input.
Christian said something that was really helpful to me, repeat yourself three times and then refactor. Because then you will see where the commonalities and the differences lie. This way you don't start to optimise something without knowing all the cases. This paradigm has been really helpful to me.
TLDR: I would leave it like that and if we see after we did all the transformations that there is a way to optimise it, we can do it then.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -443,7 +443,7 @@ def _make_list_value(value: XMLValue, iri_lookup: dict[str, str]) -> dict[str, A
 def _transform_text_prop(
     prop: XMLProperty, permissions_lookup: dict[str, Permissions], iri_resolver: IriResolver
 ) -> SerialiseProperty:
-    values: list[SerialiseValue] = []
+    values = []
```
jnussbaum, line 446: This is another case, here it is an advantage to have a type hint. By default, `[]` is treated as `list[Any]`, so it is better to type it as `values: list[SerialiseValue] = []`

src/dsp_tools/commands/xmlupload/resource_create_client.py
jnussbaum: > I don't want the serialisation classes to have any logic or transformations
That's a valid argument.

> the program would crash before even one resource was created
Also a good argument

> repeat yourself three times and then refactor
Well, that would be now, wouldn't it? 😅 

Here a suggestion how it could be done:

```
def _do_special_checks(value: XMLValue, serialiser: Callable[[str, str | None, str | None], SerialiseValue]) -> None:
    match serialiser:
        case SerialiseDecimal():
            do_special_checks_for_decimal(value)
        case SerialiseGeometry():
            do_special_checks_for_geometry(value)
        case _:
            pass


def _transform_into_serialise_value(
    value: XMLValue,
    serialiser: Callable[[str, str | None, str | None], SerialiseValue],
) -> SerialiseValue:
    _do_special_checks(value, serialiser)
    permission_str = _get_permission_str(value.permissions, permissions_lookup)
    value_str = cast(str, value.value)
    return serialiser(value_str, permission_str, value.comment)
```

PR #1276, opened 2024-11-15
refactor(validate-data): make abstract methods explicit

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/models/input_problems.py
```diff
@@ -146,12 +147,15 @@ class InputProblem(ABC):
     prop_name: str
 
     @property
+    @abstractmethod
```
Nora-Olivia-Ammann, line 150: If you use the decorator you force a child class to overwrite the function, which is what we want.

src/dsp_tools/commands/validate_data/models/data_deserialised.py
```diff
@@ -64,7 +64,7 @@ class AudioSegmentDeserialised(AbstractResource): ...
 
 
 @dataclass
-class ValueDeserialised:
+class ValueDeserialised(ABC):
```
jnussbaum, line 67: Why is this an ABC? It has no abstract methods that must be overridden in child classes

src/dsp_tools/commands/validate_data/models/validation.py
```diff
@@ -74,7 +75,7 @@ class DetailBaseInfo:
 
 
 @dataclass
-class ValidationResult:
+class ValidationResult(ABC):
```
jnussbaum, line 78: Why is this an ABC? It has no abstract methods that must be overridden in child classes

jnussbaum (approved):

PR #1278, opened 2024-11-15
refactor(xmlupload): serialise resource

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -103,32 +104,25 @@ def _make_resource(
         resource: XMLResource,
         bitstream_information: BitstreamInfo | None,
     ) -> dict[str, Any]:
-        resource_iri = resource.iri
+        migration_metadata = None
+        res_iri = resource.iri
         if resource.ark:
-            resource_iri = convert_ark_v0_to_resource_iri(resource.ark)
-        context = get_json_ld_context_for_project(self.project_onto_dict)
```
Nora-Olivia-Ammann, line 109: This is the same every time, so it should be moved out.

src/dsp_tools/commands/xmlupload/models/namespace_context.py
```diff
@@ -1,7 +1,28 @@
...
+
+def get_json_ld_context_for_project(ontos: dict[str, str]) -> JSONLDContext:
+    """Provided a dictionary of ontology names and IRIs, returns a JSON-LD context for the project."""
+    context = _get_default_json_ld_context()
+    project_context = _correct_project_context_namespaces(ontos)
+    context.update(project_context)
+    return JSONLDContext(context)
+
+
+def _get_default_json_ld_context() -> dict[str, str]:
```
Nora-Olivia-Ammann, line 25: I am keeping this out of the `JSONLDContext` so we don't have to make a dictionary update, everytime we serialise it.

src/dsp_tools/commands/xmlupload/models/serialise/serialise_resource.py
```diff
@@ -0,0 +1,46 @@
...
+            "@type": self.res_type,
+            "knora-api:attachedToProject": {"@id": self.project_iri},
+        }
+        if self.permissions:
+            serialised["knora-api:hasPermissions"] = self.permissions
+        return serialised
+
+
+@dataclass(frozen=True)
+class MigrationMetadata:
```
jnussbaum, line 35: A class with this name exists already. This class here is for serialisation, so I suggest to add the same prefix than the other classes in this file.

```suggestion
class SerialiseMigrationMetadata:
```

src/dsp_tools/commands/xmlupload/models/serialise/serialise_resource.py
```diff
@@ -0,0 +1,46 @@
...
+    iri: str
+    creation_date: DateTimeStamp | None
+
+    def to_dict(self) -> dict[str, Any]:
```
jnussbaum, line 39: normally, you call these methods `serialise()`, so I would do the same here.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -103,35 +104,28 @@ def _make_resource(
...
-                raise PermissionNotExistsError(
-                    f"Could not find permissions for resource {resource.res_id} with permissions {resource.permissions}"
-                )
-        if resource.creation_date:
-            res["knora-api:creationDate"] = {
-                "@type": "xsd:dateTimeStamp",
-                "@value": str(resource.creation_date),
-            }
+            res_iri = convert_ark_v0_to_resource_iri(resource.ark)
+        if res_iri:
```
jnussbaum, line 111: Why is this dependent on the res_iri? It should rather be dependent on whether a creation date is existent or not:

```suggestion
        if resource.creation_date:
```

jnussbaum (approved):
Looks good. In 1 place, I don't understand the logic, so I made a comment.

PR #1280, opened 2024-11-15
refactor(xmlupload): make value transformers

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/models/value_transformers.py
```diff
@@ -0,0 +1,57 @@
...
+            case FormattedTextValue() as xml:
+                raise BaseError(f"Expected string value, but got XML value: {xml.as_xml()}")
+            case _:
+                assert_never(value)
+
+
+class DateTransformer(ValueTransformer):
+    def transform(self, input_value: Any) -> Date:
+        val = self._assert_is_string(input_value)
+        return parse_date_string(val)
```
Nora-Olivia-Ammann, line 40: This function is now only used here and in tests, perhaps we can in the future move it into this, but for the moment it works and keeps everything minimal.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -444,7 +427,14 @@ def _transform_text_prop(
     for val in prop.values:
         match val.value:
             case str():
-                values.append(_transform_into_serialise_value(val, permissions_lookup, SerialiseSimpletext))
+                values.append(
```
Nora-Olivia-Ammann, line 430: I prefer to keep the text value separate, because not only do we transform, but there are also two different serialiser routes. I think this will make it too complicated to do nicely. But we can always change it in the future...

src/dsp_tools/commands/xmlupload/models/value_transformers.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class DateTransformer(ValueTransformer):
+    def transform(self, input_value: InputTypes) -> Date:
+        val = self._assert_is_string(input_value)
+        return parse_date_string(val)
+
+
+class DecimalTransformer(ValueTransformer):
+    def transform(self, input_value: InputTypes) -> str:
```
jnussbaum, line 43: I like the general idea, but not the way how you put it into practice. These classes are purely statical, because they don't hold any information. In their current state, it doesn't make sense to have several objects (instantiations) from them. You can use the function directly with `DecimalTransformer.transform(val)`, without needing an actual `DecimalTransformer` object.

In other words, the only thing that the `DecimalTransformer` class does is encapsulating a method. But the encapsulation is effectively useless. It would make more sense if the method would be a free-standing function.

To demonstrate it, I have prepared this PR: https://github.com/dasch-swiss/dsp-tools/pull/1281. Feel free to modify it, merge it, or just let you inspire by it.

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -151,30 +155,25 @@ def make_values(p: XMLProperty) -> list[dict[str, Any]]:
...
-            "geoname": SerialiseGeoname,
-            "time": SerialiseTime,
-            "uri": SerialiseURI,
+            "color": (SerialiseColor, StringTransformer()),
+            "date": (SerialiseDate, DateTransformer()),
+            "decimal": (SerialiseDecimal, DecimalTransformer()),
+            "geometry": (SerialiseGeometry, GeometryTransformer()),
+            "geoname": (SerialiseGeoname, StringTransformer()),
+            "time": (SerialiseTime, StringTransformer()),
+            "uri": (SerialiseURI, StringTransformer()),
```
jnussbaum, line 164: There are 2 things which I think are not optimal:

1. Every time that make_values() is called, this dictionary is built again. It would make more sense if it lives on the top level, or even better in the `value_transformers.py`. (See my PR https://github.com/dasch-swiss/dsp-tools/pull/1281)
2. The tuple `(SerialiseURI, StringTransformer())` is untyped, so you have to do a "blind" unpacking operation like `serialiser, transformer = str_value_to_serialiser_mapper[val_type]`. Alternatively, you could use a DTO (data transfer object), which encapsulates this info. For example like this:

```
@dataclass
class TransformationSteps:
    serialiser: ValueSerialiser
    transformer: ValueTransformer
```

Then, you could annotate this dict with `dict[str, TransformationSteps]`. But I admit this this perhaps an overkill, you can also leave it as it is.

jnussbaum (commented):
I very much like the general idea of this PR! That's definitively a step in the right direction, and you should absolutely merge this PR!

But on the technical level, I see some difficulties. Especially, I think that the ABC Value Transformer with its child classes is an overcomplification of something that could more easily be done with functions. (see my comments)

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -145,31 +139,15 @@ def make_values(p: XMLProperty) -> list[dict[str, Any]]:
...
-            "uri": SerialiseURI,
-        }
-
         for prop in resource.properties:
             match prop.valtype:
                 # serialised as dict
-                case "uri" | "color" | "geoname" | "time" as val_type:
-                    transformed_prop = _transform_into_serialise_prop(
+                case "uri" | "color" | "geoname" | "time" | "decimal" | "geometry" | "date" as val_type:
+                    transformations = value_to_serialiser_mapper[val_type]
```
jnussbaum, line 146: so now it would rather be a `value_to_transformations_mapper`

jnussbaum (approved):

PR #1281, opened 2024-11-16
chore: suggest functions instead of classes

jnussbaum:


src/dsp_tools/commands/xmlupload/models/value_transformers.py
```diff
@@ -1,56 +1,74 @@
...
+    "date": TransformationSteps(SerialiseDate, transform_date),
+    "decimal": TransformationSteps(SerialiseDecimal, transform_decimal),
+    "geometry": TransformationSteps(SerialiseGeometry, transform_geometry),
+    "geoname": TransformationSteps(SerialiseGeoname, transform_string),
+    "time": TransformationSteps(SerialiseTime, transform_string),
+    "uri": TransformationSteps(SerialiseURI, transform_string),
+}
+
+
+def assert_is_string(value: str | FormattedTextValue) -> str:
```
jnussbaum, line 67: I put it here because of a circular import

src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -154,26 +145,16 @@ def make_values(p: XMLProperty) -> list[dict[str, Any]]:
...
 
-        str_value_to_serialiser_mapper = {
-            "color": (SerialiseColor, StringTransformer()),
-            "date": (SerialiseDate, DateTransformer()),
-            "decimal": (SerialiseDecimal, DecimalTransformer()),
-            "geometry": (SerialiseGeometry, GeometryTransformer()),
-            "geoname": (SerialiseGeoname, StringTransformer()),
-            "time": (SerialiseTime, StringTransformer()),
-            "uri": (SerialiseURI, StringTransformer()),
-        }
```
jnussbaum, line 165: I moved this away, because it is not dependent of its encapsulating function. It always stays the same, so you don't have to build it again and again.

PR #1282, opened 2024-11-17
feat(xmllib): check tags of richtext values (RDU-58)

jnussbaum:


src/dsp_tools/commands/excel2json/models/input_error.py
```diff
@@ -11,15 +12,6 @@
...
 
-class Problem(Protocol):
-    """Information about input errors."""
-
-    def execute_error_protocol(self) -> str:
-        """
-        This function initiates all the steps for successful problem communication with the user.
-        """
-
-
```
jnussbaum, line 22: I moved this to dsp_tools.models.problems, because it doesn't make sense to have it duplicated in different locations

src/dsp_tools/xmllib/models/values.py
```diff
@@ -390,6 +394,62 @@ def make_element(self) -> etree._Element:
...
+        # transform named entities (=character references) to numeric entities, e.g. &nbsp; -> &#160;
+        num_ent = numeric_entities(escaped_text)
+        pseudo_xml = f"<ignore-this>{num_ent}</ignore-this>"
+        try:
+            _ = etree.fromstring(pseudo_xml)
+        except etree.XMLSyntaxError as err:
+            prob = IllegalTagProblem(orig_err_msg=err.msg, pseudo_xml=pseudo_xml)
+            warnings.warn(DspToolsUserWarning(prob.execute_error_protocol()))
+
+    def _escape_reserved_chars(self) -> str:
```
Nora-Olivia-Ammann, line 408: I would put this in the value transformers file. If you don't want, we don't have to export it, therefore I assume it would not appear on the docs... In any case I dislike functions this large in a models file.

Nora-Olivia-Ammann (approved):

src/dsp_tools/xmllib/models/values.py
jnussbaum: You're right, they better fit into `value_checkers.py`. And since all functions inside `value_checkers.py` are served on docs.dasch.swiss, it's consistent if this one here is also served.

PR #1283, opened 2024-11-18
refactor(xmlupload): move create resource out of `resource_create_client`

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/transform_resource_and_values_for_upload.py
Nora-Olivia-Ammann, line 1: The code itself did not change much.
The only real difference is that some were class methods that referred to a `self` parameter, these were not moved into the `lookups` which I had to hand around. Other than that no changes were made.

PR #1284, opened 2024-11-19
RFC: experiments in serialisation

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/resource_and_value_transformer.py
Nora-Olivia-Ammann, line 1: This would take place before the upload has started. It would be like our fail fast.

src/dsp_tools/commands/xmlupload/make_graphs.py
Nora-Olivia-Ammann, line 1: This transformation happens just before one resource is uploaded. The IRIs for richtext and link values are resolved here.

seakayone (commented):
Tests are failing, please rerequest a review once this is ready. Thanks.

Nora-Olivia-Ammann:
> Tests are failing, please rerequest a review once this is ready. Thanks.

This is just to discuss the concept, it is not integrated into the `xmlupload` nor is it my intention to do so.

jnussbaum (commented):
Thank you for preparing this code example. It helped me to understand your vision. I think that this makes sense, and I like it.

PR #1285, opened 2024-11-19
RFC xml copyright license experiment

Nora-Olivia-Ammann:


testdata/xml-data/license_copy_in_attrib.xml
Nora-Olivia-Ammann, line 1: The advantage if they have to enter the actual string that is displayed is that they don't have to keep track of a mapping and we do not need to resolve it in a step where we have the possibility that they used a reference that does not exist. However this can also be checked easily with SHACL so in the long term that is not a lot of work for us.

testdata/xml-data/license_copy_as_tag.xml
Nora-Olivia-Ammann, line 1: This is the option I like least.

testdata/xml-data/license_copy_as_tag_with_reference.xml
Nora-Olivia-Ammann, line 1: I like this option least. When we have a reference we may as well put is as an attrib.

testdata/xml-data/license_copy_as_tag_with_text.xml
Nora-Olivia-Ammann, line 1: If we do not want to make a reference then I think this is the better option that writing it into the attribs.

copyright_license/examples/license_copy_in_attrib.xml
Nora-Olivia-Ammann, line 1: I would vote against this one, because if anyone modifies the xml manually for unknown reasons, then we get an ugly result.

copyright_license/examples/license_copy_with_reference.xml
Nora-Olivia-Ammann, line 1: This is personally my preferred version.

jnussbaum: Me too, I prefer this one. Thanks for the prototype, this helps to get a clear understanding.

PR #1297, opened 2024-11-26
fix(xmllib): don't crash if comment input is `nan` (DEV-4399)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/values.py
```diff
@@ -386,10 +386,10 @@ def make_element(self) -> etree._Element:
         attribs = {"encoding": "xml"}
         if self.permissions != Permissions.PROJECT_SPECIFIC_PERMISSIONS:
             attribs["permissions"] = self.permissions.value
-        if self.comment:
-            attribs["comment"] = self.comment
+        if is_string_like(self.comment):
+            attribs["comment"] = str(self.comment)
         ele = etree.Element(f"{DASCH_SCHEMA}text", attrib=attribs, nsmap=XML_NAMESPACE_MAP)
-        ele.text = self.value
+        ele.text = str(self.value)
```
Nora-Olivia-Ammann, line 392: Since we only warn if the input is invalid it is possible to have values here that would cause a crash during the serialisation. It would also be possible in the `__post_init__`, but I thought I would not change too much... What do you think is better?
If we decide to do it in the `__post_init__` then I would check the comment there also and replace it with `None` if it is empty.

src/dsp_tools/xmllib/models/file_values.py
```diff
@@ -38,11 +38,13 @@ def __post_init__(self) -> None:
...
-        attribs = {}
-        if self.permissions != Permissions.PROJECT_SPECIFIC_PERMISSIONS:
-            attribs["permissions"] = self.permissions.value
-        if self.comment:
-            attribs["comment"] = self.comment
+        attribs = (
+            {"permissions": self.permissions.value}
+            if self.permissions != Permissions.PROJECT_SPECIFIC_PERMISSIONS
+            else {}
+        )
```
jnussbaum, line 45: I think the old syntax is easier to read:

```suggestion
        attribs = {}
        if self.permissions != Permissions.PROJECT_SPECIFIC_PERMISSIONS:
            attribs["permissions"] = self.permissions.value
```

src/dsp_tools/xmllib/models/file_values.py
```diff
@@ -60,11 +62,13 @@ def __post_init__(self) -> None:
...
```
jnussbaum, line 69: I think the old syntax is easier to read:

```suggestion
        attribs = {}
        if self.permissions != Permissions.PROJECT_SPECIFIC_PERMISSIONS:
            attribs["permissions"] = self.permissions.value
```

src/dsp_tools/xmllib/models/values.py
jnussbaum: I think crashing during serialization is not a good thing. It is then difficult for the user to understand where the error stems from. I prefer checking this in the `__post_init__`.

jnussbaum: So I think that `ele.text = str(self.value)` is actually a bad idea, because the user would end up with `<text>nan</text>`, without getting an error.

jnussbaum (changes requested):
Concerning the initial problem (crash if comment input is nan): I'm okay with this, but perhaps you could add a test.

Concerning the `ele.text = str(self.value)`: I think that's a Verschlimmbesserung. (If I got it correct. If not, please correct me).

Nora-Olivia-Ammann:
> Concerning the initial problem (crash if comment input is nan): I'm okay with this, but perhaps you could add a test.
> 
> Concerning the `ele.text = str(self.value)`: I think that's a Verschlimmbesserung. (If I got it correct. If not, please correct me).

Well from a stylistic point of view it is really not a good idea. But that is not helpful for RDU because it is possible to crash. As far as I know we decided that we should not crash the script. If we decide that we do not want to accept certain faulty values, then I would argue that we should crash with our own error.

src/dsp_tools/xmllib/models/values.py
Nora-Olivia-Ammann: They would get an error in the `__post_init__` just not during serialisation.

jnussbaum:
> As far as I know we decided that we should not crash the script. If we decide that we do not want to accept certain faulty values, then I would argue that we should crash with our own error.

If the users get a warning, I would be okay with `ele.text = str(self.value)`. It's not necessary to crash, but at least emit a warning.

jnussbaum (approved):

PR #1300, opened 2024-11-27
feat(xmllib): simplify creation of geometry-prop with classes (RDU-55)

Nora-Olivia-Ammann:


docs/assets/images/img-xml-grid-for-geometry-prop.png
Nora-Olivia-Ammann, line 1: I discussed this with Noemi, she prefers to have the added 1 and remove the salsah display.

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -213,13 +199,97 @@ def create_new(
         return RegionResource(
             res_id=res_id,
             label=label,
-            color=color,
             region_of=region_of,
-            geometry=geometry,
+            geometry=GeometryShape(res_id),
```
Nora-Olivia-Ammann, line 203: Noemi says she prefers everything to have a default value. The line width is the same default that the app uses.
As you see next, all the values can be overwritten.

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -213,13 +199,97 @@ def create_new(
...
+        A point will only be added if **both x and y** are valid floats.
+
+        Args:
+            x: point on the x-axis (width)
+            y: point on the y-axis (height)
+
+        Returns:
+            A region resource with the added point if both are valid floats.
+        """
+        if any([is_decimal(x), is_decimal(y)]):
```
Nora-Olivia-Ammann, line 289: I discussed this with Noemi, we get a warning in any case if a point is not valid, but if one of them is, it can be assumed that the second one is an error in the data, not that it should not be created. So the user should be informed. Also discussed with Noemi, she prefers that these details are not mentioned in the docstring

src/dsp_tools/xmllib/models/geometry.py
```diff
@@ -0,0 +1,102 @@
...
+            if not 0 <= float(self.y) <= 1:
+                info.append(f"y: '{self.y}'")
+            if info:
+                msg = (
+                    f"The geometry points must be larger/equal 0 and smaller/equal 1. "
+                    f"The following points of the resource with the ID '{self.resource_id}' "
+                    f"are not valid: {', '.join(info)}"
+                )
+                warnings.warn(DspToolsUserWarning(msg))
+            self.x = float(self.x)
```
Nora-Olivia-Ammann, line 98: I am "converting it" in case someone entered a float as a string. Because in the json we do not want them with quotation marks.

src/dsp_tools/xmllib/models/resource.py
```diff
@@ -136,8 +136,8 @@ def add_bool(
 
         Conversions:
          - The conversion is case-insensitive, meaning that the words can also be capitalised.
-         - "false", "0", "0.0", "no", "non", "nein" -> "false"
-         - "true", "1", "1.0", "yes", "oui", "ja" -> "true"
+         - `false`, `0`, `0.0`, `no`, `non`, `nein` -> `false`
```
Nora-Olivia-Ammann, line 139: Looks cleaner to me when rendered.

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -143,7 +147,15 @@ def serialise(self) -> etree._Element:
...
-        res_ele.append(_serialise_has_comment(self.comments, self.res_id))
+        if self.comments:
+            res_ele.append(_serialise_has_comment(self.comments, self.res_id))
+        else:
+            warnings.warn(
+                DspToolsUserWarning(
+                    f"The annotation with the ID '{self.res_id}' does not have any comments. "
+                    f"At least one comment must be provided, please note that an xmlupload will fail."
+                )
+            )
```
jnussbaum, line 158:
```suggestion
            msg = (
                f"The annotation with the ID '{self.res_id}' does not have any comments. "
                f"At least one comment must be provided, please note that an xmlupload will fail."
            )
            warnings.warn(DspToolsUserWarning(msg))
```

jnussbaum (approved):

PR #1302, opened 2024-11-28
test(xmllib): test root

Nora-Olivia-Ammann:


test/unittests/xmllib/models/test_root.py
```diff
@@ -0,0 +1,39 @@
...
+    # add_resource
+    anno_res = AnnotationResource("annoID", "label", "id1", ["comment"], Permissions.OPEN)
+    anno_res.add_comment("comment2")
+    xml_root.add_resource(anno_res)
+    assert len(xml_root.resources) == 1
+
+    # add_resource_multiple
+    many_resources = []
+    res1 = Resource.create_new("id1", ":Restype", "label")
+    res1.add_bool(":boolProp", True)
```
jnussbaum, line 19: why do you add a value, but then you don't test for its presence? Then it would be simpler not to add the value, to keep the test concise.

test/unittests/xmllib/models/test_root.py
```diff
@@ -0,0 +1,39 @@
...
+    many_resources.append(res1)
+    res2 = Resource.create_new("id2", ":Restype", "label")
+    res2.add_color(":colorProp", "#5b24bf")
```
jnussbaum, line 22: same

test/unittests/xmllib/models/test_root.py
Nora-Olivia-Ammann: Because that is extensively tested for the resource. I could actually remove it since we dont require a value.

test/unittests/xmllib/models/test_root.py
```diff
@@ -0,0 +1,35 @@
...
+    anno_res = AnnotationResource("annoID", "label", "id1", ["comment"])
+    xml_root.add_resource(anno_res)
+    assert len(xml_root.resources) == 1
+
+    # add_resource_multiple
+    many_resources = []
+    res1 = Resource.create_new("id1", ":Restype", "label")
+    many_resources.append(res1)
+    res2 = Resource.create_new("id2", ":Restype", "label")
+    many_resources.append(res2)
```
jnussbaum, line 19:
```suggestion
    many_resources = [Resource.create_new("id1", ":Restype", "label"), Resource.create_new("id2", ":Restype", "label")]
```

PR #1308, opened 2024-12-02
fix(xmllib): check if richtext input is string like before value conversion (DEV-4418)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/resource.py
```diff
@@ -1173,3 +1182,11 @@ def add_migration_metadata(
             )
         self.migration_metadata = MigrationMetadata(creation_date=creation_date, iri=iri, ark=ark, res_id=self.res_id)
         return self
+
+
+def _check_richtext_before_conversion(value: Any, res_id: str, prop_name: str) -> None:
```
Nora-Olivia-Ammann, line 1187: I don't want to make this a public function, which is why I did not add it to the value checkers.

src/dsp_tools/xmllib/models/resource.py
```diff
@@ -1173,3 +1182,11 @@ def add_migration_metadata(
...
+    if not is_string_like(value):
+        msg = (
+            f"Resource '{res_id}' has a richtext value that is not a string: " f"Value: {value} | Property: {prop_name}"
```
jnussbaum, line 1190:
```suggestion
            f"Resource '{res_id}' has a richtext value that is not a string: Value: {value} | Property: {prop_name}"
```

jnussbaum, line 1190: If the value's representation is very long, it hides the information about the property, so I would change the order:

```suggestion
            f"Resource '{res_id}' has a richtext value that is not a string: Property: {prop_name} | Value: {value}"
```

src/dsp_tools/xmllib/models/resource.py
```diff
@@ -884,6 +888,11 @@ def add_richtext_multiple(
         Returns:
             The original resource, with the added values
         """
+        # Because of the conversion, the richtext is converted to a string values such as `pd.NA`
+        # are not recognised as empty any more.
+        # Therefore, this additional check is required here.
```
jnussbaum, line 893: I don't understand this comment. You can merge now, but perhaps afterwards we can discuss what this should mean, and if we want to improve its wording.

PR #1311, opened 2024-12-02
feat(xmllib): ensure that non-empty lists are found during add multiple values (DEV-4421)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/resource.py
```diff
@@ -1188,3 +1199,18 @@ def _check_richtext_before_conversion(value: Any, res_id: str, prop_name: str) -
     if not is_string_like(value):
         msg = f"Resource '{res_id}' has a richtext value that is not a string: Value: {value} | Property: {prop_name}"
         warnings.warn(DspToolsUserWarning(msg))
+
+
+def _check_and_fix_collection_input(value: Any, res_id: str, prop_name: str) -> list[Any]:
```
Nora-Olivia-Ammann, line 1204: I am sure that at some point someone will pass a string into the multiple function. Which will not raise an error but iterate over the string and will confuse everyone for a while until we find the problem.
Therefore I would add this tolerance even if we do not advertise it.

src/dsp_tools/xmllib/models/resource.py
```diff
@@ -1188,3 +1199,18 @@ def _check_richtext_before_conversion(value: Any, res_id: str, prop_name: str) -
...
+    msg = f"The input value of the resource with the ID '{res_id}' and the property '{prop_name}'"
+    match value:
+        case set() | list() | tuple():
+            if len(value) == 0:
```
Nora-Olivia-Ammann, line 1208: Not warning if the input list is empty, is I think an oversight.

test/unittests/xmllib/models/test_resource.py
```diff
@@ -214,6 +217,11 @@ def test_add_simple_text_multiple(self) -> None:
         assert len(res.values) == 2
         assert all([isinstance(x, SimpleText) for x in res.values])
 
+    def test_add_simple_text_multiple_input_non_list(self) -> None:
+        res = Resource.create_new("res_id", "restype", "label").add_simpletext_multiple(":prop", "text1")
+        assert len(res.values) == 1
+        assert all([isinstance(x, SimpleText) for x in res.values])
```
jnussbaum, line 223: Here it would be good to also check the content (i.e. "text1")

src/dsp_tools/xmllib/models/resource.py
jnussbaum, line 1204: Why not: 

```suggestion
def _check_and_fix_collection_input(value: Collection[float | int |str], res_id: str, prop_name: str) -> list[Any]:
```

jnussbaum (approved):

PR #1315, opened 2024-12-03
feat(xmllib): expose lib through init and enable colored printing (RDU-39)

Nora-Olivia-Ammann:


src/dsp_tools/models/custom_warnings.py
```diff
@@ -19,7 +24,16 @@ class DspToolsUserWarning(Warning):
     @classmethod
     def showwarning(cls, message: str) -> None:
         """Print the warning, without context"""
-        cprint(f"WARNING: {message}", color="red", attrs=["bold"])
+        print(BOLD_RED + f"WARNING: {message}" + RESET_TO_DEFAULT)
```
Nora-Olivia-Ammann, line 29: Interestingly it does not work with cprint. You have to configure them with ANSI, but I think this way it is tenable for future use.

src/dsp_tools/xmllib/__init__.py
```diff
@@ -1,4 +1,6 @@
 from .helpers import create_label_to_name_list_node_mapping as create_label_to_name_list_node_mapping
+from .helpers import create_list_from_string as create_list_from_string
```
Nora-Olivia-Ammann, line 2: previously forgot to export them

src/dsp_tools/models/custom_warnings.py
```diff
@@ -1,7 +1,20 @@
 from typing import Protocol
 from typing import runtime_checkable
 
-from termcolor import cprint
+# These are ANSI escape codes which can be used to configure the print output on the terminal
+# https://en.wikipedia.org/wiki/ANSI_escape_code
+# The semicolon separates different configurations
+
+# All codes must start with "\033["
```
Nora-Olivia-Ammann, line 8: I have taken it apart, so that for the future it will be easier to understand the construct.

jnussbaum: Very good introduction to ANSI codes, thanks :-)

src/dsp_tools/models/custom_warnings.py
```diff
@@ -19,7 +32,16 @@ class DspToolsUserWarning(Warning):
     @classmethod
     def showwarning(cls, message: str) -> None:
         """Print the warning, without context"""
-        cprint(f"WARNING: {message}", color="red", attrs=["bold"])
+        print(BOLD_RED + f"WARNING: {message}" + RESET_TO_DEFAULT)
+
+
+class DspToolsUserInfo(Warning):
+    """Class for general user-facing warnings"""
```
jnussbaum, line 39: Identical docstring as above, perhaps you want to formulate one specific for this class?

src/dsp_tools/models/custom_warnings.py
```diff
@@ -1,7 +1,20 @@
...
+# All codes must start with "\033["
+SEQUENCE_START = "\033["
+
+# the "m" at the end signals,
+# that the configuration code is finished and after that the string that should be printed starts
+SEQUENCE_END = "m"
+
+BOLD_RED = SEQUENCE_START + "1;31" + SEQUENCE_END  # 1 (bold) ; 31 (red)
+YELLOW = SEQUENCE_START + "0;33" + SEQUENCE_END  # 0 (normal font) ; 33 (yellow)
+RESET_TO_DEFAULT = SEQUENCE_START + "0" + SEQUENCE_END  # reset to the default setting of the console
```
jnussbaum, line 17: I would use f-strings consistently, and no `+` concatenation at all.

src/dsp_tools/models/custom_warnings.py
```diff
@@ -19,7 +32,16 @@ class DspToolsUserWarning(Warning):
...
+        print(BOLD_RED + f"WARNING: {message}" + RESET_TO_DEFAULT)
+
+
+class DspToolsUserInfo(Warning):
+    """Class for general user-facing warnings"""
+
+    @classmethod
+    def showwarning(cls, message: str) -> None:
+        """Print the warning, without context"""
+        print(YELLOW + f"INFO: {message}" + RESET_TO_DEFAULT)
```
jnussbaum, line 44: Here, the mix between + concatenation and f-string is especially unelegant.

jnussbaum (approved):

src/dsp_tools/models/custom_warnings.py
Nora-Olivia-Ammann: what I like about it is that it clearly and visibly separates what the user message is and what is configuration

PR #1319, opened 2024-12-04
refactor(xmlupload): adapt intermediary resource transformation

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/models/intermediary/resource.py
```diff
@@ -11,8 +11,8 @@
 
 @dataclass
 class ResourceTransformationOutput:
-    resource_success: list[IntermediaryResource]
-    resource_failure: list[ResourceInputConversionFailure]
+    resource_success: IntermediaryResource | None
+    resource_failure: ResourceInputConversionFailure | None = None
```
jnussbaum, line 15: I don't know how much type safety you need, but in case you want to be sure that an instance of this class has always exactly one of the fields filled, and the other is None, you could write some validation logic, as e.g. 

```
def __post_init__(self):
    if bool(self.resource_success) == bool(resource_failure):
        raise ValueError("ResourceTransformationOutput must have either a success, or a failure, but not both, and not neither")
```

jnussbaum (approved):

src/dsp_tools/commands/xmlupload/models/intermediary/resource.py
Nora-Olivia-Ammann: I don't think that is necessary, because the conversion failure can only be created if an error is raised, at that time you do not have access to the intermediary one.

PR #1320, opened 2024-12-05
refactor(xmlupload): integrate intermediary class structure into upload

Nora-Olivia-Ammann:


test/unittests/commands/xmlupload/make_rdf_graph/test_make_values.py
```diff
@@ -351,74 +310,26 @@ def test_segment_of_audio(self, lookups: Lookups) -> None:
         assert value == RES_ONE_URI
 
 
-class TestMakeOnePropGraphRaises:
-    def test_permissions(self, lookups: Lookups) -> None:
```
Nora-Olivia-Ammann, line 355: It is not possible for some error to occur anymore, because permissions resolving etc. would have been noticed earlier.

src/dsp_tools/commands/xmlupload/make_rdf_graph/make_values.py
```diff
@@ -51,157 +69,142 @@ def make_values(properties: list[XMLProperty], res_bnode: BNode, lookups: Lookup
...
-                prop_name=prop_name,
-                permissions_lookup=lookups.permissions,
-                iri_resolver=lookups.id_to_iri,
+                prop_type_info=LINK_PROP_TYPE_INFO,
+                iri_resolver=iri_lookup.id_to_iri,
             )
-        case "text":
-            properties_graph = _make_text_prop_graph(
-                prop=prop,
+        case IntermediaryRichtext():
```
Nora-Olivia-Ammann, line 107: Even with the improved transformation, it is not possible for some values to be constructed through a generic route.

src/dsp_tools/commands/xmlupload/make_rdf_graph/make_values.py
```diff
@@ -51,157 +69,142 @@ def make_values(properties: list[XMLProperty], res_bnode: BNode, lookups: Lookup
...
-        resolved_permission = resolve_permission(val.permissions, permissions_lookup)
-        transformed = TransformedValue(
-            value=URIRef(iri_str),
-            prop_name=prop_name,
-            permissions=resolved_permission,
-            comment=val.comment,
-        )
-        g += _make_simple_value_graph(transformed, res_bn, RDF_PROP_TYPE_MAPPER["list"])
+    val_bn, g = _make_base_value_graph(val, prop_type_info, res_bn)
+    g.add((val_bn, prop_type_info.knora_prop, URIRef(val.value)))
```
Nora-Olivia-Ammann, line 166: `URIRef` is very different from `xsd` therefore I prefer to do it separately from the values that are `Literals`. It would be possible in a generic way, but I think it would make more work in total.

src/dsp_tools/commands/xmlupload/models/intermediary/values.py
```diff
@@ -70,6 +70,9 @@ class IntermediaryInterval(IntermediaryValue):
 class IntermediaryLink(IntermediaryValue):
     value: str
 
+    def __post_init__(self) -> None:
+        self.prop_iri = f"{self.prop_iri}Value"
```
jnussbaum, line 74: This seems a bit odd. I looked at the code and tried to understand why this is necessary, but I didn't manage.

src/dsp_tools/commands/xmlupload/models/lookup_models.py
```diff
@@ -20,12 +19,9 @@ class IntermediaryLookup:
 
 
 @dataclass
-class Lookups:
+class IRILookup:
```
jnussbaum, line 22: The name doesn't really fit the content. The name suggests that the only content of this class is the IRI resolver (=1 thing). But actually, it contains 2 lookups.

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -342,9 +347,17 @@ def _upload_one_resource(
     resource: XMLResource,
     ingest_client: AssetClient,
     resource_create_client: ResourceCreateClient,
-    lookups: Lookups,
+    iri_lookup: IRILookup,
+    intermediary_lookups: IntermediaryLookup,
     creation_attempts_of_this_round: int,
 ) -> None:
+    transformation_result = transform_into_intermediary_resource(resource, intermediary_lookups)
```
jnussbaum, line 354: I'm a bit surprised that you do this so late, immediately before the resource is uploaded. Wouldn't there be a much bigger benefit if this is done earlier, before the upload starts?

test/unittests/commands/xmlupload/make_rdf_graph/test_make_values.py
```diff
@@ -11,49 +10,58 @@
...
 RES_ONE_URI = URIRef("http://rdfh.ch/9999/res_one")
 
 
-@pytest.fixture
-def permissions_lookup() -> dict[str, Permissions]:
-    return {"open": Permissions({PermissionValue.CR: ["knora-admin:ProjectAdmin"]})}
+OPEN_PERMISSION = Permissions({PermissionValue.CR: ["knora-admin:ProjectAdmin"]})
+
+
+def onto_str(prop: str) -> str:
```
jnussbaum, line 48: The name is not intuitive enough, I suggest `dereference_prop_name()` or something similar.

src/dsp_tools/commands/xmlupload/make_rdf_graph/make_values.py
```diff
@@ -51,157 +69,140 @@ def make_values(properties: list[XMLProperty], res_bnode: BNode, lookups: Lookup
...
 
-def _make_one_prop_graph(prop: XMLProperty, res_bnode: BNode, lookups: Lookups) -> tuple[Graph, URIRef]:
-    prop_name = get_absolute_iri(prop.name, lookups.namespaces)
-    match prop.valtype:
-        case "boolean" | "color" | "decimal" | "geometry" | "geoname" | "integer" | "time" | "uri" as val_type:
-            literal_info = RDF_PROP_TYPE_MAPPER[val_type]
-            transformer = RDF_LITERAL_TRANSFORMER_MAPPER[val_type]
-            properties_graph = _make_simple_prop_graph(
-                prop=prop,
+def _make_one_prop_graph(val: IntermediaryValue, res_bnode: BNode, iri_lookup: IRILookup) -> tuple[Graph, URIRef]:
```
jnussbaum, line 79: Does this produce 1 prop graph or 1 value graph?

src/dsp_tools/commands/xmlupload/make_rdf_graph/make_values.py
```diff
@@ -51,157 +69,140 @@ def make_values(properties: list[XMLProperty], res_bnode: BNode, lookups: Lookup
...
+                val=val,
                 res_bn=res_bnode,
-                prop_name=prop_name,
-                permissions_lookup=lookups.permissions,
             )
         case _:
-            raise UserError(f"Unknown value type: {prop.valtype}")
-    return properties_graph, prop_name
+            raise UserError(f"Unknown value type: {type(val).__name__}")
+    return properties_graph, URIRef(val.prop_iri)
```
jnussbaum, line 126: what is the meaning of URIRef(val.prop_iri)? Could you quickly document (perhaps in a function docstring) what this is and what it is used for?

test/unittests/commands/xmlupload/make_rdf_graph/test_make_values.py
```diff
@@ -11,49 +10,58 @@
...
+    def test_boolean(self, lookups: IRILookup) -> None:
         res_bn = BNode()
-        xml_prop = etree.fromstring("""
-        <boolean-prop name=":isTrueOrFalse">
-            <boolean permissions="open">true</boolean>
-        </boolean-prop>
-        """)
-        prop = XMLProperty.from_node(xml_prop, "boolean", "onto")
+        prop = IntermediaryBoolean(True, onto_str("isTrueOrFalse"), None, OPEN_PERMISSION)
         result, prop_name = _make_one_prop_graph(prop, res_bn, lookups)
```
jnussbaum, line 65: What is result, and what is prop_name? In the prod code, they are called single_prop_graph and last_prop_name. If you use the same names here, that would simplify the understanding of the tests.

test/unittests/commands/xmlupload/make_rdf_graph/test_make_values.py
```diff
@@ -65,14 +73,9 @@ def test_boolean(self, lookups: Lookups) -> None:
...
+    def test_color(self, lookups: IRILookup) -> None:
         res_bn = BNode()
-        xml_prop = etree.fromstring("""
-        <color-prop name=":hasColor">
-            <color>#5d1f1e</color>
-        </color-prop>
-        """)
-        prop = XMLProperty.from_node(xml_prop, "color", "onto")
+        prop = IntermediaryColor("#5d1f1e", onto_str("hasColor"), None, None)
         result, prop_name = _make_one_prop_graph(prop, res_bn, lookups)
```
jnussbaum, line 79: same here, and multiple times below

test/unittests/commands/xmlupload/make_rdf_graph/test_make_values.py
```diff
@@ -274,14 +219,36 @@ def test_richtext(self, lookups: Lookups) -> None:
...
+        )
+        result, prop_name = _make_one_prop_graph(prop, res_bn, lookups)
+        assert len(result) == 4
+        assert prop_name == ONTO.hasRichtext
+        val_bn = next(result.objects(res_bn, prop_name))
+        rdf_type = next(result.objects(val_bn, RDF.type))
+        assert rdf_type == KNORA_API.TextValue
+        value = next(result.objects(val_bn, KNORA_API.textValueAsXml))
+        expected_text = (
+            '<?xml version="1.0" encoding="UTF-8"?>\n'
```
jnussbaum, line 236: where does the newline come from? not that it matters, just curiosity

jnussbaum (approved):
Overall, it looks very good, thanks for your efforts!
I stumbled over some namings that seemed counter-intuitive to me.
In addition, I wondered why the transformation of the XMLResources into IntermediaryResources happens so late.

src/dsp_tools/commands/xmlupload/models/intermediary/values.py
Nora-Olivia-Ammann: All links between two resources (eg. `LinkValue`) generate two link triples.

User-facing triple:
`:A :hasLinkTo :B`

1. `hasLinkTo` -> property that connects resource `:A` to resource `:B`
2.  `hasLinkToValue` -> property that connects resource `:A` to the `LinkValue` where all the metadata (permissions etc.) are stored.

Because we send metadata along, we do not send the actual `hasLinkTo` but the `hasLinkToValue`
I looked at the possible places where to do this best and because I want to save the correct data in the intermediary classes this seemed the easiest. But I am open to change this.

src/dsp_tools/commands/xmlupload/xmlupload.py
Nora-Olivia-Ammann: Absolutely, I think I mentioned it somewhere, the reason why I do not, is because the diff in this PR is already quite large. Moving it out will generate an even larger diff. I wanted to break it up into two to make the review easier.

test/unittests/commands/xmlupload/make_rdf_graph/test_make_values.py
Nora-Olivia-Ammann: it comes from copy pasting... thanks for noticing it.

src/dsp_tools/commands/xmlupload/make_rdf_graph/make_values.py
Nora-Olivia-Ammann: It creates one graph with one property which also corresponds to one value. I can rename it if it is confusing .

test/unittests/commands/xmlupload/make_rdf_graph/test_make_values.py
Nora-Olivia-Ammann: I want it short and simple. It is only in the test code.

PR #1321, opened 2024-12-05
fix(ingest-xmlupload): remove leading slash

jnussbaum:


src/dsp_tools/commands/ingest_xmlupload/create_resources/apply_ingest_id.py
```diff
@@ -57,8 +57,8 @@ def replace_filepath_with_internal_filename(
         if not etree.QName(elem).localname.endswith("bitstream") or not elem.text:
             continue
         img_path_str = elem.text
-        if img_path_str not in orig_path_2_asset_id:
-            img_path_str = "/" + img_path_str
```
jnussbaum, line 61: When I coded this, I didn't write a test, so I did it the wrong way round. Only now, when I used this, I realised I had done it wrong.

Nora-Olivia-Ammann (approved):

PR #1322, opened 2024-12-05
docs(xmllib): use doctest

jnussbaum:


jnussbaum:
@Notheturtle I have quickly told you about a new syntax for the docstrings example. Here is now how it would look like. Here are some screenshots, so that you can compare how the docstrings example would look like. "Old" means: As Nora has discussed with Noémi yesterday. "New" means: Johannes' new idea of today, not yet discussed with Noémi.

There are pros and cons of both:
- Old: perhaps easier to look at (?), but not testable -> could become outdated or be inaccurate
- New: less nice to look at (?), but testable -> are guaranteed to be always accurate

| Old | New |
| ---- | --- | 
| img | img |
| img | img |
| img | img |

src/dsp_tools/xmllib/helpers.py
```diff
@@ -60,15 +61,12 @@ def create_label_to_name_list_node_mapping(
...
-        ```python
-        result = xmllib.create_label_to_name_list_node_mapping(
-            project_json_path="project.json",
-            list_name="listName",
-            language_of_label="de",
-        )
-        # -> { "Knoten 1": "n1", "Knoten 2": "n2" }
-        ```
+        >>> xmllib.create_label_to_name_list_node_mapping(
+        ...     project_json_path="src/dsp_tools/xmllib/assets/project.json",
```
jnussbaum, line 65: This needs to reference a real file, otherwise it would fail. Alternatively, we could just mark this example with `# doctest: +SKIP`, then it wouldn't be tested.

src/dsp_tools/xmllib/assets/project.json
jnussbaum, line 1: This file is used by `create_label_to_name_list_node_mapping()`, which expects a real file being present.

src/dsp_tools/xmllib/helpers.py
```diff
@@ -210,15 +203,9 @@ def find_date_in_string(string: str) -> str | None:
...
-        # -> "GREGORIAN:CE:1849:CE:1850"
-        ```
-
-        ```python
-        result = xmllib.find_date_in_string("not a valid date")
-        # -> None
-        ```
+        >>> xmllib.find_date_in_string("1849/1850")
+        'GREGORIAN:CE:1849:CE:1850'
+        >>> xmllib.find_date_in_string("not a valid date")
```
jnussbaum, line 208: I haven't found out a nice way how to display "no output".

Nora-Olivia-Ammann:
Thank you for the mock-ups! What i think is important in any case is to visually separate the different examples. Now we have one grey box per example. In the new version you have all examples in one grey box. This makes it much less clear to me.

src/dsp_tools/xmllib/helpers.py
Nora-Olivia-Ammann: I don't like that you have to reference a real file. Now we have a file that we do not otherwise need and a path is just unnecessary noise. If we went with the docstest I would absolutely skip this.

src/dsp_tools/xmllib/helpers.py
Nora-Olivia-Ammann: `None` ?

Nora-Olivia-Ammann (commented):
On the whole i do like the idea. But I am not sure if it is necessary for us. These functions should never have behavioral changes in any case...

jnussbaum:
> These functions should never have behavioral changes in any case...
Well, that's the point. They SHOULD not have behavioral changes, but we might inadvertently introduce changes.

jnussbaum:
> Thank you for the mock-ups! What i think is important in any case is to visually separate the different examples. Now we have one grey box per example. In the new version you have all examples in one grey box. This makes it much less clear to me.

That can be easily done, I will adapt it.

Nora-Olivia-Ammann:
> > These functions should never have behavioral changes in any case...
> > Well, that's the point. They SHOULD not have behavioral changes, but we might inadvertently introduce changes.

We would notice that in the tests though...

PR #1323, opened 2024-12-05
chore: upgrade deps / fix new typing errors of types-lxml / resolve 1 warning

jnussbaum:


Nora-Olivia-Ammann (approved):
Why is this necessary?

PR #1324, opened 2024-12-05
refactor(xmllib): change boolean conversion

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/value_converters.py
```diff
@@ -1,48 +1,51 @@
 from typing import Any
 
+from dsp_tools.models.exceptions import InputError
 from dsp_tools.xmllib.models.config_options import NewlineReplacement
 
 
-def convert_to_bool_string(value: Any) -> str:
+def convert_to_bool(value: Any) -> bool:
```
Nora-Olivia-Ammann, line 7: I thought that likely the user will expect a bool from such a function and not a string that is an xml boolean...

PR #1331, opened 2024-12-11
feat(validate-data): validate `DocumentRepresentation`, `ArchiveRepresentation`, `TextRepresentation` (DEV-4334)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/make_data_rdf.py
```diff
@@ -217,10 +220,16 @@ def _map_into_correct_file_value(val: AbstractFileValueDeserialised) -> FileValu
...
     match file_extension:
+        case "zip" | "tar" | "gz" | "z" | "tgz" | "gzip" | "7z":
+            file_type = ARCHIVE_FILE_VALUE
         case "mp3" | "wav":
             file_type = AUDIO_FILE_VALUE
+        case "pdf" | "doc" | "docx" | "xls" | "xlsx" | "ppt" | "pptx":
+            file_type = DOCUMENT_FILE_VALUE
         case "mp4":
             file_type = MOVING_IMAGE_FILE_VALUE
+        case "odd" | "rng" | "txt" | "xml" | "xsd" | "xsl" | "csv" | "json":
```
seakayone, line 231: *idea*: How about introducing some central place for all`SupportedFiletypes` 

I am pretty sure I have seen these file extension already in dsp-tool on some other places.

We should think about having a central place to define the extensions so we do not run into trouble once we might have to support other file formats.

seakayone (approved):

src/dsp_tools/commands/validate_data/make_data_rdf.py
Nora-Olivia-Ammann: Yes that is true. I have been refactoring the xmlupload and introduced the anti-corruption layer. The problem currently is that in the xmlupload we still do some crashes if the data is incorrect (list node could not be resolved, etc.) which we do not want here. But the files are simpler I think so that is something we could do. My plan is (I will put it on linear) to use the already existing parser for the xml from the upload here. I have to wait though till we implemented all the in-built dsp-resources. Once I am finished with the files and all the possible mistakes I will take a look. Thanks!

PR #1335, opened 2024-12-12
refactor(upload-files): stream big files (DEV-4469)

jnussbaum:


src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
```diff
@@ -59,29 +59,27 @@ def upload_file(
...
-            with open(self.imgdir / filepath, "rb") as binary_io:
-                content = binary_io.read()
-        except OSError as e:
-            err_msg = f"Cannot bulk-ingest {filepath}, because the file could not be opened/read: {e.strerror}"
-            logger.error(err_msg)
-            return UploadFailure(filepath, err_msg)
         url = self._build_url_for_bulk_ingest_ingest_route(filepath)
         headers = {"Content-Type": "application/octet-stream"}
-        timeout = 600
+        timeout = 1000
```
seakayone, line 64: *issue*: We have to make sure that this timeout is aligned with how our load balancers (LB) are currently configured

If both values are not equal we might end up in two different non-optimal or even problematic situations:
* timeout of LB is bigger: We loose time we could have used to upload bigger files. 
* timeout of LB is smaller: The request is killed by the LB and we have hard to understand error messages on both sides dsp-tools and dsp-ingest.

I would recommend a timeout value which is a couple of seconds smaller than the LB timeout.

Please double check with infra how far of we are from the optimal value.

test/unittests/commands/ingest_xmlupload/test_bulk_ingest_client.py
```diff
@@ -45,7 +45,6 @@ def test_upload_file_success(ingest_client: BulkIngestClient, requests_mock: Moc
     assert req.method == "POST"
     assert req.headers["Authorization"] == "Bearer mocked_token"
     assert req.headers["Content-Type"] == "application/octet-stream"
-    assert req.body == file_content.encode()
```
seakayone, line 48: IMHO there is no need to remove this assertion.

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
```diff
@@ -59,29 +59,27 @@ def upload_file(
         filepath: Path,
     ) -> UploadFailure | None:
         """Uploads a file to the ingest server."""
-        try:
```
seakayone, line 62: *suggestion* Keep separate try blocks if not other reasons force you to do it.

In my experience as soon as code is raising an Error it is generally a harder thing to understand the control flow, especially when reading the code. In a `try` block with multiple lines and `except`s it is not always obvious which line could have caused the `Error`. 

I always try minimizing the number of lines in a `try` like here, the OSError can only come from the open or the read and not from something else.

seakayone (approved):
(nitpick) This is not a pure refactoring, you are changing the behaviour by changing the timeout value.   

Generally I would appreciate a short description of the motiviation of the refactoring. My comment on merging the two try blocks might miss the context of your motivation and could be void. 

Now after having reviewed this PR I think the central change in this PR is that the binaray_io is directly supported by requests. This looks good to me and my comment on the try block is void.

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
jnussbaum: @SamuelBoerlin @LukasStoeckli What is the timeout of traefik? I ran `grep -ri "timeout"` in ops-deploy, but didn't find anything.

SamuelBoerlin: @jnussbaum the [default is 60s](https://doc.traefik.io/traefik/routing/entrypoints/#transport), which is what we're currently using for all publicly exposed endpoints.

test/unittests/commands/ingest_xmlupload/test_bulk_ingest_client.py
jnussbaum: Previously, the body contained all bytes of the entire file, which is now not the case anymore. I don't think there is an easy way to test the body, because now the file is streamed.

src/dsp_tools/commands/ingest_xmlupload/bulk_ingest_client.py
jnussbaum: Thanks for your input. Resolved with https://github.com/dasch-swiss/dsp-tools/commit/d3d15d7ba90a62799c182bbedd2116eadc3d84fa

PR #1337, opened 2024-12-12
refactor(validate-data): change exclusion of properties from gui-order to `isEditable` (DEV-4281)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/sparql/cardinality_shacl.py
```diff
@@ -84,12 +83,18 @@ def _construct_1_cardinality(onto_graph: Graph) -> Graph:
...
           owl:onProperty ?propRestriction ;
-          salsah-gui:guiOrder ?order ;
           owl:cardinality 1 .
+          
+      ?propRestriction knora-api:isEditable true .
       FILTER NOT EXISTS { ?propRestriction knora-api:isLinkValueProperty true }
       FILTER (?class NOT IN (
             knora-api:Region, knora-api:AudioSegment, knora-api:VideoSegment, knora-api:LinkObj)
       )
+      FILTER (?propRestriction NOT IN (
```
Nora-Olivia-Ammann, line 93: The ttl file describing the content of the file values already contains a cardinality. If we add them here too, we get two violations reports for one cardinality violation. We could remove the cardinality restriction in the ttl file and only leave it here, but in that case the user message is not as nice. This is why I have decided to filter them out here.

seakayone (approved):

PR #1339, opened 2024-12-13
refactor(xmlupload): create resource with rdflib

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/resource_create_client.py
```diff
@@ -10,15 +14,27 @@ class ResourceCreateClient:
...
         resource_has_bitstream: bool,
     ) -> str:
         """Creates a resource on the DSP server, and returns its IRI"""
+        res_dict = _make_json(graph)
         headers = {"X-Asset-Ingested": "true"} if resource_has_bitstream else None
         res = self.con.post(route="/v2/resources", data=res_dict, headers=headers)
         return cast(str, res["@id"])
+
+
+def _make_json(rdf_graph: Graph) -> dict[str, Any]:
```
Nora-Olivia-Ammann, line 31: The result of the serialisation is a list. Although the API would accept that, our connection has the payload typed as `dict[str, Any]`. This means that I would either have to change the typing everywhere in the connection (I tried it out, it is not as simple as it sounds), or write a separate post route or not use the connection class. I deem this to be out of scope of this PR therefore, I chose to frame it again. I do this here, because the format of the payload is the business of the API call.

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -180,14 +181,13 @@ def _upload_resources(clients: UploadClients, upload_state: UploadState) -> None
     namespaces = make_namespace_dict_from_onto_names(project_onto_dict)
 
     iri_lookup = IRILookups(
-        project_iri=project_iri,
+        project_iri=URIRef(project_iri),
         id_to_iri=upload_state.iri_resolver,
         jsonld_context=project_context,
```
Nora-Olivia-Ammann, line 186: The json-ld context is no longer needed here, but it is for the stash. Therefore it remains for the time being.

test/integration/commands/xmlupload/test_resource_creation.py
```diff
@@ -86,8 +86,10 @@ def test_one_resource_without_links(ingest_client_mock: AssetClient) -> None:
         "route": "/v2/resources",
         "data": {
             "@type": "http://0.0.0.0:3333/ontology/9999/onto/v2#foo_1_type",
-            "rdfs:label": "foo_1_label",
-            "knora-api:attachedToProject": {"@id": "https://admin.test.dasch.swiss/project/MsOaiQkcQ7-QPxsYBKckfQ"},
+            "http://www.w3.org/2000/01/rdf-schema#label": "foo_1_label",
```
Nora-Olivia-Ammann, line 89: The serialisation of rdflib works with absolute IRIs

src/dsp_tools/commands/xmlupload/make_rdf_graph/make_file_value.py
```diff
@@ -22,28 +22,29 @@
...
         iiif_uri: Information about the IIIF URI
-        res_bn: Blank-node of the resource
+        res_node: Node of the resource
 
     Returns:
         Graph with the IIIF-URI Value
     """
-    g = _make_abstract_file_value_graph(iiif_uri, IIIF_URI_VALUE, res_bn, KNORA_API.stillImageFileValueHasExternalUrl)
-    return g, KNORA_API.hasStillImageFileValue
+    return _make_abstract_file_value_graph(
```
Nora-Olivia-Ammann, line 36: The propery name was needed for the framing when we only had the values. Now that we have the resource we can frame it in a more generalised way.

src/dsp_tools/commands/xmlupload/make_rdf_graph/make_resource_and_values.py
```diff
@@ -1,82 +1,93 @@
...
 
-    res = _make_resource(resource=resource, lookup=lookup)
-    res.update(_make_values_dict_from_resource(resource, bitstream_information, lookup))
-    res.update(lookup.jsonld_context.serialise())
-    return res
+    graph = Graph()
+
+    res_node: BNode | URIRef = BNode()
+    if migration := resource.migration_metadata:
+        if migration.iri_str:
```
Nora-Olivia-Ammann, line 42: We cannot guarantee that the IRI exists

test/unittests/commands/xmlupload/make_rdf_graph/test_make_values.py
```diff
@@ -120,10 +116,9 @@ def test_geometry(self, lookups: IRILookups) -> None:
     def test_geoname(self, lookups: IRILookups) -> None:
         res_bn = BNode()
         prop = IntermediaryGeoname("5416656", absolute_iri("hasGeoname"), None, None)
-        result, prop_name = _make_one_value_graph(prop, res_bn, lookups)
+        result = _make_one_value_graph(prop, res_bn, lookups)
         assert len(result) == 3
-        assert prop_name == ONTO.hasGeoname
-        val_bn = next(result.objects(res_bn, prop_name))
+        val_bn = next(result.objects(res_bn, ONTO.hasGeoname))
```
Nora-Olivia-Ammann, line 121: We do not need to assert the the property name, because if it is not the expected one, we would get a stop iteration here.

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -260,7 +259,7 @@ def _upload_one_resource(
     iri = None
     try:
         serialised_resource = create_resource_with_values(
-            resource=transformed_resource, bitstream_information=media_info, lookup=iri_lookup
+            resource=transformed_resource, bitstream_information=media_info, lookups=iri_lookup
```
jnussbaum, line 262: I still feel uneasy with the naming here. You have a variable `iri_lookup`, but the class name is `IRILookups`. But the content of the class are several fields, so the name doesn't fit the content.

test/integration/commands/xmlupload/test_resource_creation.py
```diff
@@ -86,8 +86,10 @@ def test_one_resource_without_links(ingest_client_mock: AssetClient) -> None:
...
+            "http://api.knora.org/ontology/knora-api/v2#attachedToProject": {
+                "@id": "https://admin.test.dasch.swiss/project/MsOaiQkcQ7-QPxsYBKckfQ"
```
jnussbaum, line 91: not part of this PR, but is this a correct project IRI? It looks outdated.

test/unittests/commands/xmlupload/make_rdf_graph/test_make_resource_and_values.py
```diff
@@ -0,0 +1,103 @@
...
+        label="Resource Label",
+        permissions=None,
+        values=[],
+        file_value=None,
+        iiif_uri=None,
+        migration_metadata=migration_metadata,
+    )
+    res_iri = URIRef("http://rdfh.ch/4123/DiAmYQzQSzC7cdTo6OJMYA")
+    res_graph = _make_resource(res, res_iri, PROJECT_IRI)
+    assert len(res_graph) == 3
```
jnussbaum, line 85: hmm, but the metadata should also be contained in the graph. Why have they vanished?

jnussbaum (approved):

src/dsp_tools/commands/xmlupload/xmlupload.py
Nora-Olivia-Ammann: yes

test/integration/commands/xmlupload/test_resource_creation.py
Nora-Olivia-Ammann: I have never taken a look. So it could be an outdated project iri...

test/unittests/commands/xmlupload/make_rdf_graph/test_make_resource_and_values.py
Nora-Olivia-Ammann: They have not vanished, but they are not added to the graph in this function. I will remove it, I think when I wrote the test it made more sense.

PR #1340, opened 2024-12-13
refactor(validate-data): change deserialisation in preparation for dsp in-built resource validation

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/deserialise_input.py
```diff
@@ -50,23 +51,24 @@ def deserialise_xml(root: etree._Element) -> ProjectDeserialised:
 
 
 def _deserialise_all_resources(root: etree._Element) -> DataDeserialised:
-    all_res: list[AbstractResource] = []
-    for res in root.iterchildren():
+    all_res: list[ResourceDeserialised] = []
```
Nora-Olivia-Ammann, line 54: I want to discuss the deserialisation strategy in more detail (I scheduled a meeting on monday). But for the short-term this change will make things easier in the short term

seakayone (approved):

PR #1342, opened 2024-12-16
feat(validate-data): `knora-api:LinkObj` (DEV-4463)

Nora-Olivia-Ammann:


src/dsp_tools/resources/validate_data/api-shapes-with-cardinalities.ttl
Nora-Olivia-Ammann, line 1: It auto-formatted this file... nothing changed otherwise

seakayone (approved):

PR #1344, opened 2024-12-17
fix(validate-data): message of LinkValue in SHACL (DEV-4471)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/sparql/value_shacl.py
```diff
@@ -164,6 +164,7 @@ def _construct_link_value_node_shape(onto: Graph) -> Graph:
                 knora-api:objectType ?rangeClass .
 
         BIND(IRI(CONCAT(str(?prop), "_NodeShape")) AS ?nodeShapeIRI)
+        BIND(CONCAT("Range is ", str(?rangeClass), " or a subclass.") AS ?msg)
```
Nora-Olivia-Ammann, line 167: Using an IRI in the message violates the constraint of `sh:message` is must be a string.

PR #1351, opened 2024-12-18
chore: upgrade `rdflib` dependency

Nora-Olivia-Ammann:


test/integration/commands/xmlupload/test_resource_creation.py
```diff
@@ -86,14 +86,20 @@ def test_one_resource_without_links(ingest_client_mock: AssetClient) -> None:
         "route": "/v2/resources",
         "data": {
             "@type": "http://0.0.0.0:3333/ontology/9999/onto/v2#foo_1_type",
-            "http://www.w3.org/2000/01/rdf-schema#label": "foo_1_label",
+            "http://www.w3.org/2000/01/rdf-schema#label": {
```
Nora-Olivia-Ammann, line 89: All of these changes do not at all change the content. It is just explicit typing, in the previous version the typing was implicit. Which is not actually ideal.

jnussbaum: Does that mean that the old and the new are equivalent? If yes, then I begin understanding the benefit of using a graph library. And the benefit of thinking in graph terms, rather than JSON terms. Because when thinking in pure JSON, then your change would change the content.

jnussbaum (approved):

test/integration/commands/xmlupload/test_resource_creation.py
Nora-Olivia-Ammann: Yes and no with regard to the content.
It works like this with regard to typing
`"1" `-> is like a string, we treat it like a string, but it is not explicitly a string
`1` -> is not like a string, json allows numbers like that, so we treat it like a integer, BUT, there are two [xsd integers](https://stackoverflow.com/questions/15336872/xsd-what-is-the-difference-between-xsinteger-and-xsint), you have no control which one it is.

with typing
`"1"^^xsd:float` -> this is a decimal equivalent to 1.0
`"1"^^xsd:int`-> this is an integer equivalent to 1

The reason why I say it is a little different is because in both the triplestores I know (GraphDB and Fuseki) if you don't type your strings they will be saved as `"string"` and if you type it it is saved as `"string"^^xsd:string`. 
For us I don't think very relevant because it needs to go through the API anyway. I have not experimented with this but [this](https://www.w3.org/TR/rdf12-concepts/#dfn-literal-term-equality) indicates that it would influence your query results.

So in short, be consistent and take control of your datatypes otherwise you don't know what happens...

PR #1353, opened 2024-12-19
feat(xmllib): make checks for non-empty input consistent (DEV-4435)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/internal_helpers.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Returns:
+        A richtext value
+
+    Raises:
+        Input Error if the input is a dictionary
+    """
+    # Because of the richtext conversions, the input value is cast as a string.
+    # Values such as str(`pd.NA`) result in a non-empy string.
+    # Therefore, a check must occur before the conversion takes place.
```
jnussbaum, line 43: @Nora-Olivia-Ammann Where does a conversion take place? In `replace_newlines_with_tags()`, the input is not converted into a string.

src/dsp_tools/xmllib/internal_helpers.py
```diff
@@ -0,0 +1,90 @@
...
+        newline_replacement: the replacement for the newlines in the string
+        res_id: id of the calling resource
+
+    Returns:
+        A richtext value
+
+    Raises:
+        Input Error if the input is a dictionary
+    """
+    # Because of the richtext conversions, the input value is cast as a string.
```
jnussbaum, line 41: @Nora-Olivia-Ammann Where is the "cast as a string"? I don't see it.

PR #1355, opened 2024-12-19
refactor(xmllib): simplify serialisation of values

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/resource.py
```diff
@@ -116,7 +116,7 @@ def serialise(self) -> etree._Element:
         res_ele = self._serialise_resource_element()
         if self.file_value:
             res_ele.append(self.file_value.serialise())
-        res_ele.extend(self._serialise_values())
+        res_ele.extend(serialise_values(self.values))
```
Nora-Olivia-Ammann, line 119: Making the serialisation of the resource generic is a next step.
I don't think it is a problem to remove the serialise function in the values, because it was never user-facing. If we want to keep it public for the resource we can decide later.

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -354,9 +356,9 @@ def _serialise_resource_element(self) -> etree._Element:
         return etree.Element(f"{DASCH_SCHEMA}region", attrib=attribs, nsmap=XML_NAMESPACE_MAP)
 
     def _serialise_values(self) -> list[etree._Element]:
-        return [
-            LinkValue(value=self.region_of, prop_name="isRegionOf", resource_id=self.res_id).serialise(),
-        ]
+        return serialise_values(
```
Nora-Olivia-Ammann, line 359: This is a little odd but I am planning on improving that in the future, but for now it works and allowed me to remove the `.serialise()` function completely.

jnussbaum (approved):

PR #1367, opened 2025-01-15
refactor(validate-data): simplify internal representation of file values

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/deserialise_input.py
```diff
@@ -167,12 +168,46 @@ def _get_text_as_string(value: etree._Element) -> str | None:
...
+        ValueInformation(
+            user_facing_prop=prop_str,
+            user_facing_value=file_str,
+            knora_type=file_type,
+            value_metadata=[],
+        )
+    ]
+
+
+def _get_file_value_type(file_name: str) -> tuple[KnoraValueType | None, str]:  # noqa:PLR0911 (Too many return statements)
```
Nora-Olivia-Ammann, line 196: To make it consistent with the other values, we need the knora property here already.
Previously this was handled by the same code that does the xmlupload.
I am not sure if this is the most optimal and final solution. But it makes the serialisation in an rdflib representation easier. Currently this is still better than what it replaces but also it makes it more consistent.

uv.lock
```diff
@@ -104,7 +104,7 @@ name = "click"
 version = "8.1.8"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "colorama", marker = "sys_platform == 'win32'" },
+    { name = "colorama", marker = "platform_system == 'Windows'" },
```
jnussbaum, line 107: How did this come into being? Is this necessary / beneficial?

jnussbaum (approved):

uv.lock
Nora-Olivia-Ammann: I updated in between.

PR #1370, opened 2025-01-15
chore: upgrade `rdflib` and fix typing

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/constants.py
```diff
@@ -17,6 +23,10 @@
 from dsp_tools.commands.xmlupload.make_rdf_graph.constants import URI_PROP_TYPE_INFO
 from dsp_tools.commands.xmlupload.models.rdf_models import RDFPropTypeInfo
 
+# rdflib typing
+PropertyTypeAlias: TypeAlias = Union[IdentifiedNode, Variable]
```
Nora-Olivia-Ammann, line 27: I copied the types they use internally:
https://github.com/RDFLib/rdflib/blob/6dc95f8043dcf802849bb17fbd46e3520e92fd90/rdflib/graph.py#L307

Unfortunately they don't expose type aliases...

jnussbaum (approved):

PR #1371, opened 2025-01-16
feat(validate-data): validate `hasComment` on values (DEV-4472)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/deserialise_input.py
```diff
@@ -104,16 +105,27 @@ def _deserialise_one_property(prop_ele: etree._Element) -> list[ValueInformation
             return []
 
 
+def _extract_metadata_of_value(value: etree._Element) -> list[PropertyObject]:
+    property_objects = []
+    for k, v in value.attrib.items():
+        if not (knora_prop := XML_ATTRIB_TO_PROP_TYPE_MAPPER.get(k)):
+            continue
```
Nora-Olivia-Ammann, line 112: permissions are not implemented. This will change as soon as that is done.

PR #1373, opened 2025-01-16
feat(validate-data): `seqnum` and `isPartOf` (DEV-4242)

Nora-Olivia-Ammann:


testdata/validate-data/onto_for_cardinality_checks.ttl
Nora-Olivia-Ammann, line 1: This used to be the onto.ttl file. I renamed it because it is used to check if the cardinalities are constructed correctly. For other tests I need an updated ontology, which is the onto.ttl file. But there is no benefit to update the cardinality tests every time.

src/dsp_tools/commands/validate_data/reformat_validaton_result.py
```diff
@@ -180,6 +182,13 @@ def _query_one_without_detail(
             return _query_for_non_existent_cardinality_violation(base_info, results_and_onto)
         case SH.SPARQLConstraintComponent:
             return _query_for_unique_value_violation(base_info, results_and_onto)
+        case DASH.CoExistsWithConstraintComponent:
```
Nora-Olivia-Ammann, line 185: The construction of the user message is a little hacky.
But the seqnum / isPartOf is not completely implemented yet. It is extremely complicated because the API not only allows erroneous data, but also erroneous ontologies (both of them can be used multiple times on one resource. Which sequence number is displayed is chance. This is not a hypothetical error but has caused problems with real data. Balduin knows of it but it is not high priority. Before I optimise the user information both with regard to its content and construction in code, I want to have it completely implemented.

src/dsp_tools/commands/validate_data/models/input_problems.py
Nora-Olivia-Ammann, line 1: The diff is so large because I had to update and add some test turtle files.

src/dsp_tools/commands/validate_data/models/input_problems.py
```diff
@@ -193,6 +193,29 @@ def sort_value(self) -> str:
         return self.prop_name
 
 
+#######################
```
Nora-Olivia-Ammann, line 196: See below for comment regarding user message.

testdata/validate-data/generic/project.json
```diff
@@ -689,40 +689,6 @@
                 "cardinality": "0-1"
               }
             ]
-          },
```
Nora-Olivia-Ammann, line 692: This is a larger problem, I am taking care of it in https://linear.app/dasch/issue/DEV-4517/prevent-erroneous-ontology-cardinalities-for-seqnum-and-ispartof with a separate ontology and project.

seakayone (approved):

PR #1374, opened 2025-01-17
feat(validate-data): prevent erroneous ontology for `seqnum` and `isPartOf`(DEV-4517)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/validate_ontology.py
```diff
@@ -0,0 +1,54 @@
...
+from dsp_tools.commands.validate_data.utils import reformat_onto_iri
+
+LIST_SEPARATOR = "\n    - "
+
+
+def validate_ontology(
+    onto_graph: Graph, shacl_validator: ShaclValidator, save_path: Path | None
+) -> OntologyValidationProblem | None:
+    """
+    The API accepts erroneous cardinalities in the ontology.
```
seakayone, line 21: *question*: Would it also make sense to create an issue for dsp-api to prevent accepting these cases?

seakayone (approved):

src/dsp_tools/commands/validate_data/validate_ontology.py
Nora-Olivia-Ammann: Absolutely! But the answer regarding that from Balduin was, they know it is a problem. When / if they address it, they will have to do further investigation how to solve it. So I don't expect the API to prevent this anytime soon.

PR #1378, opened 2025-01-21
improvement(validate-data): improve performance time for SHACL construction (DEV-4523)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/sparql/cardinality_shacl.py
```diff
@@ -91,9 +91,7 @@ def _construct_1_cardinality(onto_graph: Graph) -> Graph:
           
       ?propRestriction knora-api:isEditable true .
       FILTER NOT EXISTS { ?propRestriction knora-api:isLinkValueProperty true }
-      FILTER (?class NOT IN (
```
Nora-Olivia-Ammann, line 94: These are no longer necessary because the classes are not in the graph anymore.

src/dsp_tools/commands/validate_data/sparql/construct_shacl.py
```diff
@@ -8,20 +8,55 @@
...
+    cardinality = construct_cardinality_node_shapes(graph_to_query)
+    content = construct_property_shapes(graph_to_query, project_lists)
+    file_values = construct_file_value_cardinality(graph_to_query)
     cardinality += file_values
     return SHACLGraphs(cardinality=cardinality, content=content)
+
+
+def _get_all_relevant_knora_subset(knora_api: Graph) -> Graph:
+    logger.info("Getting relevant knora-api subset")
+    props = ["isLinkProperty", "isEditable"]
```
Nora-Olivia-Ammann, line 35: The aim is to filter out all relevant information from knora-api and add it to the ontology graph rather than add the entire knora-api ontology. The construct queries are extremely slow. By doing this I could reduce the time from over 20 seconds to 3.

We need this information about knora properties for cardinalities.

src/dsp_tools/resources/validate_data/api-shapes-with-cardinalities.ttl
```diff
@@ -8,6 +8,63 @@
...
+  sh:maxCount 1 ;
+  sh:path     rdfs:label ;
+  sh:severity sh:Violation .
+
+#########################################
+# DSP IN-BUILT RESOURCES
+#########################################
+
+
+###########################
```
Nora-Olivia-Ammann, line 24: Because we no longer have the resource classes in the graph that is queried these cardinalities will not be constructed. As these do not change often I decided to "hard-code" them here. The shape here is the same as if it was queried.

seakayone (approved):

PR #1380, opened 2025-01-22
feat(validate-data): validate `knora-api:Region` (DEV-4462)

Nora-Olivia-Ammann:


src/dsp_tools/resources/validate_data/api-shapes-with-cardinalities.ttl
```diff
@@ -24,7 +24,7 @@ api-shapes:rdfsLabel_Cardinality
 ###########################
 # LinkObj
 
-api-shapes:LinkObj_Cardinalities
+knora-api:LinkObj
```
Nora-Olivia-Ammann, line 27: This way of targeting the resource is now consistent with how the project resources are targeted.

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -283,6 +312,25 @@ api-shapes:hasLinkTo_NodeShape
...
+  sh:path knora-api:isRegionOf ;
+  sh:node api-shapes:LinkValue_ClassShape, api-shapes:isRegionOf_NodeShape .
+
+api-shapes:isRegionOf_NodeShape
+  a           sh:NodeShape ;
+  sh:name     "This ensures that the target of the property is of type Representation, i.e. exists in the graph." ;
+  sh:property [
+                a          sh:PropertyShape ;
+                sh:class   knora-api:Representation ;
+                sh:message "http://api.knora.org/ontology/knora-api/v2#Representation" ;
```
Nora-Olivia-Ammann, line 328: Although this is an IRI which we use for user communication, `sh:message` requires a string as value.

seakayone (approved):

PR #1385, opened 2025-01-28
fix(validate-data): include LinkValueProperty in knora-subset and remove witespaces (DEV-4547)

Nora-Olivia-Ammann:


test/e2e_validate_data/test_validate_data.py
Nora-Olivia-Ammann, line 1: I added the systematic test project so that we do not forget the validation.
Otherwise I just moved the fixtures that are only used once into the functions. Maybe that speeds things up too...

src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -52,17 +52,16 @@ def validate_data(filepath: Path, api_url: str, dev_route: bool, save_graphs: bo
         save_graphs: if this flag is set, all the graphs will be saved in a folder
 
     Returns:
-        true unless it crashed
+        true if no errors were found otherwise false
```
Nora-Olivia-Ammann, line 55: This is for future use, because in this case returning always true is quite contra-intuitive.

src/dsp_tools/cli/call_action.py
```diff
@@ -212,9 +212,12 @@ def _call_xmlupload(args: argparse.Namespace) -> bool:
 
 
 def _call_validate_data(args: argparse.Namespace) -> bool:
-    return validate_data(
+    # The boolean return value of the function indicates if validation errors were found,
+    # not the success status of the command
```
jnussbaum, line 216: Well, but these 2 things should be the same: If errors were found, the success status of the command should be false (non-zero exit code). That's a convention with command line tools: If something went wrong, they should NOT return successfully.

I would recommend to revert this change.

jnussbaum (commented):
LGTM. I just have 1 concern (see comment), which we should clarify before merging.

src/dsp_tools/cli/call_action.py
Nora-Olivia-Ammann: Well the command was successful. Just that errors were found. I changed this because I wanted to integrate it into the xmlupload (longer story why I did not finish that). But to me then either the function `validate_data` should return `None` or a boolean that indicates if errors were found.

jnussbaum: Running a command line tool should result in 2 possible outcomes:

- return code 0 (on the level of a python function, this would translate to: `return True`)
    - this means that no errors were found
- return code non-0 (on the level of a python function, this would translate to: `return False`)
    - this means that errors were found,
    - or the ontology is inconsistent, 
    - or the provided file paths do not exist
    - or the XML is faulty

There is no point in saying that the command was successful, if errors were found. What would the meaning of "successful" be in that case? That there were no software bugs? Well, we should prevent that in the development phase!

Or what are the 3 states that you would like to measure with `True`, `False`, and `None`?

Nora-Olivia-Ammann: > There is no point in saying that the command was successful, if errors were found. What would the meaning of "successful" be in that case? 

But don't all our commands do that? They all finish with True because it ran to its conclusion?

jnussbaum: > But don't all our commands do that?

No, rather they track the success status, and then return it, e.g. here: https://github.com/dasch-swiss/dsp-tools/blob/0106a351f0f69ea250b974b63cf2f8b8e47740ea/src/dsp_tools/commands/project/create/project_create.py#L151

At the end, we evaluate if an error occurred or not, and emit a return code (1 or 0) based on it: https://github.com/dasch-swiss/dsp-tools/blob/0106a351f0f69ea250b974b63cf2f8b8e47740ea/src/dsp_tools/cli/entry_point.py#L85

Nora-Olivia-Ammann: Well I'll turn it back, but I am not convinced that if validation errors were found we should have a sys.exit(1)

jnussbaum: > I think displaying that an error occurred during the validation is confusing... because in fact no error occurred...

In my understanding, the only kind of error which can happen during a validation is a validation issue. I cannot think of another type of error. What are you concretely thinking about?

Nora-Olivia-Ammann: Well I don't consider erroneous data to be an error in this case...

jnussbaum: > Well I'll turn it back, but I am not convinced that if validation errors were found we should have a sys.exit(1)

We can also discuss this in person, if you prefer.

jnussbaum (approved):

src/dsp_tools/cli/call_action.py
Nora-Olivia-Ammann: Yes but I want to merge this fix in any case. We can discuss the rest later.

PR #1402, opened 2025-02-04
chore: `just clean` removes also built documentation site

jnussbaum:


justfile
```diff
@@ -86,10 +86,11 @@ clean:
...
     -find . -name .pytest_cache -exec rm -rf {} \;
     -find . -name .mypy_cache -exec rm -rf {} \;
-    -rm -rf *id2iri_mapping*.json
-    -rm -f warnings.log
-    -rm -rf testdata/e2e/tmp-dsp-ingest/
-    -rm -rf testdata/e2e/tmp-dsp-sipi/
-    -rm -rf testdata/e2e/ingest-db/
-    -rm -rf testdata/e2e/images/
-    -rm -f mapping-????.csv
+    -rm -rf ./*id2iri_mapping*.json
```
jnussbaum, line 89: the `./` makes sure that only the CWD is targeted. This doesn't matter in most cases, but it's always better to be explicit before deleting something.

Nora-Olivia-Ammann (approved):

PR #1406, opened 2025-02-05
refactor(validate-data): class structure of `InputProblem`

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/query_validation_result.py
Nora-Olivia-Ammann, line 1: The queries and the mapping into the new input error format absolutely needs refactoring.
I expect that we will be able to make it less complex and wordy. That is planned for my next PR. I did not want to make the diff too big and focus on the message generation changes.

seakayone (approved):

PR #1410, opened 2025-02-06
chore(optimise release process): deduce api commit hash from version tag

jnussbaum:


Nora-Olivia-Ammann (approved):

seakayone (approved):

src/dsp_tools/commands/start_stack.py
```diff
@@ -90,12 +89,16 @@ def _get_url_prefix(self) -> str:
...
 
+        docker_compose_pth = importlib.resources.files("dsp_tools").joinpath("resources/start-stack/docker-compose.yml")
+        docker_compose = yaml.safe_load(docker_compose_pth.read_bytes())
+        tag = docker_compose["services"]["api"]["image"].split(":")[-1]
+
+        response = requests.get(f"https://api.github.com/repos/dasch-swiss/dsp-api/git/ref/tags/{tag}", timeout=5)
+        commit_of_used_api_version = response.json()["object"]["sha"]
+
         if self.__stack_configuration.latest_dev_version:
             return f"{url_prefix_base}/main/"
```
BalduinLandolt, line 100: Nitpick: It would probably make sense to move this guard up to the beginning of the function. It does not depend on the result of the above lines which include file system and HTTP calls, so if the condition is true, it would make sense not to have to do this work.

src/dsp_tools/commands/start_stack.py
```diff
@@ -90,12 +89,16 @@ def _get_url_prefix(self) -> str:
         """
         url_prefix_base = "https://github.com/dasch-swiss/dsp-api/raw"
 
+        docker_compose_pth = importlib.resources.files("dsp_tools").joinpath("resources/start-stack/docker-compose.yml")
+        docker_compose = yaml.safe_load(docker_compose_pth.read_bytes())
+        tag = docker_compose["services"]["api"]["image"].split(":")[-1]
+
+        response = requests.get(f"https://api.github.com/repos/dasch-swiss/dsp-api/git/ref/tags/{tag}", timeout=5)
+        commit_of_used_api_version = response.json()["object"]["sha"]
```
BalduinLandolt, line 97: Question: Maybe I'm missing something here. But from my understanding, you're getting the SHA only to construct raw links.  
But GitHub allows raw links to be constructed from tags  
(e.g. `https://raw.githubusercontent.com/dasch-swiss/dsp-api/refs/tags/v31.4.0/webapi/scripts/fuseki-repository-config.ttl.template` )  
So couldn't you simply use the tag from the docker-compose.yml to construct the links directly?

BalduinLandolt (approved):
LGTM, see my comments, though

src/dsp_tools/commands/start_stack.py
jnussbaum: Absolutely

src/dsp_tools/commands/start_stack.py
jnussbaum: The one who missed something is me :-) Thanks, this simplifies a lot

jnussbaum:
@BalduinLandolt I have already updated https://www.notion.so/dasch-swiss/Release-and-Deployment-Process-v3-dfb5b5d1fbb84b3f8896aef945cc4b3d?pvs=4#1538946b7d40802cb56ec5540999d622 🥳

PR #1418, opened 2025-02-11
feat(excel2json): allow list names to contain colons (DEV-4580)

Nora-Olivia-Ammann:


test/unittests/commands/excel2json/test_properties.py
```diff
@@ -155,9 +156,9 @@ def test_get_gui_attribute(self) -> None:
         self.assertIsNone(e2j._get_gui_attribute(df_row=cast("pd.Series[Any]", original_df.loc[0, :]), row_num=2))
 
         res_1 = e2j._get_gui_attribute(df_row=cast("pd.Series[Any]", original_df.loc[1, :]), row_num=3)
-        assert isinstance(res_1, InvalidExcelContentProblem)
-        assert res_1.excel_position.row == 3
-        assert res_1.actual_content == "max:1.4 / min:1.2"
+        assert isinstance(res_1, GuiAttributes)
```
Nora-Olivia-Ammann, line 159: Before we were able to identify if something else is suspicious (here an invalid separator is used, we expect a comma. If we implement this feature then that check is no longer possible. They will notice the problem when the `create` fails. I think that is ok, becasue erronous data is not created on the server.

PR #1425, opened 2025-02-12
refactor(create): change structure of parsed project

Nora-Olivia-Ammann:
New:
- `ParsedProject` object contains metadata and the ontologies, they are no longer separate

Changes:
- Moved code around, no changes in content or logic


src/dsp_tools/commands/project/create/project_create.py
```diff
@@ -165,107 +165,8 @@ def create_project(
     return overall_success
 
 
-def _validate_and_parse_project_json(
```
Nora-Olivia-Ammann, line 168: Code moved to parsing function, no logic changes.

PR #1430, opened 2025-02-17
fix(xmllib): don't spam user with dsp-tools internal logging/printing (DEV-4595)

jnussbaum:


src/dsp_tools/utils/xml_validation.py
```diff
@@ -66,10 +66,6 @@ def validate_xml(xml: etree._Element) -> bool:
         logger.opt(exception=True).error(err_msg)
         raise InputError(err_msg)
 
-    info_str = "The XML file is syntactically correct."
-    logger.info(info_str)
```
jnussbaum, line 70: First, I had considered to leave this logging statement in place, and then suppress logging for xmllib only. That would've been the approach recommended by https://loguru.readthedocs.io/en/stable/api/logger.html#loguru._logger.Logger.disable. (See the first commit of this PR: https://github.com/dasch-swiss/dsp-tools/pull/1430/commits/f2a636bb7bf6f24573a0f8ea882dd8008cad1cfc).

But this didn't work, because `xmllib.__init__.py` is always imported, also when calling `dsp-tools xmlupload foo.xml` (or similar) from the Terminal. I think there is no way to distinguish if xmllib is imported from a user script, or from a dsp-tools internal file that was called by a Terminal command.

Therefore, the only solution I see is to remove this logging statement.

Nora-Olivia-Ammann (approved):

PR #1434, opened 2025-02-18
refactor(xmllib): move serialisation out of the classes

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/serialise/serialise_file_value.py
```diff
@@ -0,0 +1,43 @@
...
+    Returns:
+        File Value as etree
+    """
+    if isinstance(file_value, FileValue):
+        return _serialise_file_value(file_value, "bitstream")
+    elif isinstance(file_value, IIIFUri):
+        return _serialise_file_value(file_value, "iiif-uri")
+    else:
+        never: Never = cast(Never, file_value)
+        assert_never(never)
```
jnussbaum, line 32: This seems a bit odd. If you first cast the file_value to be Never, it will of course pass the next line. To me, it would seem more natural to raise an error here:

```suggestion
        raise BaseError("file_value must be either a FileValue or a IIIFUri, but you provided {class(file_value)}")
```

jnussbaum (approved):

src/dsp_tools/xmllib/serialise/serialise_file_value.py
Nora-Olivia-Ammann: That is also an option. I saw this in some code Balduin wrote and like the approach because it demonstrated that this code should actually not be reachable...

jnussbaum: I think it is the casting which seems odd to me. It seems more natural if you do it with a match case:

```
match file_value:
    case FileValue():
        return _serialise_file_value(file_value, "bitstream")
    case IIIFUri():
        return _serialise_file_value(file_value, "iiif-uri")
    case _:
        assert_never(file_value)
```

PR #1443, opened 2025-02-19
refactor(xmlupload): use `IntermediaryFileValue` for ingest upload

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/xmlupload.py
Nora-Olivia-Ammann, line 1: In preparation for licensing etc. I want to refactor this code

src/dsp_tools/commands/xmlupload/models/ingest.py
```diff
@@ -125,37 +117,31 @@ def _ingest(self, filepath: Path) -> IngestResponse:
...
-        """Returns the BitstreamInfo of the already ingested file based on the `XMLBitstream.value`."""
-        permissions = permissions_lookup.get(bitstream.permissions) if bitstream.permissions else None
-        return True, BitstreamInfo(bitstream.value, bitstream.value, permissions)
+    def get_bitstream_info(self, file_info: IntermediaryFileValue) -> IngestResult:
+        """Returns the BitstreamInfo of the already ingested file based on the `IntermediaryFileValue.value`."""
+        return IngestResult(True, BitstreamInfo(file_info.value, file_info.value, file_info.metadata.permissions))
+
+
+@dataclass
+class IngestResult:
```
jnussbaum, line 145: The class IngestResult seems redundant. The success status doesn't need to be stored separately, it can be deduced from media_info: if it is None, then it wasn't successful.

This redundancy was not introduced by you, it was already present in the code. But I wouldn't put 2 redundant infos into a class.

jnussbaum (approved):

PR #1449, opened 2025-02-20
chore: enable darglint only in `xmllib`

Nora-Olivia-Ammann:


justfile
```diff
@@ -30,11 +30,7 @@ mypy:
 # Check completeness and correctness of python docstrings
 [no-exit-message]
 darglint:
-    find . -name "*.py" \
-    -not -path "./src/dsp_tools/commands/project/models/*" \
-    -not -path "./.git/*" \
-    -not -path "./.venv/*" \
-    | xargs uv run darglint -v 2\
+    uv run darglint -v 2 ./src/dsp_tools/xmllib/*.py
```
jnussbaum, line 33: this selects only the files directly in the xmllib folder. We would probably want to include files in subfolders as well:

```suggestion
    uv run darglint -v 2 ./src/dsp_tools/xmllib/**/*.py
```

jnussbaum (approved):

PR #1456, opened 2025-02-21
refactor: reorganise XML cleaning functions

Nora-Olivia-Ammann:


src/dsp_tools/utils/xml_parsing/xml_schema_validation.py
```diff
@@ -30,7 +31,8 @@ def validate_xml_with_schema(xml: etree._Element) -> bool:
     Returns:
         True if the XML file is valid
     """
-    cleaned = remove_namespaces_from_xml(xml)
```
Nora-Olivia-Ammann, line 33: This used to combine the two, I consolidated it because it is not worth it to have another function

jnussbaum (approved):

PR #1467, opened 2025-02-24
refactor(xmlupload): create `Stash` from `IntermediaryResource`

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/stash/stash_circular_references_from_intermediary_resource.py
```diff
@@ -0,0 +1,83 @@
...
+    for val in resource.values.copy():
+        if isinstance(val, IntermediaryLink):
+            if val.value_uuid not in stash_lookup[resource.res_id]:
+                continue
+            stashed_link_values.append(_stash_link(val, resource.res_id, resource.type_iri))
+            resource.values.remove(val)
+        elif isinstance(val, IntermediaryRichtext):
+            if val.value_uuid not in stash_lookup[resource.res_id]:
+                continue
+            # value.value is a KnoraStandoffXml text with problematic links.
```
jnussbaum, line 50:
```suggestion
            # val.value is a KnoraStandoffXml text with problematic links.
```

src/dsp_tools/commands/xmlupload/stash/stash_circular_references_from_intermediary_resource.py
```diff
@@ -0,0 +1,83 @@
...
+        permission=str(value.permissions) if value.permissions else None,
+    )
+
+
+def _stash_standoff(value: IntermediaryRichtext, res_id: str, res_type: str) -> StandoffStashItem:
+    actual_text = value.value
+    # Replace the content with the UUID
+    value.value = FormattedTextValue(value.value_uuid)
+    # It is not necessary to add the permissions to the StandoffStashItem because it is an update request
+    # If no new permissions are given during that request, the permissions of the previous value are taken
```
jnussbaum, line 76:
```suggestion
    # It is not necessary to add the permissions to the StandoffStashItem.
    # Because when no new permissions are given during an update request,
    # the permissions of the previous value are taken.
```

test/unittests/commands/xmlupload/stash/test_stash_circular_references_from_intermediary_resource.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+@pytest.fixture
+def link_value_to_resource_no_links() -> IntermediaryLink:
+    return IntermediaryLink("res_no_links", "prop", None, None, str(uuid4()))
+
+
+@pytest.fixture
+def text_value_no_links() -> IntermediaryRichtext:
+    return IntermediaryRichtext(FormattedTextValue("No links"), "prop", None, None, set(), str(uuid4()))
```
jnussbaum, line 45: This is unused

```suggestion
```

test/unittests/commands/xmlupload/stash/test_stash_circular_references_from_intermediary_resource.py
```diff
@@ -0,0 +1,187 @@
...
+    stash_circular_references,
+)
+from dsp_tools.commands.xmlupload.stash.stash_models import LinkValueStash
+from dsp_tools.commands.xmlupload.stash.stash_models import StandoffStash
+from dsp_tools.commands.xmlupload.stash.stash_models import Stash
+
+
+@pytest.fixture
+def simple_text_value() -> IntermediarySimpleText:
+    return IntermediarySimpleText("Text", "prop", None, None)
```
jnussbaum, line 23: One thought concerning these fixtures: 

- With 1 exception, they are only used once. 
- Most of them are 1-liners
- That means, most of them could be inlined.

Inlining them would greatly improve the [Locality of Behaviour](https://htmx.org/essays/locality-of-behaviour/). As the linked essay states, LoB might be in conflict with other principles. So it's up to the developers to find a trade-off between conflicting principles.

In this concrete case, I would find it easier if no fixtures are used. But I leave that decision up to you :-)

test/unittests/commands/xmlupload/stash/test_stash_circular_references_from_intermediary_resource.py
```diff
@@ -0,0 +1,187 @@
...
+    resource_1: IntermediaryResource,
+    resource_2: IntermediaryResource,
+    resource_3: IntermediaryResource,
+    resource_no_links: IntermediaryResource,
+) -> None:
+    copied_res_1 = deepcopy(resource_1)
+    copied_res_2 = deepcopy(resource_2)
+    copied_res_3 = deepcopy(resource_3)
+    resources = [copied_res_1, copied_res_2, copied_res_3, deepcopy(resource_no_links)]
+    lookup = {"res_1": ["link_to_res_2_uuid"]}
```
jnussbaum, line 118: Not part of this PR, but it struck my mind: When looking at this dict, it might be difficult to understand its purpose. One sees resource identifiers as keys, and then UUIDs as values. But what are the meaning of the UUIDs? 

I think it might be worth to make a class with some informative field names.

jnussbaum (approved):

PR #1472, opened 2025-02-25
fix(excel2xml): filter out blanks-only strings from create_json_excel_list_mapping()

jnussbaum:


src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -2097,7 +2097,7 @@ def create_json_excel_list_mapping(
     excel_values_new = []
     for val in excel_values:
         if isinstance(val, str):
-            excel_values_new.extend([x.strip() for x in val.split(sep) if x])
+            excel_values_new.extend([x.strip() for x in val.split(sep) if x.strip()])
```
Nora-Olivia-Ammann, line 2100: Even shorter:
`[found for x in val.split(sep) if (found := x.strip())]`

Nora-Olivia-Ammann (approved):

PR #1473, opened 2025-02-25
feat(xmlupload): parse legal info for files from XML (DEV-4635)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/prepare_xml_input/transform_into_intermediary_classes.py
```diff
@@ -121,18 +124,51 @@ def _transform_migration_metadata(resource: XMLResource) -> MigrationMetadata:
...
+        "http://rdfh.ch/licenses/cc-by-nc-sa-4.0",
+        "http://rdfh.ch/licenses/cc-by-nd-4.0",
+        "http://rdfh.ch/licenses/cc-by-nc-nd-4.0",
+        "http://rdfh.ch/licenses/ai-generated",
+        "http://rdfh.ch/licenses/unknown",
+        "http://rdfh.ch/licenses/public-domain",
+    ]
+    if file_metadata.license_ and file_metadata.license_ not in predefined_licenses:
+        raise InputError(
+            f"The license '{file_metadata.license_}' used for an image or iiif-uri is unknown. "
```
Nora-Olivia-Ammann, line 153: The resource id needed to identify the file is added when handling the error.

src/dsp_tools/commands/xmlupload/prepare_xml_input/transform_into_intermediary_classes.py
```diff
@@ -121,18 +124,51 @@ def _transform_migration_metadata(resource: XMLResource) -> MigrationMetadata:
...
+        "http://rdfh.ch/licenses/cc-by-4.0",
+        "http://rdfh.ch/licenses/cc-by-sa-4.0",
+        "http://rdfh.ch/licenses/cc-by-nc-4.0",
+        "http://rdfh.ch/licenses/cc-by-nc-sa-4.0",
+        "http://rdfh.ch/licenses/cc-by-nd-4.0",
+        "http://rdfh.ch/licenses/cc-by-nc-nd-4.0",
+        "http://rdfh.ch/licenses/ai-generated",
+        "http://rdfh.ch/licenses/unknown",
+        "http://rdfh.ch/licenses/public-domain",
+    ]
```
jnussbaum, line 150: Probably it would be a better idea to move these already existing constants from `xmllib/models/config_options.py` into a repo-wide constants file, so that you can import them from there.

jnussbaum (approved):

src/dsp_tools/commands/xmlupload/models/deserialise/deserialise_value.py
```diff
@@ -222,15 +222,33 @@ class IIIFUriInfo:
...
         """Factory that parses an IIIF URI node from the XML DOM"""
         if not node.text:
             raise XmlUploadError("Empty IIIF URI tag")
-        return IIIFUriInfo(node.text.strip(), node.get("permissions"))
+        return IIIFUriInfo(node.text.strip(), get_file_metadata(node))
+
+
+@dataclass
+class XMLFileMetadata:
+    license_: str | None
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: It is an existing keyword

src/dsp_tools/commands/xmlupload/prepare_xml_input/transform_into_intermediary_classes.py
Nora-Olivia-Ammann: I think then we will have again a problem with the documentation visibility in the xmllib. I don't think it is that critical here, because these will never change. We also have a lot of knora properties hardcoded at several places. I think this is an optimisation we can do in the future if needed.

src/dsp_tools/commands/xmlupload/models/deserialise/deserialise_value.py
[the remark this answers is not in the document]
jnussbaum: Interesting, didn't know that. Thanks 👍

PR #1479, opened 2025-02-26
refactor: move generic and static logging functions out from `ConnectionLive` to separate file

Nora-Olivia-Ammann:


src/dsp_tools/utils/connection_live.py
```diff
@@ -246,37 +247,14 @@ def _renew_session(self) -> None:
...
-        dumpobj: dict[str, Any] = {
-            "status_code": response.status_code,
-            "headers": _sanitize_headers(dict(response.headers)),
-        }
-        try:
-            dumpobj["content"] = response.json()
-        except JSONDecodeError:
-            dumpobj["content"] = response.text
-        logger.debug(f"RESPONSE: {json.dumps(dumpobj)}")
+        log_response(response)
```
Nora-Olivia-Ammann, line 251: I we do not create a class method, then we cannot test that it was called with these parameters in the unittest. However, I want to use the logging message also in the planned `LegalInfoClient` so that the string is consistent. This is why I created this wrapper.

jnussbaum: It is well possible to test this as standalone function, for this we don't need a class method. I have tried it out here: https://github.com/dasch-swiss/dsp-tools/pull/1481.

I find my solution cleaner, even if it's more cumbersome to do the mocking. WDYT?

jnussbaum (approved):
I generally like this PR, there's just 1 thing I'm a little unhappy with (see my separate PR).

PR #1481, opened 2025-02-26
improve mock testing

jnussbaum:


test/integration/commands/xmlupload/test_resource_creation.py
```diff
@@ -471,7 +471,6 @@ def test_interruption_if_resource_cannot_be_created_because_of_404() -> None:
     )
     con = ConnectionLive("foo")
     con._log_request = Mock()  # type: ignore[method-assign]
-    con._log_response = Mock()  # type: ignore[method-assign]
```
jnussbaum, line 474: funnily, this line was useless

test/unittests/utils/test_connection_live.py
```diff
@@ -299,23 +299,6 @@ def test_log_request() -> None:
         debug_mock.assert_called_once_with(f"REQUEST: {json.dumps(expected_output)}")
 
 
-def test_log_response() -> None:
```
jnussbaum, line 302: this function shouldn't live in this file

PR #1484, opened 2025-02-27
feat(xmlupload): Upload the legal info of the file values during the xmlupload (DEV-4638)

Nora-Olivia-Ammann:


jnussbaum (approved):
PR title: We were once told to only use "feat" for user-facing new features. If that still holds true, then this PR would rather be a refactor or chore.

Nora-Olivia-Ammann:
> PR title: We were once told to only use "feat" for user-facing new features. If that still holds true, then this PR would rather be a refactor or chore.

This is absolutely user facing.

jnussbaum:
> > PR title: We were once told to only use "feat" for user-facing new features. If that still holds true, then this PR would rather be a refactor or chore.
> 
> This is absolutely user facing.

Then I would reformulate the title in a way that the user understands what changes.

PR #1485, opened 2025-02-27
feat(xmlupload): integrate create copyright holders into xmlupload (DEV-4650)

Nora-Olivia-Ammann:


testdata/xml-data/test-data-e2e.xml
```diff
@@ -6,10 +6,17 @@
...
+    </authorship>
+
     <resource label="Resource 1"
               restype=":ImageResource"
               id="resource_1">
-        <bitstream>testdata/bitstreams/test.jpg</bitstream>
+        <bitstream license="http://rdfh.ch/licenses/cc-by-4.0"
+                   copyright-holder="DaSCH"
+                   authorship-id="authorship_1"
+        >testdata/bitstreams/test.jpg</bitstream>
```
Nora-Olivia-Ammann, line 19: I left out the legal info for the other bitstreams so that we can ensure that the upload also works without the information. As soon as it is mandatory we will change it.

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -93,17 +95,20 @@ def _get_live_clients(
     con: Connection,
     auth: AuthenticationClient,
     creds: ServerCredentials,
+    server: str,
```
jnussbaum, line 98: this is already contained in ConnectionLive

test/e2e/test_create_and_xmlupload.py
```diff
@@ -66,6 +66,19 @@ def test_xmlupload(auth_header: dict[str, str], project_iri: str) -> None:
...
 
+@pytest.mark.usefixtures("_xmlupload")
+def test_all_copyright_holders(auth_header: dict[str, str]) -> None:
+    response = _get_copyright_holders(auth_header)
+    assert response["data"] == ["DaSCH"]
+
+
+def _get_copyright_holders(auth_header: dict[str, str]) -> dict[str, Any]:
+    url = f"{CREDS.server}/admin/projects/shortcode/4125/legal-info/copyright-holders?page=1&page-size=25&order=Asc"
+    headers = auth_header | {"accept": "application/json"}
```
jnussbaum, line 77:
```suggestion
    headers = auth_header | {"Accept": "application/json"}
```

jnussbaum (approved):

src/dsp_tools/commands/xmlupload/xmlupload.py
Nora-Olivia-Ammann: Yes but it is not typed in the `Connection`, I first added it as a parameter there too but then got problems at other places. But I took it now from creds.

PR #1487, opened 2025-02-28
test(validate-data): add legal info to test data

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -372,14 +376,18 @@ def _reformat_one_validation_result(validation_result: ValidationResult) -> Inpu
...
             | ViolationType.PATTERN
             | ViolationType.UNIQUE_VALUE
             | ViolationType.VALUE_TYPE as violation
         ):
             problem = RESULT_TO_PROBLEM_MAPPER[violation]
             return _reformat_generic(result=validation_result, problem_type=problem)
+        case ViolationType.GENERIC:
+            prop_str = None
+            if validation_result.property in FILE_VALUE_PROPERTIES:
+                prop_str = "bitstream / iiif-uri"
```
Nora-Olivia-Ammann, line 389: If we don't overwrite it here, the users sees the knora property (eg. hasArchiveFileValue)

src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -298,7 +300,9 @@ def _query_pattern_constraint_component_violation(
 
 def _query_generic_violation(base_info: ValidationResultBaseInfo, results_and_onto: Graph) -> ValidationResult:
     detail_info = cast(DetailBaseInfo, base_info.detail)
-    val = next(results_and_onto.objects(detail_info.detail_bn, SH.value))
+    val = None
```
Nora-Olivia-Ammann, line 303: As opposed to the previous generic results, they all had a value, this does not as the problem is that the value is missing.

src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -238,6 +238,8 @@ def _query_one_with_detail(
     detail_info = cast(DetailBaseInfo, base_info.detail)
     match detail_info.source_constraint_component:
         case SH.MinCountConstraintComponent:
+            if base_info.result_path in FILE_VALUE_PROPERTIES:
+                return _query_generic_violation(base_info, results_and_onto)
             return _query_for_value_type_violation(base_info, results_and_onto, data_graph)
```
Nora-Olivia-Ammann, line 243: An ontology property that is missing produces a different validation result. The only other MinCountConstraintComponent that reaches this code is the wrong text ecoding, which is a value type error. To differentiate the two it is easiest to check if the property is a file value property.

src/dsp_tools/resources/validate_data/api-shapes-with-cardinalities.ttl
```diff
@@ -344,7 +383,7 @@ api-shapes:hasStillImageFileValue_PropShape
               A StillImageRepresentation requires an external IIIF-URI or a file with one of the extensions:
               'jpg', 'jpeg', 'png', 'tif', 'tiff', 'jp2'.
               """ ;
-  sh:or       ( [ sh:node api-shapes:StillImageFileValue_ClassShape ] [ sh:node api-shapes:StillImageExternalFileValue_ClassShape ] ) .
+  sh:xone     ( [ sh:node api-shapes:StillImageFileValue_ClassShape ] [ sh:node api-shapes:StillImageExternalFileValue_ClassShape ] ) .
```
Nora-Olivia-Ammann, line 386: `sh:or` was actually a bad choice, because it says at least one shape must fit. `sh:xone` means exactly one of those shapes apply.

src/dsp_tools/resources/validate_data/api-shapes-with-cardinalities.ttl
```diff
@@ -368,7 +407,7 @@ api-shapes:StillImageFileValue_ClassShape
                             'jpg', 'jpeg', 'png', 'tif', 'tiff', 'jp2'.
                             """ ;
               ] ,
-              api-shapes:hasPermissions_Cardinality;
+              api-shapes:hasPermissions_Cardinality ;
```
Nora-Olivia-Ammann, line 410: StillImageFileValues will be implemented separately because they lead to a `sh:xone` violation. This will cause further changes in the parsing of the violation results

testdata/validate-data/generic/every_combination_once.xml
```diff
@@ -21,14 +20,12 @@
 
 
         <!--
-        Our Violation Type            ContentRegexViolation
```
Nora-Olivia-Ammann, line 24: Every time I refactor our violation type I have to change this, this is in fact outdated and I forgot about it the last time.
However the information can also be found in the e2e test: `test_reformat_every_constraint_once` Therefore I think it makes sense to remove this here. The SHACL components come from the validation and are stable therefore this will not be outdated if we change our code.

src/dsp_tools/resources/validate_data/api-shapes-with-cardinalities.ttl
```diff
@@ -195,6 +195,33 @@ knora-api:VideoSegment
 # RESOURCES WITH FILES
 #########################################
 
+### Legal Info Properties
+api-shapes:hasLicense_PropShape
+  a sh:PropertyShape ;
+  sh:path knora-api:hasLicense ;
+  sh:minCount 1 ;
+  sh:maxCount 1 ;
+  sh:message "Files and IIIF-URIs require a reference to a license." ;
```
Nora-Olivia-Ammann, line 204: This is the user facing message whenever the information is missing.

testdata/validate-data/generic/dsp_inbuilt_correct.xml
```diff
@@ -14,21 +14,34 @@
         <allow group="ProjectAdmin">CR</allow>
     </permissions>
 
+    <authorship id="authorship_1">
```
Nora-Olivia-Ammann, line 17: I forgot to add the data here in the last PR, validate found the missing info!

jnussbaum: cool!

testdata/validate-data/generic/every_combination_once.xml
```diff
@@ -6,10 +6,20 @@
        shortcode="9999"
        default-ontology="onto">
 
+    <authorship id="authorship_1">
+        <author>Author One</author>
+    </authorship>
+
+    <!-- Target class that does not produce a violation -->
```
jnussbaum, line 13: why Target class?

jnussbaum (approved):
I trust you

testdata/validate-data/generic/every_combination_once.xml
Nora-Olivia-Ammann: Target class means that the resource itself does not violate anything. I prefer to have the classes that I must reference (for example for segment I need a video), to be a resource that is only used to reference but not produce an error itself.

PR #1490, opened 2025-03-03
feat(validate-data): validate if legal info exists for images (DEV-4655)

Nora-Olivia-Ammann:


testdata/validate-data/generic/dsp_inbuilt_correct.xml
Nora-Olivia-Ammann, line 1: Forgot to add the info previously. This file should not raise an error

test/e2e/test_validate_data.py
```diff
@@ -263,12 +266,13 @@ def test_inheritance_correct(self, api_con: ApiConnection, shacl_validator: Shac
     def test_inheritance_violation(self, inheritance_violation: ValidationReportGraphs) -> None:
         assert not inheritance_violation.conforms
 
-    @pytest.mark.usefixtures("_create_projects")
-    def test_systematic_correct(self, api_con: ApiConnection, shacl_validator: ShaclValidator) -> None:
-        file = Path("testdata/xml-data/test-data-systematic.xml")
-        graphs = _get_parsed_graphs(api_con, file)
-        systematic_correct = _get_validation_result(graphs, shacl_validator, None)
-        assert systematic_correct.conforms
+    #
```
Nora-Olivia-Ammann, line 269: Since legal info is not yet mandatory this test project purposefully contains files without it. For simplicity's sake I am already making it mandatory in SHACL, it is not stopping the xmlupload. Until we adjust this data set, this test will always fail. I created a ticket in the project to reactivate it once it is mandatory.

seakayone (approved):

seakayone:
@Nora-Olivia-Ammann the PR looks good to me.

The API allows to attach LegalInfo to any subclass of `FileValue`, this PR only does the check for `StillImageFileValue`, do you plan on supporting the other as well?

Nora-Olivia-Ammann:
> @Nora-Olivia-Ammann the PR looks good to me.
> 
> The API allows to attach LegalInfo to any subclass of `FileValue`, this PR only does the check for `StillImageFileValue`, do you plan on supporting the other as well?

It is actually the other way around. I did not realise that I didn't tell you:
https://github.com/dasch-swiss/dsp-tools/commit/209a1d19b818d0bd835691c0f733b0ac1c6662ad

The reason is because of the `xone` constraint for images, the modelling of the shapes hast to be different. So I made it in two steps.

PR #1491, opened 2025-03-03
chore: bump stack to 2025.03.01

siers:


jnussbaum (changes requested):
Line 31 also needs to be updated

src/dsp_tools/resources/start-stack/docker-compose.yml
```diff
@@ -28,7 +28,7 @@ services:
 
   sipi:
     # on the verge of every deployment (fortnightly), take the same tag as DSP-API
-    image: daschswiss/knora-sipi:v31.7.0
+    image: daschswiss/knora-sipi:v31.8.0-1-g9805e5d
```
jnussbaum, line 31: Why a dirty commit? Usually we take the same as "api":

```suggestion
    image: daschswiss/knora-sipi:v31.8.0
```

siers: Ah, I wasn't sure where that came from, so I opted for the last just to be sure. Yes, let's take the stable.

jnussbaum (approved):

PR #1492, opened 2025-03-03
feat(validate-data): verify that the used license is allowed (DEV-4659)

Nora-Olivia-Ammann:


src/dsp_tools/resources/validate_data/api-shapes-with-cardinalities.ttl
```diff
@@ -238,27 +238,27 @@ api-shapes:hasArchiveFileValue_PropShape
...
-                            """ ;
-              ] ,
-              api-shapes:hasPermissions_Cardinality ,
-              api-shapes:hasLicense_PropShape ,
-              api-shapes:hasAuthorship_PropShape ,
-              api-shapes:hasCopyrightHolder_PropShape ;
+  a              sh:NodeShape ;
+  sh:name        "Validates the class type" ;
+  sh:message     "ArchiveFileValue" ;
+  sh:targetClass knora-api:ArchiveFileValue ;
```
Nora-Olivia-Ammann, line 244: Using `sh:class` is actually not correct: https://www.w3.org/TR/shacl/#ClassConstraintComponent

I am not sure why I did this.

seakayone (approved):

PR #1494, opened 2025-03-04
refactor(validate-data): harmonise file value shapes

Nora-Olivia-Ammann:


test/e2e/test_validate_data.py
```diff
@@ -168,6 +168,7 @@ def test_extract_identifiers_of_resource_results(every_combination_once: Validat
         (URIRef("http://data/image_no_legal_info"), None),
         (URIRef("http://data/image_no_legal_info"), None),
         (URIRef("http://data/image_no_legal_info"), None),
+        (URIRef("http://data/image_no_legal_info"), None),
```
Nora-Olivia-Ammann, line 171: This extra line is because the StillImageRepresentation also causes an `xoneConstraint`. We filter this out for the user message as all it says is that the file value did not conform to either shape. The exact error information we get from the shapes directly.

PR #1499, opened 2025-03-05
fix(validate-data): correct the way Values are targeted in the SHACL shapes (DEV-4672)

Nora-Olivia-Ammann:
Erroneously sh:class (https://www.w3.org/TR/shacl/#ClassConstraintComponent) was used in some cases to declare which class should be targeted by a shape. The correct property to target a class is sh:targetClass. However, we do not explicitly target the classes in case of Values, because the simple and richtext value look very different. Also not every link and list value looks the same because of user defined objects. To make it consistent we remove the targeting of a specific class from all values. In the query that defines which property has which value class, we declare the sh:node , there we reference the correct value and the restrictions on that class apply through that. If we did both, we would get duplicate errors in some cases.


src/dsp_tools/commands/validate_data/sparql/value_shacl.py
```diff
@@ -74,6 +76,87 @@ def _add_property_shapes_to_class_shapes(onto: Graph) -> Graph:
     return Graph()
 
 
+def _construct_value_type_shapes_to_class_shapes(onto: Graph) -> Graph:
```
Nora-Olivia-Ammann, line 79: The previous way to validate value types worked, but as explained in the ticket the approach was not correct. This is the correct approach to check that the target class is actually the correct value type.

src/dsp_tools/commands/validate_data/sparql/value_shacl.py
```diff
@@ -74,6 +76,87 @@ def _add_property_shapes_to_class_shapes(onto: Graph) -> Graph:
...
+          knora-api:canBeInstantiated true ;
+          rdfs:subClassOf ?restriction .
+      ?restriction a owl:Restriction ;          
+          owl:onProperty ?propRestriction .
+
+      ?propRestriction knora-api:isEditable true ;
+                       knora-api:objectType ?objectType .
+
+      FILTER NOT EXISTS { ?propRestriction knora-api:objectType knora-api:TextValue }
+      FILTER NOT EXISTS { ?propRestriction knora-api:isLinkProperty true }
```
Nora-Olivia-Ammann, line 111: The object type for a link property is the resource class that should be used, not the LinkValue class. Because of this we filter it out.

src/dsp_tools/commands/validate_data/sparql/value_shacl.py
```diff
@@ -74,6 +76,87 @@ def _add_property_shapes_to_class_shapes(onto: Graph) -> Graph:
...
+    PREFIX knora-api:  <http://api.knora.org/ontology/knora-api/v2#>
+
+    CONSTRUCT {
+
+      ?class a sh:NodeShape ;
+             sh:property [
+                            a sh:PropertyShape ;
+                            sh:path ?propRestriction ;
+                            sh:class knora-api:LinkValue ;
+                            sh:message "This property requires a LinkValue" ;
```
Nora-Olivia-Ammann, line 138: Since they are filtered out we add them here

testdata/validate-data/knora-api-subset.ttl
```diff
@@ -70,3 +70,36 @@ knora-api:VideoSegment
   rdfs:label                  "VideoSegment" ;
   knora-api:canBeInstantiated true ;
   knora-api:isResourceClass   true .
+
+
+knora-api:ArchiveFileValue
```
Nora-Olivia-Ammann, line 75: We now need this information for the tests

PR #1500, opened 2025-03-05
chore(xmlupload): make error handling clearer/simpler

jnussbaum:


src/dsp_tools/commands/xmlupload/models/ingest.py
```diff
@@ -109,20 +107,15 @@ def _ingest(self, filepath: Path) -> IngestResponse:
                 elif res.status_code == STATUS_UNAUTHORIZED:
                     raise BadCredentialsError("Bad credentials")
                 else:
-                    user_msg = f"{err} See {WARNINGS_SAVEPATH} for more information."
-                    print(user_msg)
```
jnussbaum, line 113: printing interfers with the progress bar. The error message can be retrieved from the logs and from the `warnings.log`

src/dsp_tools/commands/xmlupload/models/ingest.py
```diff
@@ -109,20 +107,15 @@ def _ingest(self, filepath: Path) -> IngestResponse:
...
-                    log_msg = f"{err}. Response status code {res.status_code} '{res.text}'"
-                    logger.error(log_msg)
-                    raise PermanentConnectionError(log_msg)
+                    raise PermanentConnectionError()
             except requests.exceptions.RequestException as e:
-                raise PermanentConnectionError(f"{err}. {e}") from e
+                raise PermanentConnectionError(str(e)) from e
```
Nora-Olivia-Ammann, line 112: Will this not be empty because you did not hand over a string when you raise the exception?

jnussbaum: The string (message) inside the error class doesn't really matter here. The caller catches this exceptions and logs it (with stack trace). Nothing is done with the message. I have now removed the message altogether: https://github.com/dasch-swiss/dsp-tools/commit/b365bf081df5e577a271355b970bf334f8dde38b

Nora-Olivia-Ammann (approved):

PR #1505, opened 2025-03-05
refactor: _handle_non_ok_responses()

jnussbaum:


src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -310,6 +310,7 @@ def _handle_permanent_connection_error(err: PermanentConnectionError) -> Never:
     msg = "Lost connection to DSP server, probably because the server is down. "
     msg += f"Please continue later with 'resume-xmlupload'. Reason for this failure: {err.message}"
     logger.error(msg)
+    msg += f"\nSee {WARNINGS_SAVEPATH} for more information."
```
jnussbaum, line 313: Previously, this was further down in the call hierarchy (in the connection_live.py), and was only visible in the log file and warnings.log file, which is of course useless.

I identified this place as being more appropriate: Here, it is printed to the Terminal, which is useful.

src/dsp_tools/utils/connection_live.py
```diff
@@ -198,14 +199,16 @@ def _handle_non_ok_responses(self, response: Response, retry_counter: int) -> No
...
+        if blame == "client":
+            raise InvalidInputError(msg)
         else:
-            msg = "Permanently unable to execute the network action. "
-            if original_str := regex.search(r'{"knora-api:error":"dsp\.errors\.(.*)","@context', str(response.content)):
-                msg += f"\n{' ' * 37}Original Message: {original_str.group(1)}\n"
-                if original_str.group(1).startswith("OntologyConstraintException"):
-                    msg += f"See {WARNINGS_SAVEPATH} for more information."
-                    raise InvalidInputError(msg)
-            msg += f"See {WARNINGS_SAVEPATH} for more information."
```
jnussbaum, line 208: This is an attempt to make the code easier, so that I can do further work on it (concretely: https://linear.app/dasch/issue/DEV-4631/improvement-of-error-message-when-api-returns-notfoundexception).

Please tell me if you feel it was easier before...

Nora-Olivia-Ammann (approved):

PR #1509, opened 2025-03-06
chore: add some just linting commands

jnussbaum:


justfile
```diff
@@ -3,6 +3,28 @@ default:
...
+[no-exit-message]
+format:
+    ruff format .
+    yamlfmt .
+
+
+# Run all linters
+[no-exit-message]
+lint:
+    #!/usr/bin/env -S parallel --shebang --ungroup --jobs {{ num_cpus() }}
```
jnussbaum, line 16: This parallelizes the below commands. The [just docs](https://just.systems/man/en/running-tasks-in-parallel.html?highlight=parallel#running-tasks-in-parallel) recommend to use the lesser-known [GNU parallel](https://savannah.gnu.org/projects/parallel/) over the more portable xargs. Probably because xargs would interweave the outputs of the different commands.

Therefore, make sure to install `brew install parallel`

justfile
```diff
@@ -3,6 +3,28 @@ default:
...
+    just ruff-check
+    just ruff-format-check
+    yamlfmt -lint .
+    yamllint .
+    just markdownlint
+    just darglint
+    just mypy
+    uv run mkdocs build --strict
+    just check-links
```
jnussbaum, line 25: Not sure if `check-links` should be included. It prints the success status of every single link to the output (currently 276). It cannot be configured to print less.

Nora-Olivia-Ammann (approved):

seakayone (approved):

PR #1512, opened 2025-03-06
chore: reorganise justfile

jnussbaum:


.github/workflows/tests-on-push.yml
```diff
@@ -37,31 +37,22 @@ jobs:
       - name: Install uv, just, and dependencies
         uses: ./.github/actions/setup
       - name: Linting with ruff
-        run: just ruff-check-github
+        run: just ruff-check --output-format=github
```
jnussbaum, line 40: The extra just command `just ruff-check-github`  was only ever used in the GitHub CI, so I thought it's not worth having a extra command for it.

Nora-Olivia-Ammann (approved):

PR #1513, opened 2025-03-06
refactor(validate-data): add property shapes to Values

Nora-Olivia-Ammann:


src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -69,55 +69,94 @@ api-shapes:rdfsLabel_Shape
 ##############
 # BooleanValue
 api-shapes:BooleanValue_ClassShape
-  a           sh:NodeShape ;
-  sh:name     "Validates the class type" ;
-  sh:message  "BooleanValue" ;
-  sh:property api-shapes:valueHasComment_Shape ;
+  a              sh:NodeShape ;
+  # sh:targetClass knora-api:BooleanValue ;
```
Nora-Olivia-Ammann, line 73: I will activate this in the next PR

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -69,55 +69,94 @@ api-shapes:rdfsLabel_Shape
...
+  sh:name        "Validates the properties used with this targetClass" ;
```
Nora-Olivia-Ammann, line 74: This name feels more precise to me and clear what it exactly does.

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -69,55 +69,94 @@ api-shapes:rdfsLabel_Shape
...
-  sh:property api-shapes:valueHasComment_Shape ;
+  a              sh:NodeShape ;
+  # sh:targetClass knora-api:DateValue ;
+  sh:name        "Validates the properties used with this targetClass" ;
+  sh:message     "DateValue" ;
+  sh:property    api-shapes:valueAsString_DateValueShape ,
+                 api-shapes:valueHasComment_Shape ;
+  sh:severity    sh:Violation .
+
+# Currently it is only checked that the string value is not empty
```
Nora-Olivia-Ammann, line 121: I tried to add some explanations, if you would like any more clarifications please do write a comment.

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -69,55 +69,95 @@ api-shapes:rdfsLabel_Shape
...
 
 ##############
 # GeonameValue
 api-shapes:GeonameValue_ClassShape
-  a           sh:NodeShape ;
-  sh:name     "Validates the class type" ;
-  sh:message  "GeonameValue" ;
-  sh:property api-shapes:geonameValueAsGeonameCode_Shape, api-shapes:valueHasComment_Shape ;
-  sh:severity sh:Violation .
+  a              sh:NodeShape ;
```
Nora-Olivia-Ammann, line 154: I also did some formatting and github seems to be bad in recognizing the diff there...

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -163,65 +206,97 @@ api-shapes:IntervalValue_ClassShape
...
+  # sh:targetClass knora-api:ListValue ;
+  sh:name        "Validates the metadata properties used with this targetClass" ;
+  sh:message     "ListValue" ;
+  sh:property    api-shapes:valueHasComment_Shape ;
+  sh:severity    sh:Violation .
 
 ##############
 # TextValue
 
+# This validates the metadata of the TextValue
```
Nora-Olivia-Ammann, line 278: Since the metadata properties are the same we evaluate those together.

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -69,55 +69,95 @@ api-shapes:rdfsLabel_Shape
...
+  sh:property    api-shapes:booleanValueAsBoolean_Shape ,
+                 api-shapes:valueHasComment_Shape ;
+  sh:severity    sh:Violation .
+
+api-shapes:booleanValueAsBoolean_Shape
+  a           sh:PropertyShape ;
+  sh:message  "A BooleanValue requires exactly one valid boolean" ;
+  sh:path     knora-api:booleanValueAsBoolean ;
+  sh:datatype xsd:boolean ;
+  # sh:minCount 1 ;
```
Nora-Olivia-Ammann, line 85: Right now the `api-shapes:BooleanValue_ClassShape` is connected through the SPARQL.
Through the SPARQL we evaluate that the target must be a boolean. This shape is connected to the property that should have a boolean. We raise an error when the class type is wrong. With the cardinality check we now also raise an error if this property is missing. Since the class is not infact a boolean it is impossible for it to have a boolean. But because the shape is connected it is "activated". Once we no longer connect the `api-shapes:BooleanValue_ClassShape` through the SPARQL query. If the target class is wrong, only the error checking the class type is raised. Right now we get two errors, as soon as I target the classes directly in the ttl and no longer through the SPARQL query we will only get one error. So it is not worthwhile to filter that out for the time between this and the next PR.

seakayone: Makes sense, so if I understand correcty you will uncomment the `sh:[min|max]Count` once we are able to use the `sh:targetClass`.

seakayone (approved):

src/dsp_tools/resources/validate_data/api-shapes.ttl
Nora-Olivia-Ammann: Yes!

PR #1515, opened 2025-03-06
docs: communicate supported Python versions (DEV-4680)

jnussbaum:


README.md
```diff
@@ -1,9 +1,9 @@
-[![](https://img.shields.io/pypi/v/dsp-tools.svg)](https://pypi.org/project/dsp-tools/) 
-[![](https://img.shields.io/pypi/l/dsp-tools.svg)](https://pypi.org/project/dsp-tools/) 
+[![DSP-TOOLS version on PyPI](https://img.shields.io/pypi/v/dsp-tools.svg)](https://pypi.org/project/dsp-tools/) 
+[![Supported Python versions](https://img.shields.io/pypi/pyversions/dsp-tools.svg)](https://pypi.org/project/dsp-tools/)
```
jnussbaum, line 2: I thought it would be nice to display the supported python versions in the badges on top. For this, pyproject.toml must contain the appropriate versions.

Then, I slightly rearranged our badges:
- added an explanatory alt text for all images (mainly for us)
- got rid of 2 badges (markdownlint and markdown-link-validator), because I think they are less important. It's better if the other (more important ones) stay out.

docs/index.md
```diff
@@ -24,6 +24,13 @@ To update to the latest version run:
 pip3 install --upgrade dsp-tools
 ```
 
+!!! warning
+
+    If your Python version is older than ours,
```
Nora-Olivia-Ammann, line 29: It would be nice if you could tell them where they can find our version... otherwise they do not know to which one they should update.

Nora-Olivia-Ammann (approved):

docs/index.md
jnussbaum: what do you think about how it is now? As soon as this PR is merged, it should display 3.12 | 3.13 instead of 3:

Screenshot 2025-03-06 at 20 01 56

Nora-Olivia-Ammann: Looks nice!

PR #1517, opened 2025-03-06
test: fix diverse warnings

jnussbaum:


test/unittests/xmllib/serialise/test_serialise_file_value.py
```diff
@@ -74,28 +74,28 @@ def test_serialise_file_value_bitstream_with_comment(metadata_no_permissions: Me
...
-    val = IIIFUri("https://link.org", metadata_no_permissions, None)
+    val = IIIFUri("https://example.org/image.jpg/full/1338%2C/0/default.jpg", metadata_no_permissions, None)
     result = serialise_file_value(val, "authorship_1")
     expected = (
         b'<iiif-uri xmlns="https://dasch.swiss/schema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" '
         b'license="license" '
         b'copyright-holder="copyright" '
         b'authorship-id="authorship_1"'
-        b">https://link.org</iiif-uri>"
+        b">https://example.org/image.jpg/full/1338%2C/0/default.jpg</iiif-uri>"
```
Nora-Olivia-Ammann, line 84: You can add this, but for me in tests it is ok to put an invalid value to keep it short. That was my aim.

testdata/xml-data/test-data-e2e.xml
```diff
@@ -22,7 +25,10 @@
     <resource label="Resource 2"
               restype=":ImageResource"
               id="resource_2">
-        <bitstream>testdata/bitstreams/test.jpg</bitstream>
+        <bitstream license="http://rdfh.ch/licenses/cc-by-4.0"
```
Nora-Olivia-Ammann, line 28: This is on purpose. I want to have a test that does not have the information as long as it is still permissible through the API to leave it out. I would like to keep that.

test/unittests/xmllib/serialise/test_serialise_file_value.py
jnussbaum: My concern is that there should be no uncaught warnings in tests. That's why I want to either fix them, or catch them. In this case here, fixing makes sense. 

This means that I only agree with you opinion as far as no warnings are raised.

testdata/xml-data/test-data-e2e.xml
jnussbaum: The test-data-systematic.xml also contains bitstreams without legal info. Do you think that's enough, or would you also like to keep a unit test without legal info? I'm okay either way. 

If you'd like to keep this unit test, I can assert that the warning was raised.

jnussbaum: I have now explicitly tested that in the legacy test, the warning is raised that there are missing legal info: https://github.com/dasch-swiss/dsp-tools/pull/1517/commits/f54cfefc96a1a5eba5292058ae1f1f3bc435f658

So I think it's okay to not have this warning again here.

Nora-Olivia-Ammann (approved):

PR #1518, opened 2025-03-07
refactor(validate-data): target values in the turtle shapes

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -92,7 +92,10 @@ def _extract_base_info_of_resource_results(
     for nd in focus_nodes:
         focus_iri = nd[1]
         res_type = next(data_onto_graph.objects(focus_iri, RDF.type))
-        if res_type in resource_classes:
+        if not (found := list(results_and_onto.objects(nd[0], SH.sourceConstraintComponent))):
+            continue
+        constraint_component = found.pop(0)
+        if any([bool(res_type in resource_classes), bool(constraint_component == SH.PatternConstraintComponent)]):
```
Nora-Olivia-Ammann, line 98: The values are not in the resource classes and I do not want to add them as the filter would no longer work since that is needed at other places.

src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -168,7 +171,7 @@ def _query_one_without_detail(  # noqa:PLR0911 (Too many return statements)
     component = next(results_and_onto.objects(base_info.result_bn, SH.sourceConstraintComponent))
     match component:
         case SH.PatternConstraintComponent:
-            return _query_pattern_constraint_component_violation(base_info.result_bn, base_info, results_and_onto)
+            return _query_pattern_constraint_component_violation(base_info, results_and_onto, data)
```
Nora-Olivia-Ammann, line 174: Before the pattern was used for results with and without detail the bn therefore was not consistent, this is no longer the case. But we need the data now, since we do not get the information about the resource whose value has a pattern constraint error.

src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -290,6 +293,8 @@ def _query_one_with_detail(
             | SH.MinInclusiveConstraintComponent
         ):
             return _query_generic_violation(base_info, results_and_onto)
+        case SH.PatternConstraintComponent:
```
Nora-Olivia-Ammann, line 296: The simple and richtext now generate two errors, before, we only got the pattern constraint in the detail. We ignore the pattern constraint in a result with detail, because now we query all the pattern contstraints directly.

PR #1519, opened 2025-03-07
refactor(validate-data): change separation of result types

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -85,23 +84,26 @@ def _separate_result_types(
...
+            focus_iri=focus_iri,
+            focus_rdf_type=res_type,
+        )
+        all_res_focus_nodes.extend(_extract_one_base_info(info, results_and_onto, data_onto_graph))
     return all_res_focus_nodes
 
 
+def _get_all_main_result_bns(results_and_onto: Graph) -> set[SubjectObjectTypeAlias]:
+    all_bns = set(results_and_onto.subjects(RDF.type, SH.ValidationResult))
+    detail_bns = set(results_and_onto.objects(predicate=SH.detail))
```
Nora-Olivia-Ammann, line 103: If a bn is used as a detail of another validation result we do not want to analyse it separately. Therefore we remove those.

src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -85,23 +84,26 @@ def _separate_result_types(
 def _extract_base_info_of_resource_results(
     results_and_onto: Graph, data_onto_graph: Graph
 ) -> list[ValidationResultBaseInfo]:
-    focus_nodes = list(results_and_onto.subject_objects(SH.focusNode))
-    resource_classes = list(data_onto_graph.subjects(KNORA_API.canBeInstantiated, Literal(True)))
-    resource_classes.extend(STILL_IMAGE_VALUE_CLASSES)
```
Nora-Olivia-Ammann, line 90: This is not a good criteria anymore because when we target the classes directly we will have results that are no longer resources but still important.

PR #1520, opened 2025-03-07
test: escape regex patterns

jnussbaum:


test/unittests/commands/excel2xml/excel2xml_lib/test_make_props.py
```diff
@@ -465,7 +465,7 @@ def test_make_bitstream_prop_valid_file(self) -> None:
         assert res.text == "testdata/bitstreams/test.jpg"
 
     def test_make_bitstream_prop_invalid_file(self) -> None:
-        with pytest.warns(DspToolsUserWarning, match=".*Failed validation in bitstream tag.*"):
+        with pytest.warns(DspToolsUserWarning, match=regex.escape("Failed validation in bitstream tag")):
```
jnussbaum, line 468: Wherever we test for a regex pattern, we escape first. In this case, it doesn't matter, but I wanted it to be consistent.

Nora-Olivia-Ammann (approved):

PR #1524, opened 2025-03-07
test: configure ports of testcontainers (DEV-4688)

jnussbaum:


test/e2e/setup_testcontainers.py
```diff
@@ -93,7 +93,7 @@ def _get_fuseki_container(network: Network, version: str) -> DockerContainer:
         DockerContainer(f"daschswiss/apache-jena-fuseki:{version}")
         .with_name("db")
         .with_network(network)
-        .with_bind_ports(3030, 3030)
+        .with_bind_ports(3030, 3031)
```
jnussbaum, line 96: The order of the parameters here is: container, host

PR #1527, opened 2025-03-10
refactor(validate-data): remove duplication of file value errors

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -85,21 +83,18 @@ def _separate_result_types(
...
 ) -> list[ValidationResultBaseInfo]:
-    resource_classes = list(data_onto_graph.subjects(KNORA_API.canBeInstantiated, Literal(True)))
-    resource_classes.extend(STILL_IMAGE_VALUE_CLASSES)
     all_res_focus_nodes = []
     main_bns = _get_all_main_result_bns(results_and_onto)
     value_types = _get_all_value_classes(data_onto_graph)
     for nd in main_bns:
         focus_iri = next(results_and_onto.objects(nd, SH.focusNode))
         res_type = next(data_onto_graph.objects(focus_iri, RDF.type))
-        if res_type in resource_classes:
```
Nora-Olivia-Ammann, line 96: Because of the duplication of errors, I previously needed this clause, now since files do not cause duplicate errors anymore, we can leave this.

src/dsp_tools/resources/validate_data/api-shapes-with-cardinalities.ttl
```diff
@@ -234,8 +234,7 @@ api-shapes:hasArchiveFileValue_PropShape
   sh:message  """
               An ArchiveRepresentation requires a file with one of the following extensions
               'zip', 'tar', 'gz', 'z', 'tgz', 'gzip', '7z'.
-              """ ;
-  sh:node     api-shapes:ArchiveFileValue_ClassShape .
```
Nora-Olivia-Ammann, line 238: This caused the duplication of errors. 
Because we declare `sh:targetClass knora-api:ArchiveFileValue` on line 244, we get one error because the class does not conform and one error because the `sh:node` does not conform. The former caused an error without detail where the file value is the focus node. The later caused an error with detail where the resource was the focus node. Now we only get an error where the file value is the focus node and we query for the resource information to get that to the user.

test/unittests/commands/validate_data/test_query_validation_result.py
```diff
@@ -78,13 +78,23 @@ def test_link_target_inexistent(self, report_target_resource_wrong_type: tuple[G
...
+        assert isinstance(result, ValidationResult)
+        assert result.violation_type == ViolationType.GENERIC
+        assert result.res_iri == DATA.bitstream_no_legal_info
+        assert result.res_class == ONTO.TestArchiveRepresentation
+        assert result.property == KNORA_API.hasLicense
+        assert not result.input_value
+        assert not result.input_type
+        assert result.expected == Literal("Files and IIIF-URIs require a reference to a license.")
 
-class TestExtractBaseInfo:
```
Nora-Olivia-Ammann, line 82: This test does not make sense anymore because we no longer filter according to the criteria if a node is a resource or not.

src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -85,21 +83,18 @@ def _separate_result_types(
 def _extract_base_info_of_resource_results(
     results_and_onto: Graph, data_onto_graph: Graph
 ) -> list[ValidationResultBaseInfo]:
-    resource_classes = list(data_onto_graph.subjects(KNORA_API.canBeInstantiated, Literal(True)))
-    resource_classes.extend(STILL_IMAGE_VALUE_CLASSES)
     all_res_focus_nodes = []
     main_bns = _get_all_main_result_bns(results_and_onto)
     value_types = _get_all_value_classes(data_onto_graph)
```
Nora-Olivia-Ammann, line 88: Here we query for the value classes from knora-api, this way we know when we have to query for the resource information because it is not a focus node.

seakayone (approved):

PR #1528, opened 2025-03-10
refactor(validate-data): change targeting of `knora-api:Values` (DEV-4690)

Nora-Olivia-Ammann:


seakayone (approved):
The PR LGTM. I like that the `sh:NodeShape`s for `knora-base`values are now much closer to the definitions in the ontology.

I just noticed that for some of the `NodeShape`s we do not yet define `sh:minCount` or `sh:maxCount` on the according `sh:PropertyShape` even though the owl cardinality in the ontology is exaclty one. Like for example `api-shapes:geonameValueAsGeonameCode_Shape`.

PR #1531, opened 2025-03-10
test: parallelize e2e tests, make testcontainers use custom ports (DEV-4688)

jnussbaum:


test/e2e/conftest.py
jnussbaum, line 1: Every test run has its custom artifact directory, which is assigned dynamically. So the tidyup has to be done for every test run individually (see `artifacts.py`)

test/e2e/setup_testcontainers/artifacts.py
```diff
@@ -0,0 +1,32 @@
...
+        "ingest_db": E2E_TESTDATA / "ingest-db" / _uuid,
+    }
+    for _dir in dirs.values():
+        _dir.mkdir(parents=True)
+    return ArtifactDirs(**dirs)
+
+
+def remove_artifact_dirs(artifact_dirs: ArtifactDirs) -> None:
+    for _dir in [artifact_dirs.sipi_images, artifact_dirs.tmp_sipi, artifact_dirs.tmp_ingest, artifact_dirs.ingest_db]:
+        with contextlib.suppress(PermissionError):
```
jnussbaum, line 31: I could theoretically write some retry logic here, or dig further into why this can happen. But I think it isn't worth the effort. If tidyup is not complete, then be it so.

seakayone (approved):

Nora-Olivia-Ammann (approved):
For me the test run was the same length as on main, but it was nice that my stack was not destroyed...

jnussbaum:
> For me the test run was the same length as on main, but it was nice that my stack was not destroyed...

Interesting... How did you execute the tests? With `just e2e-tests`, there should be a difference. Could you try again? I'm interested in how this performs on different machines, and under what circumstances it is faster/slower.

Nora-Olivia-Ammann:
> just e2e-tests

Then it was my commant I did `pytest ...` with no extra flags, now it is almost one minute faster

PR #1533, opened 2025-03-11
refactor(validate-data): validate metadata of values directly on the value

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/constants.py
```diff
@@ -55,18 +55,6 @@
     KNORA_API.hasAuthorship,
 }
 
-VALUE_CLASS_SHAPES = {
```
Nora-Olivia-Ammann, line 58: They were no longer in use

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -80,14 +80,28 @@ api-shapes:Resource_Shape
 # USER FACING VALUES
 ########################
 
+api-shapes:Value_ClassShape
```
Nora-Olivia-Ammann, line 83: We have a test for this, so we know it works

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -271,40 +278,16 @@ api-shapes:intValueAsInt_Shape
 
 # The target of the LinkValue is ontology specific.
 # The Shape is constructed through the SPARQL queries
-api-shapes:LinkValue_ClassShape
```
Nora-Olivia-Ammann, line 274: When we remove the value has comment, this shape no longer does anything, same with list and text value

src/dsp_tools/resources/validate_data/api-shapes-with-cardinalities.ttl
```diff
@@ -229,8 +229,10 @@ api-shapes:hasCopyrightHolder_PropShape
 api-shapes:hasArchiveFileValue_PropShape
   a           sh:PropertyShape ;
   sh:path     knora-api:hasArchiveFileValue ;
+  sh:class    knora-api:ArchiveFileValue ;
```
Nora-Olivia-Ammann, line 232: This was missing. again for stillimage we cannot do it because it can be either type of class

src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -216,10 +216,6 @@ def _query_one_without_detail(  # noqa:PLR0911 (Too many return statements)
             )
         case SH.ClassConstraintComponent:
             return _query_class_constraint_without_detail(base_info, results_and_onto, data, msg)
-        # This component appears when an image file has any kind of problem.
```
Nora-Olivia-Ammann, line 219: See ttl comment for explanation

src/dsp_tools/resources/validate_data/api-shapes-with-cardinalities.ttl
```diff
@@ -378,8 +386,8 @@ api-shapes:hasStillImageFileValue_PropShape
   a           sh:PropertyShape ;
   sh:path     knora-api:hasStillImageFileValue ;
   sh:xone     (
-                [ sh:node api-shapes:StillImageFileValue_ClassShape ]
-                [ sh:node api-shapes:StillImageExternalFileValue_ClassShape ]
```
Nora-Olivia-Ammann, line 382: By targeting the node, we get the xone error, because the node that is validated does not conform to either of those node shapes. Now we just state that it needs one of those classes, it does not validate the shape of the class. As with all other values the class is targeted directly in the class shape (line 397), this way we could get rid of a redundant error, which we ignored anyway.

test/e2e/test_validate_data.py
```diff
@@ -167,7 +167,6 @@ def test_extract_identifiers_of_resource_results(every_combination_once: Validat
         (URIRef("http://data/image_no_legal_info"), None),
         (URIRef("http://data/image_no_legal_info"), None),
         (URIRef("http://data/image_no_legal_info"), None),
-        (URIRef("http://data/image_no_legal_info"), None),
```
Nora-Olivia-Ammann, line 170: This is the `xone` that now no longer exists

seakayone (approved):

PR #1534, opened 2025-03-11
refactor(validate-data): group property and nodeshapes better together

Nora-Olivia-Ammann:


src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -94,126 +94,107 @@ api-shapes:Value_ClassShape
...
 # BooleanValue
 api-shapes:BooleanValue_ClassShape
   a              sh:NodeShape ;
   sh:name        "Validates the properties used with this targetClass" ;
   sh:targetClass knora-api:BooleanValue ;
-  sh:property    api-shapes:booleanValueAsBoolean_Shape ;
-  sh:severity    sh:Violation ;
-  sh:message     "BooleanValue" .
-
-api-shapes:booleanValueAsBoolean_Shape
```
Nora-Olivia-Ammann, line 108: I like named propery shapes when they reduce redundancy. In this case these properties are only ever used with this value, therefore I think this is more readable

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -94,126 +94,107 @@ api-shapes:Value_ClassShape
...
 
 ##############
 # ColorValue
 api-shapes:ColorValue_ClassShape
   a              sh:NodeShape ;
   sh:name        "Validates the properties used with this targetClass" ;
   sh:targetClass knora-api:ColorValue ;
-  sh:property    api-shapes:colorValueAsColor_Shape ;
-  sh:severity    sh:Violation ;
-  sh:message     "ColorValue" .
```
Nora-Olivia-Ammann, line 125: This violation and message does not make sense, because the shape that actually does any evaluation has its own message...

src/dsp_tools/resources/validate_data/api-shapes.ttl
seakayone: I like inlining the property shape in these cases as well, makes it easier to read the class shapes.
Should we need to reuse these shapes we can still extract them.

seakayone (approved):

PR #1535, opened 2025-03-11
refactor(validate-data): move `FileValue` cardinalities

Nora-Olivia-Ammann:


src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -60,8 +60,6 @@ api-shapes:Resource_Shape
                     sh:path     rdfs:label ;
                     sh:datatype xsd:string ;
                     sh:pattern  "\\s*\\S+\\s*" ;
-                    sh:minCount 1 ;
```
Nora-Olivia-Ammann, line 63: The cardinality of the label is validated in the cardinality shapes. This would lead to a duplicate error. This did not happen for us, because it is not possible to create a resource without a label with dsp-tools

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -345,6 +343,172 @@ api-shapes:UriValue_ClassShape
                     sh:message  "A UriValue requires exactly one valid xsd:anyURI" ;
                  ] .
 
+
```
Nora-Olivia-Ammann, line 346: Validating the cardinalities of values can be done in the graph that validates the content. This is because we have not closed the `knora-api:Value`. If we did then we could not take advantage of inheritance here. I also suggest that we continue to do so for the moment. If the API will ever rely more on the SHACL for its own validation we would have to revisit this. But with dsp-tools I prefer to take advantage of inheritance and check the cardinalities here.
This is also the reason why I renamed the file "resource-cardinalities", because it really is only about resources.

seakayone (approved):
Much cleaner, thanks.

PR #1537, opened 2025-03-11
refactor(validate-data): make name property shapes for ontology properties instead of blank nodes

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/sparql/value_shacl.py
```diff
@@ -84,13 +84,10 @@ def _construct_value_type_shapes_to_class_shapes(onto: Graph) -> Graph:
 
     CONSTRUCT {
 
-      ?class a sh:NodeShape ;
```
Nora-Olivia-Ammann, line 87: Instead of BNs we create a named `_PropShape`.
So that we can just add all properties as `_PropShapes` we need to create then as named shapes (we add them to the resource on line 53). This makes more sense because we need to add the Link, List and Text property shapes anyway. Otherwise we have a lot of `_PropShapes` in our shacl graph that do not actually exist (this is the state now, but having "empty" shapes does not seem to bother...

test/unittests/commands/validate_data/sparql/test_value_shacl.py
```diff
@@ -45,11 +45,11 @@ def test_add_property_shapes_to_class_shapes(card_1: Graph) -> None:
 
 def test_construct_value_type_shapes_to_class_shapes_literal(card_1: Graph) -> None:
     res = _construct_value_type_shapes_to_class_shapes(card_1)
-    prop_bn = next(res.objects(ONTO.ClassMixedCard, SH.property))
-    assert next(res.objects(prop_bn, SH.path)) == ONTO.testBoolean
-    assert next(res.objects(prop_bn, URIRef("http://www.w3.org/ns/shacl#class"))) == KNORA_API.BooleanValue
-    assert next(res.objects(prop_bn, SH.message)) == Literal("This property requires a BooleanValue")
-    assert len(res) == 6
+    prop_iri = ONTO.testBoolean_PropShape
```
Nora-Olivia-Ammann, line 48: Here we no longer have the bn and consequently fewer triples.

seakayone (approved):

PR #1539, opened 2025-03-12
test: split up validate-data tests

jnussbaum:


test/e2e/commands/validate_data/test_custom_projects.py
Nora-Olivia-Ammann, line 1: The name "custom_projects" is not meaningful to me. Maybe something with edge cases or so would be more informative...

test/e2e/commands/validate_data/test_generic_1.py
Nora-Olivia-Ammann, line 1: I assume this runs in parallel and does not make it slower than the original one. Because the project is created twice..

test/e2e/commands/validate_data/test_generic_1.py
```diff
@@ -0,0 +1,202 @@
...
+            ("image_no_legal_info", ProblemType.GENERIC),
+        ]
+        assert not result.unexpected_results
+        assert len(result.problems) == len(expected_info_tuples)
+        sorted_problems = sorted(result.problems, key=lambda x: x.res_id)
+        for one_result, expected_info in zip(sorted_problems, expected_info_tuples):
+            assert one_result.problem_type == expected_info[1]
+            assert one_result.res_id == expected_info[0]
+
+    def test_reformat_dsp_inbuilt_violation(self, dsp_inbuilt_violation: ValidationReportGraphs) -> None:
```
Nora-Olivia-Ammann, line 174: If it is possible to create the one result for example `every_combination_once` in one file and use it in two separate ones, I would prefer that, then we could split up other tests from `reformat_validation_graph` (most of them are that). Right now it is extremely random. The split is not informative to me and I would prefer to keep the generic ones together. Another option is to keep the correct test data and the wrong one separate, then it would be a bit shorter and the split is informative...

test/e2e/commands/validate_data/test_generic_1.py
jnussbaum: Yes, the "testdata/validate-data/generic/project.json" project is created in 2 different workers. This would not make sense in a sequential test run, but in parallel, it can make sense. 

My intention was to split up the tests into groups which take approx. the same time to execute.

Nora-Olivia-Ammann: I would prefer to have some logic behind it rather than a random split, this makes looking at the test results more onerous to me...

test/e2e/commands/validate_data/test_generic_1.py
jnussbaum: Thanks for the feedback. In fact, me too, I wasn't very happy about my splitting strategy 🙈 . I have now separated into correct ones vs. violations, and it's even faster than before 🎉

test/e2e/commands/validate_data/test_generic_1.py
jnussbaum: Done (see other comment)

test/e2e/commands/validate_data/test_generic_correct.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def test_check_for_unknown_resource_classes(unknown_classes_graphs: RDFGraphs) -> None:
+    result = _check_for_unknown_resource_classes(unknown_classes_graphs)
+    assert isinstance(result, UnknownClassesInData)
+    expected = {"onto:NonExisting", "unknown:ClassWithEverything", "unknownClass"}
+    assert result.unknown_classes == expected
+
+
+def test_extract_identifiers_of_resource_results(every_combination_once: ValidationReportGraphs) -> None:
```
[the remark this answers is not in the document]
jnussbaum: I renamed it and moved it into the other file

Nora-Olivia-Ammann (approved):

PR #1544, opened 2025-03-14
chore: don't spam log file with stack traces (DEV-4607)

jnussbaum:


src/dsp_tools/models/exceptions.py
```diff
@@ -31,13 +34,12 @@ class InternalError(BaseError):
 
     def __init__(self, custom_msg: str | None = None, keep_default_msg: bool = True) -> None:
         default_msg = (
-            "\n\nAn internal error occurred.\n"
+            f"\n\n{BOLD_RED}An internal error occurred.{RESET_TO_DEFAULT}\n"
             "Please contact the dsp-tools development team with the following information:\n"
             "    - Which command was used.\n"
             "    - If applicable, any files that were used in conjunction with the command.\n"
-            "    - A file with the terminal output copied into.\n"
-            "    - The log files called 'logging.log', if there are several, include all.\n"
```
jnussbaum, line 39: I don't see a point in getting all logging files. They are huge, so I cannot imagine a case where 1 user problem spans more than 1 file.

src/dsp_tools/models/exceptions.py
```diff
@@ -31,13 +34,12 @@ class InternalError(BaseError):
...
-            f"      They can be found at: {Path.home() / Path('.dsp-tools')}\n"
+            "    - A text file with the terminal output copied into.\n"
+            f"    - The log file {Path.home() / '.dsp-tools/logging.log'}.\n"
```
jnussbaum, line 42: Like this, we get a clickable link :-)

src/dsp_tools/models/exceptions.py
Nora-Olivia-Ammann: LHTT upload or similar. Also The problem may be distributed over two log files, because it may reach its size limit during the run.

src/dsp_tools/cli/entry_point.py
```diff
@@ -71,12 +71,11 @@ def run(args: list[str]) -> None:
...
     except BaseError as err:
-        logger.exception("The process was terminated because of an Error:")
-        print("\nThe process was terminated because of an Error:")
-        print(err.message)
+        logger.error(f"The process was terminated because of an Error: {err.message}")
+        print(f"\n{BOLD_RED}The process was terminated because of an Error: {err.message}{RESET_TO_DEFAULT}")
         sys.exit(1)
     except Exception as err:  # noqa: BLE001 (blind-except)
-        logger.exception(err)
+        logger.exception(str(err))
```
Nora-Olivia-Ammann, line 78: Why `str()`?

Nora-Olivia-Ammann (approved):

src/dsp_tools/models/exceptions.py
jnussbaum: Currently, they grow up to 100 MB, and 30 of them are kept. So there may be users who have 30 x 100 MB. I think it's an overkill to ask them to send us all of them. 

In the rare cases where we need an old log file, we can still get in touch with them.

src/dsp_tools/cli/entry_point.py
jnussbaum: I quickly tested it, and it doesn't matter. I have reverted it.

PR #1554, opened 2025-03-17
refactor(validate-data): PostFiles

jnussbaum:


src/dsp_tools/utils/request_utils.py
Nora-Olivia-Ammann, line 1: I think it would be a good time now to perhaps have a newer folder structure: `utils/clients/...`
where you could also create a `models` file to add these and the `RequestParameters`.
Otherwise I like it.

jnussbaum: Agree. I will do it in a follow up. Otherwise I fear this PR gets too big and complicated.

PR #1560, opened 2025-03-21
test: make healthcheck of testcontainers instead of waiting for logs (DEV-4706)

jnussbaum:


jnussbaum:
Even with a sleep time of 2s after the API reports healthy, the e2e tests are flaky. I don't understand why, but the last thing I want is to introduce flakiness. So I'll close this.

PR #1562, opened 2025-03-24
chore: run `uv sync` before `just lint`, to prevent parallelism issues

jnussbaum:
Several tools that are called in `just lint` try to update the virtual environment. Because of the parallelization, this can result in conflicts.

Solution: call `uv sync` before `just lint`

### DIFF
--- a/justfile
+++ b/justfile
@@ -11,9 +11,15 @@ format:
     yamlfmt .
 
 
+# Rebuild the virtual environment (must be run before `just lint`, otherwise several tools will try to do it in parallel)
+[no-exit-message]
+uv-sync:
+    uv sync
+
+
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -110,7 +116,7 @@ legacy-e2e-tests *FLAGS:
 [no-exit-message]
 clean:
     -find . -name "*.pyc" -exec rm -rf {} \;
-    -find . -name .__pycache__ -exec rm -rf {} \;
+    -find . -name __pycache__ -exec rm -rf {} \;
```
jnussbaum, line 119: This is a mistake that was probably introduced by me. There will never be any folders with a leading dot. this doesn't exist.

Nora-Olivia-Ammann (approved):

PR #1564, opened 2025-03-24
refactor: reorganise package structure (DEV-4721)

jnussbaum:


Nora-Olivia-Ammann (approved):
I am not entirely happy with the folder name "error" because it does not only contain errors... I can't think of anything better though, because it is also not just "user_communication"... I leave it up to you

PR #1565, opened 2025-03-24
refactor: reorganise utils

jnussbaum:
This PR aims to group the utils into groups. The grouping criterion is "what does the code do"? 

- utils/data_formats: This groups together code that checks if some data corresponds to a certain format
- utils/json_parsing: This parses JSON
- utils/request_utils.py: These classes and functions help with doing requests. Since these are less than 200 lines, I thought I keep them in 1 file.
    - the class SetEncoder was put into this file, because it is only used here.

I did NOT apply the grouping criterion "Is this a class (-> models) or a function (-> utils)?", because I think this is not a smart way to organise code.


Nora-Olivia-Ammann (approved):
Looks good so far, I cannot tell yet if it is intuitive until I've used it but we can always tweak it later...

PR #1566, opened 2025-03-25
chore: bump dependencies

jnussbaum:


jnussbaum:
> How come there are changes in the lock but not the toml?

Because we have unpinned the dependencies in the pyproject.toml file, they don't have versions any more.

PR #1567, opened 2025-03-25
chore: remove version restriction on "testcontainers"

jnussbaum:


pyproject.toml
```diff
@@ -72,8 +72,7 @@ dev-dependencies = [
     # pytest: don't upgrade until https://github.com/pytest-dev/pytest/issues/12044 is fixed
     "requests-mock",
     "ruff",
-    "testcontainers==4.7.2",
-    # testcontainers: don't upgrade until https://github.com/dasch-swiss/dsp-tools/pull/1122#discussion_r1723055599 is fixed
```
jnussbaum, line 76: It looks as if the problem doesn't exist any more. I've run the e2e tests locally several times, and in the CI 2 times, so I guess it's safe to merge.

Nora-Olivia-Ammann (approved):

PR #1573, opened 2025-03-28
fix(xmlupload): add comment on value to shashed links and texts (DEV-4730)

Nora-Olivia-Ammann:


testdata/xml-data/test-circular-references.xml
BalduinLandolt, line 1: is there ever an assertion somewhere, that makes sure that these comments are there at the end of the upload?

PR #1575, opened 2025-03-28
refactor(xmlupload): add option to use own value IRI in graph

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/make_rdf_graph/make_values.py
```diff
@@ -237,12 +241,13 @@ def _make_interval_value_graph(
 def _make_richtext_value_graph(
```
Nora-Olivia-Ammann, line 241: I want to use this and the link value function in the stash, for the richtext I need to be able to add an existing IRI for the value, this enables that.

src/dsp_tools/commands/xmlupload/make_rdf_graph/make_values.py
```diff
@@ -103,6 +103,7 @@ def _make_one_value_graph(val: IntermediaryValue, res_node: BNode | URIRef, iri_
         case IntermediaryRichtext():
             properties_graph = _make_richtext_value_graph(
                 val=val,
+                val_node=BNode(),
```
Nora-Olivia-Ammann, line 106: Since this function will not be used in the stash upload we can always create a new bnode here

Nora-Olivia-Ammann:
> I trust you

Thanks! I trust the tests

jnussbaum:
> > I trust you
> 
> Thanks! I trust the tests

Very good, then we're on the safe side :-)

PR #1576, opened 2025-03-28
refactor(xmlupload): serialise `StandoffStashItem` with rdflib

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/stash/upload_stashed_xml_texts.py
```diff
@@ -93,72 +94,64 @@ def _make_prefixed_prop_from_absolute_iri(absolute_iri: str) -> str:
...
-        link_prop_name: prop_json,
-        "@context": context,
-    }
-    return jsonobj
+    graph = _make_richtext_update_graph(
+        stash_item=stash_item, value_iri_str=value_iri_str, res_iri_str=res_iri_str, iri_resolver=iri_resolver
+    )
+    graph_bytes = graph.serialize(format="json-ld", encoding="utf-8")
+    serialised_json: list[dict[str, Any]] = json.loads(graph_bytes)
+    json_frame: dict[str, Any] = {"@id": res_iri_str}
```
Nora-Olivia-Ammann, line 137: If we do not frame it the serialisation result is a list with dicts, our connection typing does not allow that. I tried it earlier and changing that is an enormous amount of work...

Nora-Olivia-Ammann: If we do not frame it the serialisation result is a list with dicts, our connection typing does not allow that. I tried it earlier and changing that is an enormous amount of work...

Nora-Olivia-Ammann: If we do not frame it the serialisation result is a list with dicts, our connection typing does not allow that. I tried it earlier and changing that is an enormous amount of work...

src/dsp_tools/commands/xmlupload/make_rdf_graph/make_values.py
```diff
@@ -236,12 +236,24 @@ def _make_interval_value_graph(
...
+
+    Args:
+        val: Richtext value
+        val_node: IRI or blank node of the value
+        res_node: IRI or blank node of the resource
+        iri_resolver: id to IRI resolver
+
+    Returns:
+        Graph
+    """
```
jnussbaum, line 256: I thought we are now free to omit these superfluous docstrings?

jnussbaum (approved):

src/dsp_tools/commands/xmlupload/make_rdf_graph/make_values.py
Nora-Olivia-Ammann: Yes but I wanted to make sure that `..._node` was correctly understood.

PR #1587, opened 2025-04-01
docs: create architectural design record

Nora-Olivia-Ammann:


docs/developers/architecture.md
```diff
@@ -0,0 +1,144 @@
...
+        resetree1-->transval: extract values
+        resetree1-->transfile: extract file or iiif
+        
+        state transres {
+            resetree2-->rescontent: transform special tags</br></br>resolve prefixes into absolute IRIs
+        }
+        transres-->respars
+
+        state transval {
+            valtree-->valpars: transform special tags</br></br>resolve prefixes into absolute IRIs</br></br>map value type into Python Class
```
Nora-Olivia-Ammann, line 91: This line is too long, but I was not able to disable the check here locally (internet was not helpful) do you know how @jnussbaum ?

jnussbaum: Resolved :-)

docs/developers/architecture.md
```diff
@@ -0,0 +1,147 @@
...
+        r2-->r3: xsd validation success
+        r3-->r4: make localnames
+    }
+    state transpy {
+        resetree1-->transres
+        resetree1-->transval: extract values
+        resetree1-->transfile: extract file or iiif
+        
+        state transres {
+            resetree2-->rescontent: transform special tags</br></br>resolve prefixes into absolute IRIs
```
jnussbaum, line 88: The parser seems to be forgiving, but AFAIK, `</br>` is not correct HTML. `<br/>` seems more correct to me.

docs/developers/architecture.md
```diff
@@ -0,0 +1,147 @@
...
+        }
+        transres-->respars
+
+        state transval {
+            valtree-->valpars: transform special tags</br></br>resolve prefixes into absolute IRIs</br></br>map value type into Python Class
```
jnussbaum:
```suggestion
            valtree-->valpars: transform special tags</br></br>resolve prefixes into absolute IRIs</br></br>map value type to Python Class
```

docs/developers/architecture.md
```diff
@@ -0,0 +1,147 @@
...
+state "Resource etree" as resetree2
+state "ParsedResource" as respars
+state "Value etree" as valtree
+state "IIIF/bitstream etree" as filetree
+state "ParsedValue" as valpars
+state "ParsedFileValue" as filepars
+state "Transform Resource" as transres
+state "Transform Value" as transval
+state "Transform FileValue" as transfile
+state "Transformed Content" as rescontent
```
jnussbaum, line 67: The position of this box suggests that this is a class name. But I think it is rather a state. Is there a possibility to omit this box, to avoid that it is misunderstood as a class name?

jnussbaum: My suggestion: 

```suggestion
state "ParsedResource" as rescontent
```

docs/developers/architecture.md
Nora-Olivia-Ammann: Thanks!

docs/developers/architecture.md
jnussbaum: (My above suggestion works, but you would have to adapt the naming of the variable)

docs/developers/architecture.md
```diff
@@ -0,0 +1,147 @@
...
+    }
+    state transfomrationval {
+        parsedval-->valdes: map permissions</br></br>map listnodes to IRIs</br></br>map file metadata
+    }
+    state transformfile {
+        ParsedFileValue-->IntermediaryFileValue: map permissions</br></br>map metadata
+    }
+    transformfile-->coll: return result
+    transfomrationval-->coll: return result
+    transformres-->coll: return result
```
jnussbaum, line 142: If you reverse these 3 lines, it's more consistent with the above graphic.

```suggestion
    transformres-->coll: return result
    transfomrationval-->coll: return result
    transformfile-->coll: return result
```

docs/developers/architecture.md
```diff
@@ -0,0 +1,147 @@
...
+
+state transformall {
+    parsedres-->transformfile
+    parsedres-->transfomrationval
+    parsedres-->transformres
+    state transformres {
+        ParsedResource-->Permissions: resolve permissions
+    }
+    state transfomrationval {
+        parsedval-->valdes: map permissions</br></br>map listnodes to IRIs</br></br>map file metadata
```
jnussbaum, line 135: I don't understand "map file metadata". What does it mean?

docs/developers/architecture.md
```diff
@@ -0,0 +1,147 @@
...
+    }
+    state transformfile {
+        ParsedFileValue-->IntermediaryFileValue: map permissions</br></br>map metadata
```
jnussbaum, line 138: Here you say "map permissions", but on the resource level, it's "resolve permissions". Shouldn't this be the same both times?

docs/developers/architecture.md
```diff
@@ -0,0 +1,147 @@
...
+
+state "ParsedResource" as start
+state "Transform Entire Resource" as transformall
+state "Transform Resource" as transformres
+state "Transform Value" as transfomrationval
+state "Transform FileValues" as transformfile
+state "ParsedValue" as parsedval
+state "ParsedResource" as parsedres
+state "IntermediaryValue" as valdes
+state "Collected Transformations" as coll
```
jnussbaum, line 123: I'm a bit sceptical about this box. What does it designate? Is this a real state or class? Or do you just use it to make the diagram look nicer?

jnussbaum (approved):
I very much like this, thank you!
I added some comments how to improve it further.

docs/developers/architecture.md
```diff
@@ -0,0 +1,147 @@
...
+        ParsedFileValue-->IntermediaryFileValue: map permissions</br></br>map metadata
+    }
+    transformfile-->coll: return result
+    transfomrationval-->coll: return result
+    transformres-->coll: return result
+    coll-->ResourceInputConversionFailure: mapping errors
+    ResourceInputConversionFailure-->[*]
+    coll-->IntermediaryResource: successful transformations
+}
+```
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: We can do that

docs/developers/architecture.md
Nora-Olivia-Ammann: The resource actually does the transformations, if all is good then it outputs the `IntermediaryResource` it collects the problem and if there was any it outputs a `ResourceInputConversionFailure`. I did not know how to express this differently...

docs/developers/architecture.md
jnussbaum: So, does "map file metadata" mean that the authorship IDs are resolved to the actual strings of the authors, and the permissions IDs are resolved to IRIs?

Then it's perhaps a problem of "map" versus "resolve". For me, "resolve" is better understandable.

docs/developers/architecture.md
jnussbaum: I think it's okay how it is. I was a bit confused because there is no distinction between general descriptions and code examples. I quickly checked, there is no support for marking text as code in mermaid, so I think we'll leave it as it is.

docs/developers/architecture.md
```diff
@@ -0,0 +1,133 @@
...
+### From `ParsedResource` to `IntermediaryResource` in `xmlupload`
+
+```mermaid
+---
+title: Transformations from ParsedResource for xmlupload
+---
+stateDiagram-v2
+
+state "Transform Resource" as transformres
+state "Transform Value" as transfomrationval
```
jnussbaum, line 108:
```suggestion
state "Transform Value" as transformationval
```

docs/developers/architecture.md
Nora-Olivia-Ammann: I think we can revisit it once I wrote some code, maybe we need to change things anyway...

PR #1591, opened 2025-04-02
refactor: create new classes to parse XML

Nora-Olivia-Ammann:


src/dsp_tools/utils/xml_parsing/models/parsed_resource.py
```diff
@@ -0,0 +1,81 @@
...
+    creation_date: DateTimeStamp | None = None
+
+    def any(self) -> bool:
+        return any([self.iri, self.ark, self.creation_date])
+
+
+@dataclass
+class ParsedValue:
+    prop_name: str
+    value: str | tuple[str, str] | None
```
Nora-Olivia-Ammann, line 34: I wanted to use one class for all values. for `etree.Element.text` the typing can also be `None` that why this is there. The tuple is for list and interval values. I debated creating separate classes for all Value types, but I think at this stage I prefer to keep it generic. When we transform this class into the `Intermediate` then we type all the values correctly (int, float, etc.). That is the point where the interval will be changed and the list node resolved. I think using a tuple is nicer than creating a new string, because we want to manipulate the input as little as possible here.

src/dsp_tools/utils/xml_parsing/models/parsed_resource.py
```diff
@@ -0,0 +1,76 @@
...
+
+@dataclass
+class ParsedFileValueMetadata:
+    license_iri: str | None
+    copyright_holder: str | None
+    authorship_id: str | None
+    permissions_id: str | None
+
+
+class KnoraValueType(Enum):
```
Nora-Olivia-Ammann, line 50: I wanted to recycle this, because otherwise we would have to create a mapper for the validate-data that maps to something identical

src/dsp_tools/utils/xml_parsing/models/parsed_resource.py
```diff
@@ -0,0 +1,76 @@
...
+
+from dataclasses import dataclass
+from enum import Enum
+from enum import auto
+
+
+@dataclass
+class ParsedResource:
+    res_id: str
+    res_type: str
```
jnussbaum, line 11: nitpick: In the XML, it's `restype`

src/dsp_tools/utils/xml_parsing/models/parsed_resource.py
```diff
@@ -0,0 +1,76 @@
...
+    label: str
+    permissions_id: str | None
+    values: list[ParsedValue]
+    file_value: ParsedValue | None
```
jnussbaum, line 15:
```suggestion
    file_value: ParsedFileValue | None
```

src/dsp_tools/utils/xml_parsing/models/parsed_resource.py
jnussbaum: That's fine

jnussbaum (approved):
I think I found 1 mistake that should be fixed, but apart from that it looks good.

src/dsp_tools/utils/xml_parsing/models/parsed_resource.py
Nora-Olivia-Ammann: Yes and I don't like that. because it stands for resource type which is two words...

PR #1596, opened 2025-04-02
feat(xmllib): support BCE dates in `find_date_in_string()` (RDU-83)

jnussbaum:


test/unittests/xmllib/test_helpers.py
```diff
@@ -216,6 +216,22 @@ def test_find_date_in_string_year_range(self) -> None:
...
 
+    @pytest.mark.parametrize("string", ["9 BC", "9 B.C.", "9 BCE", "9 B.C.E."])
+    def test_find_date_in_string_bc_different_notations(self, string: str) -> None:
+        assert find_date_in_string(string) == "GREGORIAN:BC:9:BC:9"
+
+    @pytest.mark.parametrize("string", ["9 CE", "9 C.E.", "9 AD", "9 A.D."])
+    def test_find_date_in_string_ce_different_notations(self, string: str) -> None:
+        assert find_date_in_string(string) == "GREGORIAN:CE:9:CE:9"
+
+    def test_find_date_in_string_bc(self) -> None:
```
Nora-Olivia-Ammann, line 227: You can have multiple variables in `mark.parameterize`, because of the test run and from a display perspective I find it preferable.

Nora-Olivia-Ammann (approved):

Notheturtle (commented):
One remark about "Christian era": we could think about replacing it by Common Era, which corresponds to the BCE and CE. It would make sense, since the language used is English.

PR #1606, opened 2025-04-07
Switch command parts in start-stack to match documentation

rfbaumgartner:
This change adapts the use of docker compose up to the format
`docker compose up [OPTIONS] [SERVICE...]`
as described in the documentation at https://docs.docker.com/reference/cli/docker/compose/up/

While docker can deal with both syntaxes, Podman as drop-in replacement relies on the format described above.

Tested on Ubuntu 24.04.1 with package `podman-docker` by adapting the code inside the python environment.

Expected behaviour:
- No change with docker
- Podman will recognise `-d`

### DIFF
--- a/src/dsp_tools/commands/start_stack.py
+++ b/src/dsp_tools/commands/start_stack.py
@@ -142,10 +142,10 @@ class StackHandler:
         Raises:
             UserError: if the database cannot be started
         """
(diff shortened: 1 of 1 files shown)


jnussbaum (approved):

jnussbaum:
Thank you very much @rfbaumgartner for your contribution 👍

PR #1611, opened 2025-04-08
fix(xmllib): replace namedentities package (DEV-4755)

jnussbaum:


Nora-Olivia-Ammann (approved):
This is a semi blind approve. The code looks reasonable, but I do not know about html entities to be able to judge if this solves the problem.

PR #1621, opened 2025-04-10
chore: consolidate duplications of is_nonempty_value() and is_string_like() (DEV-4788)

jnussbaum:


src/dsp_tools/xmllib/value_checkers.py
```diff
@@ -13,33 +13,46 @@
 
 def is_nonempty_value(value: Any) -> bool:
     """
```
jnussbaum, line 15: @Notheturtle @flo7ian Here I would be glad about your feedback: Is the docstring well understandable?

src/dsp_tools/xmllib/value_checkers.py
```diff
@@ -243,42 +256,6 @@ def is_integer(value: Any) -> bool:
             return False
 
 
-def is_string_like(value: Any) -> bool:
```
jnussbaum, line 246: @flo7ian @Notheturtle Here I would be glad about your feedback: Is it okay if we remove the function https://docs.dasch.swiss/latest/DSP-TOOLS/xmllib-api-reference/value-checkers/#xmllib.value_checkers.is_string_like?

We realised that it is not significantly different from https://docs.dasch.swiss/latest/DSP-TOOLS/xmllib-api-reference/value-checkers/#xmllib.value_checkers.is_nonempty_value.

We couldn't identify a use case where the two differed from each other in a significant way.

test/unittests/xmllib/test_internal_helpers.py
jnussbaum, line 1: I moved this test to `test_value_checkers.py`

Nora-Olivia-Ammann (approved):

src/dsp_tools/xmllib/value_checkers.py
```diff
@@ -13,33 +13,46 @@
...
-
-        ```python
-        result = xmllib.is_nonempty_value("")
-        # result == False
-        ```
-
-        ```python
-        result = xmllib.is_nonempty_value(None)
-        # result == False
+        # True values:
```
Nora-Olivia-Ammann, line 31: What I dislike about such large lists is that they contain a lot of redundancies, you could also just show one boolean and comment that all booleans are non empty. For me there is more value in an explanation than combing through lists, and then being confused why mine is not on it, because they are never entirely exhaustive anyway...

src/dsp_tools/xmllib/value_checkers.py
Notheturtle: It is ok for me!

src/dsp_tools/xmllib/value_checkers.py
Notheturtle: What is the difference to you between None and None-like ?

Notheturtle (approved):

src/dsp_tools/xmllib/value_checkers.py
flo7ian: ok for me. did not use it

src/dsp_tools/xmllib/value_checkers.py
jnussbaum: "None-like" is my self-invented term for None, pd.NA, np.nan, ...

PR #1626, opened 2025-04-11
chore: implement vulture to find dead code

Nora-Olivia-Ammann:


src/dsp_tools/utils/xml_parsing/models/text_value_validation.py
jnussbaum, line 1: I'm a bit surprised that this file is not used any more. Is this correct?

jnussbaum (approved):

PR #1631, opened 2025-04-14
refactor(validate-data): integrate get data deserialised from `ParsedResource` into code

Nora-Olivia-Ammann:


test/e2e/commands/validate_data/test_generic_violations.py
```diff
@@ -285,6 +285,7 @@ def test_reformat_file_value_violation(self, file_value_violation: ValidationRep
             ("bitstream_no_legal_info", ProblemType.GENERIC),
             ("bitstream_no_legal_info", ProblemType.GENERIC),
             ("bitstream_no_legal_info", ProblemType.GENERIC),
+            ("empty_copyright_holder", ProblemType.GENERIC),
```
Nora-Olivia-Ammann, line 288: This error existed in the test data but was not yet implemented, but because we changed the parsing, it is now implemented as a side-effect

testdata/validate-data/special_characters/project_special_characters.json
```diff
@@ -21,7 +21,7 @@
         },
         "nodes": [
           {
-            "name": " \\ backslash",
+            "name": "\\ backslash",
```
Nora-Olivia-Ammann, line 24: It is no longer possible to have leading spaces in list names, because we strip everything when we parse it. The same in the xmlupload, if this is something that should be possible we need to discuss it separately...

src/dsp_tools/commands/validate_data/get_data_deserialised.py
```diff
@@ -34,7 +34,7 @@ def _get_one_resource(resource: ParsedResource) -> ResourceDeserialised:
         PropertyObject(TriplePropertyType.RDF_TYPE, resource.res_type, TripleObjectType.IRI),
     ]
     metadata.extend(_get_all_stand_off_links(values))
-    if resource.permissions_id:
+    if resource.permissions_id is not None:
```
Nora-Olivia-Ammann, line 37: If a user enters an empty string it should get the validation result that it is missing, but that the content was an empty string. Therefore we need this comparison rather than the previous.

src/dsp_tools/commands/validate_data/get_data_deserialised.py
```diff
@@ -115,6 +115,17 @@ def _get_interval_value(value: ParsedValue) -> ValueInformation:
     )
 
 
+def _get_list_value(user_value: str | tuple[str | None, str | None] | None) -> str | None:
```
Nora-Olivia-Ammann, line 118: The original behaviour was actually not ideal. This is better

PR #1633, opened 2025-04-15
feat(xmllib): create function to remove redundant whitespaces and newlines from string (DEV-4801)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/helpers.py
```diff
@@ -1109,3 +1109,34 @@ def create_non_empty_list_from_string(
         msg += " " + details_msg + "results in an empty list."
         raise InputError(msg)
     return lst
+
+
+def remove_whitespaces_from_string(string: str) -> str:
```
Nora-Olivia-Ammann, line 1114: @Notheturtle @jnussbaum 
I am not 100% convinced about the function name.
Thoughts?

src/dsp_tools/xmllib/helpers.py
```diff
@@ -1109,3 +1109,34 @@ def create_non_empty_list_from_string(
...
+    """
+    Removes redundant whitespaces and replaces newlines (`\\n`) and carriage returns (`\\t`) with a simple space.
```
jnussbaum, line 1116: This is not correct.  
 - `\t` is a tab
 - `\r` would be carriage return

"Whitespace" is a group that contains many characters, like
- ` ` space
- `\t` tab
- `\r` carriage return
- `\n` newline
- non-breaking whitespace
- En Quad, Em Quad, Zero width space, ... and many others
```suggestion
    Remove redundant whitespaces (space, `\\n`, `\\t`, etc.) and replace them with a simple space.
```

src/dsp_tools/xmllib/helpers.py
jnussbaum: perhaps: remove_redundant_whitespaces_from_string ?

jnussbaum (approved):

src/dsp_tools/xmllib/helpers.py
Notheturtle: clean_whitespaces_from_string?

PR #1638, opened 2025-04-16
refactor(xmlupload): move create lookup functions into utils

Nora-Olivia-Ammann:


src/dsp_tools/utils/xml_parsing/get_lookups.py
jnussbaum, line 1: Should this belong to `xml_parsing`? It's not really part of the parsing process, AFAIU

jnussbaum (approved):

src/dsp_tools/utils/xml_parsing/get_lookups.py
Nora-Olivia-Ammann: Well it is extracting the information from the XML, so in that sense it belongs there for me...

PR #1640, opened 2025-04-19
feat(xmllib): helper function to parse licenses (DEV-4800)

jnussbaum:


src/dsp_tools/xmllib/helpers.py
```diff
@@ -1140,3 +1142,66 @@ def clean_whitespaces_from_string(string: str) -> str:
...
+
+def find_license_in_string(string: str) -> License | None:  # noqa: PLR0911 (too many return statements)
+    """
+    Checks if a string contains a license, and returns the first found license as `xmllib.LicenseRecommended` object.
+    Once a license has been found, subsequent licenses are ignored.
+    Returns None if no license was found.
+    See [recommended licenses](https://docs.dasch.swiss/latest/DSP-TOOLS/xmllib-api-reference/licenses/recommended/)
+    for details.
+
+    Currently, only [Creative Commons](https://docs.dasch.swiss/latest/DSP-TOOLS/xmllib-api-reference/licenses/recommended/#xmllib.models.licenses.recommended.CC)
```
Nora-Olivia-Ammann, line 1155: I would not write "supported", because when I first read it I thought about the licenses that are supported in general. Maybe something like "currently only CC licenses can be identified"

src/dsp_tools/xmllib/helpers.py
```diff
@@ -1140,3 +1142,66 @@ def clean_whitespaces_from_string(string: str) -> str:
...
+    licenses are supported.
+
+    Args:
+        string: string to check
+
+    Returns:
+        xmllib.LicenseRecommended object or None
```
Nora-Olivia-Ammann, line 1162:
```suggestion
        `xmllib.LicenseRecommended` object or `None`
```

Nora-Olivia-Ammann (approved):

Notheturtle (approved):
To me it looks very good, thanks a lot !

test/unittests/xmllib/test_helpers.py
```diff
@@ -395,3 +398,156 @@ def test_clean_whitespaces_from_string_warns() -> None:
...
+            ("text CC-BY-4.0 text", LicenseRecommended.CC.BY),
+            ("text CC_BY_4.0 text", LicenseRecommended.CC.BY),
+            ("text CC BY text", LicenseRecommended.CC.BY),
+            ("text CC-BY text", LicenseRecommended.CC.BY),
+            ("text CC_BY text", LicenseRecommended.CC.BY),
+            ("CC BY 4.0", LicenseRecommended.CC.BY),
+            ("CC BY", LicenseRecommended.CC.BY),
+            ("CC-BY", LicenseRecommended.CC.BY),
+            ("CC BY BY", LicenseRecommended.CC.BY),
+            ("CC BY NC NC", LicenseRecommended.CC.BY_NC),
```
jnussbaum, line 419: @Notheturtle : I was unsure how to handle these. Should they be recognised or not?

src/dsp_tools/xmllib/helpers.py
```diff
@@ -1140,3 +1142,90 @@ def clean_whitespaces_from_string(string: str) -> str:
...
+        # result == LicenseRecommended.CC.BY
+        ```
+
+        ```python
+        result = xmllib.find_license_in_string("unsupported: Creative Commons Developing Nations 2.0 Generic Deed")
+        # result == None
+        ```
+
+        ```python
+        result = xmllib.find_license_in_string("CC BY. Second license is ignored: CC BY SA")
```
Notheturtle, line 1178: This sentence was not clear to me at first glance. I would understand it better if it was "CC BY, CC BY SA. The second license will be ignored. "

test/unittests/xmllib/test_helpers.py
Notheturtle: Good question. the two first ones are illogical to me since it is a mix of abbreviation and full text. But who knows, could be used, and it does not hurt to leave them in. 
The last ones with BY BY and NC NC, I think honestly not. Because unlike the others, it is false, and it could be a mistake: BY NC NC for BY NC ND. So there, I would stay strict.

Nora-Olivia-Ammann (approved):

test/unittests/xmllib/test_helpers.py
```diff
@@ -395,3 +398,156 @@ def test_clean_whitespaces_from_string_warns() -> None:
...
+        assert find_license_in_string(string) == expected
+
+    @pytest.mark.parametrize(
+        ("string", "expected"),
+        [
+            ("CC BY NC SA 1.0", LicenseRecommended.CC.BY_NC_SA),
+            ("CC BY NC SA 2.0", LicenseRecommended.CC.BY_NC_SA),
+            ("CC BY NC SA 3.1", LicenseRecommended.CC.BY_NC_SA),
+            ("CC BY NC SA 10.20", LicenseRecommended.CC.BY_NC_SA),
+        ],
```
Notheturtle, line 473: Here I wonder: if it is specified, that it is CC BY 3.0 and not 4.0, I think we are legally bound to display CC BY 3.0 (the number is a remark about the country it is valid in. 4.0 is international) ...

Notheturtle (commented):
I had a look at everything, other than my remarks, I think it is good!

test/unittests/xmllib/test_helpers.py
jnussbaum: I think this is not dsp-tools specific, but concerns the entire DSP license system. AFAIK, our system doesn't distinguish between versions yet. So you should discuss this concern with PM.

test/unittests/xmllib/test_helpers.py
jnussbaum: Okay, then I'll forbid the last ones.

PR #1651, opened 2025-04-23
feat(validate-data): remove whitespaces from `TextValues` to better detect duplicates (RDU-78)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/prepare_xml_input/transform_input_values.py
```diff
@@ -108,46 +106,13 @@ def transform_geometry(value: InputTypes) -> str:
 
 def transform_simpletext(value: InputTypes) -> str:
     str_val = assert_is_string(value)
-    # replace multiple spaces or tabstops by a single space
```
Nora-Olivia-Ammann, line 111: I moved both of these functionalities and their tests to the parsing. No other changes were made. As before they are only used with simpletext or richtext. The other values do not get cleaned like this.

src/dsp_tools/commands/xmlupload/prepare_xml_input/transform_input_values.py
```diff
@@ -108,46 +106,13 @@ def transform_geometry(value: InputTypes) -> str:
...
-    str_val = regex.sub(r" {2,}|\t+", " ", str_val)
-    # remove leading and trailing spaces (of every line, but also of the entire string)
-    str_val = "\n".join([s.strip() for s in str_val.split("\n")])
-    result = str_val.strip()
-    if len(result) == 0:
+    if len(str_val) == 0:
         raise InputError("After removing redundant whitespaces and newlines the input string is empty.")
-    return result
+    return str_val
```
Nora-Olivia-Ammann, line 111: Raising an error if a value is empty is not desired for validate therefore this check only happens here.

src/dsp_tools/utils/xml_parsing/get_parsed_resources.py
```diff
@@ -94,19 +86,24 @@ def _parse_segment_values(segment: etree._Element, segment_type: str) -> list[Pa
...
-                prop = f"{KNORA_API_STR}{val.tag!s}"
+            case "hasDescription" | "hasComment":
+                val_type = KnoraValueType.RICHTEXT_VALUE
+                value = _get_richtext_as_string(val)
+            case "relatesTo":
+                val_type = KnoraValueType.LINK_VALUE
                 value = _get_etree_content_as_string(val)
+            case _:
+                val_type = KnoraValueType.SIMPLETEXT_VALUE
+                value = _get_simpletext_as_string(val)
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: It is because `relatesTo` is a link value, the other possible properties that are left over are both simple texts. So i think I did what you wanted?

PR #1655, opened 2025-04-24
refactor(xmllib): create separate warning classes (DEV-4819)

Nora-Olivia-Ammann:


src/dsp_tools/error/custom_warnings.py
```diff
@@ -1,21 +1,21 @@
-from typing import Protocol
```
Nora-Olivia-Ammann, line 1: I needed to change this from a Protocol, because when we evaluate the warnings in warnings_config everything was duck-typed because it is a protocol. Meaning that all warnings matched the pattern of the protocol and the second check if it is an xmllib warning was never evaluated. Now because it is no longer a protocol it stopped duck-typing.

PR #1672, opened 2025-04-29
refactor(xmllib): remove resource id from `Value`

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/serialise/serialise_values.py
```diff
@@ -115,16 +118,16 @@ def _serialise_complete_simple_text_prop(values: list[SimpleText], prop_name: st
...
     for val in values:
         val_ele = _serialise_generic_element(val, "text")  # produces wrong text content
         val_ele.attrib["encoding"] = "xml"
-        prop.append(_create_richtext_elements_from_string(val, val_ele))  # overwrite the wrong text content
+        prop.append(_create_richtext_elements_from_string(val, val_ele, res_id))  # overwrite the wrong text content
     return prop
 
 
-def _create_richtext_elements_from_string(value: Richtext, text_element: etree._Element) -> etree._Element:
+def _create_richtext_elements_from_string(value: Richtext, text_element: etree._Element, res_id: str) -> etree._Element:
```
Nora-Olivia-Ammann, line 130: This check needs the ID, which is why I hand it over to the function. But, I think this check and subsequent raising of error should occur when we create the richtext value class. Because of our new warnings implementation the user will be able to locate it when the value is created and not when it is serialised. But that is for another PR.

jnussbaum (approved):

PR #1673, opened 2025-04-30
feat(xmllib): support polars null values in is null checks (DEV-4841)

Nora-Olivia-Ammann:


test/unittests/xmllib/test_value_checkers.py
```diff
@@ -62,6 +63,12 @@ def test_is_nonempty_value_wrong(text: str) -> None:
     assert not is_nonempty_value(text)
 
 
+def test_is_nonempty_value_wrong_polars_types() -> None:
+    df = pl.DataFrame({"col": [float("nan"), np.nan, None]})
```
jnussbaum, line 67: I'm a bit surprised by `float("nan")`, does this result in polars null? Is there no direct way to express this, for example `pl.null` or so?

Nora-Olivia-Ammann: Polars null exists: https://docs.pola.rs/api/python/stable/refe
But I could not find out how you ever got to this type of input. rence/api/polars.datatypes.Null.html

I image polars does some conversion internally with all these inputs because when you print the df all of them are displayed as null, but they are not actually of the `pl.Null` data type but `NoneType` from python.

So I decided to copy the input of this df they explain how to deal with empty values:
https://docs.pola.rs/user-guide/expressions/missing-data/#not-a-number-or-nan-values
I think that should be sufficient because we can assume that the input value stems from a `pl.DataFrame` therefore we only need to check the null values that we get out of them.

jnussbaum: That makes sense

jnussbaum (approved):

Issue #1615, opened 2025-04-09
Add parameter for Host


rfbaumgartner:
I am very happy with the easy way to start a DSP stack with a simple command. I would like to extend the functionality with a parameter for the host IP to use it on a server to be able to test and discuss it in a project.

Before I start with transforming my suggestions into code, I want to have a discussion about the best way to implement it.

# Option 1 
An optional parameter accepts an IP or hostname. This can be used to replace the host `0.0.0.0` when copying the files `docker-compose.yml` and `dsp-app-config.json`.
This option is not very flexible but easy to understand by users and the parameters are easy to check. A drawback is the tight connection to the format of `docker-compose.yml`.

## Option 1.1
Instead of changing the yml file, a new file to overwrite the relevant values can be added, similar to how using the latest version of the containers is handled in `docker-compose.override.yml`.

# Option 2
An optional parameter accepts a path to a custom `docker-compose.yml` and `dsp-app-config.json`.
When the configurations are copied, they can be copied from this path instead of the resources of the library. An advantage would be the loose coupling to the format for docker and the high customisability. The major drawback will be the need to understand the configurations to adapt them.

# Option 2.1
Instead of changing the copy process, the working directory for the subprocess can be manipulated.

jnussbaum:
Dear Reto

Thanks for your active collaboration, and the input to improve our platform.

I'm open for any of your proposed options, but with a clear preference for Option 1.1. This seems the most future-proof. Because it already happened that our docker-compose.yml setup changed (new containers added, environment variables added, etc.). Option 1.1 is best suited to handle this scenario.

I think that a tight coupling to the format of `docker-compose.yml` is desired in our case. 

If you want, you can take some inspiration from this draft:

Add 2 new files:

### docker-compose.override.2.j2

```yaml
services:
  api:
    environment:
      - KNORA_WEBAPI_KNORA_API_EXTERNAL_HOST={{ CUSTOM_HOST }}
  sipi:
    environment:
      - KNORA_WEBAPI_KNORA_API_EXTERNAL_HOST={{ CUSTOM_HOST }}
  ingest:
    environment:
      - JWT_ISSUER={{ CUSTOM_HOST }}:3333

```

### dsp-app-config.override.j2

```json
{
  "apiHost": {{ CUSTOM_HOST }},
  "iiifHost": {{ CUSTOM_HOST }},
  "ingestUrl": "https://{{ CUSTOM_HOST }}:3340"
}
```

[...]

PR #1675, opened 2025-04-30
feat(start-stack): add parameter for a custom host (DEV-4907)

rfbaumgartner:
This PR is a follow up to issue #1615 to enable the use of `start-stack` on a test server.

It introduces a parameter `--custom-host` that accepts an IP or a domain, making it possible to start a DSP stack on a server outside of localhost. It is necessary to replace the default value in this case, so the components can communicate and the IRIs are consistent between frontend and backend.

It has been tested with Podman/Podman-compose with and without the parameter `--latest`.
Tests with `--with-test-data` have not yet been successful. This may be related to the ontology IRI constructed by the DSP-APP.


jnussbaum:
@rfbaumgartner Please tell me if you need assistance with fixing the failing tests. Normally I only examine a PR closely once I get a review request. But in this case, I can already now look at it and assist you in getting it ready for review.

src/dsp_tools/commands/start_stack.py
```diff
@@ -112,6 +118,41 @@ def _copy_resources_to_home_dir(self) -> None:
...
+            )
+
+            dsp_app_config_template_path = importlib.resources.files("dsp_tools").joinpath(
+                "resources/start-stack/dsp-app-config.override-host.j2"
+            )
+            dsp_app_config_template = Template(dsp_app_config_template_path.read_text(encoding="utf-8"))
+            dsp_app_config_rendered = dsp_app_config_template.render(CUSTOM_HOST=self.__stack_configuration.custom_host)
+            Path(self.__docker_path_of_user / "dsp-app-config.json").write_text(
+                dsp_app_config_rendered, encoding="utf-8"
+            )
```
jnussbaum, line 153: To make it clearer what happens, I would delete the existing config file directly before writing to it. Currently, you delete it on line 122. That works, but then we have 2 things at 2 different places, while they belong together.

jnussbaum (commented):
I have added a reference to a linear ticket, so that the tests should succeed.

From my comments, only 1 is a must-do, namely updating the uv.lock file. The others are optional. As soon as this is done, we can merge this PR :-)

src/dsp_tools/commands/start_stack.py
```diff
@@ -45,6 +47,10 @@ def __post_init__(self) -> None:
             raise InputError(f"max_file_size must be between 1 and {MAX_FILE_SIZE}")
         if self.enforce_docker_system_prune and self.suppress_docker_system_prune:
             raise InputError('The arguments "--prune" and "--no-prune" are mutually exclusive')
+        if self.custom_host is not None and regex.match(
+            r"^((\d{1,3}\.)^{3}\d{1,3})|(((\w|\d|-|_|~)+\.)(\w){2,})$", self.custom_host
+        ):
+            raise InputError("Invalid format for custom host")
```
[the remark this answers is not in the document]
rfbaumgartner: I made the error message more specific which should also clarify the regular expression above

src/dsp_tools/commands/start_stack.py
rfbaumgartner: Good point. Moving it even saves one `if` statement.

src/dsp_tools/commands/start_stack.py
```diff
@@ -45,6 +47,10 @@ def __post_init__(self) -> None:
             raise InputError(f"max_file_size must be between 1 and {MAX_FILE_SIZE}")
         if self.enforce_docker_system_prune and self.suppress_docker_system_prune:
             raise InputError('The arguments "--prune" and "--no-prune" are mutually exclusive')
+        if self.custom_host is not None and regex.match(
+            r"^(((\d{1,3}\.){3}\d{1,3})|(((([a-z]|\d|-|_|~)+\.)*([a-z]){2,})))$", self.custom_host
+        ):
```
jnussbaum, line 52: Now it does the opposite of what you intend. The regex should be negated. (It can also be slightly simplified, but that's not that important. The `not` is central):

```suggestion
        if self.custom_host is not None and not regex.match(
            r"^(((\d{1,3}\.){3}\d{1,3})|((([-\w_~]+\.)*([a-z]){2,})))$", self.custom_host
        ):
```

jnussbaum: BTW, have you tried out if this PR works as intended? You could install this on a server to try it out: `pip install git+https://github.com/rfbaumgartner/dsp-tools@param-for-ip`

rfbaumgartner: The `not` was missing all the time. I probably forgot to check the command after committing and after working on the regex I took a shortcut of only testing the expression itself.
Now, the tests seem to work as intended

jnussbaum: Perfect. I saw that you committed my suggestion. But then you overwrote them again during merging main.

jnussbaum (approved):

PR #1682, opened 2025-05-02
test: fix warnings in tests

jnussbaum:


src/dsp_tools/error/xmllib_warnings_util.py
```diff
@@ -86,7 +86,7 @@ def _filter_stack_frames(file_path: str) -> bool:
 
 
 def get_user_message_string(msg: MessageInfo, function_trace: str | None) -> str:
-    str_list = [f"Trace '{function_trace}'"] if function_trace else []
+    str_list = [f"File '{function_trace}'"] if function_trace else []
```
jnussbaum, line 89: The term "Trace" is IMHO not useful for users

test/unittests/xmllib/models/internal/test_values.py
```diff
@@ -42,9 +43,9 @@ def test_warns(self) -> None:
             BooleanValue.new("other", ":booleanProp", permissions=Permissions.OPEN, resource_id="res_id", comment=None)
 
     def test_warns_potentially_empty_comment(self) -> None:
-        with pytest.warns(XmllibInputWarning):
+        with pytest.warns(XmllibInputInfo):
             BooleanValue.new(
-                "other", ":booleanProp", permissions=Permissions.OPEN, resource_id="res_id", comment=str(pd.NA)
+                True, ":booleanProp", permissions=Permissions.OPEN, resource_id="res_id", comment=str(pd.NA)
```
jnussbaum, line 48: Before, this raised 2 warnings:
1. XmllibInputWarning because "other" is not a valid boolean
2. XmllibInputInfo because `str(pd.NA)` is an empty comment

(1) is already tested above. The thing that this function wants to test is (2)

PR #1683, opened 2025-05-02
refactor: deduplicate helpers from excel2xml and xmllib

jnussbaum:
Currently, there are several public helper functions from excel2xml/xmllib, for which each library has its own implementation. The longer that excel2xml lives, and the more bugs that are fixed in xmllib only, the more problematic this gets.

Therefore I suggest that excel2xml imports some helper functions, instead of having an outdated implementation of them.

This PR deletes the following functions:
- `excel2xml.find_date_in_string()` -> redirected to `xmllib.find_date_in_string()`
- `excel2xml.make_xsd_id_compatible_without_uuid()` -> redirected to `xmllib.make_xsd_compatible_id()`
- `excel2xml.make_xsd_id_compatible()` -> redirected to `xmllib.make_xsd_compatible_id_with_uuid()`

The redirections are made in a way that the excel2xml user-facing API remains unchanged.


src/dsp_tools/commands/excel2xml/excel2xml_lib.py
```diff
@@ -27,306 +24,19 @@
...
 from dsp_tools.utils.data_formats.uri_util import is_uri
 from dsp_tools.utils.xml_parsing.parse_clean_validate_xml import parse_and_validate_xml_file
+from dsp_tools.xmllib.helpers import find_date_in_string as find_date_in_string  # noqa: PLC0414 (explicit re-export)
+from dsp_tools.xmllib.helpers import make_xsd_compatible_id
+from dsp_tools.xmllib.helpers import make_xsd_compatible_id_with_uuid
 from dsp_tools.xmllib.internal.input_converters import numeric_entities
 
 # ruff: noqa: E501, UP031 (line-too-long, use f-string over percent formatting)
 
+make_xsd_id_compatible = make_xsd_compatible_id_with_uuid
```
Nora-Olivia-Ammann, line 34: Did you verify that this still works and that if you use xmllib the name is not overwritten there too?

Nora-Olivia-Ammann (approved):

PR #1684, opened 2025-05-03
fix(xmllib): prevent data loss in find_date_in_string() (DEV-4839)

jnussbaum:


src/dsp_tools/xmllib/helpers.py
```diff
@@ -969,6 +971,15 @@ def _from_year_range(year_range: Match[str]) -> str:
     return f"GREGORIAN:CE:{startyear}:CE:{endyear}"
 
 
+def _extract_already_parsed_date(string: str) -> str | None:
+    rgx_year = r"\d+(-\d{2}(-\d{2})?)?"
+    era_with_colon = r"(CE:|BC:)"
+    rgx = rf"(GREGORIAN|JULIAN|ISLAMIC):{era_with_colon}{rgx_year}:{era_with_colon}?{rgx_year}"
+    if match := regex.search(rgx, string):
```
Nora-Olivia-Ammann, line 978: Why do we return a regex result when it matches and not the original?

Nora-Olivia-Ammann (approved):

src/dsp_tools/xmllib/helpers.py
jnussbaum: Because the original could have leading/trailing chars

PR #1686, opened 2025-05-05
refactor(xmllib): move raising of errors out of richtext serialisation (DEV-4832)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/internal/values.py
```diff
@@ -282,7 +287,20 @@ def new(
...
+        try:
+            _ = etree.fromstring(pseudo_xml)
+        except etree.XMLSyntaxError as err:
+            msg_str = (
+                f"The entered richtext value could not be converted to a valid XML.\n"
+                f"Original error message: {err.msg}\n"
+                f"Potential line/column numbers are relative to this text: {pseudo_xml}"
+            )
+            msg = MessageInfo(resource_id=resource_id, prop_name=prop_name, message=msg_str)
+            raise_input_error(msg)
```
jnussbaum, line 303: There is one thing I dislike: The code duplication in `value_checkers` and here. (To be fair, the duplication existed already before). I would prefer to have a separate function which does these steps:

```
escaped_text = escape_reserved_xml_characters(converted_val)
        num_ent = numeric_entities(escaped_text)
        pseudo_xml = f"<ignore-this>{num_ent}</ignore-this>"
        _ = etree.fromstring(pseudo_xml)
```

This new function can then be called from here, and from the  `value_checkers`.

jnussbaum (approved):

PR #1692, opened 2025-05-06
feat(xmllib): enable to save warning messages in csv through an .env configuration (DEV-4853)

Nora-Olivia-Ammann:


docs/xmllib-api-reference/advanced-set-up.md
```diff
@@ -0,0 +1,19 @@
+# Advanced Set-Up Options
+
+## Save User Information to CSV
```
jnussbaum, line 3:
```suggestion
## Save Output to CSV
```

docs/xmllib-api-reference/advanced-set-up.md
```diff
@@ -0,0 +1,19 @@
...
+To save them to a csv file, you can set that up through a `.env` file through the following steps:
+
+- In the directory from where you run your code create a new file called `.env`. 
+  If you already have a `.env` file you can add the new line directly.
+- Enter the path of the csv file in quotation marks with the variable `WARNING_CSV_SAVEPATH`
+
+Example `.env` file content:
+
+  ```env
+  WARNING_CSV_SAVEPATH="my_folder/my_file.csv"
```
jnussbaum, line 15: I suggest calling it WARNINGS_CSV_SAVEPATH (plural)

docs/xmllib-api-reference/advanced-set-up.md
```diff
@@ -0,0 +1,19 @@
+# Advanced Set-Up Options
+
+## Save User Information to CSV
+
+User information, warnings and errors are printed out on the command line.
+To save them to a csv file, you can set that up through a `.env` file through the following steps:
+
+- In the directory from where you run your code create a new file called `.env`. 
+  If you already have a `.env` file you can add the new line directly.
+- Enter the path of the csv file in quotation marks with the variable `WARNING_CSV_SAVEPATH`
```
jnussbaum, line 10:
```suggestion
- In the directory from where you run your code, create a new file called `.env`. 
  If this file already exists, continue with the next step. 
- Set `WARNING_CSV_SAVEPATH` to the path of the CSV file
```

docs/xmllib-api-reference/advanced-set-up.md
```diff
@@ -0,0 +1,19 @@
+# Advanced Set-Up Options
+
+## Save User Information to CSV
+
+User information, warnings and errors are printed out on the command line.
+To save them to a csv file, you can set that up through a `.env` file through the following steps:
```
jnussbaum, line 6:
```suggestion
To additionally save them to a CSV file, you can set that up through a `.env` file through the following steps:
```

jnussbaum (approved):
This already looks good. But I think it could be made more user-friendly by providing a helper function which then handles the .env file. 

For example:

```python
# user code
from xmllib.helpers import set_up_warnings_csv
set_up_warnings_csv(Path("path/to/my_file.csv"))

# xmllib.helpers
def set_up_warnings_csv(savepath: Path) -> None:
    env = Path(".env")
    content = env.read_text()
    if "WARNING_CSV_SAVEPATH" not in content:
        env.write_text(f'WARNING_CSV_SAVEPATH="{savepath}"')
```

In `XMLRoot.write_file()` we could then read the .env file and remove the savepath, so that it doesn't stay there for subsequent runs. 

Like this, the info is only then written to csv if the helper function is called.

Nora-Olivia-Ammann:
> This already looks good. But I think it could be made more user-friendly by providing a helper function which then handles the .env file.
> 
> For example:
> 
> ```python
> # user code
> from xmllib.helpers import set_up_warnings_csv
> set_up_warnings_csv(Path("path/to/my_file.csv"))
> 
> # xmllib.helpers
> def set_up_warnings_csv(savepath: Path) -> None:
>     env = Path(".env")
>     content = env.read_text()
>     if "WARNING_CSV_SAVEPATH" not in content:
>         env.write_text(f'WARNING_CSV_SAVEPATH="{savepath}"')
> ```
> 
> In `XMLRoot.write_file()` we could then read the .env file and remove the savepath, so that it doesn't stay there for subsequent runs.
> 
> Like this, the info is only then written to csv if the helper function is called.

The problem here is that the users do not always call the script from the same file. For example jose luis and noemi run files that are for a specific function separately and then in the end they may call it from a general main file. So importing it at the beginning of a file is in my opinion not a good option. Since we declare it as an advanced option I think they can manage to create an env file and copy paste the content.

PR #1696, opened 2025-05-07
feat: include enabled licenses in json (DEV-4805)

Nora-Olivia-Ammann:


docs/file-formats/json-project/overview.md
```diff
@@ -180,6 +184,17 @@ The description is represented as a collection of strings with language tags
...
+### `enabled_licenses`
+
+(required)
+
+`"enabled_licenses": ["<license-iri>", "<license-iri>", ...]`
+
+In order for a project to be able to choose a license for their data, files and IIIF-URIs the licenses must be enabled.
+Only licenses that exist in DSP can be enabled and referenced. They must be referenced by their IRI.
+For example: `"http://rdfh.ch/licenses/cc-by-4.0"` is a valid license IRI.
+All the listed licenses will be enabled, licenses can be disabled by omitting them.
```
jnussbaum, line 196:
```suggestion
In order for files/IIIF-URIs to reference a license, the license must be enabled here.
Only licenses that exist in DSP can be enabled. They must be referenced by their IRI.
For example: `http://rdfh.ch/licenses/cc-by-4.0` is a valid license IRI.
All the licenses listed here will be enabled. Licenses can be disabled by omitting them.
```

src/dsp_tools/resources/schema/project.json
```diff
@@ -1182,8 +1189,7 @@
                 "descriptions",
                 "ontologies",
                 "keywords"
-            ],
-            "additionalProperties": false
+            ]
```
jnussbaum, line 1192: Why did you remove that? It is still possible to omit the "enabled_licenses", also when additionalProperties are disallowed. I'm reluctant about removing this line, because then, it is possible to add random stuff.

Nora-Olivia-Ammann: Because we cannot add it as required yet. I have not found a json schema option to declare something as optional. 
I created a ticket to add it again: https://linear.app/dasch/issue/DEV-4863/add-license-array-to-all-json-files-and-make-mandatory

Nora-Olivia-Ammann: No the problem is vice versa, I cannot include it in the json as long as the additional properties is set to false. I need this array to be optional for the moment...

jnussbaum: On my machine, it behaves as follows:
- set additionalProperties = false
- leave the enabled licenses in the json file
- run the validation
- validation passes

Nora-Olivia-Ammann: It did not work for me, but now it also worked twice...

jnussbaum (approved):

PR #1704, opened 2025-05-09
chore: try out using num of logical cores for pytest-xdist

jnussbaum:


jnussbaum:
There is no performance gain

PR #1705, opened 2025-05-12
test(xmlupload): add test data to check richtext parsing

Nora-Olivia-Ammann:


testdata/xml-data/generic_project_richtext.xml
```diff
@@ -0,0 +1,122 @@
...
+                    Paragraph 2 text
+                </p>
+            </text>
+        </text-prop>
+    </resource>
+
+    <resource label="res_with_escaped_characters" restype=":ClassWithEverything" id="res_with_escaped_characters">
+        <text-prop name=":testRichtext">
+            <text encoding="xml">
+                text &lt;not a tag&gt; text and with ampersand &amp;
```
jnussbaum, line 63: Not sure if syntactically, this would be a candidate for a tag. To make it more tag-like, we could put it like this:

```suggestion
                text &lt;notatag&gt; text and with ampersand &amp;
```

jnussbaum (approved):

PR #1706, opened 2025-05-12
test(xmlupload): e2e if text content is parsed correctly

Nora-Olivia-Ammann:


test/e2e/commands/xmlupload/utils.py
```diff
@@ -8,18 +8,29 @@
...
 
+from dsp_tools.utils.rdflib_constants import KNORA_API
 from dsp_tools.utils.rdflib_constants import SubjectObjectTypeAlias
 
 
 def util_request_resources_by_class(resclass_iri: str, auth_header: dict[str, str], project_iri: str, creds) -> Graph:
+    all_resources = Graph()
+    for i in range(100):
+        g = _get_single_graph_from_api(resclass_iri, auth_header, project_iri, creds, i)
+        all_resources += g
```
jnussbaum, line 19: Not sure how performant this is. I think that this creates a new copy of the graph every time, then adds the new triples to it, and then overwrites the old graph with the new one.

Perhaps it would be better to update the graph in-place.

test/e2e/commands/xmlupload/utils.py
```diff
@@ -8,18 +8,29 @@
...
+        if not list(g.triples((None, KNORA_API.mayHaveMoreResults, None))):
+            break
```
jnussbaum, line 21: Going until page number 100 is probably good enough, but not so elegant. What about this?

```suggestion
    mayHaveMoreResults = True
    i = 0
    while mayHaveMoreResults:
        g = _get_single_graph_from_api(resclass_iri, auth_header, project_iri, creds, i)
        all_resources += g
        mayHaveMoreResults = bool(list(g.triples((None, KNORA_API.mayHaveMoreResults, None))))
        i += 1
```

jnussbaum (approved):

test/e2e/commands/xmlupload/test_values.py
```diff
@@ -279,3 +289,128 @@ def test_uri(self, g_minimal, onto_iri):
...
+        )
+        assert returned_str == SPECIAL_CHARACTERS_STRING
+
+    def test_special_characters_in_label(self, g_text_parsing):
+        # Since we search the resource according to its label we know that it was parsed correctly.
+        # Otherwise, this function call would fail.
+        returned_iri = util_get_res_iri_from_label(g_text_parsing, SPECIAL_CHARACTERS_STRING)
+        assert str(returned_iri).startswith("http://rdfh.ch/9999/")
+
+    # def test_special_characters_in_footnote(self, g_text_parsing, onto_iri):
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: I have created the ticket, but I'll write the link here

test/e2e/commands/xmlupload/utils.py
Nora-Olivia-Ammann: I dislike writing `while True` loops.

test/e2e/commands/xmlupload/utils.py
Nora-Olivia-Ammann: I created a new graph because otherwise I cannot query for may have more results in the newly returned graph, because it would already be in the graph.

test/e2e/commands/xmlupload/utils.py
jnussbaum: It's not `while True`, it's `while <var>`. Me too, I rarely use while loops, but here it would be a good fit. The for loop isn't a clean solution here. However, it's not super important, so feel free to follow your preferences ;-)

PR #1710, opened 2025-05-13
feat(xmllib): add the option to overwrite the `Permissions.PROJECT_SPECIFIC_PERMISSIONS` with a custom default permissions (DEV-4877)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/root.py
```diff
@@ -149,12 +150,24 @@ def add_resource_optional(self, resource: AnyResource | None) -> XMLRoot:
             self.resources.append(resource)
         return self
 
-    def write_file(self, filepath: str | Path) -> None:
+    def write_file(self, filepath: str | Path, default_permissions: Permissions | None = None) -> None:
```
Nora-Olivia-Ammann, line 153: @Notheturtle 
Do you think more examples are necessary or is this explanation enough?

test/unittests/xmllib/models/test_root.py
```diff
@@ -187,6 +187,178 @@ def test_video_segment_with_permissions(self):
         assert bounds.attrib == {"segment_start": "1", "segment_end": "2", "permissions": "restricted"}
 
 
+class TestSerialiseOverwriteDefaultPermissions:
```
Nora-Olivia-Ammann, line 190: I copy pasted the test from above which has mixed permissions and updated the assertions for the attributes. This way we can be sure that everything comes out as expected.

src/dsp_tools/xmllib/models/root.py
```diff
@@ -149,12 +150,24 @@ def add_resource_optional(self, resource: AnyResource | None) -> XMLRoot:
...
         Args:
             filepath: where to save the file
+            default_permissions: permissions to overwrite `Permissions.PROJECT_SPECIFIC_PERMISSIONS`
+
+        Attention:
+            If the parameter `default_permission` is not filled
+             and no permissions were assigned to values and resources,
+             the default permissions (DOAP) on DSP will be applied.
+             If the default permissions (DOAP) are not changed, only project members and admins have view permissions.
+             A project is ongoing where setting custom DOAPs can be done through the project JSON.
```
Notheturtle, line 166: In the near future, custom DOAPs will be able to be set in the project JSON file (ongoing DSP-TOOLS project)

Notheturtle (approved):

src/dsp_tools/xmllib/internal/serialise_resource.py
```diff
@@ -60,6 +70,40 @@ def _serialise_one_resource(res: AnyResource, authorship_lookup: AuthorshipLooku
...
+
+def _resolve_permissions_segment_resource(
+    segment: AudioSegmentResource | VideoSegmentResource, default_permissions: Permissions | None
+) -> None:
+    _resolve_permissions_resource_and_generic_values(segment, default_permissions)
+    if default_permissions:
+        segment.segment_bounds.permissions = _resolve_default_permission(
+            segment.segment_bounds.permissions, default_permissions
+        )
+
```
jnussbaum, line 105: Isn't it also necessary to handle `RegionResource.geometry`?

src/dsp_tools/xmllib/internal/serialise_resource.py
```diff
@@ -60,6 +70,40 @@ def _serialise_one_resource(res: AnyResource, authorship_lookup: AuthorshipLooku
...
+def _resolve_permissions_of_generic_resource(resource: Resource, default_permissions: Permissions | None) -> None:
+    _resolve_permissions_resource_and_generic_values(resource, default_permissions)
+    if default_permissions:
+        if resource.file_value:
+            resource.file_value.metadata.permissions = _resolve_default_permission(
+                resource.file_value.metadata.permissions, default_permissions
+            )
+
+
+def _resolve_permissions_segment_resource(
```
jnussbaum, line 97: _resolve_permissions_of_segment_resource

src/dsp_tools/xmllib/internal/serialise_resource.py
```diff
@@ -60,6 +70,40 @@ def _serialise_one_resource(res: AnyResource, authorship_lookup: AuthorshipLooku
             )
 
 
+def _resolve_default_permission(permission: Permissions, default_permission: Permissions) -> Permissions:
+    if permission == Permissions.PROJECT_SPECIFIC_PERMISSIONS:
+        return default_permission
+    return permission
+
+
+def _resolve_permissions_resource_and_generic_values(
```
jnussbaum, line 79: _resolve_permissions_of_resource_and_generic_values

src/dsp_tools/xmllib/models/root.py
```diff
@@ -149,12 +150,25 @@ def add_resource_optional(self, resource: AnyResource | None) -> XMLRoot:
...
             filepath: where to save the file
+            default_permissions: permissions to overwrite `Permissions.PROJECT_SPECIFIC_PERMISSIONS`
+
+        Attention:
+            If the parameter `default_permission` is not filled
+             and no permissions were assigned to values and resources,
+             the default permissions (DOAP) on DSP will be applied.
+             If the default permissions (DOAP) are not changed, only project members and admins have view permissions.
+             In the near future,
+             custom DOAPs will be able to be set in the project JSON file (ongoing DSP-TOOLS project)
```
jnussbaum, line 167:
```suggestion
             custom DOAPs will be able to be set in the project JSON file (ongoing DSP-TOOLS project).
```

src/dsp_tools/xmllib/models/root.py
```diff
@@ -149,12 +150,25 @@ def add_resource_optional(self, resource: AnyResource | None) -> XMLRoot:
...
 
         Args:
             filepath: where to save the file
+            default_permissions: permissions to overwrite `Permissions.PROJECT_SPECIFIC_PERMISSIONS`
+
+        Attention:
+            If the parameter `default_permission` is not filled
+             and no permissions were assigned to values and resources,
+             the default permissions (DOAP) on DSP will be applied.
+             If the default permissions (DOAP) are not changed, only project members and admins have view permissions.
```
jnussbaum, line 165:
```suggestion
             If the default permissions (DOAP) are not changed via DSP-API, only project members and admins have view permissions.
```

jnussbaum (approved):
Just some small remarks. The only real concern that I have is: Isn't it also necessary to handle `RegionResource.geometry`? It seems to me that we forgot to assign the inherited permissions to it.

src/dsp_tools/xmllib/internal/serialise_resource.py
Nora-Olivia-Ammann: No because we do not store permissions in the geometry value, it inherits it from the resource. Because of design decisions we store permissions in the segment bounds. We could change that in the future if we want.

jnussbaum: Does it really inherit them from the resource? 
(I'm not asking because I think the geometry should have permissions that are distinct from its parent resource. Just to prevent that the geometry has no permissions at all, and inherits the DOAPs...)

Nora-Olivia-Ammann: This is exactly what I did in the last PR...
https://github.com/dasch-swiss/dsp-tools/blob/222f4f7eef02bb354b0c6c9155c67a576c511460/src/dsp_tools/xmllib/internal/serialise_resource.py#L91

PR #1711, opened 2025-05-13
fix(xmllib): values of dsp in built resources should inherit permissions of the resource if it is not possible to specify them individually (DEV-4888)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/internal/serialise_resource.py
```diff
@@ -74,11 +73,7 @@ def _serialise_generic_resource(res: Resource, authorship_lookup: AuthorshipLook
 def _serialise_region(res: RegionResource) -> etree._Element:
     ele = _make_generic_resource_element(res, "region")
     ele.extend(_serialise_geometry_shape(res))
-    props: list[Value] = [res.region_of]
```
Nora-Olivia-Ammann, line 77: We no longer need to do this because we take advantage of the generic values, see more later.

src/dsp_tools/xmllib/internal/serialise_resource.py
```diff
@@ -134,54 +128,42 @@ def _make_generic_resource_element(res: AnyResource, res_type: str) -> etree._El
     return etree.Element(f"{DASCH_SCHEMA}{res_type}", attrib=attribs, nsmap=XML_NAMESPACE_MAP)
 
 
-def _validate_segment(segment: AudioSegmentResource | VideoSegmentResource) -> None:
```
Nora-Olivia-Ammann, line 137: This is no longer required because we validate the content with the `Value` classes. This is why I wanted to use the value classes even though we cannot use the serializing mechanism.

test/unittests/xmllib/internal/test_serialise_resource.py
```diff
@@ -151,8 +151,8 @@ def test_serialised_string_no_warnings(self, link_obj_no_warnings: RegionResourc
         expected = (
             b'<link xmlns="https://dasch.swiss/schema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" '
             b'label="lbl" id="id">'
-            b'<text-prop name="hasComment"><text encoding="xml">cmt</text></text-prop>'
             b'<resptr-prop name="hasLinkTo"><resptr>link</resptr></resptr-prop>'
+            b'<text-prop name="hasComment"><text encoding="xml">cmt</text></text-prop>'
```
Nora-Olivia-Ammann, line 155: Order changed when serialising.

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -33,9 +34,8 @@
 class RegionResource:
     res_id: str
     label: str
-    region_of: LinkValue
```
Nora-Olivia-Ammann, line 36: Because we have a generic value, we no longer need this

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -664,20 +650,30 @@ def create_new(
             )
             ```
         """
+        segment_of_val = LinkValue.new(
```
Nora-Olivia-Ammann, line 653: The advantage of using the values here to is that we get the validation mechanism of the `LinkValue` which then also gives a good user message.

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -664,20 +650,30 @@ def create_new(
...
             label=label,
-            segment_of=segment_of,
-            segment_bounds=SegmentBounds(segment_start, segment_end, res_id),
+            values=[segment_of_val],
+            segment_bounds=SegmentBounds(segment_start, segment_end, permissions, res_id),
             permissions=permissions,
         )
 
-    def add_title(self, title: str) -> VideoSegmentResource:
+    def add_title(
```
Nora-Olivia-Ammann, line 664: Because these are now `Values` we can add individual metadata

jnussbaum (approved):

PR #1713, opened 2025-05-14
feat(validate-data): validate authorship string and downgrade severity of missing legal information (DEV-4896)

Nora-Olivia-Ammann:


src/dsp_tools/utils/xml_parsing/get_lookups.py
```diff
@@ -14,7 +14,6 @@ def get_authorship_lookup(root: etree._Element) -> dict[str, list[str]]:
     def get_one_author(ele: etree._Element) -> str:
         # The xsd file ensures that the body of the element contains valid non-whitespace characters
         txt = cast(str, ele.text)
-        txt = regex.sub(r"[\n\t]", " ", txt)
```
Nora-Olivia-Ammann, line 17: This changes user input too much. We generally do not remove newlines that are within the string only leading and trailing. Since we now can find the violations through shacl we can stop cleaning it.

test/e2e/commands/validate_data/test_generic_violations.py
```diff
@@ -308,6 +309,7 @@ def test_reformat_file_value_violation(self, file_value_violation: ValidationRep
             ("image_no_legal_info", ProblemType.GENERIC),
             ("image_no_legal_info", ProblemType.GENERIC),
             ("inexistent_license_iri", ProblemType.GENERIC),
+            ("unknown_authorship_id", ProblemType.GENERIC),
```
Nora-Olivia-Ammann, line 312: The validation message is not extremely detailed, just that it requires a non-empty string without newlines. But since we expect them to use the `xmllib` to create the XML an authorship ID that is not known is extremely unlikely and therefore I think the generic string is sufficient.

test/unittests/utils/xml_parsing/test_get_lookups.py
```diff
@@ -10,16 +10,17 @@ def test_get_authorship_lookup():
     <knora>
         <authorship id="authorship_1">
             <author>
-            Lukas
```
Nora-Olivia-Ammann, line 13: Since we no longer correct this I needed to remove the newline between the names

src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -41,6 +42,13 @@
 
 
 def validate_data(filepath: Path, api_url: str, save_graphs: bool) -> bool:
+    # The command `validate-data` should always return true if the CLI command ran without exceptions,
```
Nora-Olivia-Ammann, line 45: This is so that the code that does the heavy lifting can be used for the xmlupload where the result is relevant.

src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -156,6 +167,7 @@ def _print_shacl_validation_violation_message(
             _save_unexpected_results_and_inform_user(
                 sorted_problems.unexpected_shacl_validation_components, report, filepath
             )
+    return errors_found
```
jnussbaum, line 170: This is counter-intuitive for 2 reasons:
 - There is no logical connection between the name `_print_shacl_validation_violation_message`  and the returned bool. Ideally, as a caller, I want to look at the name of the function, and immediately understand the meaning of the returned bool.
 - Intuitively, "true" is rather connected to "good", but here it is reversed: `errors_found=True` is negative.

src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -132,7 +139,9 @@ def get_resource_msg(res: OntologyResourceProblem) -> str:
 
 def _print_shacl_validation_violation_message(
     report: ValidationReportGraphs, filepath: Path, save_graphs: bool
-) -> None:
+) -> bool:
+    # Not all results are violations, some are only warnings. In that case we do not want to return True
```
jnussbaum, line 143: I think I haven't understood the concept how `validate-data` behaves. AFAIU, it is somehow like this:
 - Option 1: no errors, no warnings -> return True
 - Option 2: only warnings -> return False
 - Option 3: errors -> escalate, crash

I'm not sure if we should follow that concept. Wouldn't it be easier to have only 2 options? 
 - Option 1: all good -> return True
 - Option 2: warnings or errors -> return False

src/dsp_tools/commands/validate_data/validate_data.py
jnussbaum: Why should `validate-data` return true if the CLI command ran without exceptions? (see also my other comment)

jnussbaum (approved):

src/dsp_tools/commands/validate_data/validate_data.py
Nora-Olivia-Ammann: Because every other CLI command does that too? It is a weird set-up in my opinion but that is what we have...

src/dsp_tools/commands/validate_data/validate_data.py
Nora-Olivia-Ammann: The warnings include for example if you reference absolute IRIs in your data, the user should know that the upload may fail if the IRIs are false but still it should proceed. An additional reason is, that in the next PR I want to downgrade missing legal info from Violation to Warning. Again this should not impede an xmlupload but we should inform the user.

src/dsp_tools/commands/validate_data/validate_data.py
jnussbaum: AFAIU, the other CLI commands return True if all is good, and False if an error happened. Unhandled exceptions are not intended, but if they appear, we have (as a safety net) the try-except clause in [the entry point](https://github.com/dasch-swiss/dsp-tools/blob/648879edaab0506a5e9417b5a7e0f50bc895d95d/src/dsp_tools/cli/entry_point.py#L66). 

I think that `validate-data` should return False if an error is detected.

src/dsp_tools/commands/validate_data/validate_data.py
jnussbaum: Then I would vote for this:

- Option 1: all good or only warnings -> return True
- Option 2: errors -> return False

Nora-Olivia-Ammann: It already does that.

src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -75,15 +77,17 @@ def validate_data(filepath: Path, api_url: str, save_graphs: bool) -> bool:
...
         logger.info("Validation passed.")
         print(BACKGROUND_BOLD_GREEN + "\n   Validation passed!   " + RESET_TO_DEFAULT)
-    else:
-        _print_shacl_validation_violation_message(report, filepath, save_graphs)
-    return True
+        return True
+    reformatted = reformat_validation_graph(report)
+    sorted_problems = sort_user_problems(reformatted)
+    _print_shacl_validation_violation_message(sorted_problems, report, filepath, save_graphs)
+    return any([bool(sorted_problems.unique_violations), bool(sorted_problems.unexpected_shacl_validation_components)])
```
jnussbaum, line 90: Do I understand correctly that now it returns True if errors were found? I would return False then...

Plus, I think now the docstring is outdated. It says: 

```
Returns:
        true unless it crashed
```

PR #1723, opened 2025-05-15
feat(xmllib): change .env variable name to save warnings file (RDU-91)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/internal/values.py
jnussbaum, line 1: Are the changes in this file on purpose? I'm not against them, but I also can't judge the consequences that these changes have. I just ask because the PR title doesn't mention this.

jnussbaum (approved):

PR #1725, opened 2025-05-15
feat: retrieve licenses of project and add authentication to validate-data (DEV-4906)

Nora-Olivia-Ammann:


test/integration/commands/xmlupload/authentication_client_mock.py
```diff
@@ -2,5 +2,9 @@
 
 
 class AuthenticationClientMockBase(AuthenticationClient):
+    server: str
```
Nora-Olivia-Ammann, line 5: I need access to these variables in the protocol.

.vulture_whitelist.py
```diff
@@ -17,3 +20,6 @@
 is_full_date("")
 
 check_notna("")
+
```
Nora-Olivia-Ammann, line 23: Once I use it in the SHACL validation we can remove it here.

src/dsp_tools/clients/authentication_client.py
```diff
@@ -8,5 +8,9 @@ class AuthenticationClient(Protocol):
     Protocol for a client that can authenticate with a DSP server and returns a token.
     """
 
+    server: str
```
Nora-Olivia-Ammann, line 11: I need access to this, so the protocol need to have them as parameters as well

src/dsp_tools/cli/args.py
```diff
@@ -16,5 +16,5 @@ class ServerCredentials:
 class ValidateDataConfig:
     """Contains the configuration for validate data."""
 
-    save_dir: Path
-    save_graphs: bool
+    xml_file: Path
```
Nora-Olivia-Ammann, line 19: I needed to handle it a bit different after all to get the correct directory in the right situation.

jnussbaum (approved):

PR #1726, opened 2025-05-16
refactor(validate-data): add validate config

Nora-Olivia-Ammann:


test/unittests/cli/test_cli_with_mock.py
```diff
@@ -143,32 +144,27 @@ def test_xmlupload_interrupt_after(xmlupload: Mock) -> None:
 def test_validate_data_default(validate_data: Mock) -> None:
     file = "filename.xml"
     args = f"validate-data {file}".split()
+    config = ValidateDataConfig(Path(file), False)
     entry_point.run(args)
-    validate_data.assert_called_once_with(filepath=Path(file), api_url="http://0.0.0.0:3333", save_graphs=False)
-
-
-@patch("dsp_tools.cli.call_action.validate_data")
```
Nora-Olivia-Ammann, line 150: The arg for dev route does no longer exist. Therefore this test was identical to the one above.

src/dsp_tools/cli/args.py
```diff
@@ -9,3 +10,11 @@ class ServerCredentials:
     password: str
     server: str
     dsp_ingest_url: str = "http://0.0.0.0:3340"
+
+
+@dataclass(frozen=True)
+class ValidateDataConfig:
+    """Contains the configuration for validate data."""
+
+    filepath: Path
```
Nora-Olivia-Ammann, line 19: I included the filepath in the config because when we check if we want to save the graph, we always need the path. So it was a handy place to put it.
I also added it to the args here because it truly does represent the parsed arguments even if they are specific to validate data.

jnussbaum: could you rename that to xml_filepath? First I thought this is the path where you want to save the graphs.

jnussbaum (approved):

PR #1727, opened 2025-05-16
fix(validate-data): allow values of the same property to have identical comments (DEV-4905)

Nora-Olivia-Ammann:


testdata/validate-data/generic/every_violation_combination_once.xml
```diff
@@ -153,8 +153,8 @@
         -->
         <resource label="Duplicate Value: valueAsString" restype=":ClassWithEverything" id="identical_values">
             <text-prop name=":testTextarea">
-                <text encoding="utf8">Text</text>
-                <text encoding="utf8">Text</text>
+                <text encoding="utf8" comment="comment">Text</text>
```
Nora-Olivia-Ammann, line 156: I added this to ensure that it still finds the `Text` that is duplicated. The danger in the query is that one might exclude all the values that contain a comment. Therefore we need to ensure that it still finds them.

Nora-Olivia-Ammann: I added this to ensure that it still finds the `Text` that is duplicated. The danger in the query is that one might exclude all the values that contain a comment. Therefore we need to ensure that it still finds the duplicate.

testdata/validate-data/generic/every_violation_combination_once.xml
Nora-Olivia-Ammann, line 1: I added this to ensure that it still finds the Text that is duplicated. The danger in the query is that one might exclude all the values that contain a comment. Therefore we need to ensure that it still finds the duplicate.

editio (approved):
Thank you!

PR #1730, opened 2025-05-19
feat(xmllib): enable sorting of resources and values during serialisation (RDU-89)

Nora-Olivia-Ammann:


test/unittests/xmllib/internal/test_serialise_values.py
```diff
@@ -18,296 +23,335 @@
...
+EXPECTED_TYPE_LOOKUP = {
+    ":bool": "boolean",
+    ":int": "integer",
+    ":text1": "simpletext",
+    ":text2": "simpletext",
+    ":link": "resptr",
+}
+
+
+class TestSortValue:
```
Nora-Olivia-Ammann, line 48: @jnussbaum 
Here are the actual tests, I have just moved the rest into a class to group it better.

docs/xmllib-api-reference/advanced-set-up.md
```diff
@@ -1,12 +1,23 @@
 # Advanced Set-Up Options
 
-## Save Output to CSV
+## Using an `.env` File
```
Nora-Olivia-Ammann, line 3: @Notheturtle 
Since this is the second time we use .env variables, I decided to move this instruciton to the top.
What do you think?

Notheturtle (approved):

docs/xmllib-api-reference/advanced-set-up.md
```diff
@@ -1,12 +1,23 @@
...
+Please note that if a parameter is omitted it has the same meaning as setting it to `false`.
+
+Oftentimes, `.env` files are used to store sensitive parameters, such as passwords.
+Therefore, it is best practice to add it to your `.gitignore` file if it is not already.
+
+
+## Save Warnings Output to CSV
 
 User information, warnings and errors are printed out on the command line.
 To additionally save them to a CSV file, you can set that up by following the next steps:
```
jnussbaum, line 19: Now that only 1 step is left, it doesn't make sense to have a bullet point.

```suggestion
To additionally save them to a CSV file, set `XMLLIB_WARNINGS_CSV_SAVEPATH` to the path of the CSV file.
```

docs/xmllib-api-reference/advanced-set-up.md
```diff
@@ -1,12 +1,23 @@
...
 
-- In the directory where you run your code from, create a new file called `.env`. 
-  If this file already exists, skip this step and write in the existing file.
 - Set `XMLLIB_WARNINGS_CSV_SAVEPATH` to the path of the CSV file
```
jnussbaum, line 22:
```suggestion
```

docs/xmllib-api-reference/advanced-set-up.md
```diff
@@ -17,3 +28,42 @@ Example `.env` file content:
 
 This file is set-up in append mode, meaning that if you do not delete it after one run, 
 the new warnings will be added to the existing file. If the file already exists a row containing `*` is added. 
+
+
+## Configurations for the Serialised XML
```
jnussbaum, line 33:
```suggestion
## Configurations for the Resulting XML File
```

docs/xmllib-api-reference/advanced-set-up.md
```diff
@@ -17,3 +28,42 @@ Example `.env` file content:
...
+- `XMLLIB_SORT_PROPERTIES`: sort properties by the property name, and the values within the properties
+
+Example `.env` file content:
+
+  ```env
+  XMLLIB_SORT_RESOURCES=true
+  XMLLIB_SORT_PROPERTIES=true
+  ```
+
+If you only want to have either sorted, the keyword can either be omitted or set to `false`.
```
jnussbaum, line 50:
```suggestion
If you want to enable only one of these sortings, the other can either be omitted or set to `false`.
```

docs/xmllib-api-reference/advanced-set-up.md
```diff
@@ -17,3 +28,42 @@ Example `.env` file content:
...
+
+
+**Note on the sorting order:**
+
+- The order is determined by the character number in the [Unicode chart](https://www.unicode.org/charts/)
+- This means that for example: `B` will come before `a` as capital letters have a lower number than lower case letters
+- Numbers are treated as strings, this means that for example: `01` and `1` are treated differently
```
jnussbaum, line 57:
```suggestion
- The order is determined by the character number in the [Unicode chart](https://www.unicode.org/charts/).
- This means that for example: `B` will come before `a`, as capital letters have a lower number than lower case letters.
- Numbers are treated as strings. This means that for example `01` and `1` are treated differently.
```

PR #1733, opened 2025-05-19
feat!: find_date_in_string() returns list[str] instead of Optional[str] (DEV-4900)

jnussbaum:


test/unittests/xmllib/test_helpers.py
```diff
@@ -116,197 +116,196 @@ def test_create_standoff_link_to_uri_text_empty() -> None:
...
-    def test_find_date_in_string_ignore_subsequent(self) -> None:
-        assert find_date_in_string("x 1492-10-12, 2025-01-01x") == "GREGORIAN:CE:1492-10-12:CE:1492-10-12"
-        assert find_date_in_string("first date: 2024. Second is ignored: 2025.") == "GREGORIAN:CE:2024:CE:2024"
-
     def test_find_date_in_string_accept_only_dash_as_range_delimiter(self) -> None:
-        assert find_date_in_string("01.01.1900:31.12.2000") == "GREGORIAN:CE:1900-01-01:CE:1900-01-01"
-        assert find_date_in_string("01.01.1900 to 31.12.2000") == "GREGORIAN:CE:1900-01-01:CE:1900-01-01"
-        assert find_date_in_string("1900:2000") == "GREGORIAN:CE:1900:CE:1900"
-        assert find_date_in_string("1900 to 2000") == "GREGORIAN:CE:1900:CE:1900"
+        assert find_date_in_string("01.01.1900:31.12.2000") == {
```
Nora-Olivia-Ammann, line 120: To me it is weird that `:` is considered a separator between two dates and does not signify that this is in fact the range. But since that is congruent with the previous behaviour we can leave it if it makes sense to everyone else.

Nora-Olivia-Ammann (approved):

test/unittests/xmllib/test_helpers.py
jnussbaum: @Notheturtle This is the class that contains the tests for `find_date_in_string()`: https://github.com/dasch-swiss/dsp-tools/blob/53c7d7531b6f08d5b409ba10c078182c1ab32bb2/test/unittests/xmllib/test_helpers.py#L118

Notheturtle: It is weird for me either way. I never had it but I think, if a project would give me this configuration 1900:2000 I would have to ask what they mean. 
In the test data I saw as well that / is interpreted as a span and that is not what I would spontaneously expect. 

Right now I have | as separator for multiple dates (not a span) and what I did was create a list of dates and then use the function find_date_in_string, so that I knew it would work. Maybe this could be an option ? to specify somewhere what the separator is ?

PR #1734, opened 2025-05-20
feat(validate-data): warn or prohibit based on server if dummy legal info is used (DEV-4925)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/sparql/legal_info_shacl.py
```diff
@@ -34,3 +57,32 @@ def construct_allowed_licenses_shape(license_iris: EnabledLicenseIris) -> Graph:
...
+    
+    api-shapes:DummyLicense_Shape
+        a              sh:NodeShape ;
+        sh:targetClass knora-api:FileValue ;
+        sh:sparql      [
+                       a          sh:SPARQLConstraint ;
+                       sh:select  """
+                                    PREFIX knora-api:  <http://api.knora.org/ontology/knora-api/v2#>
+                                    
+                                    SELECT $this ?path WHERE {
```
Nora-Olivia-Ammann, line 77: If I don't select the path, then the result queries crash because they always expect one. In my opinion it is cleaner to always include a path so I solved it through `BIND`

src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -69,9 +70,10 @@ def validate_data(filepath: Path, save_graphs: bool, creds: ServerCredentials) -
     graph_save_dir = None
     if save_graphs:
         graph_save_dir = _get_graph_save_dir(filepath)
+    is_production_server = is_on_production_server(creds.server)
     config = ValidateDataConfig(filepath, graph_save_dir, ValidationSeverity.INFO)
     auth = AuthenticationClientLive(server=creds.server, email=creds.user, password=creds.password)
-    graphs, used_iris = _prepare_data_for_validation_from_file(filepath, auth)
+    graphs, used_iris = _prepare_data_for_validation_from_file(filepath, auth, is_production_server)
```
Nora-Olivia-Ammann, line 76: I added the boolean here because this way I can test both ways easily in the e2e tests.

PR #1745, opened 2025-05-21
refactor(xmlupload): remove unnecessary conversion checks (DEV-4934)

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/prepare_xml_input/get_processed_resources.py
```diff
@@ -69,7 +68,7 @@ def get_processed_resources(resources: list[ParsedResource], lookups: XmlReferen
         try:
             result = _get_one_resource(res, lookups)
             processed.append(result)
-        except (PermissionNotExistsError, InputError, InvalidFileTypeError) as e:
+        except PermissionNotExistsError as e:
```
Nora-Olivia-Ammann, line 71: The goal is to remove all error code that is not actually reachable anymore because of validate.
In the case that something should have gone wrong, it will still crash, just not very gracefully anymore. But I think that is fine so that we can make the code easier. To remove the permissions, I have to implement the check first. Once that is done we can remove all conversion errors.

src/dsp_tools/commands/xmlupload/prepare_xml_input/transform_input_values.py
```diff
@@ -11,7 +11,7 @@
 from dsp_tools.commands.xmlupload.models.processed.values import IntervalFloats
 from dsp_tools.commands.xmlupload.models.processed.values import ProcessedValue
 from dsp_tools.commands.xmlupload.models.processed.values import ProcessedValueTypes
-from dsp_tools.error.exceptions import InputError
+from dsp_tools.error.exceptions import XmlInputConversionError
```
Nora-Olivia-Ammann, line 14: I created a new error so that in case this is actually ever triggered we can easily pinpoint the problem.

PR #1747, opened 2025-05-21
feat(xmlupload): show progress bar on CLI when uploading stash (DEV-4935)

Nora-Olivia-Ammann:


jnussbaum (approved):
Have you verified how it looks like?

PR #1748, opened 2025-05-22
feat(validate-data): ensure that only defined permissions are used (DEV-4939)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/sparql/construct_shacl.py
```diff
@@ -61,3 +68,25 @@ def _get_one_relevant_knora_subset(knora_api: Graph, knora_prop: str) -> Graph:
...
+
+def _get_defined_permissions_shape(permission_ids: list[str]) -> Graph:
+    g = Graph()
+    permissions = " ".join([f'"{x}"' for x in permission_ids])
+    msg = f"You must reference one of the pre-define permissions: {', '.join(permission_ids)}"
+    shape = f"""
+     @prefix sh:         <http://www.w3.org/ns/shacl#> .
+     @prefix owl: <http://www.w3.org/2002/07/owl#> .
+     @prefix knora-api:  <http://api.knora.org/ontology/knora-api/v2#> .
+     @prefix api-shapes: <http://api.knora.org/ontology/knora-api/shapes/v2#> .
```
jnussbaum, line 81:
```suggestion
     @prefix sh:         <http://www.w3.org/ns/shacl#> .
     @prefix owl:        <http://www.w3.org/2002/07/owl#> .
     @prefix knora-api:  <http://api.knora.org/ontology/knora-api/v2#> .
     @prefix api-shapes: <http://api.knora.org/ontology/knora-api/shapes/v2#> .
```

testdata/validate-data/generic/dsp_inbuilt_violation.xml
```diff
@@ -270,20 +281,8 @@
...
-              id="non_existent_permissions_iiif"
-    >
-        <iiif-uri license="http://rdfh.ch/licenses/unknown"
-                   copyright-holder="DaSCH"
-                   authorship-id="authorship_1"
-                   permissions="open"
-        >https://iiif.wellcomecollection.org/1Oi7mdiLsG7-FmFgp0xz2xU.jp2/full/max/0/default.jpg
-        </iiif-uri>
-    </resource>
-
```
jnussbaum, line 288: why delete this?

jnussbaum (approved):

testdata/validate-data/generic/dsp_inbuilt_violation.xml
Nora-Olivia-Ammann: Because it is the same type of class as a file value, if that works this will too. So it does not add any value.

PR #1750, opened 2025-05-22
feat: migrate warning about angular brackets in simpletext from `xmplupload` to `xmllib` (DEV-4943)

Nora-Olivia-Ammann:


test/unittests/xmllib/internal/test_checkers.py
```diff
@@ -36,3 +37,29 @@ def test_check_and_warn_potentially_empty_string_potentially_empty(in_val, type_
     )
     with pytest.warns(XmllibInputInfo, match=expected):
         check_and_warn_potentially_empty_string(value=in_val, res_id="res", expected="input")
+
+
+@pytest.mark.parametrize(
```
Nora-Olivia-Ammann, line 42: I copied the test cases from the original checker function.

src/dsp_tools/xmllib/internal/checkers.py
```diff
@@ -101,3 +101,24 @@ def check_and_warn_if_a_string_contains_a_potentially_empty_value(
...
+def check_and_inform_about_angular_brackets(value: Any, res_id: str | None, prop_name: str | None = None) -> None:
+    """
+    Checks if a string value contains angular brackets.
+
+    Args:
+        value: String value
+        res_id: resource id
+        prop_name: property name
+    """
+    if bool(regex.search(r'<([a-zA-Z/"]+|[^\s0-9].*[^\s0-9])>', str(value))):
```
Nora-Olivia-Ammann, line 115: This is the identical regex

.vulture_whitelist.py
```diff
@@ -11,6 +12,7 @@
 PermissionValue.CR
 
 BOLD_GREEN
+YELLOW
```
jnussbaum, line 15: Instead of keeping the unused YELLOW in the code, it might make more sense to keep the unused DspToolsUserInfo. Of course only if we anticipate it being used in the future. 

If you think we won't use DspToolsUserInfo in the future, I would remove YELLOW too, because there is no reason to keep it.

src/dsp_tools/xmllib/internal/checkers.py
```diff
@@ -101,3 +101,24 @@ def check_and_warn_if_a_string_contains_a_potentially_empty_value(
...
+        msg_info = MessageInfo(
+            message=(
+                f"Your input '{value}' contains angular brackets. "
+                f"Since this is a simpletext the brackets will be displayed as is."
```
jnussbaum, line 119:
```suggestion
                f"Since this is a simpletext, the brackets will be displayed as is."
```

jnussbaum, line 119: I would even be a bit more verbose. I liked the previous text: Please note that these will not be recognised as formatting in the text field, but will be displayed as-is.

jnussbaum (approved):

PR #1752, opened 2025-05-23
feat(validate-data): warn if bitstreams/iiif-uri are duplicated (DEV-4665)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/get_user_validation_message.py
```diff
@@ -115,11 +123,12 @@ def get_user_message(sorted_problems: SortedProblems, file_path: Path) -> UserPr
...
             violation_body = _get_problem_print_message(sorted_problems.unique_violations)
         violation_header = (
-            f"During the validation of the data {len(sorted_problems.unique_violations)} errors were found."
+            f"During the validation of the data {len(sorted_problems.unique_violations)} errors were found. "
+            f"Until they are resolved an xmlupload is not possible."
         )
         violation_message = MessageStrings(violation_header, violation_body)
     if sorted_problems.user_warnings:
-        if len(sorted_problems.unique_violations) > 50:
+        if len(sorted_problems.user_warnings) > 50:
```
Nora-Olivia-Ammann, line 131: This was wrong.

src/dsp_tools/commands/validate_data/get_user_validation_message.py
```diff
@@ -129,27 +138,22 @@ def get_user_message(sorted_problems: SortedProblems, file_path: Path) -> UserPr
...
+        )
+        info_message = MessageStrings(info_header, info_body)
     if sorted_problems.unexpected_shacl_validation_components:
         unexpected_header = "The following unknown violation types were found!"
         unexpected_body = LIST_SEPARATOR + LIST_SEPARATOR.join(sorted_problems.unexpected_shacl_validation_components)
         unexpected_violations = MessageStrings(unexpected_header, unexpected_body)
     return UserPrintMessages(violation_message, warning_message, info_message, unexpected_violations)
 
 
-def _get_referenced_iri_info(problems: list[InputProblem]) -> MessageStrings:
```
Nora-Olivia-Ammann, line 140: Before referenced IRIs was the only thing in the info section. Therefore constructing the string here was easier. This is no longer the case.

src/dsp_tools/commands/validate_data/get_user_validation_message.py
```diff
@@ -115,11 +123,12 @@ def get_user_message(sorted_problems: SortedProblems, file_path: Path) -> UserPr
         else:
             violation_body = _get_problem_print_message(sorted_problems.unique_violations)
         violation_header = (
-            f"During the validation of the data {len(sorted_problems.unique_violations)} errors were found."
+            f"During the validation of the data {len(sorted_problems.unique_violations)} errors were found. "
+            f"Until they are resolved an xmlupload is not possible."
```
Nora-Olivia-Ammann, line 127: I think it makes sense to include the consequences of the violation severity here.

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -353,6 +353,13 @@ api-shapes:FileValue_ClassShape
   a              sh:NodeShape ;
   sh:targetClass knora-api:FileValue ;
   sh:property    [
+                    a                        sh:PropertyShape ;
+                    sh:path                  knora-api:fileValueHasFilename ;
```
Nora-Olivia-Ammann, line 357: I have to use the actual property, because the super-property is `knora-api:valueHas`. This is the super-property of other properties used in a `FileValue` therefore it would lead to an enormous amount of false positives.
This however means that we have to add it to IIIF-URI separately.

src/dsp_tools/commands/validate_data/query_validation_result.py
```diff
@@ -232,7 +233,13 @@ def _query_one_without_detail(  # noqa:PLR0911 (Too many return statements)
             | SH.MinInclusiveConstraintComponent
             | DASH.SingleLineConstraintComponent
         ):
-            return _query_generic_violation(base_info.result_bn, base_info, results_and_onto)
+            return _query_general_violation_info(
+                base_info.result_bn, base_info, results_and_onto, ViolationType.GENERIC
+            )
+        case DASH.UniqueValueForClassConstraintComponent:
```
Nora-Olivia-Ammann, line 239: As described in the user validation messages, we may get multiple violations per resource. It is easiest to filter out duplicate messages if it is it's own ViolationType. While the queries for the violation is the same as `_query_generic_violation` the violation type is not, therefore I change the function a bit.

testdata/validate-data/generic/dsp_inbuilt_correct.xml
```diff
@@ -27,21 +27,21 @@
         <bitstream license="http://rdfh.ch/licenses/unknown"
                    copyright-holder="DaSCH"
                    authorship-id="authorship_1"
-        >this/is/filepath/file.jpg</bitstream>
```
Nora-Olivia-Ammann, line 30: Previously it did not matter that I called everything the same. Now obvisouly it does. Therefore I too the resource ID as the file name

seakayone (approved):

PR #1754, opened 2025-05-26
feat(validate-data): warnings give a validation failure if it is on a prod server (DEV-4948)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -118,13 +119,22 @@ def _validate_data(
...
+            bool(all_problems.unique_violations),
+            bool(all_problems.unexpected_shacl_validation_components),
         ]
     )
-    return no_problems
+    if violations:
+        return False
+    if is_on_prod:
+        if all_problems.user_warnings:
+            return False
```
jnussbaum, line 136:
```suggestion
    if is_on_prod and all_problems.user_warnings:
        return False
```

src/dsp_tools/cli/args.py
```diff
@@ -22,6 +22,7 @@ class ValidateDataConfig:
     xml_file: Path
     save_graph_dir: Path | None
     severity: ValidationSeverity
+    is_on_prod_server: bool = False
```
jnussbaum, line 25: Better don't use a default. We might forget setting the real value.

src/dsp_tools/utils/data_formats/uri_util.py
```diff
@@ -1,6 +1,10 @@
 import regex
 
 
+def is_prod_server(server: str) -> bool:
```
jnussbaum, line 4: is_prod_like_server ?

jnussbaum (approved):
The default value in the ValidateDataConfig seems a bit dangerous to me. Apart from that, it looks good.

PR #1775, opened 2025-06-04
feat: reformat and write large xsd schema validation results into a file (DEV-4998)

Nora-Olivia-Ammann:


test/integration/utils/xml_parsing/test_parse_clean_validate_xml.py
```diff
@@ -114,5 +117,84 @@ def test_beautify_err_msg() -> None:
         parse_and_validate_xml_file("testdata/invalid-testdata/xml-data/duplicate-res-id.xml")
 
 
+class TestReformatErrorMessage:
+    def test_empty_label(self):
+        in_msg = (
+            "Element '{https://dasch.swiss/schema}resource', attribute 'label': [facet 'minLength'] "
+            "The value '' has a length of '0'; this underruns the allowed minimum length of '1'."
+        )
+        result = _reformat_error_message_str(in_msg, 1)
```
jnussbaum, line 126: Here, a named argument would be helpful for understanding.

test/integration/utils/xml_parsing/test_parse_clean_validate_xml.py
```diff
@@ -114,5 +117,84 @@ def test_beautify_err_msg() -> None:
...
+    def test_invalid_tag(self):
+        in_msg = (
+            "Element '{https://dasch.swiss/schema}resource', attribute 'invalidtag': "
+            "The attribute 'invalidtag' is not allowed."
+        )
+        result = _reformat_error_message_str(in_msg, 1)
+        assert result.line_number == 1
+        assert result.element == "resource"
+        assert result.attribute == "invalidtag"
+        assert result.message == "The attribute 'invalidtag' is not allowed."
```
jnussbaum, line 153:
```suggestion
            "Element '{https://dasch.swiss/schema}resource', attribute 'invalidattr': "
            "The attribute 'invalidattr' is not allowed."
        )
        result = _reformat_error_message_str(in_msg, 1)
        assert result.line_number == 1
        assert result.element == "resource"
        assert result.attribute == "invalidattr"
        assert result.message == "The attribute 'invalidattr' is not allowed."
```

jnussbaum (approved):

src/dsp_tools/utils/xml_parsing/parse_clean_validate_xml.py
```diff
@@ -88,3 +99,58 @@ def _beautify_err_msg(err_msg: str) -> str:
...
+def validate_root_emit_user_message(root: etree._Element, save_path: Path) -> None:
+    validation_error = _validate_root_get_validation_messages(root)
+    if validation_error:
+        _emit_validation_errors(validation_error, save_path)
+
+
+def _validate_root_get_validation_messages(data_xml: etree._Element) -> list[XSDValidationMessage] | None:
+    if errors := _validate_xml_tree_against_schema(data_xml):
+        return _reformat_validation_errors(errors.error_log)
+    return None
```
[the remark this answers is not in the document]
Nora-Olivia-Ammann: There is a reason for that. When we use the function in other places. we will want to raise an error. In order to do that we need to know the result. In the previous code we only raised it to handle it right away. In this case we do not need to raise an error. A print out / saving of a file is enough. But at other places in the code (validate / xmlupload) we need to know about the error and raise it. This way we can reformat it, know about the result, print the message and raise the error.

PR #1777, opened 2025-06-04
chore(start-stack): add more logging

jnussbaum:


src/dsp_tools/commands/start_stack.py
```diff
@@ -323,28 +333,39 @@ def _start_remaining_docker_containers(self) -> None:
...
         for _ in range(6 * 60):
             try:
-                response = requests.get(f"{self.__localhost_url}:3333/health", timeout=1)
-                if not response.ok:
+                params = RequestParameters("GET", f"{self.__localhost_url}:3333/health", timeout=1)
+                log_request(params)
+                response = requests.get(params.url, timeout=params.timeout)
+                log_response(response)
+                if response.ok:
+                    break
```
jnussbaum, line 358: I'm very puzzled about this. This fixes an obvious bug: As soon as the container is healthy, the loop must break. But why did it work before on the main branch? The legacy-e2e tests never took 6 mins (as I would have expected), but rather 2-3 mins.

This means that in fact the loop has never taken 6 mins. But I don't see a possibility in the code how this could happen!

jnussbaum: For future reference: Before, the healthcheck route was called 6 * 60 times. As long as the API was not healthy yet, it slept 1 s, but as soon as the API became healthy, it didn't sleep any more. This means that the remaining healthcheck calls were very fast.

PR #1789, opened 2025-06-11
fix(xmllib): remove duplicate warnings for empty input when creating a new `Resource` (DEV-5057)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/res.py
```diff
@@ -50,17 +49,6 @@ class Resource:
     file_value: AbstractFileValue | None = None
     migration_metadata: MigrationMetadata | None = None
 
-    def __post_init__(self) -> None:
```
Nora-Olivia-Ammann, line 53: Usually we have these checks on `create_new` this is left over when we put it here. I prefer it in the `create_new` because it is consistent this way. I think it is ok, because the user is responsible to use it as advertised.

src/dsp_tools/xmllib/models/dsp_base_resources.py
```diff
@@ -39,10 +39,6 @@ class RegionResource:
     permissions: Permissions = Permissions.PROJECT_SPECIFIC_PERMISSIONS
     migration_metadata: MigrationMetadata | None = None
 
-    def __post_init__(self) -> None:
```
Nora-Olivia-Ammann, line 42: Not having it in the post init is consistent with the other checks

src/dsp_tools/xmllib/models/res.py
jnussbaum: Good that it is consistent now. But I wonder if it wouldn't be better to move all validations into the __post_init__. It would be safer, wouldn't it?

jnussbaum (approved):

PR #1794, opened 2025-06-17
fix(xmlupload): allow more special characters

jnussbaum:


editio (approved):
If the Spanish ñ is already present, we could also add the Spanish accented vowels:  á í ó ú  (é is already there)

PR #1796, opened 2025-06-17
feat: allow that images are part of multiple compound objects (DEV-5070)

jnussbaum:


BalduinLandolt (approved):
On the SHACL changes I trust you blindly, I don't know the validate-data well enough to reasonably review this. But it seems reasonable at least.

PR #1799, opened 2025-06-18
chore: bump dependencies

jnussbaum:


pyproject.toml
```diff
@@ -161,9 +161,9 @@ ignore = [
     "S105",   # flake8-bandit: hardcoded password
     "S106",   # flake8-bandit: hardcoded password
     "S603",   # flake8-bandit: subprocess-without-shell-equals-true
-    "S320",   # flake8-bandit: xml parsing vulnerable to XML attacks
     "UP007",  # forbid use of typing.Union and typing.Optional
     "UP015",  # Unnecessary mode="r" argument in open()
+    "UP045",  # Forced use of "|" instead of typing.Optional and typing.Union
```
jnussbaum, line 166: While I prefer the new syntax, I think it's overkill if our CI forbids the old one.

pyproject.toml
```diff
@@ -18,7 +18,7 @@ requires-python = ">=3.12"
 dependencies = [
     "argparse",
     "dotenv",
-    "jinja2>=3.1.6",
+    "jinja2",
```
jnussbaum, line 21: We stopped pinning dependencies a while ago.

test/unittests/commands/excel2json/lists/test_compliance_checks.py
```diff
@@ -262,11 +262,11 @@ def test_missing_node_column(self, cols_en_list_only: Columns) -> None:
...
-            assert isinstance(res, ListSheetComplianceProblem)
-            assert res.excel_name == "excel"
-            assert res.sheet_name == "sheet"
-            assert len(res.problems) == 1
-            assert isinstance(res.problems[0], MissingNodeColumn)
+        assert isinstance(res, ListSheetComplianceProblem)
+        assert res.excel_name == "excel"
+        assert res.sheet_name == "sheet"
+        assert len(res.problems) == 1
+        assert isinstance(res.problems[0], MissingNodeColumn)
```
jnussbaum, line 269: The pytest.warns() block should only contain what is necessary.

test/unittests/commands/validate_data/fixtures/validation_result.py
```diff
@@ -19,8 +17,7 @@
 from test.unittests.commands.validate_data.constants import PREFIXES
 
 
-@cache
-@pytest.fixture
+@pytest.fixture(scope="module")
```
jnussbaum, line 20: A strange behaviour of the new pytest version cannot handle a `@cache` decorator on fixtures. But it's preferrable anyways to use pytest's own caching mechanism.

pyproject.toml
BalduinLandolt: But it's easy enough to do, and if you prefer it, why not just enforce it? :)  
(I prefer it too, as you may have guessed)

pyproject.toml
BalduinLandolt: what do we even use jinja for?

jnussbaum: For injecting custom hosts into docker compose files, see https://github.com/dasch-swiss/dsp-tools/blob/39dbf00afd62e2a9f05b359e9eb7bdfddb49aa44/src/dsp_tools/commands/start_stack.py#L140

BalduinLandolt: ah, makes sense

PR #1809, opened 2025-06-25
fix(xmlupload): retry during ca. 15min (DEV-5089)

jnussbaum:


BalduinLandolt (approved):
👍 

This reminds me, we discussed this idea once already, a long time back, but never acted upon it: I think it would make sense to extract things like this into a global configuration singleton.  
For one, it would become much clearer, where information like this is located. And as a bonus, it would make things a lot easier, if we ever decide to make things like this configurable (through CLI flags or environment variables).  
Not something we need to do immediately - but probably worth doing at some point.

PR #1810, opened 2025-06-25
fix(start-stack): prevent sipi data from filling up storage (DEV-5065)

jnussbaum:


jnussbaum:
> I'm not sure about this change. I feel uncomfortable about removing data from the user's computer without asking, or at least giving the ability to configure it otherwise. Could this be tried to the `--no-prune` flag, at least?

I can understand your concern in principle. But in this case, there is really no reason why the ingested assets should be persisted in the first place. This folder is created by the `start-stack` command, so it is kind of reasonable to also tidy it up with the `stop-stack` command. 

This is also in line with how `stop-stack` works: It deletes everything which has been done, and removes all data. 

There is no reason why ingested files from previous stacks should survive, because there is nothing you can do with them. Even if you wanted to, you could do nothing with them.

Or do I miss something?

BalduinLandolt (approved):

PR #1814, opened 2025-06-30
docs(xmllib): fix misspelling of convert_to_bool()

jnussbaum:


src/dsp_tools/xmllib/helpers.py
```diff
@@ -356,19 +356,16 @@ def get_list_nodes_from_string_via_list_name(
...
     Args:
-        string_with_list_labels: the string containing the labels
+        string_with_list_labels: the string containing list labels
         label_separator: separator in the string that contains the labels
         list_name: name of the list
         list_lookup: `ListLookup` of the project
 
     Returns:
-        The name of the list and a list of node names.
+        A list of node names. If the string is empty, it returns an empty list.
```
jnussbaum, line 368: @Notheturtle Here we had the problem that the Fliesstext repeated the information of the "Args" and "Returns" paragraphs. At some point of time, the arguments and return types were changed, but it was forgotten to change the Fliesstext. So now, we have a wrong Fliesstext.

In order to prevent that in the future, I'd like to shorten the Fliesstext, and not have any repetition of information that is elsewhere. WDYT?

Notheturtle: Good for me!

PR #1823, opened 2025-07-03
feat(xmllib): remove functionality to set default permissions on `write_file` (DEV-5105)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/root.py
```diff
@@ -214,13 +204,17 @@ def write_file(self, filepath: str | Path, default_permissions: Permissions | No
...
-            root.write_file("xml_file_name.xml", Permissions.PUBLIC)
-            ```
         """
-        root = self.serialise(default_permissions)
+
+        if default_permissions:
+            msg = (
+                "You added a default permission, this functionality is deprecated "
+                "as project wide default permissions are set in the json. "
+                "This parameter will be removed soon."
```
jnussbaum, line 213:
```suggestion
                "You added a default permission. This has no effect. This functionality is deprecated, "
                "because project wide default permissions are set in the JSON project file. "
                "This parameter will be removed soon."
```

src/dsp_tools/xmllib/models/root.py
```diff
@@ -195,17 +195,7 @@ def write_file(self, filepath: str | Path, default_permissions: Permissions | No
             default_permissions: permissions to overwrite `Permissions.PROJECT_SPECIFIC_PERMISSIONS`
```
jnussbaum, line 195:
```suggestion
            default_permissions: This parameter is deprecated and has no effect.
              Default permissions can be set in the JSON project file.
```

src/dsp_tools/xmllib/models/root.py
```diff
@@ -195,17 +195,7 @@ def write_file(self, filepath: str | Path, default_permissions: Permissions | No
...
-             the default permissions (DOAP) on DSP will be applied.
-             If the default permissions (DOAP) are not changed via DSP-API,
-             only project members and admins have view permissions.
-             In the near future,
-             custom DOAPs will be able to be set in the project JSON file (ongoing DSP-TOOLS project).
-             At that time this parameter will be deprecated.
-
-             **This will not influence individually assigned permissions, for example
-             `Permissions.PRIVATE` will stay private even if your default is set to `Permissions.PUBLIC`.**
+            Parameter `default_permissions` is deprecated as the default permissions can be set on project level.
```
jnussbaum, line 198: The "Attention" block only for a deprecated parameter which doesn't do anything, is a bit of an overkill :-)

PR #1827, opened 2025-07-04
feat(validate-data): run validation locally with a docker container (DEV-5062)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/models/validation.py
```diff
@@ -32,6 +33,13 @@ class RDFGraphStrings:
     content_shapes: str
 
 
+@dataclass
+class ValidationFilePaths:
+    data_file: Path
+    shacl_file: Path
+    report_file: Path
```
seakayone, line 40: Currently you need to instantiate the cli validator with a directory which will be mounted to the container. 
The container expects all files to be validated in this directory and will also create the report in it.

IMHO it would be cleaner if the ValidationFilePaths would look something like this:

```python 
class ValidationFilePaths:
    directory: Path
    data_filename: str
    shacl_filename: str
    report_filename: str
```

This captures the fact that all files are in the same directory better than having to instantiate the cli validator with a directory every time. The validator class would be stateless and gets the directory to be mounted to the container from `ValidationFilePaths.data_directory`.
It would also allow you to reuse the validator without having to create it with different directories.

For example:
```python
validator=ShaclCliValidator()

paths = ValidationFilePaths(...,...,...,...)
result = validator.validate(paths)

paths2 = ValidationFilePaths(...,...,...,...)
result2 = validator.validate(paths2)
```

.github/actions/setup/action.yml
```diff
@@ -14,3 +14,6 @@ runs:
     - name: Install Python dependencies (uv will install python automatically, if no appropriate version is installed)
       run: uv sync --all-extras --dev
       shell: bash
+    - name: Create validate-data directory
+      run: mkdir -p /home/runner/.dsp-tools/validate-data
```
jnussbaum, line 18: @Nora-Olivia-Ammann This is coupled to `src/dsp_tools/commands/validate_data/constants.py`. If you modify it there, and forget to modify it here, you will run into hard-to-debug errors. I'd suggest to remove this, and instead add this to the python file:

```python
TURTLE_FILE_PATH..touch(exist_ok=True)
```

PR #1834, opened 2025-07-09
test(validate-data): separate validate-data e2e from the others

Nora-Olivia-Ammann:


.github/workflows/tests-on-push.yml
```diff
@@ -115,6 +114,32 @@ jobs:
...
+  validate-data-e2e-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout dsp-tools repo
+        uses: actions/checkout@v4
+      - name: Install uv, just, and dependencies
+        uses: ./.github/actions/setup
+      - name: docker info
+        run: docker info
+      - name: Check if the required directory exists
```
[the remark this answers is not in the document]
BalduinLandolt: I think you can remove it, unless you foresee that you might need it again soon, then you can just as well leave it in

.github/workflows/tests-on-push.yml
```diff
@@ -115,7 +114,37 @@ jobs:
       - name: e2e-tests
         run: just e2e-tests
 
+  validate-data-e2e-tests:
+    if: ${{ always() }}
+    needs: [e2e-tests]
```
BalduinLandolt, line 119: is there a particular reason for this dependency. Like this they will run _after_ the e2e-tests. And for quick feedback it would probably be nicer to have them run right away.

.github/workflows/tests-on-push.yml
```diff
@@ -115,7 +114,37 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: e2e-validate-data-logging.log
+          path: /home/runner/.dsp-tools/e2e-validate-data-logging.log
+          compression-level: 0
+
   legacy-e2e-tests:
+    if: ${{ always() }}
+    needs: [validate-data-e2e-tests]
```
BalduinLandolt, line 147: same as above

BalduinLandolt (approved):
Looks good. See my comment on the workflow order, though. I understand that un-parallelizing the tests of each test run is needed. But I would assume that jobs run in different VMs, so I don't think the jobs need to run in sequence. Or have you tried and verified that this is needed?

.github/workflows/tests-on-push.yml
Nora-Olivia-Ammann: I did not consider that they run on different vms that have their own directory. I was operating under the assumption that they share it, but this is not the case. I will remove it. Thanks!

BalduinLandolt: I mean... this is just my assumption - so it needs to be checked as well. 😉   
But it would make the test runs a bit faster, so let's hope I'm not wrong ^^

Nora-Olivia-Ammann: Copilot agreed that it is a separate vm with the directory and it does make sense. If it starts to become flaky again we can always introduce it again. Because, yes it is very slow like that.

PR #1840, opened 2025-07-10
refactor(validate-data): restructure function calls internally

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -79,8 +79,19 @@ def validate_data(
         is_on_prod_server=is_prod_like_server(creds.server),
     )
     auth = AuthenticationClientLive(server=creds.server, email=creds.user, password=creds.password)
-    graphs, used_iris = prepare_data_for_validation_from_file(filepath, auth, config.ignore_duplicate_files_warning)
-    return _validate_data(graphs, used_iris, config)
+
+    parsed_resources, shortcode, authorship_lookup, permission_ids = get_info_and_parsed_resources_from_file(
```
Nora-Olivia-Ammann, line 83: The reason why this used to be a separate function is that I need it for the tests often. But now it is not helpful because I want to do the pythonic check for duplicate filepaths before creating the graphs. Because when creating the graphs we add the shape that causes the problems. So we need to know before. 
So to help the tests I copied this function into test utils.

src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -79,8 +79,19 @@ def validate_data(
...
+        file=filepath,
+        api_url=auth.server,
+    )
+    return validate_parsed_resources(
```
Nora-Olivia-Ammann, line 87: In that function I plan to implement the check for too many duplicate files.

test/e2e_validate_data/util.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import Path
+
+from dsp_tools.clients.authentication_client import AuthenticationClient
+from dsp_tools.commands.validate_data.models.validation import RDFGraphs
+from dsp_tools.commands.validate_data.prepare_data.prepare_data import get_info_and_parsed_resources_from_file
+from dsp_tools.commands.validate_data.prepare_data.prepare_data import prepare_data_for_validation_from_parsed_resource
+
+
+def prepare_data_for_validation_from_file(
```
Nora-Olivia-Ammann, line 9: It is important for the tests that I get access to this datastructure, because this is the input of the function that we test most, `_get_validation_report`.
While these tests are e2e, they focus on individual results rather than the boolean did it pass or not.
To test an overall pass or not, we have other tests, there we will also test if the file duplication code behaves as expected with regards to server.

BalduinLandolt (approved):

PR #1851, opened 2025-07-14
refactor: reorganise GitHub workflow files (DEV-5154)

Nora-Olivia-Ammann:


.github/workflows/check-docs.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: markdownlint
+        run: just markdownlint
+
+  build-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout dsp-tools repo
+        uses: actions/checkout@v4
+      - name: Install uv, just, and dependencies
+        uses: ./.github/actions/setup
```
jnussbaum: From an economical and ecological stand point, it doesn't make sense to spin up an entire VM, and install everything, just for building the docs. I would argue that this step can also be done inside the first job.

```suggestion
```

jnussbaum (approved):

PR #1852, opened 2025-07-14
test(validate-data): add test data for date range validation

Nora-Olivia-Ammann:


testdata/validate-data/generic/content_correct.xml
```diff
@@ -147,4 +147,52 @@
         </uri-prop>
     </resource>
 
+    <!--
+    A check is conducted that ensures that the start date is equal or before the end date.
+    This is implemented for all date CE, BCE will be ignored.
```
jnussbaum, line 152:
```suggestion
    This is implemented for CE dates only. BCE will be ignored.
```

testdata/validate-data/generic/content_violation.xml
```diff
@@ -121,4 +121,45 @@
         </uri-prop>
     </resource>
 
+    <!-- Date with start later than end date -->
+
+    <resource label="wrong_date_range_full_JULIAN" restype=":ClassWithEverything" id="wrong_date_range_full_JULIAN">
+        <date-prop name=":testSubDate1">
+            <date>JULIAN:CE:2000-01-01:CE:1900-01-01</date>
```
jnussbaum, line 128: You only have one very simple error category: the start year is bigger than the end year.

I would add some cases where the year is equal, and only the months are wrong, and some where only the days are wrong.

jnussbaum (approved):

PR #1854, opened 2025-07-14
test(e2e): run all e2e tests of commands separately (DEV-5139)

Nora-Olivia-Ammann:


justfile
```diff
@@ -103,21 +103,35 @@ integration-tests *FLAGS:
...
-    # "--dist=loadfile" guarantees that all tests in a file are executed by the same worker
-    # see https://pytest-xdist.readthedocs.io/en/latest/distribution.html
-    uv run pytest -n=auto --dist=loadfile test/e2e/ {{FLAGS}}
+e2e-tests:
+    just e2e-tests-ingest-xmlupload
+    just e2e-tests-validate-data
+    just e2e-tests-xmlupload
+    just e2e-test-create-xmlupload
 
+# Run the end-to-end tests for the ingest-xmlupload command (with testcontainers)
```
jnussbaum, line 114: Most commands have a comment which explains what it does, but in most cases it's just happy talk. If you prefer consistency, you can leave the comments, but IMHO, it would be better to remove them, because they don't convey any relevant information.

Nora-Olivia-Ammann: I would leave them because they say that it is with containers while there is another that requires a running stack.

PR #1857, opened 2025-07-15
refactor(validate-data): allow `InputProblems` without resource ID and type

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/models/input_problems.py
```diff
@@ -91,7 +91,7 @@ class Severity(Enum):
 class ProblemType(StrEnum):
     GENERIC = "generic"
     FILE_VALUE = "file problem"
-    FILE_DUPLICATE = "file used several times"
+    FILE_DUPLICATE = "Duplicate Filepath / IIIF-URI"
```
Nora-Olivia-Ammann, line 94: Currently we ignore the string of this enum (the important message comes from SHACL). In the future usage we will use the string message, therefore I prefer to fix it already.

src/dsp_tools/commands/validate_data/process_validation_report/get_user_validation_message.py
```diff
@@ -115,11 +115,17 @@ def _filter_out_multiple_duplicate_file_value_problems(problems: list[InputProbl
     return result
 
 
-def _group_problems_by_resource(problems: list[InputProblem]) -> dict[str, list[InputProblem]]:
+def _group_problems_by_resource(
+    problems: list[InputProblem],
+) -> tuple[dict[str, list[InputProblem]], list[InputProblem]]:
     grouped_res = defaultdict(list)
+    problem_no_res_id = []
```
Nora-Olivia-Ammann, line 122: Currently there are no problems without resource ID but this will be the case after the check is only done in python.

src/dsp_tools/commands/validate_data/process_validation_report/get_user_validation_message.py
```diff
@@ -189,13 +195,20 @@ def _are_there_too_many_to_print(sorted_problems: SortedProblems, severity: Vali
...
+    messages.extend(messages_with_ids)
     return GRAND_SEPARATOR.join(messages)
 
 
 def _get_message_for_one_resource(problems: list[InputProblem]) -> str:
-    start_msg = f"Resource ID: {problems[0].res_id} | Resource Type: {problems[0].res_type}"
+    if problems[0].res_id:
+        start_msg = f"Resource ID: {problems[0].res_id} | Resource Type: {problems[0].res_type}"
+    else:
+        start_msg = ""
```
Nora-Olivia-Ammann, line 211: I printed the result out, it looks fine to me like that.

src/dsp_tools/commands/validate_data/validate_data.py
```diff
@@ -217,7 +217,7 @@ def _print_shacl_validation_violation_message(
 def _save_message_df_get_message_body(df: pd.DataFrame, severity: str, file_path: Path) -> str:
     out_path = file_path.parent / f"{file_path.stem}_validation_{severity}.csv"
     msg = f"Due to the large number of violations the information was saved at '{out_path}'"
-    df.to_csv(out_path)
+    df.to_csv(out_path, index=False)
```
Nora-Olivia-Ammann, line 220: I forgot that when I refactored it.

test/unittests/commands/validate_data/test_get_user_validation_message.py
```diff
@@ -403,6 +403,30 @@ def test_get_message_for_one_resource_several_problems(file_value, inexistent_li
     assert result == expected
 
 
+def test_get_message_duplicate_files_no_res_id():
+    file1 = InputProblem(
```
Nora-Olivia-Ammann, line 407: This is how the input problems for the duplicate files will look like

jnussbaum (approved):

PR #1861, opened 2025-07-15
refactor(validate-data): improve warnings and info e2e tests

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/models/input_problems.py
```diff
@@ -14,7 +14,7 @@
 
 @dataclass
 class ValidateDataResult:
-    passed: bool
+    no_problems: bool
```
Nora-Olivia-Ammann, line 17: This naming was confusing, because it only signifies that problems were found. But not if the upload may continue. For example if only problems with severity info are present the validation will pass, but this boolean is still `False` since there is something that must be communicated to the user.

There is a dedicated function that creates the validation status based on the severity levels.

test/e2e/commands/validate_data/test_generic_warning_and_info.py
```diff
@@ -153,19 +151,31 @@ def test_do_not_ignore_files_on_prod_env(self, parsed_resources_51_duplicate_ima
 
 class TestGetCorrectValidationResult:
     def test_no_violations_with_warnings_not_on_prod(self, no_violations_with_warnings):
-        result = _get_validation_status(no_violations_with_warnings, is_on_prod=False)
+        # this boolean signifies that there are problems of any severity level, but not if the validation will pass
```
Nora-Olivia-Ammann, line 154: These tests, test if the boolean that signifies the validation status is correctly created.

PR #1868, opened 2025-07-16
feat(validate-data): raise validation severity of duplicate files from `INFO` to `WARNING` (DEV-5167)

Nora-Olivia-Ammann:


test/e2e/commands/validate_data/test_generic_warning_and_info.py
```diff
@@ -58,105 +53,18 @@ def no_violations_with_info(
     return _validate_data(graphs, used_iris, parsed_resources, CONFIG)
 
 
-@pytest.fixture(scope="module")
```
Nora-Olivia-Ammann, line 61: This is no longer necessary because we do not have a cut-off point for a certain number of duplicates. The logic about prod and test env is also not necessary to test, because that is done in other tests.

test/e2e/commands/validate_data/test_generic_warning_and_info.py
```diff
@@ -220,23 +140,6 @@ def test_no_violations_with_info(self, no_violations_with_info):
             assert one_result.problem_type == expected_info[1]
             assert one_result.res_id == expected_info[0]
 
-    def test_no_violations_with_info_ignore_duplicate_files_warning(self, authentication, shacl_validator):
```
Nora-Olivia-Ammann, line 223: Because the severity is warning, this test does not make sense anymore.

jnussbaum (approved):

PR #1883, opened 2025-07-18
fix(xmlupload): use slightly shorter timeout than traefik's 10 min

jnussbaum:


jnussbaum:
@Nora-Olivia-Ammann This is a follow-up to this discussion: https://www.notion.so/Weird-behavior-when-uploading-2-7-GB-video-file-from-EKWS-via-dsp-tools-xmlupload-2318946b7d40800ebabdd52d4bbeea3a?d=2328946b7d4080cf8501001c6ded426d&source=copy_link#2318946b7d4080e9aab1e9b3be940b4b

PR #1895, opened 2025-07-25
feat: change organisation of log files (DEV-5228)

Nora-Olivia-Ammann:


jnussbaum (commented):
There are some problems with this. Unfortunately I don't have a simple idea how to solve them...

jnussbaum (approved):

PR #1902, opened 2025-08-07
fix(xmllib): don't crash if the listnode lookup does not find a label with the specified language (DEV-5247)

Nora-Olivia-Ammann:


testdata/json-project/test-list-lookup.json
Nora-Olivia-Ammann, line 1: I only want one warning

PR #1904, opened 2025-08-07
test: use tempdir for xmlupload

jnussbaum:


test/e2e/commands/xmlupload/conftest.py
```diff
@@ -56,11 +56,13 @@ def _xmlupload_minimal_correct(create_generic_project, creds) -> None:
...
-    yield
-    if found := list(Path.cwd().glob(f"id2iri_{PROJECT_SHORTCODE}_localhost_*.json")):
-        found[0].unlink()
+def _xmlupload_text_parsing(create_generic_project: None, creds: ServerCredentials) -> None:
+    absolute_xml_path = Path("testdata/xml-data/generic_project_text_parsing.xml").absolute()
+    original_cwd = Path.cwd()
+    with TemporaryDirectory() as tmpdir:
+        with pytest.MonkeyPatch.context() as m:
+            m.chdir(tmpdir)
+            assert xmlupload(absolute_xml_path, creds, str(original_cwd))
```
jnussbaum, line 65: I just copied the same pattern from above. Currently it doesn't matter, but as soon as we use this fixture in another file, we could run into the same problem as above.

PR #1906, opened 2025-08-08
feat(xmllib): helper function create list if the input is non-empty (DEV-5249)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/helpers.py
```diff
@@ -1264,6 +1264,49 @@ def create_list_from_string(string: str, separator: str) -> list[str]:
     return [strpd for x in string.split(separator) if (strpd := x.strip())]
 
 
+def create_list_if_input_is_non_empty(input_value: Any, separator: str) -> list[str]:
```
jnussbaum, line 1267: The function name explains implementation details (like "if input is non-empty") instead of describing the problem to solve. As a user, I want that the name tells me what the function does for me. For example: `split_string()`.

src/dsp_tools/xmllib/helpers.py
```diff
@@ -1264,6 +1264,49 @@ def create_list_from_string(string: str, separator: str) -> list[str]:
...
+    """
+    If the input value is a non-empty string it returns the separated list.
```
jnussbaum, line 1269: Here you directly start with special cases. Instead, I would expect 1 simple introductory sentence which describes the problem that this function solves.

For example: Split up a string (or other input value) into a list of strings, using the provided separator.

src/dsp_tools/xmllib/helpers.py
```diff
@@ -1264,6 +1264,49 @@ def create_list_from_string(string: str, separator: str) -> list[str]:
...
+    Args:
+        input_value: input value to check and convert
+        separator: The character that separates the different values in the string.
+            For example, a comma or newline.
+
+    Returns:
+        The list that results from splitting the input string.
+
+    Examples:
+        ```python
```
jnussbaum, line 1283: It's good that you include special cases, but the first example should be a straightforward one, like e.g. 

```python
result = xmllib.create_list_if_input_is_non_empty("one, two, three", ",")
# result == ["one", "two", "three"]
```

jnussbaum (approved):

src/dsp_tools/xmllib/helpers.py
Nora-Olivia-Ammann: How about `create_list_from_input`? That would be more precise and consistent. Although I personally would always wonder what happens if the input is empty. I would not expect it to be able to handle that by default

jnussbaum: There are already 2 similar functions that I wasn't aware of:
- create_list_from_string
- create_non_empty_list_from_string

I'm a bit sceptical about having too many functions that do similar things. Is it really necessary to have 3 such functions?

If yes, then I understand why you named it that way, and then I agree with your original naming.

PR #1921, opened 2025-08-18
fix: empty log files written during long-lived debugging sessions (DEV-5281)

jnussbaum:


src/dsp_tools/config/logger_config.py
```diff
@@ -1,65 +1,106 @@
...
+                format=text_format,
+                backtrace=False,
+                diagnose=False,
+                rotation=rotation_size,
+                retention=2,  # when starting a new one, the old one shouldn't be deleted immediately
+            )
+
+    def get_log_file(self) -> Path:
+        """Get the path to the log file. As a fallback, return the parent directory."""
+        if self._log_file is None:
```
Nora-Olivia-Ammann, line 74: I find this implementation very confusing, not only do you have a calling function (Line 89) that shares the same name, but should it not be an error if the log file you wish to retrive does not exist. So this use case should never exist and in my opinion raise an Internal Error because we did not use the LoggerService as intended.

src/dsp_tools/config/logger_config.py
```diff
@@ -1,65 +1,106 @@
...
+            return self._make_and_get_logs_directory()
+        return self._log_file
+
+    def get_warnings_file(self) -> Path:
```
Nora-Olivia-Ammann, line 78: Same as above.

src/dsp_tools/error/exceptions.py
```diff
@@ -1,7 +1,7 @@
 from dataclasses import dataclass
 from pathlib import Path
 
-from dsp_tools.config.logger_config import LOGGER_SAVEPATH
+from dsp_tools.config.logger_config import get_log_file
```
Nora-Olivia-Ammann, line 4: Why did you need to write another function to call a public function on an object? It creates a lot of lines of code that can be eliminated if we can call the function directly on the object...

src/dsp_tools/config/logger_config.py
```diff
@@ -1,65 +1,106 @@
...
 
 load_dotenv()
 
 
-def _make_and_get_logs_directory() -> Path:
-    """Get the base .dsp-tools directory, creating it if it doesn't exist."""
-    base_dir = Path.home() / ".dsp-tools" / "logs"
-    base_dir.mkdir(exist_ok=True, parents=True)
-    return base_dir
+@dataclass
```
Nora-Olivia-Ammann, line 14: In my opinion this design pattern is a bit overkill and creates unnecessary complexity. It very much looks like AI generated code, which in my opinion often creates a log of boilerplate code.

src/dsp_tools/config/logger_config.py
```diff
@@ -1,65 +1,106 @@
...
-WARNINGS_SAVEPATH = Path("warnings.log")
+    def configure(self) -> None:
+        """
+        Configure the log files. Currently, there are three sinks:
+          - timestamp_logging.log in ~/.dsp-tools/logs/ contains the entire stack-trace
+          - warnings.log in the cwd only with level warning and higher for the user (no stack-trace)
+            OR a complete logging.log file with the stack-trace if configured in the .env
+          - print output on the terminal, formatted the same as the warnings.log
+        """
+        logger.remove()  # If this is not removed, the default formatting is also printed out on the terminal
```
Nora-Olivia-Ammann, line 37: From what I read online (eg. https://medium.com/@balakrishnamaduru/mastering-logging-in-python-a-singleton-logger-with-dynamic-log-levels-17f3e8bfa2cf) the singleton logging makes sense when you want to configure it globally. Logguru already takes care of this for us...

PR #1923, opened 2025-08-19
test(xmlupload): fix mypy warning about wrongly escaped characters

Nora-Olivia-Ammann:


jnussbaum (approved):
Good catch, thank you. I have been seeing this for a long time, but didn't know how to fix.

PR #1935, opened 2025-08-26
fix(excel2json): put "default_permissions_overrule" at correct position (DEV-5345)

jnussbaum:


src/dsp_tools/commands/excel2json/project.py
```diff
@@ -194,11 +195,12 @@ def _old_create_project_json(
     if not success:
         overall_success = False
     project = get_json_header(Path(data_model_files) / "json_header.xlsx").to_dict()
+    if default_permissions_overrule.non_empty():
+        project["project"]["default_permissions_overrule"] = default_permissions_overrule.serialize()
```
jnussbaum, line 199: Moving these 2 lines up has turned out to be ineffective. I had to write a dedicated sorting function. But I thought that it makes anyway more sense to have these 2 lines up here.

PR #1954, opened 2025-09-10
fix(excel2json): enable references to other ontologies in resource cardinalities (DEV-5416)

Nora-Olivia-Ammann:


docs/data-model/excel2json.md
```diff
@@ -275,6 +275,14 @@ The expected columns are:
 
 - [`Property`](./json-project/ontologies.md#resource-cardinalities) 
   (mandatory): The unique identifier of the property
+    - If the property references the ontology you are working on
```
Nora-Olivia-Ammann, line 278: @Notheturtle @jnussbaum 
I had to consider the special case of seqnum and isPartOf. With the current code it would not be possible to use `knora-api:senum` in the cardinalities section even though we allow this. Because it always adds a `:`. To fix that and to allow the creation of properties in the ontology called `seqnum` and `isPartOf` I added this special behaviour. It is a bit complex but it produces the correct results. Disallowing properties of the same name is in my opinion not an option because that would be a breaking change. I am always open to suggestions to clarify this further here.

PR #1958, opened 2025-09-15
feat: remove `template` CLI command (DEV-5385)

Nora-Olivia-Ammann:


Nora-Olivia-Ammann:
> I assume the following directory should be deleted too: https://github.com/dasch-swiss/dsp-tools/tree/main/src/dsp_tools/resources/0100-template-repo Also, I think the docs need to be updated.

Yes, thanks!

BalduinLandolt (approved):
No documentation to remove, then?

PR #1963, opened 2025-09-25
debug failing excel2json integration test - sorting of lists.xlsx

jnussbaum:


src/dsp_tools/commands/excel2json/project.py
```diff
@@ -148,7 +148,7 @@ def _validate_folder_structure_and_get_filenames(data_model_files: Path) -> tupl
     processed_files.extend(processed_onto)
     listfolder, processed_lists = _get_and_validate_list_folder(data_model_files)
     processed_files.extend(processed_lists)
-    print("The following files will be processed:")
+    print("The following files will be processed:")  # MacOS: lists.xlsx, list3.xlsx  |  GitHub: list3.xlsx, lists.xlsx
```
jnussbaum, line 151: I think this comment should be removed again

```suggestion
    print("The following files will be processed:")
```

PR #1973, opened 2025-10-06
feat(xmllib)!: raise error when using `create_list_from_string` (DEV-5253)

Nora-Olivia-Ammann:


mkdocs.yml
```diff
@@ -97,6 +97,9 @@ plugins:
             show_signature: false
             show_docstring_raises: false
             show_root_toc_entry: false
+            filters:
```
Nora-Olivia-Ammann, line 100: Follow-up ticket to remove this after the function is deleted: https://linear.app/dasch/issue/DEV-5460/delete-function-create-list-from-string

PR #1980, opened 2025-10-07
feat: add functionality to replace ids with iris from mapping dynamically (DEV-5473)

Nora-Olivia-Ammann:


.vulture_whitelist.py
```diff
@@ -7,6 +7,7 @@
 from dsp_tools.utils.ansi_colors import YELLOW
 from dsp_tools.utils.data_formats.date_util import is_full_date
 from dsp_tools.utils.data_formats.shared import check_notna
+from dsp_tools.utils.replace_id_with_iri import use_id2iri_mapping_to_replace_ids
```
Nora-Olivia-Ammann, line 10: I need to integrate this into `xmlupload`, `validate-data` and `ingest-xmlupload`. To keep it organised I did not use the functionality right now. After it is integrated we can remove it here.

src/dsp_tools/utils/json_parsing.py
Nora-Olivia-Ammann, line 1: I wanted to have a nice json parsing method. It would be nicer for the existing function to already use this function, therefore, I integrated it already. It did not seem worth it to me to have a separate PR. Because I used better errors, I also had to change some tests that are unrelated to id2iri.

PR #1985, opened 2025-10-08
feat(excel2json)!: deprecate gui-attributes that do not have an effect in the APP (DEV-5482)

Nora-Olivia-Ammann:


src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -208,15 +209,26 @@ def _check_missing_values_in_row(df: pd.DataFrame) -> None | MissingValuesProble
 
 
 def _check_compliance_gui_attributes(df: pd.DataFrame) -> dict[str, pd.Series[bool]] | None:
-    mandatory_attributes = ["Spinbox", "List"]
+    mandatory_attributes = ["List"]
```
Nora-Olivia-Ammann, line 212: It would add a lot of complexity to do a warning and then deprecation. Since this is not used by many, and you can quickly remove the attributes, I decided to completely deprecate the other attributes.

src/dsp_tools/commands/excel2json/properties.py
```diff
@@ -269,49 +281,34 @@ def _get_gui_attribute(
...
-        gui_attrib[attrib_key] = attrib_val
-    return gui_attrib
+    if not gui_list:
+        return GuiAttributes({})
+    if len(gui_list) > 1:
+        raise InvalidGuiAttributeError(err_msg)
+    attrib_key, attrib_val = _extract_information_from_single_gui_attribute(gui_list.pop(0))
+    if attrib_key != "hlist":
+        raise InvalidGuiAttributeError(err_msg)
+    return GuiAttributes({attrib_key: attrib_val})
```
jnussbaum, line 304: In the class `GuiAttributes`, you can now change the type from `dict[str, int | str | float]`  to `dict[str, str]`

test/unittests/commands/excel2json/test_properties.py
```diff
@@ -132,38 +133,36 @@ def test_rename_deprecated_hlist() -> None:
...
 
 def test_get_gui_attribute() -> None:
-    original_df = pd.DataFrame({"gui_attributes": [pd.NA, "max=1.4, min:1.2", "hlist:", "234345", "hlist: languages,"]})
+    original_df = pd.DataFrame(
+        {"gui_attributes": [pd.NA, "hlist:", "234345", "size: 32, maxlength: 128", "hlist: languages,"]}
+    )
     assert e2j._get_gui_attribute(df_row=cast("pd.Series[Any]", original_df.loc[0, :]), row_num=2) is None
 
-    res_1 = e2j._get_gui_attribute(df_row=cast("pd.Series[Any]", original_df.loc[1, :]), row_num=3)
+    res_1 = e2j._get_gui_attribute(df_row=cast("pd.Series[Any]", original_df.loc[1, :]), row_num=4)
```
jnussbaum, line 149: row_num should be 3

test/unittests/commands/excel2json/test_properties.py
```diff
@@ -132,38 +133,36 @@ def test_rename_deprecated_hlist() -> None:
...
     assert isinstance(res_1, InvalidExcelContentProblem)
-    assert res_1.excel_position.row == 3
-    assert res_1.actual_content == "max=1.4, min:1.2"
+    assert res_1.expected_content == "The only valid gui-attribute is 'hlist' for the gui-element 'List'."
+    assert res_1.excel_position.row == 4
+    assert res_1.actual_content == "hlist:"
 
-    res_2 = e2j._get_gui_attribute(df_row=cast("pd.Series[Any]", original_df.loc[2, :]), row_num=4)
+    res_2 = e2j._get_gui_attribute(df_row=cast("pd.Series[Any]", original_df.loc[2, :]), row_num=5)
```
jnussbaum, line 155: row_num should be 4

test/unittests/commands/excel2json/test_properties.py
```diff
@@ -132,38 +133,36 @@ def test_rename_deprecated_hlist() -> None:
...
     assert isinstance(res_2, InvalidExcelContentProblem)
-    assert res_2.excel_position.row == 4
-    assert res_2.actual_content == "hlist:"
+    assert res_2.expected_content == "The only valid gui-attribute is 'hlist' for the gui-element 'List'."
+    assert res_2.excel_position.row == 5
+    assert res_2.actual_content == "234345"
 
     res_3 = e2j._get_gui_attribute(df_row=cast("pd.Series[Any]", original_df.loc[3, :]), row_num=5)
```
jnussbaum, line 161: ...because here the row_num is 5

src/dsp_tools/commands/excel2json/properties.py
jnussbaum: I think there is a misunderstanding about the term "deprecation". Deprecation means issueing a warning, but the feature stays in place. That's not what your PR does: You remove the feature. So you should rename the PR title to "remove" instead of "deprecate".

jnussbaum: So in fact you decided to skip the deprecation step, and remove immediately. Personally I don't mind this, because it makes life easier for us. But some time ago it was decided that we must not remove any features without a period of deprecation.

Probably you're right and not many people used the gui_attributes in their Excel sheets, so the 2 of us could now decide to ignore the official policy that we have 🙈

jnussbaum (approved):

src/dsp_tools/commands/excel2json/properties.py
Nora-Olivia-Ammann: Right I forgot about the naming nuance, thanks.
While I am a big fan of softly and slowly making big changes, I think in this case the effort is not worth it. Because hardly anyone uses it, and then they also would have to use those gui-attributes. Additionally the change is very quickly done. I think Ivan was in the throw it away spirit, so for this feature I would remove it. I would not do that for create, nor would I change the json schema right away. But I think for excel2json we can break the rule.

PR #2049, opened 2025-11-11
fix(xmlupload): allow duplicate files/iiif links also on prod (DEV-5591)

jnussbaum:


Nora-Olivia-Ammann (commented):
I don't understand why this is necessary. You are allowed to ignore it on prod. Therefore, the warning will never appear and an upload is not blocked.

jnussbaum:
Good point, thanks for mentioning this. I just blindly believed what RDU told me, without checking this myself. I will inform RDU that they should use the `--ignore-duplicate-files-warning` flag. Then I will close this PR.

Nora-Olivia-Ammann:
> Good point, thanks for mentioning this. I just blindly believed what RDU told me, without checking this myself. I will inform RDU that they should use the `--ignore-duplicate-files-warning` flag. Then I will close this PR.

Perhaps we could talk to them and see if we can improve the wording so that they will understand this in the future too?

PR #2276, opened 2026-04-24
fix(validate-data): change Docker Volume Mount Path to make it SELinux save (DEV-6303)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/shacl_cli_validator.py
```diff
@@ -18,8 +18,10 @@ def validate(self, file_paths: ValidationFilePaths) -> SHACLValidationReport:
         try:
             self._run_validate_cli(file_paths)
         except subprocess.CalledProcessError as e:
-            logger.exception(f"Docker command failed with {e.returncode}: {e.stderr or ''}")
-            raise ShaclValidationCliError(e.returncode, e.stderr or "")
+            stdout = e.stdout or ""
+            stderr = e.stderr or ""
+            logger.exception(f"Docker command failed with {e.returncode}: stdout={stdout!r}, stderr={stderr!r}")
```
seakayone, line 23: This line is a good and important change. 

Before stdout was ignored when logging. Currently the shacl-cli logs out to stdout in the case that the input files for shapes and/or data were not found. This will allow for better diagnostics.

seakayone (approved):

PR #2278, opened 2026-04-27
docs: add permissions guide (RDU-169)

jnussbaum:
## Summary

- In the VRE Permissions Workshop on April 27, we decided to write a user-friendly introduction to the DSP two-layer permission system (project-wide JSON defaults + per-resource XML overrides).
- This guide documents the Status Quo. Improvements to the DSP permissions system that were discussed in the workshop will be added later. 
- Get a local preview with `mkdocs serve` 


docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+A group can hold exactly one right, and each right includes all the rights below it:
+
+| Right | Name                  | What it allows                                                                   |
+| ----- | --------------------- | -------------------------------------------------------------------------------- |
+| `CR`  | Change Rights         | Change the access permissions on a resource or value                             |
+| `D`   | Delete                | Mark a resource/value as deleted (original is preserved)                         |
+| `M`   | Modify                | Edit a resource/value                                                            |
+| `V`   | View                  | View at full quality                                                             |
+| `RV`  | Restricted View       | Images: view at reduced resolution or with a watermark; Video/Audio: No download |
+| —     | _(no right assigned)_ | No access at all                                                                 |
```
gNahcab: it is writen "each right includes all the rights below it", while CR does include D, E, M and V, V does not include RV and _. This is stated below, but then it might be reasonable not to write "each right includes all the rights below it" above.

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+
+> `CR` implies `D`, which implies `M`, which implies `V`.
+> `RV` is a separate, more restrictive access level — it does not follow from `V`.
+
+### The Three Access Tiers
+
+Most DSP projects are configured around three conventional access tiers:
+
+| Tier | Who can access it |
```
gNahcab: I guess this refers to the built-in user groups mentioned above? If so I would refer to them explicitly here (so instead of "everyone", something like "SystemAdmin & ProjectAdmin &Project Member & KnownUser & UnknownUser")

mkdocs.yml
Nora-Olivia-Ammann, line 1: A link to the document, at the place where you set the permissions in the JSON documentation would be helpful.

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+
+## The Building Blocks
+
+### Built-in User Groups
+
+Every user on DSP belongs to exactly one built-in group:
+
+| Group           | Who is in it                                           |
+| --------------- | ------------------------------------------------------ |
+| `SystemAdmin`   | System-level administrators                            |
```
Nora-Olivia-Ammann: perhaps note that these are dasch employees only

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+
+### Rights Hierarchy
+
+A group can hold exactly one right, and each right includes all the rights below it:
+
+| Right | Name                  | What it allows                                                                   |
+| ----- | --------------------- | -------------------------------------------------------------------------------- |
+| `CR`  | Change Rights         | Change the access permissions on a resource or value                             |
+| `D`   | Delete                | Mark a resource/value as deleted (original is preserved)                         |
+| `M`   | Modify                | Edit a resource/value                                                            |
```
Nora-Olivia-Ammann: If you note above that the original is preserved you should include that here too, because they are.

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+| `V`   | View                  | View at full quality                                                             |
```
Nora-Olivia-Ammann: Is this also for audio?, because then it should be noted.
Perhaps you could say `access to image / video / audio without restrictions`

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+    ],
+    "limited_view": "all"
+}
+```
+
+The two available restriction levels are:
+
+- **`private`**: A list of class or property names.
+    - For a **class**: resources of that class are invisible to `UnknownUser` and `KnownUser`.
+    - For a **property**: the property's content is hidden, but the rest of the resource remains public.
```
gNahcab: "remains public" means 'visible to Unknown User and Known User', right? I would then write this instead of "remains public" ( I would clearly differentiate between "default_permissions"-mode "public" and using the term "public", if it makes things difficult, better rename "public" to "public_default" and "private" to "private_default")

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+- **`limited_view`**: A list of multimedia classes whose files are restricted for outsiders.
```
gNahcab: who are those "outsiders"? "Unknown User" and "Known User"? I would write it explicitly

docs/permissions-guide.md
Nora-Olivia-Ammann: Maybe say that it is not only assets but also data?

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+
+### The Three Access Tiers
+
+Most DSP projects are configured around three conventional access tiers:
+
+| Tier | Who can access it |
+| -------------- | ------------------------------------------------------------------------------- |
+| `public` | Everyone, including anonymous visitors |
+| `limited_view` | Everyone, but outsiders see images at reduced quality (blurred or watermarked) and cannot download audio or video files |
+| `private` | Project members and project admins only — invisible to all outsiders |
```
Nora-Olivia-Ammann: Since you listed it all above perhaps you could include a mapping which permissions are the result?

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+```
+
+The two available restriction levels are:
+
+- **`private`**: A list of class or property names.
+    - For a **class**: resources of that class are invisible to `UnknownUser` and `KnownUser`.
+    - For a **property**: the property's content is hidden, but the rest of the resource remains public.
+- **`limited_view`**: A list of multimedia classes whose files are restricted for outsiders.
+  For images: blurred or watermarked. For audio/video: streaming only, no download.
+  Only the multimedia file is affected — the rest of the resource stays public.
```
gNahcab: "stays public": same here, what does that mean? If it means, the "default_permissions" "public" is still applied, should we maybe write this explicitly? Otherwise I have the impression that unfamiliar people with the topic could think that "public" means something different here.

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+
+```json
+"default_permissions": "private"
+```
+
+- **`public`**: `UnknownUser` and `KnownUser` can view everything. This is the open-access default.
+- **`private`**: Only `ProjectMember` and `ProjectAdmin` can view anything.
+  Everything is locked down by default for `UnknownUser` and `KnownUser`.
+
+### `default_permissions_overrule` (only when `default_permissions` is `"public"`)
```
Nora-Olivia-Ammann: I would write 
```
### `default_permissions_overrule` (optional)

and then in bold for the rest.
```

docs/permissions-guide.md
Nora-Olivia-Ammann: I would make a list of the multimedia files and the options.

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+
+```json
+"default_permissions": "public",
+"default_permissions_overrule": {
+    "private": [
+        "my-onto:InternalCorrespondence",
+        "my-onto:hasSSN"
+    ],
+    "limited_view": [
+        "my-onto:HighResPhoto",
```
Nora-Olivia-Ammann, line 122: perhaps replace one of these with `"all"`

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+        "my-onto:ArchiveScan"
+    ]
+}
+```
+
+## Layer 2: Per-Resource Overrides (XML)
+
+The XML data file lets you override the project defaults for individual resources, values,
+and multimedia assets. This layer uses two constructs: **permission IDs** and the **`permissions` attribute**.
```
gNahcab: Why two constructs? Better:"This layer uses **permission IDs** defined in the **`permissions` attribute**."?

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+
+### Defining Permission IDs With `<permissions>`
+
+At the top of the XML file, before any `<resource>` elements,
+you can define named permission sets using `<permissions>` elements.
+
+The three canonical IDs cover the most common access patterns:
+
+```xml
+<permissions id="public">
```
gNahcab, line 142: I would use not the values from "default_permissions" here to avoid misunderstandings; maybe use "custom_public" , "custom_private", "custom_limited_view" here?

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+    <text-prop name=":hasTitle">
+        <text encoding="utf8" permissions="public">Board Meeting, January 2024</text>
+    </text-prop>
+    <text-prop name=":hasMinutes">
+        <text encoding="utf8" permissions="private">... confidential content ...</text>
+    </text-prop>
+</resource>
+```
+
+Notice that `<resource permissions="public">` only sets the resource's own permission.
```
gNahcab: maybe add here that this means that  "default_permissions" are still applied to the members of the resource even though the resource's "permissions" is set to some custom-permissions.

docs/permissions-guide.md
gNahcab: Does Modify include "create resource"? As far as I can see the right to create resources is not explicitly mentioned here, can we add this?

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+**Properties do not inherit from their resource** — each value
+must have its own `permissions` attribute if you want to control access at that level.
+
+If you omit the `permissions` attribute entirely, the project defaults apply.
+
+## Custom Groups
+
+In addition to the built-in groups, you can define project-specific groups in the JSON project file.
+Custom groups are additional membership tiers that sit between `ProjectMember` and `ProjectAdmin`:
+they let you grant finer-grained rights to a subset of project members.
```
gNahcab, line 195: as an example: student assistants which should be able to modify resources, but not delete them

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+
+A custom group cannot hold fewer rights than `ProjectMember`, and cannot hold more rights than
+`ProjectAdmin`.
```
gNahcab, line 198: Add: "Custom Groups are defined in the JSON project file, but they can also be used in XML-files."

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+```
+
+The `name` and `descriptions` fields are mandatory.
+
+### Using Custom Groups in XML
+
+Reference a custom group as `project-shortname:groupname` in a `<permissions>` block:
+
+```xml
+<permissions id="editors_only">
```
Nora-Olivia-Ammann, line 228: It is weird that in the json example the name is not the same as it is here. I assume this should be the same string?

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+"groups": [
+    {
+        "name": "editors",
+        "descriptions": {"en": "Team members with editing rights"}
+    }
+]
+```
+
+After project creation, the `ProjectAdmin` adds the actual team to the `editors` group
+via DSP-APP, and adds external academics as `ProjectMember`.
```
Nora-Olivia-Ammann, line 328: Can't you do that in the json?

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+adjustment. Write a Python script using the `dsp-permissions-script` template that makes the
+following changes:
+
+- **Administrative Permissions**: remove `ProjectResourceCreateAllPermission` from `ProjectMember`.
+- **DOAPs**: degrade `ProjectMember` from `D` to `V`.
+- **DOAPs**: grant `D` to the `my-project:editors` custom group.
+
+**Layer 2 — XML** (per-resource overrides):
+
+After running the permissions-scripts, no `permissions` attribute is required on resources or values in the XML.
```
Nora-Olivia-Ammann, line 345: I would note in bold that this must happen before the xmlupload

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+```
+
+### 4. Public Project With Watermarked Images
+
+**Situation**: All metadata is freely accessible, but image files are served watermarked or at
+reduced resolution for outsiders because the images are still under copyright.
+
+**Access goals**:
+
+- `UnknownUser` / `KnownUser` → Restricted View on images (watermark or reduced resolution)
```
Nora-Olivia-Ammann, line 363: I think you should not call it restricted here but limited?

Nora-Olivia-Ammann (approved):

Nora-Olivia-Ammann (approved):

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+> `RV` is a separate, more restrictive access level — it does not follow from `V`.
+
+### The Three Access Tiers
+
+Most DSP projects are configured around three conventional access tiers:
+
+| Tier | Who can access it |
+| -------------- | ------------------------------------------------------------------------------- |
+| `public` | Everyone, including anonymous visitors |
+| `limited_view` | Everyone, but outsiders see images at reduced quality (blurred or watermarked) and cannot download audio or video files |
```
flo7ian: maybe change the term "outsiders"? and work with "Non-Project-Members" / "Known/Unknown Users"
in the line below it could also be skipped

docs/permissions-guide.md
```diff
@@ -0,0 +1,515 @@
...
+        <text encoding="utf8" permissions="public">Board Meeting, January 2024</text>
+    </text-prop>
+    <text-prop name=":hasMinutes">
+        <text encoding="utf8" permissions="private">... confidential content ...</text>
+    </text-prop>
+</resource>
+```
+
+Notice that `<resource permissions="public">` only sets the resource's own permission.
+**Properties do not inherit from their resource** — each value
```
flo7ian: i would move the following lines up before the example. and might change it a bit? because for me it's unclear if you just set the premission on a specific property, the property with unset permissions inherits default values

PR #2280, opened 2026-04-27
refactor(validate-data): use `oxrdflib` backend in `rdflib` to improve performance speed (DEV-6310)

Nora-Olivia-Ammann:


Nora-Olivia-Ammann:
Closed because oxigraph converts primitive data types. For example `xsd:dateTimeStamp` becomes `xsd:dateTime`, because the first is a derivative of the second. This breaks our validation because we require the original data type. As this is a implementation of the database it uses, it is not possible to prevent this.

PR #2290, opened 2026-05-05
fix(validate-data): add user ID to shacl-cli command to enable file access on some systems (DEV-6431)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/shacl_cli_validator.py
SamuelBoerlin, line 51: Unrelated to this PR's changes, but constructing a string and then splitting seems prone to command injection and breaking depending on the string template's arguments. E.g. a path like `file_paths.directory` containing a whitespace would probably break this command as it'd be interpreted as multiple arguments instead of one. It'd be safer to construct an array explicitly.

SamuelBoerlin (approved):
LGTM. Other comment is unrelated to PR changes and not a blocker, but may want to look into that.

src/dsp_tools/commands/validate_data/shacl_cli_validator.py
```diff
@@ -39,8 +40,10 @@ def _run_validate_cli(self, file_paths: ValidationFilePaths) -> None:
         docker_spec = yaml.safe_load(docker_file.read_bytes())
         docker_image = docker_spec["image"]
 
+        uid = os.getuid() if hasattr(os, "getuid") else None
+        user_flag = f"--user {uid} " if uid is not None else ""
         d_cmd = (
-            f"docker run --rm -v {file_paths.directory.absolute()}:/data:z {docker_image} "
+            f"docker run --rm {user_flag}-v {file_paths.directory.absolute()}:/data:z {docker_image} "
```
jnussbaum: `{user_flag}-v`  looks as if the whitespace in between was forgotten. I saw that `user_flag` has a trailing whitespace. it works, but looks confusing. The other way round, it would look more natural to me.

Nora-Olivia-Ammann: ok thanks!

PR #2299, opened 2026-05-13
feat(validate-data): prevent resources from linking to themselves (DEV-6323)

Nora-Olivia-Ammann:
The API does not allow resources to link to themselves in a link value. Stand-off links in text values are permitted.

### DIFF
--- a/src/dsp_tools/commands/validate_data/models/input_problems.py
+++ b/src/dsp_tools/commands/validate_data/models/input_problems.py
@@ -118,4 +118,3 @@ class ProblemType(StrEnum):
     LINK_TARGET_IS_IRI_OF_PROJECT = "Linked resource is an IRI of the project"
     LINK_TARGET_NOT_FOUND_IN_DB = "Linked resource is an IRI and could not be found in the database."
     INEXISTENT_LINKED_RESOURCE = "Linked Resource does not exist"
-    DUPLICATE_VALUE = "Your input is duplicated"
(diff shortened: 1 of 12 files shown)


src/dsp_tools/commands/validate_data/models/input_problems.py
```diff
@@ -118,4 +118,3 @@ class ProblemType(StrEnum):
     LINK_TARGET_IS_IRI_OF_PROJECT = "Linked resource is an IRI of the project"
     LINK_TARGET_NOT_FOUND_IN_DB = "Linked resource is an IRI and could not be found in the database."
     INEXISTENT_LINKED_RESOURCE = "Linked Resource does not exist"
```
Nora-Olivia-Ammann, line 120: This moved to the GENERIC type. This type uses the message specified in the ttl for the user.

src/dsp_tools/commands/validate_data/process_validation_report/query_validation_result.py
```diff
@@ -209,8 +209,6 @@ def _query_one_without_detail(  # noqa:PLR0911 (Too many return statements)
             return _query_for_less_than_or_equal_violation(base_info, results_and_onto, data, msg)
         case DASH.ClosedByTypesConstraintComponent:
             return _query_for_non_existent_cardinality_violation(base_info, results_and_onto, data)
-        case SH.SPARQLConstraintComponent:
```
Nora-Olivia-Ammann, line 212: Before the duplicate value was the only sparql constraint so this made sense. Now we have two and it makes sense to move them to the generic type.

src/dsp_tools/resources/validate_data/api-shapes.ttl
```diff
@@ -53,7 +53,31 @@ api-shapes:UniqueValue_Shape
             HAVING (COUNT(?value) > 1)
                     """ ;
                    sh:severity sh:Violation ;
-                   sh:message  "A resource may not have the same property and value more than one time." ;
+                   sh:message  "Your input is duplicated." ;
```
Nora-Olivia-Ammann, line 56: This message is the one from the original enum

jnussbaum (approved):

PR #2303, opened 2026-05-19
feat: dispatch dsp-docs bump on release publish (DEV-6438)

mpro7:
[DEV-6438](https://linear.app/dasch/issue/DEV-6438)

## Summary

- Adds `versions.env` as the structured source of truth for stack versions in `src/dsp_tools/resources/start-stack/`. `bump_stack_versions.py` now only writes this file; `docker-compose.yml` interpolates `${API}`/`${APP}`/`${DB}` from it at `docker compose --env-file versions.env` time.
- Adds a `dispatch-dsp-docs-bump` job to `publish-release-to-pypi.yml` that fires `bump-release.yml` in `dsp-docs` after every successful PyPI publish.

### `versions.env` source of truth (`refactor(start-stack)`)

- `src/dsp_tools/resources/start-stack/versions.env`: structured POSIX-shell-sourceable file (`RELEASE`, `API`, `APP`, `DB`), seeded with the current PROD values.
- `src/dsp_tools/resources/start-stack/docker-compose.yml`: 5 image tags become `${API}`/`${APP}`/`${DB}` interpolation. The file is no longer rewritten by CI.
- `scripts/bump_stack_versions.py`: only writes `versions.env`. `IMAGE_SUBSTITUTIONS`, `_update_compose_content`, and the header logic (`_ensure_header`, `_HEADER_RE`, `HEADER_BLOCK`) are gone. ~70 lines smaller.
- `src/dsp_tools/commands/start_stack/start_stack.py`: `_get_url_prefix` reads the API tag from `versions.env` via a new `_read_env_var` helper. All four `docker compose` subprocess invocations gain `--env-file versions.env` so the interpolation resolves.
- Tests: 7 `TestUpdateComposeContent` cases dropped (no more compose rewriting); 5 new `TestReadEnvVar` cases for the start-stack helper.

### Dispatcher (`ci`)

- `.github/workflows/publish-release-to-pypi.yml`: new `dispatch-dsp-docs-bump` job. Sparse-checkout `versions.env`, then `grep '^[A-Z]+=' versions.env >> $GITHUB_ENV` so every subsequent step in the job sees `RELEASE`, `API`, `APP`, `DB`, `TOOLS` as plain env vars — no `steps.parse.outputs.X` indirection or per-step env bindings. Mints a DaSCH Bot App token scoped to `actions:write` on `dsp-docs` only, preflights against already-running bumps for the same DSP, then `gh workflow run bump-release.yml`. Failure step posts an internal Chat alert. Gated by `vars.DSP_DOCS_DISPATCH_ENABLED`.

## Test plan

- [x] `pytest test/unittests/scripts/test_bump_stack_versions.py test/unittests/commands/start_stack/` — 20 passed.
- [x] `docker compose --env-file versions.env config` resolves to the correct literal image tags (`v35.8.1`, `v13.3.0`, `5.5.0-3`).
- [x] `docker compose config` without `--env-file` produces the expected warnings + empty tags (confirms the flag is required).
- [x] End-to-end bump simulation: `versions.env` is the sole file written by the script; subsequent `docker compose config` resolves the new tags.
- [x] Dispatcher `$GITHUB_ENV` pattern simulated locally — `RELEASE`/`API`/`APP`/`DB`/`TOOLS` available to subsequent steps.
- [x] `actionlint` (with `shellcheck`) on the changed workflow — 0 errors in modified blocks.
- [ ] On merge: next `dsp-tools` release publish should dispatch `bump-release.yml` in `dsp-docs` within ~30 s.
- [ ] Negative: set `vars.DSP_DOCS_DISPATCH_ENABLED=false` → dispatcher job is skipped.

## Prerequisites (operator)

- ✅ DaSCH Bot App installed on `dsp-docs`, `dsp-tools`, `dsp-api`, `dsp-app`, `dsp-meta`.
- ✅ Org-level `secrets.DASCH_BOT_APP_ID` and `secrets.DASCH_BOT_APP_PRIVATE_KEY`.
- ✅ `vars.DSP_DOCS_DISPATCH_ENABLED=true`.
- ✅ `secrets.GOOGLE_CHAT_DSP_RELEASE_INTERNAL_WEBHOOK_URL` (failure alerts).

Companion PR in `dsp-docs`: <https://github.com/dasch-swiss/dsp-docs/pull/207>

### DIFF
--- a/.github/workflows/publish-release-to-pypi.yml
+++ b/.github/workflows/publish-release-to-pypi.yml
@@ -43,10 +43,14 @@ jobs:
   dispatch-dsp-docs-bump:
     name: Dispatch dsp-docs bump
     needs: [publish-release-to-pypi]
-    if: github.event_name == 'release' && github.event.action == 'published' && vars.DSP_DOCS_DISPATCH_ENABLED != 'false'
+    # Kill-switch comparisons enumerate case variants since GitHub Actions
+    # expressions don't have a tolower() function.
+    if: >-
+      github.event_name == 'release' && github.event.action == 'published'
+      && !contains(fromJSON('["false","False","FALSE"]'), vars.DSP_DOCS_DISPATCH_ENABLED)
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
         with:
           sparse-checkout: src/dsp_tools/resources/start-stack/versions.env
           sparse-checkout-cone-mode: false
@@ -62,9 +66,10 @@ jobs:
           grep -E '^[A-Z]+=' src/dsp_tools/resources/start-stack/versions.env >> "$GITHUB_ENV"
           echo "TOOLS=${{ github.event.release.tag_name }}" >> "$GITHUB_ENV"
 
+      # SHA-pinned (v3 == bcd2ba4) since this action handles private-key material.
       - name: Mint DaSCH Bot App token
         id: app-token
-        uses: actions/create-github-app-token@v3
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1  # v3
         with:
           app-id: ${{ secrets.DASCH_BOT_APP_ID }}
           private-key: ${{ secrets.DASCH_BOT_APP_PRIVATE_KEY }}
--- a/.github/workflows/publish-release-to-pypi.yml
+++ b/.github/workflows/publish-release-to-pypi.yml
@@ -7,6 +7,9 @@ on:
     types: [published]
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 env:
   UV_PUBLISH_TOKEN: ${{ secrets.PYPI_DASCHBOT_TOKEN }}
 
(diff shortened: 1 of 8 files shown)


jnussbaum (approved):
LGTM

.github/workflows/publish-release-to-pypi.yml
```diff
@@ -36,3 +39,85 @@ jobs:
...
+          echo "dsp=$DSP"        >> "$GITHUB_OUTPUT"
+          echo "api=$API"        >> "$GITHUB_OUTPUT"
+          echo "app=$APP"        >> "$GITHUB_OUTPUT"
+          echo "tools=$TOOLS_TAG" >> "$GITHUB_OUTPUT"
+
+      - name: Mint DaSCH Bot App token
+        id: app-token
+        uses: actions/create-github-app-token@v3
+        with:
+          app-id: ${{ vars.DASCH_BOT_APP_ID }}
```
SamuelBoerlin:
```suggestion
          app-id: ${{ secrets.DASCH_BOT_APP_ID }}
```

`DASCH_BOT_APP_ID` is registered as an org secret, not variable.

SamuelBoerlin (changes requested):
One issue, otherwise seems fine.
I've set up the `GOOGLE_CHAT_DSP_RELEASE_INTERNAL_WEBHOOK_URL` secret.

.github/workflows/publish-release-to-pypi.yml
mpro7: Very good catch! Fixed in [aa8a40d](https://github.com/dasch-swiss/dsp-tools/pull/2303/commits/aa8a40d09942bd5e0deac9c2e0dd3776ca4c3fcb). Thanks!

SamuelBoerlin (approved):

BalduinLandolt (approved):
We can merge this as is, but I see two red flags in terms of maintainability: 
Both the sed-parsing for the versions, and the whole thing with writing a version number in a comment in a file seem extremely brittle to me.  
And DSP-TOOLS is already doing some work to override versions, and the version bumping script is overly complex; both pointing to the same underlying issue: `docker-compose.yml` is simply not the right place to store the version numbers.

In my view, this PR can be merged, but should spawn a follow-up task (probably for @jnussbaum or @Nora-Olivia-Ammann ), which addresses this issue more fundamentally.

I'm proposing the following target state, but that's just my idea, and may change when looking into it some more:

Create a `versions.env` file, containing roughly this:

```
DSP_RELEASE=2026.05.01
API_VERSION=v35.8.1
APP_VERSION=v13.3.0
DB_VERSION=5.5.0-3
INGEST_VERSION=v35.8.1
SIPI_VERSION=v35.8.1
```

1. This is much easier to write than modifying the docker compose file.
2. docker compose has built-in support for .env files (needs a flag `--env-file`, though, if the file is called anything other than just `.env`, but I'm assuming people are already using .env files in this repo, so we should pick another name).
3. Any other process that needs access to the variables does not have to parse it, but can just `source` it, and they will be available.

And when versions are in environment variables, perhaps there is even an easier way of overriding them, so the override files are no longer needed.

Does that make sense?

jnussbaum:
Thank you @BalduinLandolt for raising the maintainability concerns, which I agree with. In case @mpro7 wants to merge this as-is, I'm more than willing to look into simplifications like the one you have suggested.

mpro7:
@BalduinLandolt thanks to your suggestion, I will do the simplification round in both PRs.

mpro7:
@SamuelBoerlin @BalduinLandolt 
This commit reduces complexity and simplifies entire workflow: [706d2c6](https://github.com/dasch-swiss/dsp-tools/pull/2303/commits/706d2c6d684f11ad6cfd42359216bb61cc727a7a).
What changed:
  - feat/dev-6438-automate-dsp-docs-release-bump → +53 / -125 lines
  - New `src/dsp_tools/resources/start-stack/versions.env` — structured source of truth
  - `bump_stack_versions.py`: dropped `_ensure_header` / `_HEADER_RE` / `HEADER_BLOCK`; added  `_write_versions_env`
  - `test_bump_stack_versions.py`: dropped 14 `TestEnsureHeader*` cases (38 → 16 tests)
  - `docker-compose.yml`: # release: line removed; static auto-managed comment kept
  - `publish-release-to-pypi.yml` dispatcher: source `versions.env` replaces sed-on-yaml; `sparse-checkout` reduced to one file.

.github/workflows/publish-release-to-pypi.yml
```diff
@@ -36,3 +39,101 @@ jobs:
...
+          source src/dsp_tools/resources/start-stack/versions.env
+          set +a
+          if [[ -z "${RELEASE:-}" || -z "${API:-}" || -z "${APP:-}" ]]; then
+            echo "::error::Missing key in versions.env (RELEASE=$RELEASE, API=$API, APP=$APP)"
+            exit 1
+          fi
+          echo "dsp=$RELEASE"     >> "$GITHUB_OUTPUT"
+          echo "api=$API"         >> "$GITHUB_OUTPUT"
+          echo "app=$APP"         >> "$GITHUB_OUTPUT"
+          echo "tools=$TOOLS_TAG" >> "$GITHUB_OUTPUT"
```
BalduinLandolt: this I'm not sure about - but I feel there should be a simpler way to load a .env as environment variables in a github action run. and if this could be put directly into the `env`, you wouldn't need to read new env variables from the previous output.

BalduinLandolt (approved):
This is already much nicer, but doesn't cover all I was hoping for: Since we now have `versions.env`, I feel the docker-compose file should read from there, so we'd no longer need to write to the compose file, which in turn would allow for drastically simplifying the bump stack versions script.  
I leave it up to you, if you want to address that, or if that's for a follow-up.

One detail comment below.

.github/workflows/publish-release-to-pypi.yml
mpro7: Fair push back. Here are the changes: [a5c7f0b](https://github.com/dasch-swiss/dsp-tools/pull/2303/commits/a5c7f0b15e8a44a18c5f3e9c4ef72b338d1885ef). Let me know this is what you though about.

BalduinLandolt (approved):
It's starting to look very reasonable to me. I hope it also works... I find that a bit hard to see from the diff. But I suppose the best way of finding out, will be trying it out. 👍

SamuelBoerlin (approved):

PR #2304, opened 2026-05-19
chore(xmlupload): log ingest server response on upload failure

jnussbaum:


src/dsp_tools/clients/ingest.py
```diff
@@ -109,6 +109,7 @@ def _ingest(self, filepath: Path) -> IngestResponse:
                 )
                 log_response(res)
             except requests.exceptions.RequestException as e:
+                logger.exception("Ingest request to {} failed", url)
```
Nora-Olivia-Ammann: weird, inconsistent formatting

Nora-Olivia-Ammann (approved):

PR #2322, opened 2026-06-02
feat: parse placeholder value from xml (DEV-6581)

Nora-Olivia-Ammann:


src/dsp_tools/utils/xml_parsing/models/parsed_resource.py
```diff
@@ -55,6 +55,13 @@ class ParsedFileIiifUri(ParsedFileValueValue):
     """Content of the iiif-uri tag, i.e. the IIIF-URI."""
 
 
+class ParsedFilePlaceholder(ParsedFileValueValue):
```
Nora-Olivia-Ammann, line 58: I need to expose the `value` parameter in the `ABC` so that I can access it from the other classes without having to cast them. This seemed like the easiest solution to me.

jnussbaum (approved):

PR #2323, opened 2026-06-02
feat(xmlupload): use new built-in urn placeholder for missing legal info (DEV-6571)

Nora-Olivia-Ammann:


src/dsp_tools/clients/legal_info_client.py
Nora-Olivia-Ammann, line 1: I tried it with the latest API locally and it works. The e2e test should work as soon as it is deployed to prod.

test/e2e/commands/test_create_get_xmlupload.py
```diff
@@ -506,6 +506,5 @@ def test_compare_license_after_data_upload(
         project_returned = json.load(f)
 
     orig_licenses = set(project_original["project"]["enabled_licenses"])
```
Nora-Olivia-Ammann, line 508: systematic project has missing legal info. Previously we enabled the unknown license which is not used, and had to add it as it was enabled on the fly. This is now no longer necessary.

jnussbaum (approved):

PR #2333, opened 2026-06-04
feat(xmllib): add `PlaceholderFile` option (DEV-6570)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/models/internal/file_values.py
```diff
@@ -116,23 +117,36 @@ class AbstractFileValue(Protocol):
...
+        match value:
+            case Path():
+                if str(value) == ".":
+                    msg_info = MessageInfo(
+                        message=f"Your input '{value}' is empty. Please enter a valid file path.",
+                        resource_id=resource_id,
+                        field="bitstream",
+                    )
+                    emit_xmllib_input_warning(msg_info)
+                    value = ""
```
jnussbaum, line 137: Previously you called `check_and_warn_potentially_empty_string()` after setting value to "". Now you skip this check, which is probably a regression.

src/dsp_tools/xmllib/models/res.py
```diff
@@ -1705,6 +1706,16 @@ def add_file(
                 permissions=xmllib.Permissions.LIMITED_VIEW,
             )
             ```
+
+            ```python
+            # a file with a placeholder file
```
jnussbaum:
```suggestion
            # when the real file is not available yet, use a placeholder file
```

jnussbaum (approved):

src/dsp_tools/xmllib/models/internal/file_values.py
Nora-Olivia-Ammann: The difference is that previously we did not inform the user here but through the mechanism you describe. Now we inform the user right above on 136. Apart from that we have a test case that would cover a regression.

PR #2336, opened 2026-06-04
feat: allow and parse value order from xml (DEV-6597)

Nora-Olivia-Ammann:


test/integration/utils/test_replace_id_with_iri.py
```diff
@@ -19,8 +19,8 @@
 
 def test_with_replacement():
     orig_list_val = ("list", "node")
-    link_val = ParsedValue(HAS_PROP, "test_thing_1", KnoraValueType.LINK_VALUE, None, None)
-    list_val = ParsedValue(HAS_PROP, orig_list_val, KnoraValueType.LIST_VALUE, None, None)
+    link_val = ParsedValue(HAS_PROP, "test_thing_1", KnoraValueType.LINK_VALUE, None, None, None)
+    list_val = ParsedValue(HAS_PROP, orig_list_val, KnoraValueType.LIST_VALUE, None, None, None)
```
jnussbaum, line 23: Nitpick: You could get rid of the "None, None, None" if you add defaults to the dataclass.

jnussbaum (approved):

test/integration/utils/test_replace_id_with_iri.py
Nora-Olivia-Ammann: I consciously did not do that, because I want to prevent that this is accidentally not parsed somewhere.

jnussbaum: Fair point

PR #2339, opened 2026-06-08
feat(xmllib): enable value ordering (DEV-6598)

Nora-Olivia-Ammann:


src/dsp_tools/xmllib/internal/checkers.py
```diff
@@ -160,3 +161,17 @@ def check_and_inform_about_angular_brackets(value: Any, res_id: str | None, prop
...
+
+def check_raise_if_input_value_for_value_order_is_incorrect(
+    input_values: Any, prop_name: str, value_type: str, res_id: str
+) -> None:
+    """Raises an error if the input for ...multiple() method is not an ordered collection (list or tuple)."""
+    if not isinstance(input_values, (list, tuple)):
+        msg_str = (
+            f"You set the parameter 'add_value_order' to true when using the 'add_{value_type}_multiple()' method. "
+            f"The only acceptable input of the values if the order is wished is a list or tuple. "
+            f"Your input with the type '{type(input_values).__name__}' is not acceptable."
```
jnussbaum:
```suggestion
            f"This is only possible if the values are in an ordered collection (like a list or a tuple). "
            f"Your input with the type '{type(input_values).__name__}' is not acceptable, because it is unordered."
```

jnussbaum (approved):

PR #2340, opened 2026-06-09
docs: add explanation to value order

Nora-Olivia-Ammann:


docs/data-file/xml-data-file.md
```diff
@@ -226,6 +226,49 @@ The authorship can be defined in the following way:
...
 
+## Value Order
+
+The optional `order` attribute on value elements specifies the position at which a value is displayed in the DSP app,
+relative to other values of the same property on the same resource.
+
+Without the `order` attribute, the display order of values is not guaranteed.
+Use `order` when a consistent, meaningful display sequence matters.
+
+<!-- markdownlint-disable MD036 -->
```
Nora-Olivia-Ammann: It spams up the sidebar if we use headers here.

src/dsp_tools/xmllib/models/res.py
```diff
@@ -223,8 +223,9 @@ def add_color(
             value: value to add
             permissions: optional permissions of this value
             comment: optional comment
-            order: Position at which this value is displayed in the app, relative to other values of the same property.
-                   If None, the display order is not guaranteed.
```
Nora-Olivia-Ammann, line 227: I removed the sentence `If None, the display order is not guaranteed.` because the display was overly long and now they should just follow the link. But I though it was important to note that it must start at 0.

docs/data-file/xml-data-file.md
```diff
@@ -226,6 +226,49 @@ The authorship can be defined in the following way:
 These identifiers can later be referenced in the `<bitstream>` and `<iiif-uri>` elements.
 
 
+## Value Order
```
jnussbaum: I would rather put this paragraph at the very bottom of the page, right before "Complete Example"

jnussbaum (approved):

PR #2344, opened 2026-06-10
experiment: increase DB limits

Nora-Olivia-Ammann:


src/dsp_tools/resources/start-stack/config.alloy
Nora-Olivia-Ammann, line 1: @SamuelBoerlin I think this is where all the action is. The rest is set-up so that we can use it with a cli flag.
Does it look good to you?

SamuelBoerlin: The alloy pipeline and compose file look correct as far as I can tell, though I imagined you'd pass the OTLP endpoint as argument to dsp-tools when starting the stack so you wouldn't have to edit the config file. 

I'm not sure we should bother with supporting all kinds of OTLP endpoints (gRPC vs. HTTP vs. with/without auth) and instead just pick one protocol, e.g. gRPC, without auth. Then you'd only have one configurable parameter, being the exporter `endpoint`.

jnussbaum:
@Nora-Olivia-Ammann  Can this PR be closed?

PR #2350, opened 2026-06-21
feat: support resource-side legal metadata (DEV-6666)

danielasubotic:
Fixes DEV-6666
Fixes DEV-6665

## Motivation

The dsp-tools side of resource-side legal metadata (spec: dasch-specs `03-feat-legal-metadata-resource-side-plan.md` §4 · dsp-tools). It round-trips the new project-level data-side legal fields and adds a per-resource authorship element to the XML data format. Integrates against the dsp-api PRs: project fields + endpoint ([dsp-api#4156](https://github.com/dasch-swiss/dsp-api/pull/4156)), and `knora-base:hasResourceAuthorship` ([dsp-api#4157](https://github.com/dasch-swiss/dsp-api/pull/4157)).

## Summary

- **tools-1** — project `data_license` / `data_copyright_holder` / `data_authorship` round-trip through `get` and `create`.
- **tools-2** — a per-resource `<data-authorship>` XML element, mapped to `knora-api:hasResourceAuthorship` on xmlupload, and validated via SHACL in `validate-data`.

## Key Changes

- **Get** (`legacy_models/project.py`): reads `dataLicense`/`dataCopyrightHolder`/`dataAuthorship` from the admin project API and emits `data_license`/`data_copyright_holder`/`data_authorship` into the project definition file (only when set). The create POST body is unchanged.
- **Create** (`parsed_project.py`, `parse_project.py`, `create_on_server/project.py`, `clients/legal_info_client*.py`): parses the three fields and, after project creation, sends them via the dedicated **`LegalInfoClient`** — `PUT /admin/projects/shortcode/{shortcode}/legal-info/resource-side` (only when at least one is set). The create POST does not carry legal fields.
- **Schema** (`resources/schema/project.json`): adds the three optional fields beside `enabled_licenses`.
- **XML** (`data.xsd`): adds a repeatable `<data-authorship>` element on `resource_type`, strictly separate from the asset-level `authorship-id` IDREF and the top-level `<authorship>` lookup. Threaded through parsing → processed model → `make_resource_and_values`, which emits one `knora-api:hasResourceAuthorship` literal per author on the resource node.
- **SHACL** (`commands/validate_data/...`, `resources/validate_data/api-shapes.ttl`): `data_authorship` is threaded through the `validate-data` RDF-like pipeline (`get_rdf_like_data` → `TriplePropertyType` → `mappers`), and a property shape on the generic `Resource_Shape` validates `hasResourceAuthorship` as a non-empty single-line string.
- **Docs** (`docs/`): the three project fields and the `<data-authorship>` element.

## Challenges and Decisions

- **Legal info via its own client** (LT-2): the project-create POST doesn't accept legal fields, so they're set after creation via a second call. This goes through the dedicated `LegalInfoClient` (which already owns the `/legal-info/*` endpoints), not `ProjectClient` — keeping the endpoint URL convention (`…/projects/shortcode/{code}/legal-info/…`) consistent. (This also corrected the URL, which had been missing the `/shortcode/` segment.)
- **SHACL shape for `hasResourceAuthorship`**: mirrors the existing FileValue authorship shape (free string, non-empty, single-line, Violation) and reuses the generic `Resource_Shape` (`sh:targetClass knora-api:Resource`). The field is optional, so the shape is format-only. The validate-data pipeline didn't previously carry `data_authorship`, so it's threaded through to the RDF-like data first.

## Gotchas

- **e2e / round-trip needs a running api** carrying the dsp-api changes (the new endpoint + fields + resource property). Locally: run the integration api branch; in CI it's gated on the dsp-api PRs deploying. Hence draft.
- The repo's pinned `uv` (0.4.25) can't install the dev deps (pre-PEP 735); a newer `uv` is needed to run the test suite locally. `uv.lock` is unchanged in this PR.

## Test Plan

- `ruff check` + `ruff format --check` + `mypy` clean; **2597 unit tests pass**, including new tests: project get/create round-trip incl. the second-call body (sent-when-set / not-sent-when-unset) now via `LegalInfoClient`, the `LegalInfoClient.set_resource_side_legal_info` request/error handling, `<data-authorship>` parsing, the `hasResourceAuthorship` RDF emission, and the SHACL thread-through. XSD validated incl. negative cases.
- **SHACL shape verified locally** against the Oxigraph Docker SHACL CLI: a clean authorship conforms, an empty authorship produces exactly one violation on `hasResourceAuthorship`. `mkdocs build` succeeds.
- Follow-up (needs the deployed/local api): e2e round-trip `get` → `create` → diff.


BalduinLandolt:
Notice an upstream change we did in the API, which will likely break this and needs a small adjustment: https://github.com/dasch-swiss/dasch-specs/pull/137

Nora-Olivia-Ammann:
@BalduinLandolt but you would like me to review it now anyway rather than wait?

BalduinLandolt:
> @BalduinLandolt but you would like me to review it now anyway rather than wait?

probably easiest to wait until the API is merged, but if you like you can have a look at it already now

Nora-Olivia-Ammann:
In that case i'll wait till it is finalised, thanks.

BalduinLandolt:
@Nora-Olivia-Ammann API is merged, and all changes are applied here. 

Tests fail expectedly, because they run against the released/deployed version of the API, not against main.

docs/data-file/xml-data-file.md
```diff
@@ -287,6 +288,29 @@ The following property elements exist:
 - `<uri-prop>`: contains URI values
 
 
+### `<data-authorship>`
```
Nora-Olivia-Ammann: Is further legal metadata on resources planned? In that case I wonder if it would make sense to wrap this.

Further I would say that it is more consistent and less verbose if it is styled the same as authorship on assets, which is done via key. This way it could also be put into an attrib which is a leaner and also more consistent implementation.

docs/data-model/json-project/overview.md
```diff
@@ -125,6 +125,9 @@ The `project` object contains the basic metadata about the project. The followin
 The following fields are optional (if one or more of these fields are not used, they should be omitted):
 
 - default_permissions_overrule (only if default_permissions = public)
+- data_license
+- data_copyright_holder
+- data_authorship
```
Nora-Olivia-Ammann: perhaps we could add "default_" so that it is consistent with the permissions and to make it clearer that it can be overwritten, this way we just have the same name in multiple places without clarification in the name how they are distinguished.

docs/data-model/json-project/overview.md
```diff
@@ -205,6 +208,39 @@ All the licenses listed here will be enabled. Licenses can be disabled by omitti
...
 
+### `data_license`
+
+(optional)
+
+`"data_license": "<license-iri>"`
+
+The project-wide license that applies to every resource record (the data itself), as opposed to the
+file/IIIF-URI licenses controlled by `enabled_licenses`.
+Only Creative Commons licenses are allowed, referenced by their IRI,
```
Nora-Olivia-Ammann, line 219: If only creative commons are allowed, it should be implemented as an enum in the json schema, this makes it easier for parsing and validation. This is a hard requirement.

src/dsp_tools/commands/create/create_on_server/project.py
```diff
@@ -21,11 +24,23 @@ def create_project(project: ParsedProjectMetadata, auth: AuthenticationClient, e
     try:
         project_iri = client.get_project_iri(project.shortcode)
         _exit_if_create_should_not_continue(project.shortcode, exit_if_exists)
-        return project_iri
-
     except ProjectNotFoundError:
         logger.debug("No project with the shortcode exists. Continuing creating the project.")
-        return _create_project_on_server(project, client)
+        project_iri = _create_project_on_server(project, client)
```
Nora-Olivia-Ammann, line 30: Why was the return statement removed from the top? and why is this not part of the top most `_create_project_on_server`

src/dsp_tools/commands/get/legacy_models/project.py
```diff
@@ -279,20 +294,30 @@ def fromJsonObj(cls, con: Connection, json_obj: Any) -> Project:
...
             "descriptions": self._description.createDefinitionFileObj(),
             "keywords": list(self._keywords),
             "enabled_licenses": list(self._enabled_licenses),
         }
+        if self._data_license:
+            proj["data_license"] = self._data_license
+        if self._data_copyright_holder:
+            proj["data_copyright_holder"] = self._data_copyright_holder
+        if self._data_authorship:
+            proj["data_authorship"] = list(self._data_authorship)
```
Nora-Olivia-Ammann: Why is this turned into a list? It should already be one...?

src/dsp_tools/utils/xml_parsing/get_parsed_resources.py
```diff
@@ -155,6 +162,9 @@ def _parse_values(
                 asset_value = _parse_bitstream_tag(val)
             case "iiif-uri":
                 asset_value = _parse_iiif_uri(val)
+            case "data-authorship":
```
Nora-Olivia-Ammann: Why is this parsed separately when we iterate over it anyway?

test/e2e/commands/test_resource_side_legal.py
```diff
@@ -0,0 +1,102 @@
...
+from dsp_tools.cli.args import ServerCredentials
+from dsp_tools.commands.create.create import create
+from dsp_tools.commands.get.get import get_project
+from dsp_tools.commands.xmlupload.upload_config import UploadConfig
+from dsp_tools.commands.xmlupload.xmlupload import xmlupload
+from dsp_tools.utils.rdf_constants import KNORA_API
+from test.e2e.commands.xmlupload.utils import util_get_res_iri_from_label
+from test.e2e.commands.xmlupload.utils import util_request_resources_by_class
+
+PROJECT_SHORTCODE = "0099"
```
Nora-Olivia-Ammann: This does not follow the convention we have for assigning shortcodes in the test data, we also keep a register

testdata/json-project/resource-side-legal-0099.json
Nora-Olivia-Ammann, line 1: we include all possible features in systematic test data. For the json this should be included there.

I think we could also delete this so that we don't have an abundance of test files.

test/unittests/commands/create/serialisation/test_project.py
```diff
@@ -57,6 +57,9 @@ def test_with_license(self):
             descriptions={"en": "english", "de": "german"},
             keywords=["test"],
             enabled_licenses=["license"],
+            data_license=None,
```
Nora-Olivia-Ammann, line 60: Why is there no test case with legal metadata?

test/unittests/commands/validate_data/test_make_data_graph.py
```diff
@@ -148,6 +148,21 @@ def test_with_props(self, rdf_like_boolean_value_corr):
...
+            property_objects=[
+                *UNREIFIED_TRIPLE_OBJECTS,
+                PropertyObject(TriplePropertyType.KNORA_RESOURCE_AUTHORSHIP, "Author One", TripleObjectType.STRING),
+            ],
+            values=[],
+            migration_metadata=MigrationMetadata(),
+        )
+        g = Graph(store="Oxigraph")
+        _add_one_resource(g, res)
+        assert len(g) == 3
```
Nora-Olivia-Ammann, line 163: We only know now that one triple has the metadata. It is more complete to assert all triples.

test/unittests/commands/xmlupload/make_rdf_graph/test_make_resource_and_values.py
```diff
@@ -80,6 +80,28 @@ def test_make_resource_permissions() -> None:
     assert permissions == Literal("CR knora-admin:ProjectAdmin", datatype=XSD.string)
 
 
+def test_make_resource_data_authorship() -> None:
```
Nora-Olivia-Ammann, line 83: Same

testdata/xml-data/resource-side-legal-0099.xml
Nora-Olivia-Ammann, line 1: All paths (happy and error) must also be included in validate e2e test data

justfile
```diff
@@ -150,7 +150,7 @@ e2e-test-mapping *FLAGS:
 # Run the end-to-end tests for the create, get and xmlupload command (with testcontainers)
 [no-exit-message]
 e2e-test-create-get-xmlupload *FLAGS:
-    uv run pytest test/e2e/commands/test_create_get_xmlupload.py {{FLAGS}}
+    uv run pytest test/e2e/commands/test_create_get_xmlupload.py test/e2e/commands/test_resource_side_legal.py {{FLAGS}}
```
Nora-Olivia-Ammann: I would not conflate several e2e into one command. Really annoying when something goes wrong.

Nora-Olivia-Ammann (changes requested):

Nora-Olivia-Ammann:
is it on purpose that the xmllib is not included?

BalduinLandolt:
> is it on purpose that the xmllib is not included?

Not necessarily on purpose. But I'd prefer to add that in a separate PR, if that's alright?

docs/data-model/json-project/overview.md
BalduinLandolt: this is only true for authorship (license and copyright holder for resources are project-wide and can not be overwritten), but your'e right, there it should be "default".  
Currently it's in line with how it's called in the API, so I'd leave it here as is for now, and then I'll change the naming both in the API and here to "default" in a separate PR. Ok?

BalduinLandolt:
@Nora-Olivia-Ammann  I hope I addressed all issues: https://github.com/dasch-swiss/dsp-tools/pull/2350/changes/d115a616554fb0ab666599e851b164d2edbc8668
Please have another look.

Tests still fail due to CI running against the deployed API version.  
Locally I ran the tests against a more recent API image, all tests pass, except one little issue, which I need to fix in the API.

I also added a conventions and a review file for future agent work, because some of the comments you remarked contained repo knowledge which was not clearly encoded, so that it was unavailable to Claude at the time of working on the PR.  
(the pattern of removing things from the claude.md file and putting it into conventions.md and review.md is already established in other repos)

Nora-Olivia-Ammann (approved):

Nora-Olivia-Ammann:
Thanks for adding the conventions and review file! @BalduinLandolt

BalduinLandolt:
> Thanks for adding the conventions and review file! @BalduinLandolt

Gladly! I hope they make sense.  
I think Claude mostly based it on the Claude.md file, so it shouldn't be too wrong.

Nora-Olivia-Ammann:
> > Thanks for adding the conventions and review file! @BalduinLandolt
> 
> Gladly! I hope they make sense. I think Claude mostly based it on the Claude.md file, so it shouldn't be too wrong.

No it makes sense to me. I read them.

docs/data-model/json-project/overview.md
```diff
@@ -125,6 +125,9 @@ The `project` object contains the basic metadata about the project. The followin
 The following fields are optional (if one or more of these fields are not used, they should be omitted):
 
 - default_permissions_overrule (only if default_permissions = public)
+- data_license
+- data_copyright_holder
+- default_data_authorship
```
Nora-Olivia-Ammann, line 130: @BalduinLandolt I had a thought. After you told me that the authorship does not work like a real default (as in if none is provided this applies) this may not actually serve us. Because if I understood your explanation correctly this will never apply automatically but will be a suggestion to the user primarily in the front-end but when doing an xmlupload will never have an effect (?). If this is correct then I think we will create more confusion, while I do not advocate for the naming to be reversed perhaps we could call it `suggested_...` (or something similar) . Because now we have the default permissions and the default authorship with completely different mechanisms and repercussions... which will not be easy to intuitively explain to the users.

docs/data-file/xml-data-file.md
```diff
@@ -287,6 +292,33 @@ The following property elements exist:
...
+```
+
+This is **distinct** from:
+
+- the per-file authorship referenced with the same [`authorship-id`](#bitstream) attribute on
+  `<bitstream>`/`<iiif-uri>`, which describes the multimedia asset, not the resource record;
+- the project-wide [`default_data_authorship`](../data-model/json-project/overview.md#default_data_authorship) default
+  set in the project definition file.
+
+
```
Nora-Olivia-Ammann, line 334: A side note, I can tell you @BalduinLandolt from experience with the permissions that not having a default option for something you want to apply identically to everything (like authorship for one import xml) will be widely unpopular for xmllib users. We could solve this by for example allowing the `XmlRoot` to have a new parameter specifying the default which we apply to all resources that are not specified otherwise during serialisation. This way we do not have to apply a default during the xmlupload while giving that ease to the users. This I think will be a solution that is appreciated.

docs/data-model/json-project/overview.md
BalduinLandolt: We discussed this quite a bit, and came to the following conclusion: From the App's perspective, the naming that fits best, is "default". And having a drift between the naming of the same property between different components, would be even worse.  
So I think the way we should frame this, is "Projects have a property `defaultDataAuthorship`, and to set this property when creating a project in dsp-tools, you just have to set the property with the same name in the JSON file".

I'll try to make this clear enough in the documentation, then we'll just have to make sure people understand. But I think seeing it in the app will actually help with understanding this.

docs/data-file/xml-data-file.md
BalduinLandolt: After a long discussion with Claude, I/we came up with the following:

```python
  # literal default
  root = xmllib.XMLRoot.create_new(
      "0000", "onto",
      apply_default_resource_authorship=["Daisy Duck"],
  )

  # the project's default, resolved at upload
  root = xmllib.XMLRoot.create_new(
      "0000", "onto",
      apply_default_resource_authorship=xmllib.PROJECT_DEFAULT,
  )
```

What I like about this, is that it makes the project default actually useful to dsp-tools, which in turn should help make it clear what the meaning of this field is.  
Furthermore, it essentially replicates the app's behaviour: The default is a value that users can opt into using.

What do you think?

Nora-Olivia-Ammann: Good but one problem, if the API does not apply the "default" then we somehow have to add it to the data. The xmllib does not know the json and I do not want it to. So that would mean that we declare that if nothing is written the default applies (same as permissions), and then somehow during the xmlupload we will have to add this default to all the resources that have nothing. It is doable but a bit more complex...

BalduinLandolt: Even worse: It is also possible to not add authorship attributions, so we can not say "if there's nothing, apply the project default". Instead, we have to add the information to the XML, that the default must be applied. And it opens the error case, that this property may be used for a project that doesn't have default authorship defined.  
But the error is easy enough to catch early, and communicate; and the complexity shouldn't be a problem.

I'll go ahead and implement that, then.

BalduinLandolt:
@Nora-Olivia-Ammann added three more commit. Do you want to have another look?

docs/data-model/json-project/overview.md
```diff
@@ -235,12 +235,22 @@ The copyright holder that applies to every resource record in the project.
...
+The default authorship for the resource records in the project.
 Each entry is the name of one author.
-This is distinct from the authorship referenced via `authorship-id` in the XML data file
-(on `<bitstream>`/`<iiif-uri>` for the asset, or on
-[`<resource>`](../../data-file/xml-data-file.md#resource-authorship-authorship-id-on-resource)
-for the resource record).
+
+Unlike `data_license` and `data_copyright_holder`, this is **not** applied to your data automatically.
+It only serves as the value that DSP-APP proposes when a resource is created manually in the web interface.
+An `xmlupload` gives a resource record an authorship only if it is set explicitly via the `authorship-id`
```
Nora-Olivia-Ammann, line 243: This contradicts the next paragraph.

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -104,6 +111,28 @@ def xmlupload(
...
+    The marker is set by `xmllib` when a user passes `apply_default_resource_authorship=PROJECT_DEFAULT`.
+    Returns None if the marker is absent, aborts if it is present but the project defines no default.
+    """
+    if root.attrib.get("use-project-default-resource-authorship") not in ("true", "1"):
+        return None
+    default_authorship = project_client.get_default_data_authorship(shortcode)
+    if not default_authorship:
+        raise MissingProjectDefaultAuthorshipError(
+            f"The XML data requests the project's default resource authorship "
+            f"(apply_default_resource_authorship=PROJECT_DEFAULT), but project {shortcode} has no "
```
Nora-Olivia-Ammann: I am not sure it is a good idea to cite the xmllib code here. Because the user input has nothing to do with it. And it is not even really correct code.

src/dsp_tools/commands/xmlupload/xmlupload.py
```diff
@@ -104,6 +111,28 @@ def xmlupload(
...
+            f"'default_data_authorship' defined. Either define it on the project, "
```
Nora-Olivia-Ammann: I would explicitly state that this must be in the "project json".

Nora-Olivia-Ammann (approved):

testdata/xml-data/test-data-systematic-4123.xml
```diff
@@ -15,6 +15,10 @@
     <authorship id="authorship_2">
```
Nora-Olivia-Ammann, line 16: We should include the default authorship = true here as well. Alternatively you could include it in the minimal test data. This way we ensure that it worked.

test/unittests/commands/xmlupload/get_processed_resources/test_get_processed_resources.py
```diff
@@ -161,6 +161,60 @@ def test_resource_with_permissions(self, lookups: XmlReferenceLookups):
         assert not result.file_value
         assert not result.migration_metadata
 
+    def test_resource_with_authorship(self, lookups: XmlReferenceLookups):
```
Nora-Olivia-Ammann, line 164: This is not a deal-breaker for me, but we should note that the project default authorship is never applied in the shacl validation which does not use the processed resource. So the validated data diverges from the uploaded one. But again not a deal-breaker because the validated data looks different anyway.

docs/data-model/json-project/overview.md
BalduinLandolt: changed the phrasing, thanks!

src/dsp_tools/commands/xmlupload/xmlupload.py
BalduinLandolt: removed

src/dsp_tools/commands/xmlupload/xmlupload.py
BalduinLandolt: done

testdata/xml-data/test-data-systematic-4123.xml
BalduinLandolt: done

PR #2359, opened 2026-07-02
feat(xmllib): support resource-record authorship on Resource (DEV-6666)

BalduinLandolt:
## Summary

Adds xmllib support for resource-record authorship (`hasResourceAuthorship`), deferred from #2350.

`Resource.create_new()` accepts an `authorship` parameter, mirroring the asset authorship on `add_file()`/`add_iiif_uri()`: the user passes a plain list of names, ids stay internal.

- The authorship validation logic from `Metadata.new` is extracted into a shared `check_and_fix_authorship_input()` so files and resources behave identically.
- `_make_authorship_lookup` also collects resource-level author tuples — a resource and a file with the same authors share one id in the top-level `<authorship>` pool.
- The serialiser emits `authorship-id` on the `<resource>` element when set.
- Docs: new section in `overview.md` + `create_new` docstring.

## Notes

- **Stacked on #2350** (`feature/dev-6666-tools-resource-side-legal`) because the XSD change (`authorship-id` on `resource_type`) lives there. Retarget to `main` once #2350 merges.
- End-to-end upload of resource authorship needs dasch-swiss/dsp-api#4173 deployed; this PR only concerns XML generation.

## Tests

TDD: 8 new tests covering input validation, lookup dedup across files and resources, serialisation, and a full `XMLRoot.serialise()` round-trip validated against `data.xsd`. All unit (2606), integration (128), ruff, mypy, vulture, markdownlint green.

https://claude.ai/code/session_014HCy4paZicL9wJhQJYwWeb

### DIFF
--- a/docs/xmllib-docs/overview.md
+++ b/docs/xmllib-docs/overview.md
@@ -66,6 +66,10 @@ resource = xmllib.Resource.create_new(
 This is distinct from the authorship of a file (see [below](#adding-a-file-to-a-resource)),
 which describes the multimedia asset, not the resource record.
 
+If your project defines a default authorship, this default is not applied to your data automatically:
+it only serves as the value that DSP-APP proposes when a resource is created manually in the web interface.
+A resource record has an authorship only if you set it explicitly with `authorship`, as shown above.
+
 ### Adding Values to a Resource
 
 Once you created a `Resource` you can add values to it.
--- a/docs/xmllib-docs/overview.md
+++ b/docs/xmllib-docs/overview.md
@@ -50,6 +50,22 @@ resource = xmllib.Resource.create_new(
 )
 ```
 
+### Setting the Authorship of a Resource
+
+You can state who authored the resource record itself, i.e. the data describing the object:
+
+```python
+resource = xmllib.Resource.create_new(
+    res_id="general_resource",
+    restype=":ResourceType",
+    label="Title in the DSP-APP",
+    authorship=["Minnie Mouse"],
+)
+```
+
+This is distinct from the authorship of a file (see [below](#adding-a-file-to-a-resource)),
+which describes the multimedia asset, not the resource record.
+
 ### Adding Values to a Resource
 
 Once you created a `Resource` you can add values to it.
--- a/src/dsp_tools/xmllib/models/res.py
+++ b/src/dsp_tools/xmllib/models/res.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from typing import Any
 
 from dsp_tools.xmllib.internal.checkers import check_raise_if_input_value_for_value_order_is_incorrect
+from dsp_tools.xmllib.internal.input_converters import check_and_fix_authorship_input
 from dsp_tools.xmllib.internal.input_converters import check_and_fix_collection_input
 from dsp_tools.xmllib.internal.input_converters import check_and_fix_is_non_empty_string
 from dsp_tools.xmllib.internal.xmllib_warnings import MessageInfo
@@ -49,6 +50,7 @@ class Resource:
     permissions: Permissions = Permissions.PROJECT_SPECIFIC_PERMISSIONS
     file_value: AbstractFileValue | None = None
     migration_metadata: MigrationMetadata | None = None
+    authorship: tuple[str, ...] | None = None
 
     @staticmethod
     def create_new(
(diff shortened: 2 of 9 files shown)


docs/xmllib-docs/overview.md
```diff
@@ -50,6 +50,22 @@ resource = xmllib.Resource.create_new(
 )
 ```
 
+### Setting the Authorship of a Resource
```
Nora-Olivia-Ammann, line 53: Perhaps it is worth noting that this overwrites the default set in the project and is optional and only necessary if no default is set or if it differs.

test/unittests/xmllib/models/test_root.py
```diff
@@ -395,6 +396,32 @@ def test_make_authorship_lookup() -> None:
...
+    xml_root = XMLRoot.create_new("0000", "test")
+    res_with_file = Resource.create_new("id_1", ":ResType", "lbl", authorship=["auth1"]).add_file(
+        "file.jpg", LicenseRecommended.DSP.UNKNOWN, "copy", ["auth1"]
+    )
+    xml_root.add_resource(res_with_file)
+    serialised = xml_root.serialise()
+    resource_ele = next(serialised.iterdescendants(tag=f"{DASCH_SCHEMA}resource"))
+    bitstream_ele = next(serialised.iterdescendants(tag=f"{DASCH_SCHEMA}bitstream"))
+    assert resource_ele.attrib["authorship-id"] == bitstream_ele.attrib["authorship-id"]
+    schema_res = importlib.resources.files("dsp_tools").joinpath("resources/schema/data.xsd")
```
Nora-Olivia-Ammann: it is not necessary to go over importlib because this is a test file that is not shipped and is only run when you have the entire code, in that case the filepath is ordinary.

test/unittests/xmllib/internal/test_serialise_resource.py
```diff
@@ -57,6 +57,28 @@ def test_one_value(self) -> None:
...
+            result = _serialise_one_resource(res, AUTHOR_LOOKUP)
+            assert len(caught_warnings) == 0
+        expected = (
+            b'<resource xmlns="https://dasch.swiss/schema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" '
+            b'label="lbl" id="id" restype=":Type" authorship-id="authorship_1"/>'
+        )
+        assert etree.tostring(result) == expected
+
+    def test_resource_authorship_unknown_author(self) -> None:
+        res = Resource.create_new("id", ":Type", "lbl", authorship=["unknown"])
```
Nora-Olivia-Ammann, line 73: This makes no sense to me. It is not possible to have an unknown author because we create the lookup for the ids. So if it is possible for an authorship ID not to exist, we created a bug and the user is not able to fix it.

src/dsp_tools/xmllib/models/res.py
```diff
@@ -69,6 +72,9 @@ def create_new(
             restype: resource type
             label: resource label
             permissions: optional permissions of this resource
+            authorship: optional authorship of the resource record itself,
```
Nora-Olivia-Ammann, line 75: I would mention that it overwrites the default

Nora-Olivia-Ammann (approved):

docs/xmllib-docs/overview.md
BalduinLandolt: This is actually a misunderstanding of what's happening (and it is admittedly a bit confusing - and makes me second guess the "default" naming):  
The default authorship is not actually applied to the data (we decided against that because it might end up mis-attributing data to the wrong person, if someone's not aware of the default.  
Instead, what the default is used for, is for dsp-app to propose an authorship value in the form for creating new resources.

I'll try to clarify that in the documentation some more.

src/dsp_tools/xmllib/models/res.py
BalduinLandolt: see my response above

test/unittests/xmllib/internal/test_serialise_resource.py
BalduinLandolt: Agreed, this should never happen. The test was created in analogy to what's already there for asset authorship, which equally makes no sense. The right thing to do would be to raise an `UnreachableCodeError`, I suppose.  
I can leave it as is, assuming it will never happen; I can change it only here; or I can change it for the asset case too. What do you think?

Nora-Olivia-Ammann: I would change it, we did not have the UnreachableCodeError at that time, but it makes absolutely sense. Thanks

docs/xmllib-docs/overview.md
Nora-Olivia-Ammann: Ah ok, that really was not clear to me. I think it makes sense to explain that extensively and then refer to it at those points.

PR #2362, opened 2026-07-09
docs(validate-data): add new SHACL shape creation playbook

Nora-Olivia-Ammann:


Nora-Olivia-Ammann:
My plan is to validate that this works well when I create the new value type. So far I created it conceptually, so expect that modifications and perhaps even a more comprehensive example will be added in the future, but it did not make sense to me to fake this right now. Especially because I would like to do a reference to an API PR and then help update the shapes based on that information, but that is best done with new things...

BalduinLandolt (commented):
I didn't look at everything in detail.

Firstly, a formal thing: Instead of a markdown document that's mentioned in prose, you could make this a repo-local skill. In practice it wouldn't change much, but it would be more "Claude native".  
A nice thing about making it a skill would be that you could use anthropic's "skill-creator" skill to have a look at it. This is designed to create or improve skills and it tends to have good findings.

More importantly for my review, I didn't have to read much of the document to come to the conclusion that this very different from how I think that Agents should be driven and skills should be written.  
In the end it has to work for you first and foremost, so I don't want to block this strictly, but I'm reluctant to approve it in that form, so I may not be the right person to ask for review; and I would encourage to make this opt-in, because I would not want to have it in my workflow.  
Here are the high-level points that struck me immediately:

1. Prompting the agent to ask everything and not be able to assume or research essentially forces it into helplessness, and degrades it to a chat-bot. I explicitly want my agents to be able to be as autonomous as possible, encouraging them to be as autonomous as possible, and preferring research (online or in the code base) over asking me; I want them to only ask me questions they genuinely can not resolve themselves.
2. Similarly, I want my agent instructions to specify the intent of the action, i.e. what I want the agent to achieve. That way it can self-correct by checking if its actions are in line with the intent. This prompt here is a step-by-step instruction, which means the agent may or may not follow the steps in a satisfying manner, but it has no way of adjusting when it goes off rails. In my view, the most important thing is to specify what the context of the work in question is, and what it's supposed to achieve; any step-by-step instructions should be very light and only serve to give the agent hints of how he may reach the goal most easily.
3. Any workflow like this, which is clearly distinct from other work, can easily - and will benefit from - adding a reviewing step, where one or many subagents are spawned, to check if the work makes sense for an agent that did not implement the thing.
4. Telling an agent that it only needs limited knowledge of a domain that's important for its work, seems dangerous (presumably that will immediately make the agent dumber in that field) and unnecessary. Also, providing it with the RDF/SHACL primer probably doesn't benefit at all - most of these things it will already know, and if it has a knowledge gap, I'm assuming that allowing it to do online research will make it more effective.

Nora-Olivia-Ammann:
> I didn't look at everything in detail.
> 
> Firstly, a formal thing: Instead of a markdown document that's mentioned in prose, you could make this a repo-local skill. In practice it wouldn't change much, but it would be more "Claude native". A nice thing about making it a skill would be that you could use anthropic's "skill-creator" skill to have a look at it. This is designed to create or improve skills and it tends to have good findings.


--> the reason why I did it like that is so that it is consistent with all the other claude.md files, which has also better visibility for people...
> 
> More importantly for my review, I didn't have to read much of the document to come to the conclusion that this very different from how I think that Agents should be driven and skills should be written. In the end it has to work for you first and foremost, so I don't want to block this strictly, but I'm reluctant to approve it in that form, so I may not be the right person to ask for review; and I would encourage to make this opt-in, because I would not want to have it in my workflow. Here are the high-level points that struck me immediately:
> 
>     1. Prompting the agent to ask everything and not be able to assume or research essentially forces it into helplessness, and degrades it to a chat-bot. I explicitly want my agents to be able to be as autonomous as possible, encouraging them to be as autonomous as possible, and preferring research (online or in the code base) over asking me; I want them to only ask me questions they genuinely can not resolve themselves.

--> I only required 1 absolute ask (the severity), the others I only said that if it is not clear from research it must ask and not guess... Do you think it will not be able to handle that?

> 
>     2. Similarly, I want my agent instructions to specify the intent of the action, i.e. what I want the agent to achieve. That way it can self-correct by checking if its actions are in line with the intent. This prompt here is a step-by-step instruction, which means the agent may or may not follow the steps in a satisfying manner, but it has no way of adjusting when it goes off rails. In my view, the most important thing is to specify what the context of the work in question is, and what it's supposed to achieve; any step-by-step instructions should be very light and only serve to give the agent hints of how he may reach the goal most easily.

--> the reason I did it like that is based on the omissions I saw in Daniela's claude code. Additionally, AI has a tendency to create new code rather than reusing it, I am afraid that it will just continue to add to the querying for validation response which is pretty convoluted and slow already... So this was an attempt to make it try it out first, before adding a ton of new code and test cases if it is not necessary.

> 
>     3. Any workflow like this, which is clearly distinct from other work, can easily - and will benefit from - adding a reviewing step, where one or many subagents are spawned, to check if the work makes sense for an agent that did not implement the thing.

--> where would you do that?

> 
>     4. Telling an agent that it only needs limited knowledge of a domain that's important for its work, seems dangerous (presumably that will immediately make the agent dumber in that field) and unnecessary. Also, providing it with the RDF/SHACL primer probably doesn't benefit at all - most of these things it will already know, and if it has a knowledge gap, I'm assuming that allowing it to do online research will make it more effective.

--> Thanks, I will remove that.

BalduinLandolt:
> > Firstly, a formal thing: Instead of a markdown document that's mentioned in prose, you could make this a repo-local skill. In practice it wouldn't change much, but it would be more "Claude native". A nice thing about making it a skill would be that you could use anthropic's "skill-creator" skill to have a look at it. This is designed to create or improve skills and it tends to have good findings.
> 
> --> the reason why I did it like that is so that it is consistent with all the other claude.md files, which has also better visibility for people...

I'm not sure visibility is the main thing to optimise for in an agent workflow. Also, as long as it's a repo-local skill, not in the plugins, it's also fairly visible, it's just in a different folder.  
But I don't think that matters much

> 
> > More importantly for my review, I didn't have to read much of the document to come to the conclusion that this very different from how I think that Agents should be driven and skills should be written. In the end it has to work for you first and foremost, so I don't want to block this strictly, but I'm reluctant to approve it in that form, so I may not be the right person to ask for review; and I would encourage to make this opt-in, because I would not want to have it in my workflow. Here are the high-level points that struck me immediately:

> > 1. Prompting the agent to ask everything and not be able to assume or research essentially forces it into helplessness, and degrades it to a chat-bot. I explicitly want my agents to be able to be as autonomous as possible, encouraging them to be as autonomous as possible, and preferring research (online or in the code base) over asking me; I want them to only ask me questions they genuinely can not resolve themselves.

> 
> --> I only required 1 absolute ask (the severity), the others I only said that if it is not clear from research it must ask and not guess... Do you think it will not be able to handle that?

If you start the instructions with "**Ask, don't guess — decisions that are not yours to make**" then the agent will be very hesitant to take any decisions, and even on things where it actually knows the answer, it will be more likely to ask rather than assume the right thing.  
But more fundamentally, I just don't think it's a good idea to have the agent interrupt and wait for input _on every item_, which the severity check does. This means that the agent can never go off and and do the whole thing, and either you have to watch it constantly, or it will spend time idling while you're not seeing that it needs your input.  
What I would much prefer, is the agent to do the whole thing, and then presenting you with a list of all decisions it took, where you can simply point out where it got it wrong once.  
And I would assume that an agent also can get the severity right more often than not; so why not let it take a guess?

Now, if you really don't want the agent to take choices, the better pattern in my view, is to give it the list of things it has to know from the user up front, and tell it to first clarify these things. That way it will ask up-front rather than when questions come up.

> > 2. Similarly, I want my agent instructions to specify the intent of the action, i.e. what I want the agent to achieve. That way it can self-correct by checking if its actions are in line with the intent. This prompt here is a step-by-step instruction, which means the agent may or may not follow the steps in a satisfying manner, but it has no way of adjusting when it goes off rails. In my view, the most important thing is to specify what the context of the work in question is, and what it's supposed to achieve; any step-by-step instructions should be very light and only serve to give the agent hints of how he may reach the goal most easily.

> --> the reason I did it like that is based on the omissions I saw in Daniela's claude code. 

I don't think a step-by-step instruction is the right way to solve that. If an agent has omited certain things, then a checklist to verify that the task is completed is at least as effective in making sure that this doesn't happen, and will not obstruct the agent in its way of working.

> Additionally, AI has a tendency to create new code rather than reusing it, I am afraid that it will just continue to add to the querying for validation response which is pretty convoluted and slow already... So this was an attempt to make it try it out first, before adding a ton of new code and test cases if it is not necessary.

This could easily be solved by a clause in the review instructions, stating that it's not allowed to add to the query, where the same thing can be solved by re-using existing code.

> > 3. Any workflow like this, which is clearly distinct from other work, can easily - and will benefit from - adding a reviewing step, where one or many subagents are spawned, to check if the work makes sense for an agent that did not implement the thing.

> --> where would you do that?

In the end.  
As stated above, I wouldn't structure this as a step-by-step guide. I would probably structure it roughly as follows: Contextual Knowledge (what is this validation that's being added, what purpose does this serve), code knowledge (where in code does it live, what structure does it have, where are new ones added), success criteria (how do you know that the implementation is done, what's good enough), probably with a check list.   
And after that, a paragraph along those lines:  
"When you're done implementing the validation, before handing back to the user, start a review agent to check the the implementation for quality and simplicity. The review must consider, but is not limited to, the following criteria."  
Or something along those lines.


Also, something entirely marginal, which makes review unnecessarily harder:  
Tripple-backtick code-blocks with indention, as part of bullet-points, break github's syntax highlighting. If you could get Claude to not do that, this would make review more pleasant, which I would appreciate. :)

Nora-Olivia-Ammann:
(created with claude grilling skill)

Thanks @BalduinLandolt — this was genuinely useful, and you were right on the mechanics. Digging into *why* I wanted the guardrails, the thing I was actually protecting against isn't "the agent deciding things" — it's **rework**: wrong assumptions cascading into code and tests before anyone notices. Once I framed it that way, most of your points slotted in cleanly. Here's what changed, keyed to your numbers:

**1. "Ask, don't guess" → "front-load, then run autonomously."**
You were right that opening with *"decisions that are not yours to make"* poisons the whole thing — removed it, and there's no more per-item mid-flow pinging. Only two things get asked, both up front:
- **Severity** stays a required ask — not out of distrust, but because it's genuinely un-derivable *and* it routes downstream work: it decides which test file and which e2e assertion list the resource goes into, so getting it wrong late means redoing the test wiring.
- **Purpose**, only if it isn't clear from the code.

Everything else (datatype, target, cardinality) is now "research from the ontology / `knora-api` / XML and *propose*, don't ask" — surfaced only if it stays genuinely ambiguous (e.g. `knora-api` concepts like `valueHasOrder`, where datatype/cardinality is a DSP-behaviour call rather than an ontology lookup).

One place I nudged back on "do the whole thing, then present decisions at the end": for *correctness* I didn't want wrong code to exist and only be caught by a reviewer. So I leaned on the empirical loop the doc already had — test-data-first + run + read the firing constraint component — as the primary, objective catch *before* any Python. The review agent is a quality net on top, not the correctness check.

**2. Step-by-step → intent-first.**
Fair — the intent was buried under the steps. Added a **"What you are building, and when you are done"** header (goal + the one mental model that drives everything + a concrete "done looks like"), and reframed the steps as *the most efficient, fool-proof path — deviate with reason*, not law. I did keep the ordering, and I'd defend that: it's a fail-fast sequence (cheap test/shape/run before expensive Python + e2e), which is exactly what stops the "build a lot on a wrong assumption" failure I was worried about. Your checklist point is in there too — the Quick checklist stayed and grew.

**3. Review subagent.**
Added as **Step 7 — Review before handing back**, positioned explicitly as the quality/reuse net (correctness is the empirical loop). And you're right this is the honest home for my reuse worry: instead of a step telling the *implementing* agent "don't add to the query" (the one agent tempted to), the "reuse over addition / don't grow the validation-result query when an existing path works" rule is now a criterion the *fresh reviewer* checks. Kept it generic (no hard-coded agent names) so it works regardless of anyone's plugin setup.

**4. Domain knowledge / RDF primer.**
Removed the "you only need limited knowledge" framing and the generic RDF/SHACL primer — agreed, it either knows it or can look it up. I kept the *codebase-specific* part from that section though: the `sh:sourceConstraintComponent` model, the dispatch table, and the "wrong graph = never runs" / parser / "update both lists" gotchas. That's un-researchable repo knowledge, so it now leads as the mental model instead of reading like a lecture.

**On the skill suggestion:** kept it as a markdown doc + a one-line pointer from `CLAUDE.md` for now — matches the 10 existing `CLAUDE.md` files, and it's effectively opt-in already (the pointer only pulls the doc in when someone's actually adding a shape, so it shouldn't intrude on your workflow). I'll run `skill-creator` over it to harvest its findings, taking the substance of your suggestion. If a family of these subsystem runbooks emerges, I'll revisit converting them — and a shared review skill.

**Formatting:** fixed — de-indented the code blocks so GitHub highlights them again. :)

Would love another look when you have a moment.

src/dsp_tools/commands/validate_data/SHACL_PLAYBOOK.md
```diff
@@ -0,0 +1,403 @@
+# SHACL Shape Playbook
+
+This is a playbook for **adding a new validation ("SHACL shape")** to the `validate-data` command.
+Everything you need is here or linked. The ordered steps below are not a rigid procedure — they are the
+**most efficient, fool-proof path** to the goal, sequenced so the cheapest feedback comes first. Deviate
+if you have a good reason, but the ordering is what keeps you from building a lot on a wrong assumption.
+
+If you want to know *how the existing code is structured*, read [`CLAUDE.md`](./CLAUDE.md) first.
+This file is the *how do I add a new check* companion to it.
```
BalduinLandolt: This is not necessary, claude.md is always read.

.claude/skills/add-shacl-shape/SKILL.md
```diff
@@ -0,0 +1,403 @@
...
+**most efficient, fool-proof path** to the goal, sequenced so the cheapest feedback comes first. Deviate
+if you have a good reason, but the ordering is what keeps you from building a lot on a wrong assumption.
+
+If you want to know *how the existing code is structured*, read [`CLAUDE.md`](./CLAUDE.md) first.
+This file is the *how do I add a new check* companion to it.
+
+## What you are building, and when you are done
+
+**Goal.** A new SHACL shape that makes wrong data fail validation with a clear, actionable message and
+lets correct data pass — with the smallest possible change to the report-processing code.
```
BalduinLandolt, line 34: I like this, but I would add another paragraph before it, stating the even broader context: This is about uploading data to the DSP using dsp-tools; and the SHACL validation is run before the upload to catch issues in the data early, to give feedback without havng tho rely on lengthy uploads which may fail after a long time. Hence the purpose is to catch these issues and give informative and actionable error messages, so that the data can be corrected.

src/dsp_tools/commands/validate_data/SHACL_PLAYBOOK.md
```diff
@@ -0,0 +1,403 @@
...
+- **Severity — always confirm with the developer.** Whether a check is a violation, warning, or info is
+  a policy decision you cannot derive, and it *routes the rest of your work*: it decides which test file
+  and which e2e assertion list the data goes into (violations vs. warnings/info, see Step 6). Getting it
+  wrong late means redoing the test wiring. Add the chosen value to the shape's `sh:severity`:
+    - `sh:Violation` — **always blocks** upload on every server. For wrong data.
+    - `sh:Warning` — **non-blocking on a test server, blocks on production**. Primarily for data that is
+      only temporarily missing.
+    - `sh:Info` — just a nudge; the upload will work on **all** servers.
+- **Purpose, if it is not clear from the code.** If you cannot tell from the ontology and the code what
+  the check is *for*, ask — a shape built on a misunderstood intent is the classic source of wasted work.
```
BalduinLandolt: I would add a clause to that along the lines of "unless this is already specified". For two reasons:

1. I could start a Claude session with a prompt like "Please add a info-level check to the SHACL validation, for the case XYZ.
2. I could be working on dsp-tools in the context of a broader feature, where the PRD/Implementation plan already specifies that a certain thing is a violation.

In either of these cases, the level is already defined and I would not want to be asked again.

src/dsp_tools/commands/validate_data/SHACL_PLAYBOOK.md
```diff
@@ -0,0 +1,403 @@
...
+
+Everything else you **research and propose, you do not ask**: derive the datatype, the target
```
BalduinLandolt: this could be softened a bit: "you do not ask, unless absolutely necessary" because if the agent genuinely doesn't know, it should not be forbidden to ask.

src/dsp_tools/commands/validate_data/SHACL_PLAYBOOK.md
```diff
@@ -0,0 +1,403 @@
...
+**Key pattern — combine both.** Dynamic shapes usually do **not** duplicate logic; they *reference a
+reusable static NodeShape by IRI*. For example, for each project text property whose gui element is
+`SimpleText`, `value_shacl.py` generates a `<prop>_PropShape` PropertyShape
+(`sh:path <prop> ; sh:node api-shapes:SimpleTextValue_ClassShape`), where that NodeShape is defined
+statically in `api-shapes.ttl` (`Textarea` and `Richtext` map to other class shapes). So a very
+common move is: **add a reusable static shape, then add/extend a SPARQL generator that binds the
+project's properties to it.**
+
+**Write a good `sh:message` in the shape.** For many components the shape's message is shown verbatim
+to the user (see the table in Step 4). A clear message here often means **zero or little Python changes**.
```
BalduinLandolt: Maybe better "write a useful `sh:message`? "good" is such an under-specified term. Also, I would consider elaborating some more, what makes the message good.

src/dsp_tools/commands/validate_data/SHACL_PLAYBOOK.md
```diff
@@ -0,0 +1,403 @@
...
+    - `just e2e-test-validate-data` — **requires Docker**; spins up a real DSP stack (testcontainers),
+      creates the project, and validates the data end-to-end.
+    - `just lint` — before committing.
+
+---
+
+## 8. Step 7 — Review before handing back
+
+When the implementation is complete and both suites are green, **spawn a review subagent** before you
+hand back — the agent that wrote the code is the least reliable judge of its own reuse and complexity.
```
BalduinLandolt: I'm not sure this last half-sentence adds any value. Here you want the agent to focus on the task at hands, and not get side tracked reasoning about its own capabilities. Though, admittedly, it probably also doesn't hurt.

BalduinLandolt (approved):
I much prefer this approach!

I still added some comments, but nothing substantial anymore. 

I still think, this would fit better as a repo-defined skill (which Claude then would also auto-detect) than just a markdown document next to the code.

And if you have not done so, I would definitely ask Claude with the `skill-creator` skill loaded, to have a look at it and see if it can suggest any improvements. I think that should work, no matter if it's actually defined s a skill, or if it's just a markdown file.

PR #2367, opened 2026-07-15
fix: guard interactive prompts against non-interactive stdin (DEV-6744)

Nora-Olivia-Ammann:
[DEV-6744](https://linear.app/dasch/issue/DEV-6744)

## Summary
- `_check_version()` reached an interactive `input()` prompt during a version-drift window; on a non-interactive stdin (CI pipe, cron, nohup, Docker without a TTY, pytest capture) `input()` raised an `OSError`/`EOFError` that escaped `run()` uncaught. Under pytest, repeated failures across many tests ballooned RSS into multi-GB territory and could crash the machine.
- The reported "memory leak" was **two** distinct things: a diagnostic-harness artifact (`tracemalloc.start(25)` + per-iteration `take_snapshot()`) that overstated the growth, and this genuine uncaught-`OSError`-under-pytest crash. There is **no** native memory leak in product code — verified by A/B reproduction. Fixing the uncaught exception resolves the crash.
- Guards **every** `input()` prompt in dsp-tools against a non-interactive stdin, not just the version check.

## Changes
**Shared helper** (`src/dsp_tools/utils/interactive.py`)
- `stdin_is_interactive()` — isatty-based, robust to a closed/replaced/None stdin
- `prompt_until_valid_answer(...)` — y/n-style prompt; returns a caller-chosen default and prints a notice when non-interactive
- `prompt_for_required_value(...)` — free-text prompt; raises `UserError` naming the CLI flag when non-interactive

**Guarded prompt sites** (agreed non-interactive defaults)
- version check → continue with a printed warning
- `xmlupload` / `ingest-xmlupload` skip-validation & ignore-duplicate-files → honor the explicit flag (proceed)
- `create` project-exists → continue (use `--exit-if-exists` to abort)
- `resume-xmlupload` empty-pending → continue with stashed properties
- `start-stack` docker prune → do **not** prune (use `--prune` to force)
- `mapping` / `migration config` missing shortcode/ontology → raise an actionable `UserError`

**Docs / conventions**
- `--suppress-update-prompt` help text + docs (it now suppresses a warning, not a crash)
- `utils/CLAUDE.md` documents the new module; CLI docs note the non-interactive behavior

## Test Plan
- New tests: `test/unittests/utils/test_interactive.py`, `test_entry_point_version_check.py`, plus prompt/error-path tests for mapping config, start-stack prune, and create project-exists (interactive + non-interactive pairs).
- `just unittests` — **2599 passed**.
- `just lint` — clean (ruff, ruff-format, mypy, vulture, yamllint, markdownlint).
- End-to-end: reproduced the original ballooning (drift forced, real `_check_version`, non-interactive) → `>1.5GB`/killed before the fix; after the fix the same scenario is `15 passed`, flat ~130MB, and the full `test/unittests` suite runs at ~230MB peak.

## Screenshots
[Attach manually if UI changes]


BalduinLandolt (approved):
👍

PR #2376, opened 2026-07-23
fix(validate-data): exclude deleted resources from the list of resources that exist in the DB (DEV-6837)

Nora-Olivia-Ammann:


src/dsp_tools/commands/validate_data/prepare_data/prepare_data.py
```diff
@@ -121,10 +121,20 @@ def _get_metadata_info(
...
     retrieval_status, metadata = metadata_client.get_resource_metadata(shortcode)
-    formatted_metadata = [InfoForResourceInDB(x["resourceIri"], x["resourceClassIri"]) for x in metadata]
+    formatted_metadata = _format_metadata_export(metadata)
     return retrieval_status, formatted_metadata
 
 
+def _format_metadata_export(metadata: list[dict[str, str | None]]) -> list[InfoForResourceInDB]:
+
+    def is_deleted(single_resource_info: dict[str, str | None]) -> bool:
+        return bool(single_resource_info.get("Deletion Date (if available)"))
```
jnussbaum: this will silently fail if the column is ever renamed in the backend. Is the column always present? If yes, you could consider this to make it fail loudly:


```suggestion
        return bool(single_resource_info["Deletion Date (if available)"])
```

jnussbaum (approved):

src/dsp_tools/commands/validate_data/prepare_data/prepare_data.py
Nora-Olivia-Ammann: I tried it out and found out that the json response has different keys, and if a resource is not deleted the key is absent... So while your comment was good, I had to solve it differently.

PR #2377, opened 2026-07-24
chore(clients): set log level based on HTTP response code

Nora-Olivia-Ammann:


Nora-Olivia-Ammann:
I decided to implement this on the log level function because in general I like to be able to log before we deal with the different branches of the status code (less likely to forget) and so that it is applied everywhere the same. It is also possible to extend it in the future if we need to do more fine grained things. I think the level warning is also appropriate for all response types (but 200).

jnussbaum (approved):

PR #2382, opened 2026-07-29
fix(create): exit with code 3 from --exit-if-exists when the project already exists (DEV-6878)

jnussbaum:
## What

`dsp-tools create --exit-if-exists` now exits with a dedicated exit code **`3`** when the project
already exists on the server, instead of `sys.exit(0)`.

## Why

Previously, when the project already existed, `--exit-if-exists` called `sys.exit(0)` — indistinguishable
from a normal successful create. A consuming CI job (the daschland-scripts `create-on-stage` workflow)
could therefore not tell "already exists" apart from "created successfully", so it proceeded to the
subsequent `xmlupload` step and **created duplicate resources** in the Alice showcase project.

With a distinct exit code, a caller can tell the three outcomes apart and skip subsequent steps gracefully:

| Code | Meaning |
| ---- | ------- |
| `0`  | project did not exist yet and was created |
| `3`  | project already existed → aborted without further uploads |
| `1`  | any other failure |

## Changes

- `create_on_server/project.py`: new module constant `EXIT_CODE_PROJECT_ALREADY_EXISTS = 3`; the
  `--exit-if-exists` branch now exits with `3` (and the printed message states the code).
- `create_parsers.py`: updated the `--exit-if-exists` help text.
- Docs: added an `--exit-if-exists` section to the user-facing CLI docs, and an exit-code reference table
  to the developer error-handling guideline.
- Updated the unit test `test_project_exit_through_flag` to assert exit code `3`.

## Reviewer notes

- The interactive "project exists → user answers `n`" abort is intentionally left at exit `1` — it is a
  human-initiated abort, distinct from the `--exit-if-exists` case, and out of scope here.
- `SystemExit(3)` propagates cleanly: `entry_point.run()` catches `KeyboardInterrupt` / `UserError` /
  `Exception`, none of which match `SystemExit`, so the process exits with `3`.
- **Consumer:** the daschland-scripts `create-on-stage.yml` / `create-on-dev.yml` workflows are updated in
  a separate PR to read exit code `3` and skip `xmlupload`. Those workflows run `uv add --upgrade
  dsp-tools` at runtime, so the new behavior takes effect once this change is released to PyPI; the
  consumer change is forward-compatible and cannot break before then.

## Verification

`pytest` on the affected unit tests (create/project + CLI wiring), `ruff`, `mypy`, and `markdownlint` all pass.

### DIFF
--- a/docs/data-model/data-model-cli.md
+++ b/docs/data-model/data-model-cli.md
@@ -19,6 +19,22 @@ The most frequently used options are:
 
 To see all possible options, type `dsp-tools create --help`.
 
+### The `--exit-if-exists` flag
+
+By default, if a project with the same shortcode already exists on the server,
+`create` continues and adds the definitions from the JSON file to the existing project.
+Pass `--exit-if-exists` to abort instead, as soon as an existing project is detected,
+without uploading anything.
+
+This is useful in automated pipelines that must not modify an already-existing project.
+Depending on the outcome, the command exits with one of these codes:
+
+- `0`: the project did not exist yet and was created.
+- `3`: the project already existed, so nothing was uploaded.
+- `1`: the command failed for another reason.
+
+A calling script can check for exit code `3` to skip subsequent steps (such as a data upload) gracefully.
+
 The defaults are intended for local testing:
 
 ```bash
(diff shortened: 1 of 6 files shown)


SamuelBoerlin (approved):
LGTM

PR #2386, opened 2026-08-04
feat(mapping)!: replace existing external mappings instead of adding to them (DEV-6897)

jnussbaum:
## Why

`dsp-tools mapping add` uploads external ontology mappings, but only additive, without option to remove unwanted mappings.

In the RDU team meeting of Aug 3 we found out that this behaviour does not support real RDU workflows in an optimal way. There are more and more cases where the old mapping should be removed. 

The previous idea was to have a separate Excel column "to remove", but it turned out that this overcomplicates things. We decided on the more radical approach to take the Excel as source of truth, remove all mappings from the server, and then add the new ones.

**Breaking change:** the Excel file is now the single source of truth for the external mappings of the target
ontology. The additive behaviour is gone; there is no flag to get it back.

### Discovery — `src/dsp_tools/commands/mapping/existing_mappings.py` (new)

Three pure functions, no I/O:

- `get_existing_mappings()` parses the ontology turtle with rdflib and collects, per entity, the external IRIs
  that may be deleted. The predicate (`rdfs:subClassOf` vs `rdfs:subPropertyOf`) discriminates classes from
  properties, so no `rdf:type` lookup is needed. `owl:Restriction` blank nodes from cardinalities and subjects
  outside the target ontology's namespace are filtered out.
- `is_deletable_mapping_iri()` is the whole safety story. It mirrors the DSP-API's own forbidden hosts
  (`knora.org`, `dasch.swiss`) and forbidden namespaces (RDF, RDFS, OWL, XMLSchema, SHACL, dash), **plus one
  rule the API does not have**: an IRI whose host equals the DSP server's host is never deleted. That rule is
  load-bearing — on a local or testcontainer stack the project's own entity IRIs live at
  `http://0.0.0.0:<port>/…`, whose host contains neither forbidden substring, so a same-project super-class
  would otherwise look deletable and the server would not stop us. Only the host is compared, not the port,
  because testcontainers pick a random one.
- `select_mappings_to_delete()` computes `existing − desired`, where an entity's desired IRIs are the union over
  all its Excel rows. A re-run with an unchanged Excel therefore issues zero DELETEs, and a mapping present in
  both the old and new state is never briefly removed.

Note the deliberate absence of `is_iri()`: its pattern rejects a dot inside a path segment, which would make us
silently skip a legitimately stored mapping. Everything classified here came out of a turtle the server
serialized.

### HTTP — `mapping_client.py` / `mapping_client_live.py`

`delete_class_mapping` / `delete_property_mapping` and a shared `_delete`. The DSP-API `mapping` query parameter
is single-valued, so the fan-out is one request per (entity, external IRI) pair. `quote_plus` on the query value
is functional, not cosmetic: many external IRIs contain `#` (`http://www.w3.org/ns/prov#Entity`), and a raw `#`
becomes a client-side fragment that never reaches the server. The query is built into the URL string rather than
passed via `params=`, so `log_request` logs the real, fully encoded URL.

### Orchestration — `mapping_add.py`

- `_check_project_and_get_ontology_ttl()` replaces `_check_if_project_and_ontology_exists()`. Same checks and
  same escalating exceptions, but it returns the turtle it already downloaded instead of discarding it, so
  discovery adds **zero** HTTP requests.
- `_send_one_request_with_retry()` holds the retry logic once, instead of the four near-duplicate copies the
  add/delete phases would otherwise have needed. `functools.partial` binds the loop variables eagerly.
- `MappingAction` is threaded through every message, so a failure always says whether an add or a delete failed
  and, for a delete, which IRI it was removing.
- Before deleting anything, the command prints how many mappings will go and which entities are affected. There
  is no interactive confirmation — that would break CI and scripted runs.
- The add phase runs even if some deletes failed: aborting in between would leave entities with fewer mappings
  and no replacements. All failures land in one report, which now ends with re-run advice.

### Fail-fast on an empty mapping cell — `resolve_parsed_mappings.py`

`parse_excel` filters empty `;`-segments, so a cell containing only `";"` yields an empty IRI list, and
`dropna(how="any")` does not drop it. Previously the PUT just got a harmless 400. Under replace semantics the
entity would be wiped first and the add would then fail, leaving it bare. Such an entity now produces a
`NO_MAPPING_IN_INPUT` problem during parsing, before any request is sent.

## Reviewer notes

- **A 403 fails fast with nothing destroyed.** `BadCredentialsError` is raised on the first request, before any
  mutation.
- **The e2e replace tests deliberately establish their own pre-state.** They call `mapping_add` with the good
  config before running the replace config. This is not redundant: the earlier
  `4124_testonto_Mapping_inexistent_references.xlsx` does not list `otherResource` / `hasOtherText`, so under
  replace semantics those two are already stripped by the time the new tests run — inheriting that state would
  make the `otherResource` assertion pass even if the wipe ignored un-listed entities entirely.
  `otherResource.subClassOf == {knora-api:Resource}` is the assertion that proves whole-ontology scope, so it
  had to be made load-bearing.
- **`4124_testonto_Mapping_good.xlsx` shrank 11.9 KB → 7.3 KB.** It was rewritten with openpyxl to append the
  `otherResource` / `hasOtherText` rows, which drops Excel-specific metadata. Sheet and cell contents are
  unchanged apart from the two new rows.
- **Every DELETE takes the server's ontology-cache lock and triggers `refreshCache()`**, so a run is serialized
  server-side and per-call cost grows with ontology size. Fine for realistic Excel files (tens of entities × 1–3
  IRIs), and a steady-state re-run sends none. A bulk-delete endpoint in dsp-api would be the real fix — worth
  raising with the backend team.
- **Whole-ontology scope is a footgun.** A shorter-than-intended Excel silently strips mappings from entities the
  author never named. Mitigated by the printed summary only; there is no `--dry-run` and no undo. The selection
  logic is already a separate pure function, so a `--dry-run` is cheap to add later.
- **`existing` is a snapshot** read before the write phases, so a concurrent ontology edit can make a DELETE a
  no-op. Documented; no client-side locking exists and none is needed for the writes themselves.
- Extending `minimal-project-4124.json` is safe for its other consumers: the distribution test only runs
  `create --validate-only` / `xmlupload --validate-only`, and `test-data-minimal-4124.xml` uses only
  `:minimalResource`.

## Testing

`ruff check`, `ruff format`, `mypy` (564 files), `vulture`, `markdownlint` and the markdown link validator all
pass. 2758 unit tests and 133 integration tests pass, including 25 new tests for the filter/discovery/selection
functions, 14 new client tests (incl. the `#`-encoding regression), new delete-phase and reporting tests, and the
12 existing `test_mapping_add.py` tests that this change necessarily rewrites.

`just e2e-test-mapping` was **not** completed locally — the run was still pulling the ~3.2 GB of `v37.4.0`
testcontainer images when this PR was opened. The e2e job in `tests-e2e.yml` covers it. Its two new tests assert
whole-set equality after the replace run: `cidoc:E22` and `schema:Book` gone, `bibo:Book` (listed in both Excel
files) surviving, `knora-api` supers untouched, and `otherResource` / `hasOtherText` wiped although absent from
the replace Excel.


docs/data-model/data-model-cli.md
```diff
@@ -99,7 +99,9 @@ The `mapping` command links DSP ontology classes and properties to IRIs from ext
 (e.g. schema.org, CIDOC-CRM), enabling interoperability with other knowledge systems,
 through definitions from an Excel file.
 It has two subcommands: `mapping config` generates a configuration template,
-and `mapping add` uploads the mappings to a DSP server.
```
Nora-Olivia-Ammann, line 102: Should we perhaps change the name to `mapping update` ? Because add is now misleading

docs/data-model/data-model-cli.md
```diff
@@ -189,4 +191,38 @@ The available arguments are:
...
+- External mappings of **every class and property of the ontology**,
+  including those that the Excel file does not mention at all.
+  A class that was mapped in an earlier run, but is absent from the Excel file now, loses its mappings.
+- Only mappings of the ontology named in the config file.
+  Other ontologies of the same project are not touched.
+
+The following super-classes and super-properties are always preserved:
+
+- DSP's own ones, e.g. `knora-api:Resource` and `knora-api:hasValue`.
+- Those from DSP shared ontologies.
```
Nora-Olivia-Ammann: This is not really informative how a shared ontology differs from dsp own ontologies.

docs/data-model/data-model-cli.md
```diff
@@ -189,4 +191,38 @@ The available arguments are:
...
+- Those from other ontologies of the same project, and from the same ontology.
+
+A mapping that is listed in both the old and the new state is never removed and re-added,
+so re-running the command with an unchanged Excel file sends no delete requests at all.
```
Nora-Olivia-Ammann: This is too technical, we don't care about requests in the user documentation just that the state does not change.

src/dsp_tools/clients/mapping_client_live.py
```diff
@@ -63,3 +73,32 @@ def _put(self, url: str, external_iris: list[str]) -> ResponseCodeAndText | None
                 )
             case _:
                 return parse_api_v3_error(response)
+
+    def _delete(self, url: str, external_iri: str) -> ResponseCodeAndText | None:
+        # The query value is built into the URL so that `log_request` logs the fully encoded URL.
+        # Encoding is mandatory: a raw '#' in an external IRI would become a client-side fragment.
```
Nora-Olivia-Ammann, line 79: This is a bit noisy

Nora-Olivia-Ammann (approved):

docs/data-model/data-model-cli.md
jnussbaum: I have rewritten it to be more outcome-centric and targeting the user instead of us devs

docs/data-model/data-model-cli.md
jnussbaum: Thanks for noticing. I have removed this point.

docs/data-model/data-model-cli.md
jnussbaum: I have renamed it to "update"

PR #2392, opened 2026-08-12
chore(ci): auto-merge the dsp-tools release-train PRs via the DaSCH Bot App (DEV-6956)

mpro7:
[DEV-6956](https://linear.app/dasch/issue/DEV-6956)

## Summary

- The two release-train follow-on PRs in this repo — `chore(start-stack): bump versions to <DSP>` and the release-please `chore: release x.y.z` — were merged by hand. They are now merged by the DaSCH Bot App once their required checks are green, mirroring `dsp-docs/.github/workflows/bump-release.yml`.
- The release PR is merged **only** when the push that produced it is the start-stack bump commit. Releases outside the release train stay a deliberate human merge, exactly as today.
- Both merges are guarded by a kill switch, an idempotency check, a concurrency group, and a failure notification to the internal Google Chat webhook.
- Neither leg searches for the PR it merges: each is told which PR it is by whatever created it.

## Changes

**New composite action** — `.github/actions/merge-pr-when-green/action.yml`

Retries `gh pr merge --squash --admin` until the required checks pass (30 min budget, 30 s poll), refreshing the branch only when the merge output says it is behind base. GitHub's native auto-merge is unusable here: it waits for a real approval that never comes on a bot-authored PR. The App is a bypass actor for the *review* only — required status checks have no per-actor bypass, so the merge is rejected by the server until CI is green and can never happen before it.

**`bump-stack-versions.yml`** — `scripts/bump_stack_versions.py` now writes the URL that `gh pr create` returns to `$GITHUB_OUTPUT`, and the workflow merges exactly that PR. The script also opens the PR with the App token instead of `DASCHBOT_PAT`: an App installation token triggers workflow runs just like the PAT does (only the default `GITHUB_TOKEN` does not), so the required checks still run on the bump PR — and this workflow no longer touches the long-lived PAT. No `pr_url` output means the script no-opped, which is treated as a no-op, not an error. Added a per-release concurrency group and a failure notification.

**`release-please.yml`** — new `auto-merge-release-pr` job, gated on `startsWith(head_commit.message, 'chore(start-stack): bump versions to')`. The release PR comes from the release-please action's own `pr` output, whose number is then resolved with `gh pr view`; the pre-existing curl lookup that searched open PRs by title prefix (and its PAT usage) is gone with it. Before merging, the job verifies that the PR title, `.github/release-please-manifest.json` and `pyproject.toml` agree on the version; release-please derives all three from the same conventional commits, so disagreement means the PR was hand-edited or a run was interrupted. The job has its own concurrency group so the merge wait never serialises the release-please job above it.

**`README.md`** — documents the automated behaviour, the `RELEASE_AUTO_MERGE_ENABLED` kill switch, and that a manual fallback merge of the bump PR must be a squash merge.

## Required configuration (not in this diff)

1. Add `dsp-tools` to the DaSCH Bot App installation (contents + pull-requests, write) — org owner, GitHub UI.
2. Add the App as a review-bypass actor in `dsp-tools` branch protection on `main`, mirroring `dsp-docs`. Review bypass only — the 13 required status checks are never bypassed, and `enforce_admins` stays on.

Step 1 must come first: GitHub rejects a bypass allowance for an App that is not installed on the repo. Since the App now also opens the bump PR, step 1 is a hard prerequisite for `bump-stack-versions.yml` as a whole — without it the run fails at token minting, before a PR exists.

## Test Plan

- `just yamllint` clean (only the pre-existing `publish-release-to-pypi.yml` line-length warning remains); `just ruff-check`, `just ruff-format-check` and `mypy` clean.
- `pytest test/unittests/scripts/` — unit tests cover the new PR-URL parsing and the `$GITHUB_OUTPUT` write.
- Every embedded shell block extracted and syntax-checked with `bash -n`.
- The version-agreement check was run against the real `release-please-manifest.json` / `pyproject.toml` (`19.0.0`), and against a deliberately mismatched title — matches and rejects as expected.
- The `gh pr view` resolution and its not-open guard were run read-only against the live merged release PR #2389.
- Neither workflow runs on `pull_request`, so the automation itself is first exercised on the next release train. The raw release-please `pr` output is echoed to the log, so its exact shape is visible on that first run. `RELEASE_AUTO_MERGE_ENABLED=false` reverts to the current manual merging without a revert PR.

### DIFF
--- a/.github/workflows/bump-stack-versions.yml
+++ b/.github/workflows/bump-stack-versions.yml
@@ -26,6 +26,12 @@ permissions:
   contents: write
   pull-requests: write
 
+# Per-release group lets unrelated releases run in parallel; same-release
+# dispatches serialise, so two runs can't race on the same bump PR.
+concurrency:
+  group: bump-stack-versions-${{ inputs.release }}
+  cancel-in-progress: false
+
 jobs:
   bump:
     runs-on: ubuntu-latest
@@ -53,3 +59,64 @@ jobs:
           API: ${{ inputs.api }}
           APP: ${{ inputs.app }}
           DB: ${{ inputs.db }}
+
+      - name: Find the bump PR to merge (kill switch, idempotency guard)
+        id: find_pr
+        env:
+          GH_TOKEN: ${{ secrets.DASCHBOT_PAT }}
+          RELEASE: ${{ inputs.release }}
+          # Normalize case so `False` / `FALSE` typed in the GitHub UI don't silently no-op.
+          AUTO_MERGE_ENABLED: ${{ vars.RELEASE_AUTO_MERGE_ENABLED }}
+        run: |
+          set -euo pipefail
+          if [[ "${AUTO_MERGE_ENABLED,,}" == "false" ]]; then
+            echo "::notice::RELEASE_AUTO_MERGE_ENABLED=false, leaving the bump PR for a human to merge"
+            exit 0
+          fi
+          # Branch name is the one scripts/bump_stack_versions.py creates. No open
+          # PR means either versions.env was already current (script exited early)
+          # or a previous run already merged it — both are no-ops, not errors.
+          branch="chore/bump-version-$RELEASE"
+          pr_url=$(gh pr list --head "$branch" --state open --json url --jq '.[0].url // empty')
+          if [[ -z "$pr_url" ]]; then
+            echo "::notice::no open PR on $branch, nothing to merge"
+            exit 0
+          fi
+          echo "pr_url=$pr_url" >> "$GITHUB_OUTPUT"
+
+      # contents+pull-requests WRITE on dsp-tools ONLY. The App is a review-bypass
+      # actor on main, which is what lets a bot-authored PR merge without a human
+      # approval. SHA-pinned (v3 == bcd2ba4) since this action handles private-key material.
+      - name: Mint DaSCH Bot App token (write — dsp-tools only)
+        id: app-token
+        if: steps.find_pr.outputs.pr_url != ''
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1  # v3
+        with:
+          app-id: ${{ secrets.DASCH_BOT_APP_ID }}
+          private-key: ${{ secrets.DASCH_BOT_APP_PRIVATE_KEY }}
+          owner: dasch-swiss
+          repositories: dsp-tools
+          permission-contents: write
+          permission-pull-requests: write
+
+      - name: Merge the bump PR once its checks are green
+        if: steps.find_pr.outputs.pr_url != ''
+        uses: ./.github/actions/merge-pr-when-green
+        with:
+          pr-url: ${{ steps.find_pr.outputs.pr_url }}
+          token: ${{ steps.app-token.outputs.token }}
+
+      - name: Notify on failure (internal)
+        if: failure()
+        env:
+          INTERNAL_WEBHOOK: ${{ secrets.GOOGLE_CHAT_DSP_RELEASE_INTERNAL_WEBHOOK_URL }}
+          RELEASE: ${{ inputs.release }}
+          REPO: ${{ github.repository }}
+          RUN_ID: ${{ github.run_id }}
+        run: |
+          set -euo pipefail
+          run_url="https://github.com/$REPO/actions/runs/$RUN_ID"
+          text="❌ bump-stack-versions.yml failed for DSP=$RELEASE · <$run_url|view run>"
+          curl -fsSL -X POST -H 'Content-Type: application/json' \
+            -d "$(jq -n --arg text "$text" '{text:$text}')" \
+            "$INTERNAL_WEBHOOK"
--- a/.github/workflows/bump-stack-versions.yml
+++ b/.github/workflows/bump-stack-versions.yml
@@ -47,12 +47,13 @@ jobs:
           git config user.email "github-actions[bot]@users.noreply.github.com"
 
       - name: Bump versions
+        id: bump
         run: python scripts/bump_stack_versions.py
         env:
           # Do not use the default token (GITHUB_TOKEN) for this action,
           # as any actions triggered by the user of this token will not trigger further actions
           # (i.e. tests on the version bumping PR).
-          # See https://github.com/googleapis/release-please-action and 
+          # See https://github.com/googleapis/release-please-action and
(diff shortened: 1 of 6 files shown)


.github/workflows/bump-stack-versions.yml
```diff
@@ -53,3 +59,64 @@ jobs:
...
+          set -euo pipefail
+          if [[ "${AUTO_MERGE_ENABLED,,}" == "false" ]]; then
+            echo "::notice::RELEASE_AUTO_MERGE_ENABLED=false, leaving the bump PR for a human to merge"
+            exit 0
+          fi
+          # Branch name is the one scripts/bump_stack_versions.py creates. No open
+          # PR means either versions.env was already current (script exited early)
+          # or a previous run already merged it — both are no-ops, not errors.
+          branch="chore/bump-version-$RELEASE"
+          pr_url=$(gh pr list --head "$branch" --state open --json url --jq '.[0].url // empty')
```
SamuelBoerlin: I'm not sure this is completely reliable, i.e., whether an immediately subsequent `gh pr list` is guaranteed to find a PR that has just before been created by `gh pr create`. It might be eventually consistent. At least on the Web UI there's often a few seconds of delay until a PR appears, and I wouldn't be too surprised if this were also the case here. Ideally the bump script would return the ID of the PR it created so we could use it as input for "Merge the bump PR once its checks are green".

.github/workflows/bump-stack-versions.yml
```diff
@@ -53,3 +59,64 @@ jobs:
           API: ${{ inputs.api }}
           APP: ${{ inputs.app }}
           DB: ${{ inputs.db }}
+
+      - name: Find the bump PR to merge (kill switch, idempotency guard)
+        id: find_pr
+        env:
+          GH_TOKEN: ${{ secrets.DASCHBOT_PAT }}
```
SamuelBoerlin: This should use a DaSCH Bot App token, not the PAT (which we want to move away from). Or wouldn't even just `GITHUB_TOKEN` be sufficient here?

.github/workflows/release-please.yml
```diff
@@ -70,3 +70,111 @@ jobs:
...
+          AUTO_MERGE_ENABLED: ${{ vars.RELEASE_AUTO_MERGE_ENABLED }}
+        run: |
+          set -euo pipefail
+          if [[ "${AUTO_MERGE_ENABLED,,}" == "false" ]]; then
+            echo "::notice::RELEASE_AUTO_MERGE_ENABLED=false, leaving the release PR for a human to merge"
+            exit 0
+          fi
+          # No open release PR means release-please found nothing to release, or a
+          # peer run already merged it — both are no-ops, not errors.
+          release_pr=$(gh pr list --repo "$REPO" --state open --json title,url,headRefName \
```
SamuelBoerlin: Same potential issue as other `gh pr list`. The release-please action apparently outputs a PR object output which I suspect already has the needed information: https://github.com/googleapis/release-please-action#outputs

.github/workflows/release-please.yml
```diff
@@ -70,3 +70,111 @@ jobs:
...
+          # disagreement means the PR was hand-edited or a run was interrupted —
+          # never something to merge unattended.
+          title_version="${PR_TITLE#chore: release }"
+          manifest_version=$(jq -r '."."' .github/release-please-manifest.json)
+          pyproject_version=$(grep -m1 '^version = ' pyproject.toml | sed -E 's/^version = "(.*)"/\1/')
+          echo "title=$title_version manifest=$manifest_version pyproject=$pyproject_version"
+          if [[ "$title_version" != "$manifest_version" || "$title_version" != "$pyproject_version" ]]; then
+            echo "::error::version mismatch on the release PR — refusing to merge"
+            exit 1
+          fi
```
SamuelBoerlin, line 158: Just thinking out loud here, but would it maybe make sense to instead check if all commits were authored by our DaSCH Bot App? Less things to break if we move around files.

SamuelBoerlin (changes requested):
We shouldn't use DASCHBOT_PAT anymore if possible. The rest looks fine, but I think the PR search right after PR creation could be improved as pointed out in the other comments.

.github/workflows/release-please.yml
mpro7: I'd keep the version check: the release PR's commits are authored by `daschbot`, not the App (#2389, `ce4c177` + `7d0030d`), so that check would block every release train until the PAT migration lands. The coupling it removes is also small, `manifest-file:` already hardcodes the manifest path, and `pyproject.toml`'s version line is fixed by `release-type: python`. The two guard different things: yours catches a human touching the branch, this one catches the version that gets tagged and published disagreeing with itself. Happy to add yours alongside it.

.github/workflows/release-please.yml
mpro7: Adopted: `id` on the release-please step, PR number as a job output, resolved with `gh pr view` so title/branch/state stay authoritative. That also let me delete the pre-existing curl-by-title lookup and one more PAT usage. `pr` is unset when the action neither created nor updated a PR, so the job is gated on it (fail-closed), and the raw output is echoed to the log so its shape is visible on the first live run.

.github/workflows/bump-stack-versions.yml
mpro7: That step is gone, and `gh pr create` now uses the App token too, an App installation token triggers workflow runs just like the PAT (only the default `GITHUB_TOKEN` doesn't), so the checks still run on the bump PR. Consequence: the App is now needed to open the PR, not just merge it, so a broken token fails before a PR exists. Say the word if you'd rather keep that leg on the PAT.

.github/workflows/bump-stack-versions.yml
mpro7: Done your way: the script now writes the URL `gh pr create` returns to `$GITHUB_OUTPUT` and the workflow merges exactly that PR, so the lookup and the duplicated branch name are gone. No output means no PR was opened, which stays a no-op. Unit tests cover the parsing and the output write.

mpro7:
@SamuelBoerlin Addressed in 3 commits: `bump-stack-versions.yml` is now PAT-free, and neither leg searches for the PR it merges. The PAT stays in `release-please.yml` (action token + uv.lock push) for a separate PR, swapping it changes who authors the release commit and tag and touches the `release`-triggered PyPI publish, which I don't want to do untested in here.

SamuelBoerlin (approved):
👍

mpro7:
@BalduinLandolt FYI - related to deployment skill

PR #2395, opened 2026-08-13
docs: add missing ontology documentation for `RegionPreviewValue` (DEV-6962)

Nora-Olivia-Ammann:


docs/data-model/json-project/ontologies.md
```diff
@@ -650,6 +650,36 @@ Example:
 ```
 
 
+#### `RegionPreviewValue`
```
jnussbaum: For consistency with the other titles:

```suggestion
#### `hasRegionPreview` Property
```

jnussbaum (approved):

docs/data-model/json-project/ontologies.md
Nora-Olivia-Ammann: Claude put it in the wrong section and i did not notice. It is not classified as a link value therefore I put it in the normal properties section.

PR #2400, opened 2026-08-18
chore(skills): add skill to identify blast-radius of feature work

Nora-Olivia-Ammann:
The PR to implement legal info on data missed some part of the repo. This skill is intended to identify all the locations that changes need to happen. It is intentional that it only identifies the locations without doing any implementation planning.


jnussbaum (commented):
Frontmatter: three real bugs

1. Task is not a tool name (SKILL.md:12). The canonical name is Agent; there's no Task alias. Permission rules "match the canonical name only," and unlike deny/ask rules, allow-grants get no startup warning — so this fails silently and every Phase 3 fan-out prompts for approval mid-run. Since Phase 3 specifically wants Explore agents, the precise grant is Agent(Explore).

2. Read-only is unenforced. The skill's central promise — "It is read-only: it reads code and git history and reports; it never edits" (:24) — rests entirely on prose. Per the docs, allowed-tools "does not restrict which tools are available: every tool remains callable." The field that actually enforces it is disallowed-tools, which removes tools from the pool. For a skill whose whole contract is don't touch anything, that's the fix that makes the contract real.

3. Blanket Bash is a wider self-grant than the skill uses. Its only Bash use is git diff / git status / git show (:79). Those are already in the built-in read-only set that runs unprompted, so the blanket grant buys ~nothing and costs the guarantee — and the docs specifically warn to review allowed-tools on skills checked into a repo.

Recommended block (YAML list form, because comma-separated parsing gets ambiguous with commas inside Bash(...)):

```
allowed-tools:
  - Read
  - Grep
  - Glob
  - Agent(Explore)
  - Bash(git diff:*)
  - Bash(git status:*)
  - Bash(git show:*)
  - Bash(git log:*)
disallowed-tools: Edit, Write, NotebookEdit
```

version: "1.0" (:13) is in neither Claude Code's frontmatter table nor the six-field Agent Skills spec — inert here, and a hard error if the skill is ever packaged or uploaded. Neither sibling has it. Drop it, or move it under metadata:.

Body: the issues worth fixing

4. "Hard stop / do not implement" will fight you later in the session. A skill body "enters the conversation as a single message and stays there for the rest of the session," and the docs explicitly say to write guidance as standing instructions rather than one-time steps. So Phase 4's absolute "Do not implement" (:72) doesn't expire — it's still arguing when you read the table and say "ok, now do row 3." Scope it to the invocation: this invocation ends with the report; if the user then asks for the edits, that's a new task.

    - Suggestion how to fix:  Add context: fork to your frontmatter when you want a skill to run in isolation. The skill content becomes the prompt that drives the subagent. It won't have access to your conversation history. **The one real cost**: "It won't have access to your conversation history." Phase 2 says to infer the change "from the working-tree diff if one exists, otherwise from the described intent" — and described intent is precisely what lives in the history a fork can't see. 

5. The no-duplication claim is false, and the maintenance rule cements the drift. Lines 26-29 and :174 both insist the skill only points at the inventory. But Playbooks A/B/C carry ~30 leaf paths the inventory doesn't have (it stops at directories/pipelines). That's fine — leaf hints are useful — but "when the suite moves or renames a subsystem, update CONVENTIONS.md… not this file" tells a future maintainer not to fix stale paths in the one file that has them. Split it honestly: subsystem list → CONVENTIONS.md; leaf-path hints → here, re-verified live in Phase 3.

6. The daschland flag is restated four times with three different trigger sets. Phase 4 (:79) says all four subsystems; Playbook A (:117) says create/excel2json/xmllib and omits xmlupload; Playbook B (:139) says xmlupload/xmllib; Playbook C (:154) says all four. Someone working Playbook A on an xmlupload-touching change won't flag the companion PR. This is exactly the miss-class the skill exists to prevent, reproduced inside the skill. Phase 4 already owns the rule — the playbook rows should say "see Phase 4" and nothing more.

7. Error-handling row 1 hard-stops when it could proceed. If the inventory section is missing, the skill still holds three usable (narrower) playbooks; stopping dead serves nobody. Better: proceed, and report the missing inventory so it gets fixed. The title match is also fragile — CLAUDE.md:81 and REVIEW.md:22 cite the shorter "dsp-tools subsystem inventory," so matching that substring survives a parenthetical edit.

8. Applies? and Status encode the same fact (:74) — Status already admits n/a. ith Status ∈ {TODO, done, MISSING, n/a}, and fill in one example row: thetemplate body is empty, so column semantics are currently left to invention. 
                                                                                                        
9. Phase 3 fan-out is unbounded. "Fan out… in parallel" with no agent count and no return contract. One agent per precedent per applicable playbook (so 2, or 4 when A+B) and "returnpath list, not prose" would make it predictable and cheap.

10. "New value type" is advertised but homeless. The description leads with it  archetype A+B+xmllib case and lands in Playbook C, "anything else" — the vaguest bucket. It's also the most common real change in dsp-tools. It deserves the explicit A+B+xmllib callout.               

12. Minor: path prefixes mix src/dsp_tools/... with paths relative to src/dsp_tools/ (api-shapes.ttl differs from CONVENTIONS.md:38 this way), which can defeat a sub-agent grepping the literal string — pick one and say so once. And the trigger-phrase block could m_to_use field, leaving description for what the skill does; at 772 chars you'rewell under the 1,536 cap, so that's style, not necessity.

jnussbaum (approved):

PR #2403, opened 2026-08-20
chore(ci): take external link checking off the release-train critical path

mpro7:
## Final shape

**External links are still checked on every PR — they just can't block a merge any more. Everything that depends only on this repository stays required.**

| What is checked | Where | On | Required |
|---|---|---|---|
| Internal links, `#heading` anchors, `nav` (`mkdocs build --strict`) | `check-docs` | every PR | ✅ yes |
| markdownlint, xmllib docstring links | `check-docs` | every PR | ✅ yes |
| 227 external URLs across ~40 hosts (lychee) | `check-external-links` | every PR **except** `release-please--*` and `chore/bump-version-*` | ❌ **no** |

A dead external link is a red ✗ on the PR with the URLs in the job summary — visible to the author, who is best placed to fix a link they just added — but the merge stays available, so one third-party outage cannot stall every open PR. Anchor validation *improved*: MkDocs uses the site's real slug algorithm, where the removed tool used `uslug`.

Two release-automation fixes ride along: the merge loop re-runs a terminally-failed required check **once** instead of polling it to the 30-minute deadline, and a `merge_release_pr` dispatch input lets a stranded release train be resumed without a hand-merge.

<sub>The commit history below tracks the review discussion (weekly cron → GitHub issue → Chat alert → per-PR non-required). It squash-merges; the table above is the end state.</sub>

---

Fixes the failure of [Release-Please run 32334300996](https://github.com/dasch-swiss/dsp-tools/actions/runs/32334300996), which stranded the 19.0.1 release PR (#2393).

## What happened

`check-docs` was one of 13 required status checks on `main`, and its `check external links` step validated **227 external URLs across ~40 hosts** on every PR. On 2026-08-20 at 05:07 UTC it declared a perfectly healthy `https://hynek.me/articles/testing-packaging` invalid — a transient bot-block. That failed `check-docs`, which blocked the auto-merge job; the job re-issued the same merge call for its full 30-minute budget, could not possibly succeed, and alerted at 05:37. A manual re-run of the identical commit passed at 06:35, an hour after the merge window had closed.

Not a one-off: at 07:02 the same check failed again on the same PR, this time on a `github.com` link — GitHub itself rate-limiting. **Two unrelated links flaked within two hours.**

[...]

PR #2404, opened 2026-08-20
chore(ci): skip the e2e suite on release-please PRs

mpro7:
## Final shape

**A release-please PR runs only the two checks that read a file it actually changed.** Everything else is skipped — not to save time, but because it reads none of the four changed files.

| Check | On a release-please PR | Why |
|---|---|---|
| `check-docs` | ✅ **runs** | `CHANGELOG.md` is rendered into the site via `docs/changelog.md` → `{% include-markdown "../CHANGELOG.md" %}`, and it's in the mkdocs nav |
| `Check PR Title` | ✅ **runs** | validates the `chore: release X.Y.Z` title — 6s |
| 7 × `Tests e2e` | ⏭️ skipped | no source changed |
| `unittests`, `integration`, `benchmarking` | ⏭️ skipped | no source changed |
| `lint` | ⏭️ skipped | no Python, no YAML changed |
| `distribution` | ⏭️ skipped | checks packaging, which a version bump doesn't alter — **moved to publish time**, see below |
| `latest-backend-components` | ⏭️ skipped | no source to integrate; also non-required |
| `upload-coverage` | ⏭️ skips itself | `needs: [unittests, integration]` |
| CodeQL × 3 | runs | GitHub **default setup**, not a workflow — cannot be skipped by an `if:`. Non-required, so it never gates the merge |

**Required-check critical path: 5m53s → ~42s.** Roughly 22 runner-minutes saved per release.

## What a release-please PR actually changes

| File | Change | Read by any check? |
|---|---|---|
| `pyproject.toml` | version string | the built wheel — now checked at publish |
| `.github/release-please-manifest.json` | version string | nothing in CI |
| `uv.lock` | same version + reordered resolution markers | nothing — plain `uv lock` cannot move a dependency |
| `CHANGELOG.md` | new release section | **`check-docs`** |

## The one real safety change: the wheel is now checked where it is published

`distribution` never tested business logic. Its own docstrings:

> *"This does NOT test the xmllib. This only tests if all dependencies used in the xmllib are shipped in the distribution."*
> *"Test if the CLI entry point works for the end user… all dependencies have been installed… the resource files are accessible on the end user's machine."*

Those failure modes arrive with a **source** change, and are caught on the PR that introduces one — where the job still runs. On a release PR it re-validated packaging main had already validated.

Meanwhile `publish-release-to-pypi.yml` was `uv build` straight into `uv publish`, so the artifact that actually reaches PyPI was never inspected — and a PyPI version cannot be re-uploaded. This PR adds `check-wheel-contents` between those two steps, so the wheel that is genuinely published gets checked rather than an equivalent one built earlier. Verified against the real 19.0.1 wheel: `dsp_tools-19.0.1-py3-none-any.whl: OK`.

Net: the release PR gets faster **and** the publish path gets a check it never had.

## Why this is safe

**`main` is always fully tested.** Branch protection is `strict`, so every merge's CI ran against the combined tree — main is never in a state that hasn't passed the full suite. A release PR is up to date with main and source-identical, so it re-tests a tree that already passed.

**`uv.lock` cannot smuggle in a dependency change.** release-please runs plain `uv lock`; per uv's docs, *"locking will have no effect"* when dependencies are unchanged. There is a comment at that call site warning against adding `--upgrade`.

**Skipped ≠ missing.** A required check skipped by a job-level `if:` reports Success, so the release PR stays mergeable and no branch-protection rule is edited. (The guard is on the *jobs* for this reason: a `branches`/`paths` filter on the trigger would leave the check `Pending` forever and block the PR permanently.) Same pattern as dsp-api's `build-and-test.yml`, DEV-6678.

## ⚠️ The predicate keys on the head branch — do not change it to a bot-author check

Both release PRs are bot-authored (`daschbot` and `app/dasch-bot-app`), so an author- or `[bot]`-based condition would also match the **start-stack bump** — the one PR that must run the full e2e suite.

`test/e2e/setup_testcontainers/setup.py:52` reads the DSP-API / Sipi / Ingest / Fuseki image tags from `src/dsp_tools/resources/start-stack/versions.env`:

```python
versions_env_pth = Path("src/dsp_tools/resources/start-stack/versions.env")
...
return ImageVersions(fuseki=versions["DB"], sipi=versions["API"], ingest=versions["API"], api=versions["API"])
```

…and `versions.env` is the **only** file a start-stack bump changes. That e2e run is the sole validation that dsp-tools works against a newly released DSP stack, immediately before merging it auto-releases dsp-tools against exactly that stack. `!startsWith(github.head_ref, 'release-please--')` cannot match `chore/bump-version-*`; a bot-author predicate could. There is a comment in `tests-e2e.yml` saying so.

## Verification

`actionlint` and `yamllint` clean on all four workflows. This PR's own head branch isn't `release-please--*`, so every guarded job ran and passed here — the guard is proven not to over-match.


Nora-Olivia-Ammann (changes requested):
This does not target the job you want to disable. This lives in linting, basically you now removed all the code safeguards.

Nora-Olivia-Ammann:
If you would like to skip the tests why not also skip the others? Could there not be a more elegant way to skip entire files?

mpro7:
> This does not target the job you want to disable. This lives in linting, basically you now removed all the code safeguards.

I think two separate things may be going on, so let me address both.

**This is not the fix for the release-train failure.** That's #2403. What broke the 19.0.1 merge was the external link check in `check-docs`, and #2403 moves it off the required gate. This PR is an unrelated optimisation and doesn't touch `check-docs` or `lint-code.yml` at all — its only file is `.github/workflows/tests-e2e.yml`. Sorry, the description leaned on #2403 for context instead of saying that up front.

**The skip applies to release-please PRs only, not to all PRs.** The guard is `if: ${{ !startsWith(github.head_ref, 'release-please--') }}` on each job. On any normal PR the condition is true and every e2e job runs exactly as today — this PR's own checks are the proof: head branch `chore/skip-e2e-on-release-please-prs`, all 7 e2e jobs ran and passed.

**What still runs on a release-please PR** — 7 of the 13 required checks, ~1m46s: `unittests`, `integration`, `benchmarking`, `distribution` (builds the wheel at the release version), `lint`, `check-docs`, `Check PR Title`. So the unit and integration suites and the package build still gate the merge.

**Why skipping e2e there is safe:** a release-please PR changes 4 files — `CHANGELOG.md`, `pyproject.toml` (version), the release-please manifest, and `uv.lock` — with no source changes. `main` is strict, so every merge's CI ran against the combined tree and main is always in a state that passed the full suite; the release PR re-tests that same tree. The one file that could break the argument is `uv.lock`, and it can't: release-please runs plain `uv lock`, and per uv's docs "locking will have no effect" when dependencies are unchanged. I've added a comment at that call site so nobody adds `--upgrade` later without seeing the consequence.

Gain: 5m53s → 1m46s on the release train, ~18 runner-minutes per release. Same pattern dsp-api uses in `build-and-test.yml` (DEV-6678).

Happy to drop this PR if you'd rather not have it — #2403 is the one that actually unblocks the release, and the two are independent. But if the concern was one of the two readings above, hopefully this clears it up. Which job did you have in mind as the one that should be disabled?

Nora-Olivia-Ammann (approved):

jnussbaum (commented):
@mpro7 You could even go further. The release-please PR actually needs almost no testing at all. Since no code files are touched, you also don't need the unit tests and integration tests.

mpro7:
> @mpro7 You could even go further. The release-please PR actually needs almost no testing at all. Since no code files are touched, you also don't need the unit tests and integration tests.

Dropped all I could, but CodeQL config lives in repo settings, so can't do it via GH workflows. If you never use it actually it could be disabled at all, it was a case in app and api too in the past.

PR #2405, opened 2026-08-20
fix(xmlupload): distinguish expected from failed interruptions (DEV-6183)

jnussbaum:
Resolves [DEV-6183](https://linear.app/dasch/issue/DEV-6183/fix-inconsistencies-in-error-handling-of-xmlupload).

## Why

Every interruption of a running `xmlupload` raised `XmlUploadInterruptedError` and exited with code `0`,
regardless of the reason. Three consequences:

- A server going down mid-upload looked like a **successful** run to shells and CI.
- A deliberate `--interrupt-after` printed *"An internal error occurred. Please contact the dsp-tools
  development team"*, because `XmlUploadInterruptedError` is an `InternalError` subclass.
- `Ctrl + C` exited `0` instead of the `130` that the error-handling guideline reserves for it.

Independently, a `BadCredentialsError` did **not** halt the upload: the `raise err from None` in
`_execute_one_resource_data_upload()` was a no-op, and the blind `except Exception` in
`_execute_one_resource_upload()` caught the error, registered the resource as failed and continued with
the next one. Since the auth token is cached and never refreshed, a token expiring during a long upload
made *every remaining resource fail* instead of aborting with a clear credentials error.

## What changed

`sys.exit()` is gone from the xmlupload code. `entry_point.py` decides every exit code, as the
error-handling guideline prescribes:

| Interruption reason                            | Exit code | Mechanism                             |
| ---------------------------------------------- | --------- | ------------------------------------- |
| `--interrupt-after` limit reached              | `0`       | return value, not an exception        |
| `Ctrl + C`                                     | `130`     | `KeyboardInterrupt` propagates        |
| Permanent connection loss or permanent timeout | `1`       | `XmlUploadInterruptedError`           |
| Rejected credentials                           | `1`       | `BadCredentialsError`                 |

- Reaching the `--interrupt-after` limit is an expected outcome, so `interrupt_if_indicated()` became the
  predicate `interruption_is_indicated()` and `_upload_all_resources()` returns `True` for it — the
  guideline's "no exceptions for expected control flow" rule.
- `handle_upload_error()` is replaced by `persist_state_for_resume()`, called once in
  `_upload_all_resources()` for *all* interruption reasons. Previously a `BadCredentialsError` from the
  ingest client escaped without the state being saved; now every path stays resumable with
  `resume-xmlupload`.
- `handle_permanent_timeout_or_keyboard_interrupt()` is split into `handle_permanent_timeout()` and
  `handle_keyboard_interrupt_during_creation()`, so no `isinstance` branching on the exception type is
  needed to pick the exit code. Both keep the "unclear if the resource was created / use
  `--skip-first-resource`" message.
- A `BadCredentialsError` now propagates before the blind `except Exception`. The resource deliberately
  stays in `pending_resources`, so `resume-xmlupload` retries it once the credentials are fixed.
- The retry loop no longer logs the traceback of a `DspToolsRequestException` a second time — the client
  already logged it with `logger.exception` when converting the `requests` error.

All three commands that go through `execute_upload()` benefit: `xmlupload`, `ingest-xmlupload` and
`resume-xmlupload`.

## Notes for the reviewer

- `execute_upload()` returning `True` for a deliberate interruption is what produces exit code `0`. This
  is the behaviour a script driving a chunked upload relies on, and it matches what `sys.exit(0)` did
  before. The docstring says so explicitly.
- The two `Ctrl + C` handlers emit their message as a `DspToolsUserWarning` (which prints) because
  `entry_point.py` prints nothing for a `KeyboardInterrupt`, whereas it prints the DSP-TOOLS exceptions
  itself. There is a comment in `handle_errors.py` recording that asymmetry.
- The integration tests no longer patch `handle_upload_error` to stop it from exiting. They now assert on
  the real outcome instead: the return value, the pickled upload state read back from a `tmp_path`, and
  the exception that escapes. The repeated `UploadClients(...)` construction moved into a `clients`
  fixture, which is what makes the diff in that file large.
- Verified by driving the real CLI entry point for each reason: exit codes `0` / `1` / `1` / `130`, with
  bad credentials now printing a single clean line instead of the "contact support" block, and a planned
  interruption printing only the reason plus the resume hint.
- Not related to this change, but found while running the suites together: three xmlupload tests fail
  when `test/unittests` and `test/integration` run in one pytest session, because loguru records stop
  reaching `caplog`. Reproduced on `main` with everything stashed; CI does not see it because
  `just unittests` and `just integration-tests` are separate processes.


Nora-Olivia-Ammann:
## Review notes (DEV-6183 — distinguish expected from failed interruptions)

Reviewing only — not pushing any changes to this branch. Flagging one docs inconsistency and a couple of doc-precision nits for the author to decide on.

**The code itself looks correct and is well covered.** I verified end-to-end that each interruption reason maps to the documented exit code, with no interceptor swallowing the exception before `entry_point.py`:

| Reason | Raises | Exit code |
| --- | --- | --- |
| `--interrupt-after` reached | — (returns `True`) | `0` |
| `Ctrl + C` | `KeyboardInterrupt` | `130` |
| connection lost / timeout | `XmlUploadInterruptedError` (`InternalError`) | `1` |
| bad credentials | `BadCredentialsError` (`UserError`) | `1` |

Also confirmed: `BadCredentialsError` is not a `DspToolsRequestException`, so dropping the explicit catch in the retry loop can't cause a retry; `except BadCredentialsError: raise` correctly precedes the blind `except Exception`; all renames/splits are complete with no stale references in `src/`, `test/`, or `docs/`; and `just ruff-check`, `mypy`, and the affected tests (34) all pass locally. Moving `sys.exit()` out of `handle_errors.py` into `entry_point.py` is a nice cleanup.

### Docs plan (recommended before merge)

1. **Reconcile the `Ctrl + C` state-saving guarantee (the one real inconsistency).**
   - This PR adds to `docs/developers/architecture/error-handling.md`: *"In all four cases the upload state is saved, so the upload can be continued with `resume-xmlupload`."* — and one of the four cases is `Ctrl + C`.
   - But the pre-existing (untouched) caveat in `docs/data-file/data-file-commands.md` still reads: *"We do not guarantee that the state of an xmlupload is cleanly saved after `Ctrl + C`. We only guarantee this for `dsp-tools xmlupload --interrupt-after`."*
   - These read as contradictory for the same feature. The code now *attempts* a save on `Ctrl + C` (`_upload_all_resources` catches `KeyboardInterrupt` → `persist_state_for_resume`), but the `--skip-first-resource` flow exists precisely because a `Ctrl + C` mid-creation can leave the current resource in an unknown state — so the caveat is still substantively true.
   - Suggestion: in `error-handling.md`, say the save is *attempted* in all four cases, and keep the *clean-boundary* guarantee scoped to `--interrupt-after` (or cross-reference the caveat), so the two pages agree.

2. **Tighten two claims in `src/dsp_tools/commands/xmlupload/CLAUDE.md`.**
   - *"All handlers that end a running upload live in `handle_errors.py`"* overstates: the `BadCredentialsError` short-circuit lives in `execute_upload.py` (`except BadCredentialsError: raise`), and the `handle_errors.py` module comment lists `BadCredentialsError` among what "the handlers below convert" even though no handler there produces it — it's raised natively by the client and merely re-raised.
   - *"Every one of these paths calls `persist_state_for_resume()` once, in `_upload_all_resources()`"* — for the `--interrupt-after` path the call is one level down, in `_report_planned_interruption()`. Minor wording tweak (e.g. "in or via `_upload_all_resources()`").

### Optional, non-blocking (not docs)

- `test/unittests/commands/xmlupload/test_handle_errors.py` calls the handlers directly, with no exception in flight, so `logger.exception()` inside them logs no real traceback — the tests verify the raised type and the warning, but not the traceback logging. No defect (every production call site is inside an `except`); wrapping the calls in `try/raise/except` would make the tests faithful.
- The interrupt-after message `"Interrupted: Maximum number of resources was reached (N)"` (built in `_report_planned_interruption`) is no longer asserted by any test — a `caplog` check would close that small gap.

*Reviewed with python-reviewer, consistency-reviewer, and code-simplicity-reviewer; findings verified against the source.*

Nora-Olivia-Ammann (approved):

PR #2419, opened 2026-09-01
feat: add resource-side legal metadata to json_header.xlsx (DEV-7000)

jnussbaum:
## What

`excel2json` now reads an optional **"Data Legal Settings"** sheet from `json_header.xlsx` and writes three project-level legal fields into the generated JSON project definition:

- `data_license` — single license IRI, read from the first data row.
- `data_copyright_holder` — single string, read from the first data row.
- `default_data_authorship` — a list, one author per row in that column.

## Why

The JSON project schema and the `create` command already support these three fields, but `excel2json` never produced them. A user could not set them from Excel. This closes that gap (DEV-7000).

## Design decisions

- The sheet and all three fields are optional. An absent sheet emits no legal keys.
- `default_data_authorship` uses a multi-row column (one author per row), so multiple authors are possible. Order is preserved.
- The license value is not validated in `excel2json`. The `create` command's JSON-schema check catches an invalid license.

## Changes

- `models/json_header.py` — new `DataLegalSettings` dataclass whose `to_dict()` emits only non-empty fields. `Project` gained a `data_legal_settings` field.
- `json_header.py` — `_check_data_legal_settings` (requires the three columns) and `_extract_data_legal_settings`, both wired into the optional-sheet path.
- `project.py` — added the three keys to the `_sort_project_dict` allow-list, so the sort does not raise.
- `docs/data-model/excel2json.md` — documented the new sheet.

## Tests

- Unit: new check, extract, and model tests, plus fixtures.
- Integration: added the sheet to the fixture `json_header.xlsx`; updated `expected_json_header.json` and `excel2json-expected-output.json`.
- All 248 excel2json unit + integration tests pass. `ruff`, `ruff format`, and `mypy` are clean on the changed files.

## Reviewer notes

- The fixture `json_header.xlsx` is a binary file, re-saved via openpyxl to add the sheet. Its byte size dropped because openpyxl does not preserve the original styling, but every cell value is preserved (tests confirm).


src/dsp_tools/commands/excel2json/models/json_header.py
```diff
@@ -105,6 +107,23 @@ def to_dict(self) -> list[str]:
         return sorted(self.licenses)
 
 
+@dataclass
+class DataLegalSettings:
+    data_license: str | None
+    data_copyright_holder: str | None
+    default_data_authorship: list[str]
+
+    def to_dict(self) -> dict[str, Any]:
```
Nora-Olivia-Ammann, line 116: I think it would be nice if it produced the empty fields, legal metadata should be added, therefore while I do think we should accept it being left empty it should be noted.

Nora-Olivia-Ammann (approved):

src/dsp_tools/commands/excel2json/models/json_header.py
jnussbaum: You mean, it should produce a JSON with these keys present, but no values in it? Currently, the JSON schema forbids `data_license": ""`, so I would have to relax it there.

But I don't see much value in having these empty keys in the file. Or do you have a use case?

Nora-Olivia-Ammann: Right, relaxing the schema is not a good option. Perhaps we can solve it by producing a warning for the user? I would just like to 1. be backwards compatible. 2. remind users that they need to fill this out

jnussbaum: So what is the status of these fields? Are they optional but highly recommended? If I understand you correctly, you want to nudge users into filling it out. But then we should also have a warning in `create`, shouldn't we? 

I just don't want to overcomplicate things.

Nora-Olivia-Ammann: true, perhaps you could merge it as is, and ask PM what behaviour they expect.

PR #2420, opened 2026-09-01
refactor(xmlupload): mint the value blank-nodes in the outer most function

Nora-Olivia-Ammann:


src/dsp_tools/commands/xmlupload/make_rdf_graph/make_resource_and_values.py
```diff
@@ -70,8 +70,10 @@ def _make_values_graph_from_resource(
 
     if file_found := resource.file_value:
         abstract_value = _get_abstract_file_value(file_found, bitstream_information)
+        file_bn = BNode()
```
Nora-Olivia-Ammann, line 73: Easier to inject from outside

src/dsp_tools/commands/xmlupload/make_rdf_graph/make_values.py
```diff
@@ -69,12 +69,22 @@ def make_values(values: list[ProcessedValue], res_node: BNode | URIRef, lookups:
     """
     properties_graph = Graph()
     for val in values:
-        single_prop_graph = _make_one_value_graph(val=val, res_node=res_node, iri_lookups=lookups)
-        properties_graph += single_prop_graph
+        val_node = BNode()
```
Nora-Olivia-Ammann, line 72: This single line is the reason for the change. We need to provide all the value IRIs in the bulk import but that is not allowed in the regular xmlupload. So we have this one place where through a boolean we will know if we create a b-node (for regular upload) or a value IRI for bulk import.

test/unittests/commands/xmlupload/make_rdf_graph/test_make_values.py
Nora-Olivia-Ammann, line 1: There is a large diff here, because the tests are re-written in a more readable style that is also preferred by claude. I had to change them anyway because of the val-bnode, other than that no changes.

jnussbaum (approved):

PR #2427, opened 2026-09-11
chore: remove pre-Claude-5 ceremony from CLAUDE.md

mpro7:
## Why

Claude 5 family models follow instructions literally, so approval and ceremony blocks written to compensate for weaker models now cause per-step prompting and stalled sessions instead of safety. The Claude Code permission system enforces the real gates for every model version (mutating tools are rule-gated and prompt on anything not allowlisted), so removing these lines is safe for the remaining 4.x users too.

Team usage is already ~69% Claude 5+, and the sub-5 share shrinks daily.

## Triage

Every candidate had to answer: *what concretely breaks or degrades in a Claude 5 session if this stays?* No nameable failure meant the text was left alone — the DROPPED rows below are part of the record.

| # | Line | Quote | Class | Status | Reason |
|---|------|-------|-------|--------|--------|
| 1 | CLAUDE.md:199 | "Ask me one question after another, and provide a structured form where I can submit my answers." | C | **REMOVED** | A blanket interview protocol with no output contract. Read literally, "fix the failing validate-data test" opens a one-question-at-a-time interview and emits a fill-in form instead of running pytest. The `eng:grilling` skill still does one-question-at-a-time deliberately when it is invoked — removing this blanket rule does not touch that. |
| 2 | CLAUDE.md:220 | "ALWAYS ask for clarification rather than making assumptions." | B | **REMOVED** | A restated model default, inverted into harm by the literal "ALWAYS": it overrides routine judgement, so the model asks which test directory to use instead of reading the answer out of `CONVENTIONS.md`. |
| 3 | CLAUDE.md:233-240 | `### Learning-Focused Error Response` — "Treat each failure as a learning opportunity, not an obstacle" / "Explain what you learned about the tool/codebase" / "Build competence with development tools rather than avoiding them" | C | **REMOVED** | Reflection ritual. A `ruff` failure produces a paragraph about what the model learned about ruff before (or instead of) the fix; two of the four bullets carry no behaviour at all. See the note below about the fourth bullet. |
| 4 | CLAUDE.md:195-198 | "always think critically … If you think it doesn't make sense, then you should push back, but you should also cite evidence." | — | DROPPED — kept | Real advisory discipline, and the evidence requirement gives it teeth. |
| 5 | CLAUDE.md:207 | "You MUST ask permission before reimplementing features or systems from scratch" | A-shaped | DROPPED — kept | A scope gate on a large, hard-to-reverse action. Rare trigger, high value — not per-step ceremony. |
| 6 | CLAUDE.md:221 | "If you're having trouble with something, it's ok to stop and ask for help." | — | DROPPED — kept | Permissive escape hatch, not a mandate; costs nothing per turn. |
| 7 | CLAUDE.md:231 | "If you want to adjust a test for any reason, ask for confirmation, even if you suspect the test to contain the error." | A-shaped | DROPPED — kept | A targeted guard against the classic failure of "fixing" a red test by weakening it. Concrete, narrow, worth the interruption. |
| 8 | CLAUDE.md:194 | "You are a Senior Software Developer who supports me in my daily work." | B-shaped | DROPPED — kept | Persona framing with no behavioural mandate; nothing breaks if it stays. |
| 9 | CLAUDE.md:214 | "Whenever you modify the codebase, make sure to also update all CLAUDE.md files (if necessary)" | — | DROPPED — kept | Repo convention with a real upkeep purpose. |
| 10 | 9 nested module files (2781 lines): `cli/`, `clients/`, `utils/`, `xmllib/`, `commands/{create,excel2json,update_legal,validate_data,xmlupload}/` | — | — | Clean | Architecture and domain docs. The only pattern-sweep hits were `Authorization:` headers inside code examples and the description of the `prompt_until_valid_answer` helper. |

**One judgement call worth flagging:** the fourth bullet of the removed section, "Research the specific error before attempting fixes", went with it. On its own it is a restated default (class B), and keeping a section titled "Learning-Focused Error Response" over that single generic line would have left the heading describing content that no longer exists. If the team wants it, it belongs as a one-liner under **Code quality** — say so and I will add it back there.

**Blast radius:** grepped the repo for `Learning-Focused` and `learning opportunity` across all file types. No other references. `just markdownlint` ignores `**/CLAUDE.md`, so the docs CI gate is unaffected.

## Questions for the team

Nothing was left in place for unclear intent — the kept rows above are kept on their merits. The single open decision is the "Research the specific error" bullet noted above.


jnussbaum (approved):
Thanks, @mpro7 , it's a good idea to critically assess old agent instructions. This inspired me to investigate further, perhaps we have some more stuff that should be fixed.

PR #2428, opened 2026-09-14
fix(upload-files): handle placeholder bitstreams via ParsedResource

Nora-Olivia-Ammann:
Fixes DEV-7255.

## Problem

`dsp-tools upload-files` aborts whenever the XML data file contains a `<placeholder-file>`. The XSD allows a `<bitstream>` to hold either a file path or a `<placeholder-file>` child, so this input is valid. `upload-files` could not handle it and failed with `InvalidIngestInputFilesError` (or `AttributeError` when the tags have no whitespace between them).

## Root cause

`_get_validated_paths` re-implemented the bitstream contract with an ad-hoc xpath:

```python
paths = {Path(x.text.strip()) for x in root.xpath("//bitstream")}
```

A placeholder bitstream has no file path. `x.text.strip()` yields `""`, so `Path("")` becomes `Path(".")`, which `check_files` rejects as an unsupported extension. The canonical parser `get_parsed_resources` already classifies a placeholder correctly, but `upload-files` duplicated the logic instead of reusing it. No test covered the placeholder path, so the drift was silent.

## Change

Route `upload-files` through `get_parsed_resources`, the single source of truth for the bitstream contract. `_get_validated_paths` now takes a `list[ParsedResource]` and keeps only real bitstream file values, excluding placeholder and `<iiif-uri>` by type — not by re-parsing XML.

The code is split into two levels for the planned bulk-import, which builds `ParsedResource` once and passes it down:

- Level 1 (public, standalone): `upload_files(xml_file, creds, imgdir)` parses the XML, then calls level 2.
- Level 2 (private now): `_get_validated_paths(resources)`. The bulk-import will hold `resources` already, so it only needs this function made public — no re-parse.

`get_parsed_resources` is CPU-only, no network call. At 50,000 resources it adds ~1.4 s over the xpath, under 1 % of the HTTP upload time.

## Tests

Both were missing, which is why the drift went unnoticed.

- Unit: real bitstream included, placeholder skipped, iiif skipped, no-file skipped, mixed, unsupported-file raises, plus a parser-seam test.
- E2e: creates project 4123 and uploads the systematic project (`test-data-systematic-4123.xml`, which holds a placeholder resource), asserting success and 41 uploaded files.


src/dsp_tools/commands/ingest_xmlupload/upload_files/CLAUDE.md
```diff
@@ -0,0 +1,43 @@
...
+## Conventions
+
+- The bitstream contract lives in `get_parsed_resources` (utils), not here. Do not re-parse
+  `<bitstream>` with xpath. Extract file paths from `ParsedResource` and filter by type
+  (`ParsedFileBitstream`). Placeholder and `<iiif-uri>` values are then excluded by type.
+  A duplicated xpath parser caused the placeholder crash in DEV-7255.
+- Two-level design: `upload_files` takes the XML and runs standalone. `_get_validated_paths` takes
+  `list[ParsedResource]`. The planned bulk-import builds `ParsedResource` once and reuses level 2 —
+  make it public then. If the bulk-import is a separate command package, move it to utils first,
+  because commands must not import from other commands.
```
jnussbaum: I feel a bit unwell about mentioning future implementation plans... I think CLAUDE.md should describe the current state of the code.

src/dsp_tools/commands/ingest_xmlupload/upload_files/CLAUDE.md
```diff
@@ -0,0 +1,43 @@
...
+- Two distinct error types: `FileProblems` for local problems before upload (missing or unsupported
+  files). `UploadFailures` for files the ingest server rejects. Both expose `execute_error_protocol()`,
+  which prints the list or writes a CSV past `maximum_prints`.
+- `creds.server` is the DSP-API URL (passed to `get_parsed_resources`). `creds.dsp_ingest_url` is the
+  ingest server (passed to `BulkIngestClient`).
+
+## Tests
+
+Unit: `test/unittests/commands/ingest_xmlupload/`. E2e: `test/e2e/commands/ingest_xmlupload/`.
+Keep a placeholder bitstream in the covered cases — it guards the DEV-7255 regression.
```
jnussbaum: I like that this is mentioned, but I wonder if it's the right place. The placeholder lives in "testdata/xml-data/test-data-systematic-4123.xml", so it should be mentioned there.

jnussbaum (commented):
Good catch, thank you. It looks well, but see my 2 comments.

jnussbaum (approved):

PR #2429, opened 2026-09-14
fix: remove duplicate error logging from intermediate exception handlers (DEV-6167)

jnussbaum:
[DEV-6167](https://linear.app/dasch/issue/DEV-6167/fix-duplicate-error-logging)

## Summary

Several intermediate exception handlers logged an error and then let the same error surface a second time when `entry_point.py`'s top-level handler caught it. This removes the redundant intermediate logging, per the propagating-vs-converting rule in `docs/developers/architecture/error-handling.md`.

The ticket is about six months old and named three specific violations. Re-checking each against current code found that two no longer hold as described (already fixed, or never actually duplicated). A from-scratch audit of every `logger.error()`/`logger.exception()` call site in `src/` found six real, currently-present violations across six files instead — including one not named anywhere in the source ticket/document.

## Changes

- **`cli/utils.py`** — removed 3 redundant `logger.error()` calls before fresh raises in `_check_api_health()`, `_check_ingest_health()`, and `get_canonical_server_and_dsp_ingest_url()`.
- **`excel2json/lists/compliance_checks.py`** — removed 5 redundant `logger.error()` calls before aggregated-problem raises.
- **`start_stack/start_stack.py`** — folded response/process diagnostics into `FusekiStartUpError`'s message (curated to named fields — status code, text, return code — rather than a raw `vars()` dump, per review) instead of logging them separately; removed the 4 now-redundant `logger.error()` calls; fixed a stray-quote typo surfaced along the way.
- **`utils/request_utils.py`** — removed the redundant `logger.error()` in what is now `print_and_raise_timeouts()` (renamed from `log_and_raise_timeouts()` — it no longer logs, only prints and re-raises the same caught exception unchanged).
- **`validate_data/validation/validate_ontology.py` and `.../get_validation_report.py`** — split a broad `except Exception` so an already-logged `ShaclValidationCliError` (from a failed Docker command) propagates without being re-logged, while a genuinely fresh, not-yet-logged `ShaclValidationError` still gets its one log. Caught by TDD: an earlier draft used the broader `except BaseError`, which would have silently swallowed that fresh error's own legitimate log too.
- **`validate_data/shacl_cli_validator.py`** — added a missing `from None`, found during review: without it, the "already logged once" guarantee the two files above rely on wasn't actually watertight, since the original traceback would re-appear via exception chaining wherever the error is eventually logged upstream. Added the first direct unit test for `ShaclCliValidator.validate()`.

Cross-referenced but deliberately left out of scope (belongs to the sibling ticket [DEV-6166](https://linear.app/dasch/issue/DEV-6166/fix-exception-chain-patterns-converting-vs-propagating), "chain hygiene"): a handful of `logger.error()`-instead-of-`logger.exception()` sites in `json_parsing.py`, `mapping/config_file.py`, `create_on_server/onto_utils.py`, and `create/project_validate.py` that already log exactly once and don't duplicate.

## Test plan

- [x] `just lint` and `just mypy` clean on every changed file
- [x] All new/extended tests pass (each written to fail first, confirming the described bug, then pass after the fix)
- [x] All 17 touched files' tests pass together (135 passed)
- [x] Manually triggered two of the fixed failure paths end-to-end (`dsp-tools create` against an invalid server URL, `dsp-tools excel2lists` against a missing directory) via `DSP_TOOLS_SAVE_ADDITIONAL_LOG_FILE_IN_CWD=true`, confirming exactly one `ERROR` log line per failure
- [ ] Full plan and audit trail: [dasch-specs PR #201](https://github.com/dasch-swiss/dasch-specs/pull/201)


test/unittests/commands/excel2json/lists/test_compliance_checks.py
```diff
@@ -443,8 +458,10 @@ def test_all_problem(self, cols_en_1: Columns, cols_en_1_2: Columns) -> None:
             "The Excel sheet 'sheet2' has the following problem(s):\n"
             "    - Row Number: 6, Column(s) that must be empty: en_2"
         )
-        with pytest.raises(InvalidFileContentError, match=expected):
-            _check_for_erroneous_entries_all_excels(all_sheets)
+        with caplog.at_level(logging.ERROR):
```
Nora-Olivia-Ammann, line 461: I think there should be a pytest option for that too...?

Nora-Olivia-Ammann (approved):

PR #2433, opened 2026-09-15
docs: adopt the shared comment convention and apply it (DEV-7123)

mpro7:
Adopts the shared DaSCH comment convention (DEV-7123) in `CONVENTIONS.md` and applies it.

Re-done from `main` under the rule as it now stands: dasch-claude-plugins#145 replaced "rewrite a constraint, never delete it" with delete-by-default, and #146 added the API-surface clause. The earlier version of this branch was classified under the withdrawn rule, so the force-push replaces the whole trim rather than adding to it.

## The convention

`### Naming and comments` splits into `### Naming`, bullets unchanged, and `### Comments`, which carries the seven core bullets from `eng/references/conventions-md-conventions.md` verbatim — diffed after whitespace normalisation, all seven byte-identical. **Nine bullets, not eight:** this repo needs two doc-comment lines, one for Google-style docstrings (`pyproject.toml` already sets `convention = "google"`) and one for the formats with no doc tool.

Three notes stay beneath the core because it does not carry them: the phrasings that pass the test routinely, the dated outside-world claim, and the warning that a tense flip is not a fix. The first is load-bearing here, since every trigger-word hit in the sweep is `used to` meaning *employed to*.

Three instruction files held their own comment rules and now point at the one in `CONVENTIONS.md`: `REVIEW.md`, which carried a second copy including the withdrawn "flag as rewrite, never as delete"; and `commands/xmlupload/CLAUDE.md` and `xmllib/CLAUDE.md`, which asked for "descriptive" and "comprehensive" docstrings against the core's one-or-two-sentence bound.

## The trim

CI and config `40 -> 32` blocks, Python comments `28 -> 21`, docstrings `136 -> 133`. 247 candidates classified, 0 uncovered: **184 keep, 50 compress, 13 delete.**

Beyond compression, thirteen deletions. Nine comments that restate the line below them, under core bullet 4 — `# happy path` above `if response is None`, `# retry if it is a retriable status code` above `if should_retry_request(...)`, `# non retriable error` above its `else`, `# update` above `update_parser`, a placeholder constant's own name, two assertion labels repeating the asserted strings, a duplicated escalate note, and a block restating the workflow step directly below it. One requirement id is stripped from a test comment, under the same bullet.

**The four docstring deletions are commit `14699d6f`, kept separate so it can be dropped on its own.** Two classes documented as "Placeholder type", one implementation repeating its protocol's docstring verbatim, and an 18-line `:param` list restating `Project.__init__`'s signature in a style `convention = "google"` contradicts.

Left alone, deliberately: **61 `xmllib` docstrings that mkdocstrings renders into `docs/xmllib-docs/`** — trimming one edits the docs site; 27 more that are mechanically fixed (12 held by `D103`, 5 whose deletion would orphan a `# noqa`, 10 that are their function's entire body); the 32 `excel2xml` and `dsp_base_resources` docstrings over 12 lines, whose length is `Args:`/`Returns:` tables and which are API surface a caller reads; and `tests-e2e.yml:13-24`, now 12 lines from 18, because it still carries three distinct invariants.

No `Args:`, `Returns:` or `Raises:` block is touched anywhere, and no parameter documentation is removed.

## Measurement

One basis: every tracked file of the type, `test/` and `scripts/` included.

| | before | after |
| -- | -- | -- |
| Python `#` comment lines | 935 / 84245 | 912 / 84179 |
| yaml + toml `#` comment lines | 238 / 1972 | 179 / 1913 |
| Python docstring lines, AST-counted | 8205 | 8164 |

## Verification

The scope guard ran on `3b31e9a1..HEAD`, the four trim commits; the convention commit is deliberately outside it, since a range containing `CONVENTIONS.md` reports clean for the wrong reason.

The grep guard in the rollout prompt cannot verify this trim: it reports 34 violations on that range, every one a docstring line, because a docstring is a string literal and `+    Args:` reads to it as code. An AST-aware guard resolves each changed line against the file's own tokens and syntax tree instead, and passes. It ships with 43 tests, which caught two defects in it — most importantly that a change to `X = 2  # comment` was passing as a comment change.

Also clean: `ruff check . --ignore=D101,D102,PLR2004` and `ruff format --check` (the exact CI commands), `python -m compileall` over `src test scripts`, and `yamllint`, whose one long-line warning pre-dates this branch.

Not covered: nothing in the measured languages. `.xml`, `.ttl`, `.xsd` and the binary fixtures have no comment-marker set and were never in scope. The `justfile`s reach no `--lang` flag in the rollout tooling, so their 35 comment lines were read whole instead.

One item the convention rules on but this branch cannot fix: four undated present-tense claims about other products survive (`xmllib/general_functions.py:38,89`, `xmllib/models/dsp_base_resources.py:167,214` — "cannot **currently** be displayed in the dsp-app"). `CONVENTIONS.md` says such a claim takes a date or a link; what is missing is the fact, not the ruling, and two of the four are published docs. Worth a follow-up issue.


.github/workflows/check-external-links.yml
```diff
@@ -1,15 +1,10 @@
...
-# cannot stall every open PR in the repo.
-#
-# Internal links and `#heading` anchors are NOT checked here: `mkdocs build
-# --strict` owns those (see the `validation:` block in mkdocs.yml), and it is a
-# required check precisely because it depends on nothing outside this repository.
+# Deliberately NOT a required status check on main: it reaches ~40 third-party hosts, so as a merge
+# gate it fails for reasons unrelated to the pull request under test. A dead link still shows as a
+# red ✗ to the author. Internal links and `#heading` anchors belong to `mkdocs build --strict` (the
+# `validation:` block in mkdocs.yml), which is required precisely because it depends on nothing
+# outside this repository.
```
jnussbaum: This still narrates too many things which can go stale, or are outside of this repo:
- "NOT a required status check on main" - This is a GitHub UI setting, outside of this codebase
- the number ~40 can go stale, better remove it
- Internal links belong to mkdocs build - True, but I wonder if this is the right location to document this.
- "which is required" - GitHub setting, not in the codebase

lychee.toml
```diff
@@ -1,14 +1,9 @@
-# Configuration for `just check-links` (see .github/workflows/check-external-links.yml).
+# Configuration for `just check-links`. Why this is deliberately not a required status check is in
+# .github/workflows/check-external-links.yml.
```
jnussbaum: If a status check is required or not doesn't belong here.


```suggestion
# Configuration for `just check-links` (see .github/workflows/check-external-links.yml).
```

lychee.toml
```diff
@@ -1,14 +1,9 @@
...
-# This checks that the links in ./docs still resolve. It is deliberately NOT a
-# required status check: it depends on ~40 third-party hosts staying up and not
-# rate-limiting GitHub runner IPs, so as a merge gate it fails for reasons that
-# have nothing to do with the pull request under test.
-#
-# Internal links and `#heading` anchors are NOT validated here — `mkdocs build
-# --strict` owns those (see the `validation:` block in mkdocs.yml), and it uses
-# the site's real slug algorithm. Leaving `include_fragments` off keeps this
-# file from second-guessing it with a different one.
+# Internal links and `#heading` anchors are not validated here: `mkdocs build --strict` owns those
```
jnussbaum: This here is the good place to state that internal links are validated by mkdocs build. Remove it from the other file where I commented on it.

Nora-Olivia-Ammann (approved):

REVIEW.md
```diff
@@ -26,16 +26,10 @@ Flag anything below that a change violates.
...
-      this file tell whether it is still true?* Flag as **rewrite**, never as **delete** — the
-      rationale must survive, and a bare tense-flip usually does not fix it
-- [ ] Expected non-violations, do not flag: `used to` meaning *employed to*; a marker whose baseline is
-      stated nearby; `now`/`currently`/`before` describing execution order or live server state
-- [ ] A claim about another service, library, or product carries a date or a source link — an undated
-      `currently` cannot be checked from the file at all
+- [ ] Added comments follow `CONVENTIONS.md` -> "Comments", in every format and not only `.py`: each states
+      a standing constraint rather than what the change discovered, what belongs in the commit body, `docs/`
+      or the tracker is not left in the source, an outside-world claim carries a date or a source link, and
+      the three carve-outs listed there are not findings. Flag as **rewrite**, never as **delete**
```
jnussbaum: This kind of super-condensed summaries are often cryptic. This one here is grammatically incoherent.

mpro7:
As this repo was the first one in the commit convention chain, so there were few things that improved over next few ones, so the repo need to be re-done. Therefore I am asking Johannes for re-review as he pointed few things out, but @Nora-Olivia-Ammann fell free to re-review too.

@jnussbaum all four addressed, but by a rewrite rather than a patch: the rule this was classified under has been withdrawn (dasch-claude-plugins#145 made the trim delete-by-default), so the force-push replaces the whole trim. `lychee.toml` opens with your suggested line, the mkdocs note lives only there, and `REVIEW.md` is now a plain pointer.

One thing I could not close, if you want it: `xmllib/general_functions.py:38,89` and `models/dsp_base_resources.py:167,214` say "cannot **currently** be displayed in the dsp-app". The convention says such a claim needs a date or a link, but I can't establish the dsp-app side. Two of them are also rendered into the docs.

jnussbaum (approved):

PR #2434, opened 2026-09-15
docs: point the merge-pr-when-green comment at branch protection (DEV-7123)

mpro7:
Comment-only, stacked on #2433 (DEV-7123). **Review #2433 first** — it is the convention this change is measured against.

## What this is

dsp-tools is the calibration repo for the rollout, and the sweep found almost nothing: 590 Python files, 7220 added lines since the 2026-06-01 cutoff, a 1% comment share, zero comment blocks and zero narrative hits. The whole candidate set is three comment blocks in `.github`, and two of them are keeps. This PR is the one edit that survived classification.

| Candidate | Verdict | Reason |
|---|---|---|
| `.github/workflows/tests-e2e.yml:13-30` | keep | States why the predicate keys on the head branch and must not become a bot-author check, and why a job-level `if:` is used rather than a trigger filter. Both are constraints a reader would otherwise break. |
| `.github/workflows/tests.yml:16-28` | keep | Two separate five-line rationales, not one over-long one. Cutting the `distribution` half would remove the reason the skip is safe. |
| `.github/actions/merge-pr-when-green/action.yml:24-37` | rewrite | Below. |

## The change

The comment opened with "`main` requires 13 status checks + 1 review". The count is a claim about another system's live configuration, and `CONVENTIONS.md` -> "Comments" requires those to carry a date or a pointer to the authority that settles them. It is correct as of today (the branch-protection API returns exactly 13 contexts, one required review, `strict`, `enforce_admins`), and it rots silently the next time a required check is added, in a settings change nobody would think to mirror in a comment. It now points at the branch-protection settings instead.

`Note that ...` also loses its opening, which named nothing and is on the convention's narration list. The GitHub bypass semantics the comment exists for are untouched, and the block is the same 14 lines: this is a rewrite, not a trim.

## Verification

Every added and removed line in `git diff <base>...HEAD` is a comment or blank (step 6 of the sweep prompt, clean). No behaviour, rename or dead-code change.

After #2433 merges this branch needs `git rebase --onto origin/main <old-parent-tip> docs/dev-7123-dsp-tools-comment-trim` and a lease-force-push: dsp-tools is squash-merge, so the parent's commit lands on main under a new SHA and GitHub's retarget alone would leave the convention diff showing here.


mpro7:
Folded into #2433 as its second commit: the trim is 13 lines in one file, and the step-6 scope check runs just as mechanically on the commit range as on a PR diff. Branch deleted.

PR #2438, opened 2026-09-18
feat: support the geolocation value type (DEV-7296)

derschnee68:
Resolves [DEV-7296](https://linear.app/dasch/issue/DEV-7296/implement-in-dsp-tools).

The dsp-tools half of the geolocation work, after dsp-api ([DEV-7285](https://linear.app/dasch/issue/DEV-7285), dasch-swiss/dsp-api#4340, released in v40.0.0) and dsp-js/dsp-app ([DEV-7286](https://linear.app/dasch/issue/DEV-7286), released in v13.15.0, so the release-order constraint is met).

**Reworked** onto the revised plan: design in [dasch-specs#210](https://github.com/dasch-swiss/dasch-specs/pull/210) (plan 04), execution delta in [dasch-specs#218](https://github.com/dasch-swiss/dasch-specs/pull/218) (plan 05). This addresses the review comments on the plan: named ordinates, mandatory `crs`, excel2json, a DMS converter, and systematic testdata.

## The XML surface

```xml
<geolocation-prop name=":hasFindspot">
    <geolocation crs="LV95" easting="2600000" northing="1200000"/>
    <geolocation crs="CRS84" longitude="8.550" latitude="47.37"/>
</geolocation-prop>
```

- `crs` is **required** and is one of `CRS84`, `LV95`, `LV03` (`xs:enumeration`), so a missing or unlisted CRS, including `EPSG:4326`, fails local schema validation.
- The ordinates are **named attributes**, so they cannot be transposed. The WKT literal dsp-api stores (`<iri> POINT(x y)`) is composed inside dsp-tools, X first, with each ordinate's written form preserved (`8.550` stays `8.550`).

## Validation

- XSD: `crs` present and allowlisted; ordinates are plain decimals (ASCII digits, no exponent).
- Python, one CRS table in `xmllib/internal/geolocation.py`, which mirrors dsp-api's `Geolocation.scala`:
  - **the pair must match the CRS**. XSD 1.0 cannot make one attribute depend on another's value, so e.g. `crs="LV95" longitude="…"` is reported as *expected the attributes 'easting' and 'northing'. Found 'longitude' and 'latitude', which belong to a geographic CRS.*
  - **bounds per CRS**, naming the attribute, its range and its CRS.
- validate-data runs this in `python_checks.check_geolocation_values`, and xmlupload runs it through the same validation step. `transform_geolocation` repeats the check, so `--skip-validation` cannot send an invalid value.
- The SHACL shape keeps `datatype`/`minCount`/`maxCount` and drops the regex. Dropping it avoids reporting the same problem twice.

## xmllib

- `add_geolocation_geographic(prop, "CRS84", *, longitude, latitude)` and `add_geolocation_projected(prop, "LV95" | "LV03", *, easting, northing)`, each with `_multiple` and `_optional` variants. The ordinates are keyword-only.
- `GeographicCoordinates` and `ProjectedCoordinates` are `kw_only` value classes for the `_multiple` methods.
- `xmllib.dms_to_decimal_degrees(47, 22, 13.2, "N") == "47.37033"`, which keeps enough decimals to recover the seconds at their input precision.
- Removed, never released: `add_geolocation*`, `geolocation_from_lat_long`.

## Also

- `create`, JSON schemas, and excel2json (`Geolocation` takes no gui attributes).
- Fixed: validate-data raised `KeyError` on any geolocation value, because the value type was missing from `VALUE_INFO_TRIPLE_OBJECT_TYPE`.
- Docs: `xml-data-file.md#geolocation` and an xmllib page for the coordinate classes.

## Test plan

- `just unittests` and `just integration-tests`: 3096 passed
- ruff, ruff format, mypy (strict, 583 files), vulture, and the xmllib docstring-link check are clean
- e2e against the v40.0.0 stack: `test/e2e/commands/validate_data/` and `test_create_get_xmlupload.py`, 32 passed. This includes the restored systematic fixtures (one value per CRS) and new `core_validation` correct/violation resources.


BalduinLandolt (commented):
If I see it correctly, this does not address the comments I left on the plan.  
Other than that, it looks good to me, but I think it makes more sense to have @jnussbaum review the code in this repo.

src/dsp_tools/xmllib/internal/geolocation.py
```diff
@@ -0,0 +1,190 @@
...
+    kind=CrsKind.PROJECTED,
+    x_name="easting",
+    y_name="northing",
+    x_min=Decimal("484273.3"),
+    x_max=Decimal("837939.88"),
+    y_min=Decimal("73150.16"),
+    y_max=Decimal("299970.97"),
+)
+
+ALL_CRS: tuple[Crs, ...] = (CRS84, LV95, LV03)
```
Nora-Olivia-Ammann: An enum class in `src/dsp_tools/xmllib/models/config_options.py` would be more user friendly to set the kind. Why in any case does this have such a complicated class construction?
None of this information makes it into the xml? This is not the place for background documentation in my opinion. If it is background knowledge it belongs in the documentation that goes for the app as well. We should then link to that.

src/dsp_tools/commands/validate_data/validation/python_checks.py
```diff
@@ -62,6 +65,44 @@ def _create_input_problems(duplicates: dict[str, int]) -> list[InputProblem]:
     return all_duplicates
 
 
+def check_geolocation_values(parsed_resources: list[ParsedResource]) -> list[InputProblem]:
+    """
+    Check each geolocation against the pair of ordinates and the bounds of its CRS.
+
+    The SHACL validation cannot do this:
```
Nora-Olivia-Ammann: This is not correct, there are several ways you could do this. The easiest is probably with `sh:or` and `sh:minInclusive` and `sh:maxInclusive`. If this is insufficient for claude tell me and I can give a more detailed comment.

src/dsp_tools/xmllib/models/res.py
```diff
@@ -623,6 +630,392 @@ def add_decimal_optional(
             self.add_decimal(prop_name, value, permissions, comment)
         return self
 
+    #######################
+    # GeolocationValue
+    #######################
+
+    def add_geolocation_geographic(
```
Nora-Olivia-Ammann: In my opinion it would bloat it less and be more consistent with the way we map it to the json if we instead of having different function to use a parameter called `geolocation_type` or so, which then takes the enum class I mentioned.

PR #2440, opened 2026-09-22
feat(xmllib): add source provenance to value builders (DEV-7359)

jnussbaum:
[DEV-7359](https://linear.app/dasch/issue/DEV-7359)

## Summary

- Adds an optional `SourceProvenance` carrier (`source_file`, `sheet`, `row`, `cell`) to
  `xmllib`'s value API, threaded from every `add_...` builder method through the internal value
  factories into the warning system.
- Emits four new trailing columns (`Source File`, `Sheet`, `Row`, `Cell`) in
  `xmllib_warnings.csv` and the terminal warning message.
- Additive and backward compatible: the parameter defaults to `None` everywhere, and an existing
  caller that omits it sees no behaviour change.

## Changes

- **New public API**: `SourceProvenance` (`src/dsp_tools/xmllib/models/provenance.py`),
  re-exported from `dsp_tools.xmllib`.
- **Warning carrier**: `MessageInfo` gains a `provenance` field; the CSV writer and terminal
  message builder append its four fields as trailing columns.
- **Value factories**: all 13 `Value.new()` classmethods, `FileValue.new`, `IIIFUri.new`, and the
  `emit_xmllib_input_type_mismatch_warning` helper accept and forward `provenance`.
- **Builder surface**: all 43 value `add_...` methods on `Resource`, plus the in-scope value
  methods on `RegionResource`/`LinkResource`/`VideoSegmentResource`/`AudioSegmentResource`, gain
  the optional kwarg. A `_multiple` variant applies one `provenance` to the whole collection.
  `RegionResource`'s geometry methods (`add_rectangle`/`add_polygon`/`add_circle`) and
  `Metadata.new` (license/copyright/authorship) are deliberately excluded — no value-factory
  check exists to attach provenance to, and this is marked with an in-repo comment.
- **Secondary helpers**: comment correction, value-order validation, empty-collection info,
  angular-brackets info, and the `ListLookup` warning path all carry `provenance` too, closing a
  completeness gap an adversarial review found after the initial pass.
- **Docs**: new `docs/xmllib-docs/provenance.md` stub, `mkdocs.yml` nav entry, a short note in
  `advanced-set-up.md`'s CSV section, and an update to `src/dsp_tools/xmllib/CLAUDE.md`.

## Test Plan

- Full unit test suite: 2826 passed (`just unittests`).
- `just ruff-check`, `just ruff-format-check`, `just mypy`, `just markdownlint`: all clean.
- New tests: `SourceProvenance` construction/defaults/immutability; end-to-end CSV round-trip
  through `Resource.add_integer`, `RegionResource.add_comment`, and `ListLookup.get_node_via_list_name`.
- Adversarial multi-dimension code review (consistency, simplicity, dune, python) run against the
  full diff; all confirmed findings were fixed in follow-up commits.


Nora-Olivia-Ammann (commented):
What I don't see is a check if the input to the provenance class is correct, we must expect that sometimes this input could be wrong itself, then it is vital that it doesn't crash. Otherwise looks good!

jnussbaum:
> What I don't see is a check if the input to the provenance class is correct, we must expect that sometimes this input could be wrong itself, then it is vital that it doesn't crash. Otherwise looks good!

Thanks for the feedback. Fixed in https://github.com/dasch-swiss/dsp-tools/pull/2440/commits/45eb68360aafabf7c0961f38c657a5595735f336

Nora-Olivia-Ammann (approved):

PR #2441, opened 2026-09-22
docs: add ARCH-MAP.md, CONTEXT.md, ADRs, and eng.yaml (DEV-7256)

jnussbaum:
## Summary

Bootstraps the `dune` plugin's architecture map and domain glossary, and the `eng` plugin's path-to-binding file, for dsp-tools (DEV-7256).

- **ARCH-MAP.md** — 15 components plus a cross-cutting-concerns section, built from the existing per-module `CLAUDE.md` files and the `CONVENTIONS.md` subsystem inventory. Documents several confirmed boundary violations (commands importing each other; `utils`/`clients` importing back from `commands`; `xmllib` importing dsp-tools internals despite its own no-internals rule).
- **CONTEXT.md** — the project's domain vocabulary (ontology, resource, cardinality, permission/DOAP, identifiers including ARK vs. IRI, legal metadata, the Group A/B command split).
- **docs/adr/** — five ADRs: two new (the Group A/B error-handling split, the split xmlupload/ingest-xmlupload workflow) and three converted from the repo's old, informal `docs/developers/decision-log.md`, which is removed.
- **eng.yaml** — binds conventions, commands, skills, and reviewers to paths, validated against the `eng` plugin's resolver.

A follow-up code review (multi-perspective: python, consistency, and dune reviewers, plus a full skill-review of the one skill edit) found and fixed:

- A confirmed CI break: the 5 new `docs/adr/*.md` files weren't in `mkdocs.yml`'s nav, and `omitted_files: warn` becomes a build error under the `--strict` flag `check-docs` CI runs. Fixed with `not_in_nav`, verified by running `mkdocs build --strict` locally.
- A completeness gap in `ARCH-MAP.md` itself (`eng.yaml`/`CONTEXT.md` were added after its last completeness check and were unmapped).
- Two components (`get`, `ingest-xmlupload`) whose "Used by" field contradicted their own "Public interface"/"Boundary rules" text.
- A mislabeled enforcement claim (`excel2json`'s `ordered_keys` guard, labeled "static-analysis" though nothing tests it).
- Several missing reciprocal Depends-on/Used-by edges.
- Smaller fixes: stale cross-references, `eng.yaml`'s `dune-reviewer` trigger being too narrow, the `update-docs` skill's missing `Write` permission and `docs/adr/` entry.

## Not included

`ARCH-MAP.md`'s `utils` entry flags the `utils`→`commands` import inversion as the map's own highest-value candidate for an ADR. Writing that ADR needs a dedicated rationale-confirmation session with the user (per how ADRs are authored here), so it's tracked as a follow-up rather than done in this PR.

## Test plan

- [x] `/dune:dune-map check` — 0 unmapped, 0 multiply-mapped files
- [x] `uv run eng_yaml.py validate` and `resolve --files` against representative paths
- [x] `mkdocs build --strict` — passes locally (confirms the CI-break fix)
- [x] `markdownlint` / `yamllint` on all new and changed files


mkdocs.yml
```diff
@@ -16,6 +16,11 @@ validation:
   unrecognized_links: warn
   anchors: warn
 
+# ADRs are a git-native decision log, not a published page — excluded from the nav-completeness
+# check above rather than given a nav entry.
```
jnussbaum, line 20:
```suggestion
```

docs/developers/decision-log.md
jnussbaum, line 1: The contents of these earlier ADRs have been moved to the new format

docs/adr/0001-split-error-handling-fail-fast-vs-aggregate.md
```diff
@@ -0,0 +1,15 @@
+---
+status: accepted
+date: 2026-09-22
+---
+
+# Split error handling into fail-fast Group A and aggregate-all Group B
+
+Commands that run in a controlled environment before production use (`create`, `get`, `xmlupload`,
+`upload-files`, `ingest-files`, `ingest-xmlupload`, `resume-xmlupload`) may fail fast with a full traceback,
```
Nora-Olivia-Ammann: `upload-files`, `ingest-files` will soon be user-facing

docs/adr/0003-split-xmlupload-workflow.md
```diff
@@ -0,0 +1,12 @@
+---
+status: accepted
+date: 2026-09-22
+---
+
+# Split xmlupload into upload-files, ingest-files, ingest-xmlupload, and resume-xmlupload
```
Nora-Olivia-Ammann: I would not do that, we will have the bulk import that will cover that, until we decide that we deprecate the xmlupload I would just leave it alone. i do not see the value investing time in splitting it up now.

CONTEXT.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Identifiers
+
+**Shortcode**:
+A project's unique 4-hex-character identifier, assigned by DaSCH.
+
+**Shortname**:
+A project's unique, human-readable slug (`xsd:NCNAME` form).
+
+**IRI**:
```
Nora-Olivia-Ammann: This is incorrect, IRI a standard like a URL, all the properties and classes in the graph are IRIs as well. this is only the convention for a resource IRI. the others are different.

docs/adr/0001-split-error-handling-fail-fast-vs-aggregate.md
```diff
@@ -0,0 +1,15 @@
+---
+status: accepted
+date: 2026-09-22
```
jnussbaum: should be February 2026

docs/adr/0001-split-error-handling-fail-fast-vs-aggregate.md
jnussbaum: I think that's not the point here. I understand group A as commands that are tested on localhost/testserver, so if something goes wrong, the devs or infra can be consulted, before going to prod. 

For group B, the users expect a higher user-friendliness, because they don't want to interrupt their workflow and consult a dev.

I will instruct Claude to frame that more clearly.

docs/adr/0003-split-xmlupload-workflow.md
jnussbaum: Sorry, this ADR is formulated in a misleading way. The correct framing of this ADR should be: At some point of time in the past (claude should find out via git history when it was) we decided to expose a second way to do an xmlupload, with separate steps. I will reformulate it.

CONTEXT.md
jnussbaum: Thanks for catching this, I will fix it.

Nora-Olivia-Ammann (approved):

